PR #5, opened 2022-01-27
added a request approval reset button to the action drop down of mode…

ARADDCC002:
…l and deployment overviews


pages/model/[uuid].tsx
```diff
@@ -148,6 +149,10 @@ const Model = () => {
     }
   }
 
+  const requestApprovalReset = async () => {
+    await postEndpoint(`/api/v1/version/${uuid}/reset-approvals`, {}).then((res) => res.json())
```
Comment (a3957273) [inline pages/model/[uuid].tsx:153]: We need to work out how we're handling errors, automatic retrying, etc.  Could you raise a ticket for this?

Comment (a3957273) [reply pages/model/[uuid].tsx:153]: (Can just be a one-liner to remind us on the elaboration board)

pages/model/[uuid].tsx
```diff
@@ -224,6 +229,15 @@ const Model = () => {
                   </ListItemIcon>
                   <ListItemText>Upload new version</ListItemText>
                 </MenuItem>
+                <MenuItem
+                  onClick={requestApprovalReset}
+                  disabled={version?.managerApproved === 'No Response' && version?.reviewerApproved === 'No Response'}
```
Comment (a3957273) [inline pages/model/[uuid].tsx:234]: If this logic starts getting any more complicated we should think about moving it above into an intermediary variable.  It's fine for now, :)

pages/deployment/[uuid].tsx
```diff
@@ -111,6 +132,13 @@ export default function Deployment() {
 
   const deploymentTag = `${uiConfig?.registry.host}/${currentUser!.id}/${tag}`
 
+  const requestApprovalReset = async () => {
+    await postEndpoint(
+      `/api/v1/deployment/${deployment?.metadata?.highLevelDetails.modelID}/reset-approvals`,
+      deployment
```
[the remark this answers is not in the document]
Comment (a3957273) [reply pages/deployment/[uuid].tsx:138]: I've just noticed that `deployment` here is the full deployment object including all the metadata, etc?  I think we should actually only be posting `{ uuid: deployment.uuid }` in retrospect.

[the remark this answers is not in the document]
Comment (a3957273) [reply pages/deployment/[uuid].tsx:138]: Alternatively, post no data at all.  Just an empty object.  We already have the deployment uuid.

server/models/Deployment.ts
```diff
@@ -13,7 +13,7 @@ const DeploymentSchema = new Schema(
     managerApproved: { type: String, required: true, enum: approvalStates, default: 'No Response' },
 
     logs: [{ timestamp: Date, level: String, msg: String }],
-    built: { type: String, required: true, enum: approvalStates, default: 'No Response' },
+    built: { type: Boolean, required: true, default: false },
```
a3957273, line 16: 👍

server/routes/v1/deployment.ts
```diff
@@ -111,3 +113,28 @@ export const postDeployment = [
...
+
+export const resetDeploymentApprovals = [
+  ensureUserRole('user'),
+  bodyParser.json(),
+  async (req: Request, res: Response) => {
+    const user = req.user
+    const body = req.body as any
+    const deployment: Deployment = await DeploymentModel.findOne({ uuid: body.uuid })
+    if (user?.id !== deployment.metadata.contacts.requester) {
+      throw UnAuthorised({}, 'User is not authorised to do this operation.')
```
a3957273, line 125: I'd like to see the deployment that the user is requesting here.  So the first parameter should be `{ deploymentUuid: body.uuid }`.

server/routes/v1/deployment.ts
```diff
@@ -111,3 +113,28 @@ export const postDeployment = [
...
   },
 ]
+
+export const resetDeploymentApprovals = [
+  ensureUserRole('user'),
+  bodyParser.json(),
+  async (req: Request, res: Response) => {
+    const user = req.user
+    const body = req.body as any
+    const deployment: Deployment = await DeploymentModel.findOne({ uuid: body.uuid })
```
a3957273, line 123: You _cannot_ cast to `Deployment` here, as it's a `Deployment | null` type.  You _must_ check that Deployment is a thing, or this function will error when it's not found.

server/routes/v1/deployment.ts
a3957273: Also, the right hand side of these error messages are user facing, so maybe make it more descriptive (e.g. `You cannot reset the approvals for a deployment you do not own.`).

server/routes/v1/deployment.ts
```diff
@@ -111,3 +113,28 @@ export const postDeployment = [
...
+    if (user?.id !== deployment.metadata.contacts.requester) {
```
a3957273, line 124: I'm worried about id changes, whenever that becomes a requirement.  Please could we compare the database id with the deployment owner (e.g. `user?._id !== deployment.owner`).

server/routes/v1/deployment.ts
```diff
@@ -111,3 +113,28 @@ export const postDeployment = [
...
+      throw UnAuthorised({}, 'User is not authorised to do this operation.')
+    }
+    const version = await VersionModel.findOne({
+      model: body.model,
+      version: body.metadata.highLevelDetails.initialVersionRequested,
+    })
+    if (!version) {
+      throw BadReq({}, 'Unabled to find version for requested deployment')
```
a3957273, line 132: Include details like deploymentId, what version was requested.

server/routes/v1/deployment.ts
a3957273, line 123: `{ uuid: body.uuid }`, mentioned above it makes no sense to send entire document.  Instead, take this from `req.params.uuid`.

a3957273 (changes requested):

a3957273:
I think this should be disabled on models / deployments that are _already_ approved.  I can't think of a situation where a double approved model would want to restart the process?

pages/deployment/[uuid].tsx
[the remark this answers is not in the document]
Comment (ARADDCC002) [reply pages/deployment/[uuid].tsx:138]: I think I did this so we didn't need to make a DB call in order to get the deployment data on the server-side, as we already have the information we need in the client-side. I'll make the changes, though.

server/routes/v1/deployment.ts
ARADDCC002: Resolved

server/routes/v1/deployment.ts
ARADDCC002: Improved the error text.

server/routes/v1/deployment.ts
ARADDCC002: Resolved

ARADDCC002:
> I think this should be disabled on models / deployments that are _already_ approved. I can't think of a situation where a double approved model would want to restart the process?

The problem with this is that deployments only have one approval, if there's no approvals then you can't reset, and if it has been approved then in this scenario you wouldn't be able to reset it either. Unless of course it's rejected? But if it's rejected I assume you'd need to edit it that would reset the approvals anyway. 

I think there might be edge cases whereby people want to reset approvals in the case that a mistake has been spotted after it's been approved.

a3957273 (approved):
Looks great to me, 👍

PR #6, opened 2022-01-28
disabled model card version field in upload form in edit mode

ARADDCC002:
When editing a version the model card version field has been disabled.

### DIFF
--- a/src/common/FormTabs.tsx
+++ b/src/common/FormTabs.tsx
@@ -70,12 +70,20 @@ export default function FormTabs({
 
   useEffect(() => {
     formatUiWidgets()
-    if (mode === 'edit' || mode === 'newVersion') {
+    if (mode === 'edit') {
       let updatedSchema = cloneDeep(selectedSchema)
-      // Name should be disabled
+      // Version and name should be disabled in edit mode
+      updatedSchema.schema.properties.highLevelDetails.properties.name.readOnly = true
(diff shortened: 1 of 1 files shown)


at55425 (approved):

src/common/FormTabs.tsx
```diff
@@ -70,12 +70,20 @@ export default function FormTabs({
...
       let updatedSchema = cloneDeep(selectedSchema)
-      // Name should be disabled
+      // Version and name should be disabled in edit mode
       updatedSchema.schema.properties.highLevelDetails.properties.name.readOnly = true
+      updatedSchema.schema.properties.highLevelDetails.properties.modelCardVersion.readOnly = true
       setUpdatedSelectedSchema(updatedSchema)
-    } else {
+    }
+    if (mode === 'newVersion') {
+      let updatedSchema = cloneDeep(selectedSchema)
```
a3957273, line 81: You can deduplicate the code by moving the following two calls outside of the if statement.  They'll remain the same in both edit and newVersion mode?

```
       let updatedSchema = cloneDeep(selectedSchema)
       updatedSchema.schema.properties.highLevelDetails.properties.name.readOnly = true
```

a3957273 (dismissed):

src/common/FormTabs.tsx
a3957273: We discussed that this wasn't possible out of band.

PR #19, opened 2022-02-05
Ensure database state is consistent for uploads by putting all changes into a transaction

a3957273:
If the function fails at any time it will rollback all changes made to the database.


a3957273:
This PR only changes the following lines:

https://github.com/gchq/Bailo/pull/19/files#diff-afe061ba1783d701a25f12c0b4e12e018e72029ae7405bbe583670156d83b71eR39-R40

The other changes are part of PR #18, which should be merged before this.

PR #24, opened 2022-02-08
Feature/bai 89

ARADDCC002:
* Improved the styling on the submission step in the model upload form
* When creating a custom step in a form (ie, Files, Submission...) you can now specify to render a basic version of the component instead
* Added a basic file upload component - this is for when you are on the "upload existing" tab in the upload page.
* Added some enhanced styling to the file upload step of the "Designer" tab.


utils/formUtils.ts
```diff
@@ -24,6 +27,8 @@ export function createStep({
   type: StepType
   section: string
   render: Function
+  renderBasic: Function
```
a3957273, line 30: Make this optional and default to the value of `render` if it's not set?

ARADDCC002: I've just been changing it so that it returns a function with a react fragment whenever a new step is called as that's how steps without a "render" call function - should both be optional?

pages/upload.tsx
```diff
@@ -149,34 +148,38 @@ function Upload() {
   const onSubmit = async () => {
     setError(undefined)
 
-    const data = getStepsData(steps, true)
-    const form = new FormData()
+    if (steps.filter((e) => !e.isComplete(e)).length === 0) {
```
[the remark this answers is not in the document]
a3957273: I'd prefer if the layout was kept to the `if (error) { returnEarly }` approach, e.g:

```
if (!steps.every(e => e.isComplete(e)) {
  return setError("Ensure all steps are complete before submitting")
}

// otherwise, submit...
```

src/Form/RenderFileTab.tsx
```diff
@@ -30,7 +29,8 @@ export default function RenderFileTab(step: Step, steps: Array<Step>, setSteps:
   }
 
   const displayFilename = (filename: string) => {
-    return filename.length > 12 ? filename.split('.')[0].substring(0, 12) + '...' + filename.split('.')[1] : filename
+    let filenameSplit = filename.split('.')
+    return filenameSplit[0].length > 12 ? filenameSplit[0].substring(0, 12) + '...' + filenameSplit[1] : filename
```
a3957273, line 33: This breaks if given a filename like `model.tar.gz` (or any string with multiple dots).  Maybe rewrite as:

```
const displayFilename = (filename: string) => {
  const parts = filename.split('.')
  const ext = parts.pop()
  const base = parts.join('.')

  return base.length > 12 ? `${base}...${ext}` : filename
}

ARADDCC002: Much better, changed!

PR #34, opened 2022-02-24
added a check to make sure that when changing schemas, the form isn't…

ARADDCC002:
… replaced with just the duplicated steps from the previous schema


utils/formUtils.ts
```diff
@@ -53,15 +53,17 @@ export function createStep({
 }
 
 export function setStepState(steps: Array<Step>, setSteps: Function, step: Step, state: any) {
-  const index = steps.findIndex((iStep) => step.section === iStep.section)
+  if (step.schemaRef !== steps[0].schemaRef) {
```
a3957273, line 56: I think we only want to keep on setting the step state when the schemas _do match up_, not when they don't.

a3957273 (changes requested):

a3957273 (approved):
Approved in chats with @ARADDCC002

PR #39, opened 2022-03-06
Add 'token' location to pull instructions

a3957273:


pages/deployment/[uuid].tsx
```diff
@@ -193,7 +194,9 @@ export default function Deployment() {
         <DialogContent>
           <DialogContentText sx={{ backgroundColor: 'whitesmoke', p: 2 }}>
             <pre>
-              <span># Login to Docker</span>
+              <span>
+                # Login to Docker (your token can be found on the <Link href='/settings'>/settings</Link> page)
+              </span>
               <br />
```
Comment (ARADDCC002) [inline pages/deployment/[uuid].tsx]: Can we use MUI Boxes here and add styling instead of using breaks? Can't remember if we can use MUI components inside the <pre> tags or not!

ARADDCC002 (approved):

PR #60, opened 2022-03-14
Bugfix/bai 127 fix booleans in schemas

ARADDCC002:


utils/formUtils.ts
```diff
@@ -151,6 +151,9 @@ export function setStepsData(splitSchema: SplitSchema, setSplitSchema: Function,
 export function validateForm(step: Step) {
   const validator = new Validator()
   const sectionErrors = validator.validate(step.state, step.schema)
+  // This is temporay validation fix - dependency questions that rely on yes or no answers in the schema
+  // will have their data persist in the form object, even if the parent question's answer has been changed
+  sectionErrors.errors.filter((e) => e.name !== 'dependencies')
```
a3957273, line 156: I see the following errors:

```
0: Object { name: "additionalProperties", property: ".endDate['date']", message: "is an invalid additional property", … }
​​1: Object { name: "enum", property: ".endDate.hasEndDate", message: "should be equal to one of the allowed values", … }
​​2: Object { name: "oneOf", property: ".endDate", message: "should match exactly one schema in oneOf", … }
```

It still does not let me click 'next'.
Screenshot 2022-03-14 at 16 25 13

a3957273 (changes requested):

.github/workflows/node.js.yml
```diff
@@ -64,6 +64,7 @@ jobs:
       - run: |
           chmod +x ./server/scripts/waitForIt.sh
           ./server/scripts/waitForIt.sh localhost:8080 -t 10
+          sleep 60
```
a3957273, line 67: This should be removed.

server/utils/smtp.ts
```diff
@@ -29,13 +30,16 @@ export async function sendEmail({
...
+    const info = await transporter.sendMail({
+      from: config.get('smtp.from'), // sender address
+      to,
+      subject,
+      text,
+      html,
+    })
+    logger.info({ messageId: info.messageId }, 'Email sent')
+  } catch (err) {
+    throw GenericError({}, 'Error sending email to reviewers', 500)
```
a3957273, line 43: I think you should _also_ run `logger.error()` here with the original `err` to avoid losing that information.  Bunyan has custom loggers to help you log errors:

```

log.info(err);  // Special case to log an `Error` instance to the record.
                // This adds an "err" field with exception details
                // (including the stack) and sets "msg" to the exception
                // message.
log.info(err, 'more on this: %s', more);
                // ... or you can specify the "msg".
```

utils/formUtils.ts
```diff
@@ -7,6 +7,7 @@ import { Validator } from 'jsonschema'
 import { SplitSchema, Step, StepType } from '../types/interfaces'
 import RenderForm from '../src/Form/RenderForm'
 import RenderButtons from '../src/Form/RenderButtons'
+import { compose } from '@mui/system'
```
a3957273, line 10: Unused import?

a3957273 (changes requested):

.github/workflows/node.js.yml
ARADDCC002: Forgot about this! Removed.

utils/formUtils.ts
ARADDCC002: Not even sure how that got added, sometimes if you make a typo VS thinks it's an actual dependency and adds it!! :(

server/utils/smtp.ts
a3957273: Since this is likely to be a non-fatal error, you should use `logger.warn()` instead of `logger.error()`.

a3957273 (approved):
Looks good! :)

PR #61, opened 2022-03-21
Update helm chart

at55425:


helm/bailo/templates/config.yaml
```diff
@@ -5,6 +5,42 @@ metadata:
   labels:
     {{- include "bailo.labels" . | nindent 4 }}
 data:
+  registry.conf: |
+    version: 0.1
+    log:
+      level: debug
```
a3957273, line 11: In production we probably want to increase this log level to info.

helm/bailo/templates/deployment.yaml
```diff
@@ -57,9 +195,11 @@ spec:
                   key: mongodb-passwords
             - name: MONGO_CONNECTION_URI
               value: {{ include "bailo.mongoConnectionURI" . | quote }}
+            - name: SCHEMA_SETUP_SCRIPT
+              value: dist/server/scripts/exampleSetAllSchemas.js
```
a3957273, line 199: I think this should default to an empty script in production.

helm/bailo/templates/service.yaml
```diff
@@ -13,3 +13,37 @@ spec:
       name: http
   selector:
     {{- include "bailo.selectorLabels" . | nindent 4 }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: registry
```
a3957273, line 20: All names should be prefixed with the deployment name ala the existing pods.

helm/bailo/templates/service.yaml
```diff
@@ -13,3 +13,37 @@ spec:
...
+spec:
+  ports:
+    - port: {{ .Values.registry.port }}
+      targetPort: http
+      protocol: TCP
+      name: http
+  selector:
+    name: {{ .Values.registry.host }}
+--- 
+apiVersion: v1
```
a3957273, line 32: With the removal of `maildev`, I also don't think this is needed.

helm/bailo/values.yaml
```diff
@@ -22,12 +26,12 @@ autoscaling:
 podSecurityContext: {}
   # fsGroup: 2000
 securityContext:
-  capabilities:
-    drop:
-      - ALL
-  readOnlyRootFilesystem: true
+  # capabilities:
+  #   drop:
+  #     - ALL
```
a3957273, line 31: All permissions should be dropped by default and only the required capabilities added back in via `add` (at the same level of `drop`).

helm/bailo/values.yaml
```diff
@@ -49,15 +53,16 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
-resources: {}
+resources: #{}
```
a3957273, line 56: What's with the hash here?

helm/bailo/values.yaml
```diff
@@ -116,7 +121,29 @@ redis:
...
+  host: "registry"
   port: 5000
-  insecure: true
\ No newline at end of file
+  insecure: true
+  certDir: "/certs"
+  certFile: "cert.pem"
+  keyFile: "key.pem"
+
+smtp:
```
a3957273, line 134: Again, should be removed with the removal of `maildev`.

a3957273 (changes requested):

PR #67, opened 2022-03-24
Use chown under 'copy' to speed up build

a3957273:


ARADDCC002:
Seems much faster to me!

ARADDCC002 (approved):

PR #72, opened 2022-04-10
Remove redis as a dependency

a3957273:


Dockerfile.prod
```diff
@@ -43,6 +43,7 @@ RUN apk update \
 # You only need to copy next.config.js if you are NOT using the default configuration
 # COPY --from=builder /app/next.config.js ./
 COPY --from=builder --chown=nextjs:nextjs /app/dist ./dist
+COPY --from=builder --chown=nextjs:nextjs /app/lib/p-mongo-queue ./dist/lib/p-mongo-queue
```
a3957273, line 46: Required whilst `p-mongo-queue` is an external module.  Only `ts` files are copied across currently during a production build.  Won't be necessary if we can shift this to a published module elsewhere.

PR #76, opened 2022-04-27
made some minor adjustments to review and deployment page

ARADDCC002:
- Added some checks to the review page so any missing data wouldn't cause the page to crash
- Added a helpful warning alert to tell users that data was missing from models/deployments needing a review
- Altered the filter section of the "My Deployments"

### DIFF
--- a/pages/deployment/[uuid].tsx
+++ b/pages/deployment/[uuid].tsx
@@ -194,7 +194,7 @@ export default function Deployment() {
           <DialogContentText sx={{ backgroundColor: 'whitesmoke', p: 2 }}>
             <pre>
               <div>
-                # Login to Docker (your token can be found on the <Link href='/settings'>/settings</Link> page)
+                # Login to Docker (your token can be found on the <Link href='/settings'>settings</Link> page)
               </div>
               <CodeLine line={`docker login ${uiConfig?.registry.host} -u ${currentUser!.id}`} />
               <br />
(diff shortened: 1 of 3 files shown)


pages/deployment/[uuid].tsx
```diff
@@ -194,7 +194,7 @@ export default function Deployment() {
           <DialogContentText sx={{ backgroundColor: 'whitesmoke', p: 2 }}>
             <pre>
               <div>
-                # Login to Docker (your token can be found on the <Link href='/settings'>/settings</Link> page)
+                # Login to Docker (your token can be found on the <Link href='/settings'>settings</Link> page)
```
Comment (a3957273) [inline pages/deployment/[uuid].tsx:197]: This was intentional due to it being a URL reference.  Pointing to the `/settings` url.  Can be removed if confusing though.

pages/review.tsx
```diff
@@ -116,36 +121,50 @@ const ApprovalList = ({ type, category }: { type: RequestType; category: ReviewF
...
-                        {requestObj.deployment.model.currentMetadata.highLevelDetails.name}
+                        {requestObj?.deployment?.model?.currentMetadata?.highLevelDetails?.name}
                       </MuiLink>
                     </Link>
                   </Typography>
+                  {requestObj?.deployment === undefined ||
+                    (requestObj?.deployment === null && (
+                      <Alert sx={{ mt: 2 }} severity='warning'>
+                        This deployment appears to have data missing - check with the requester to make sure it was
+                        requestded correctly
```
a3957273, line 165: typo in 'requested'

a3957273 (approved):

pages/deployment/[uuid].tsx
Comment (ARADDCC002) [reply pages/deployment/[uuid].tsx:197]: Makes sense, but I personally prefer this way as most people don't navigate to a page by manually typing out an address and instead go by the page name and a hyperlink - happy to concede this, though, just my thoughts!

pages/review.tsx
ARADDCC002: Haha whoops, will resolve.

PR #78, opened 2022-04-28
added a configurable alert and checkbox to appear on upload page

ARADDCC002:


a3957273:
Screenshot 2022-05-03 at 13 20 56

The alert looks a little odd with it's current size.  Would it be better if it were full width?

a3957273 (approved):

PR #79, opened 2022-04-28
Support mongoose connection to pMongoQueue

a3957273:


lib/p-mongo-queue/pMongoQueue.js
```diff
@@ -23,9 +23,9 @@ module.exports = class Queue extends EventEmitter {
     if (!db) {
       throw new Error('p-mongo-queue: provide a mongodb.MongoClient.db')
     }
-    if (db instanceof EventEmitter) {
-      throw new Error('p-mongo-queue: provide a mongodb.MongoClient.db from mongodb@4')
-    }
+    // if (db instanceof EventEmitter) {
```
ARADDCC002, line 26: Should this still be commented out or removed? Are we likely to use it again in the future?

ARADDCC002 (approved):

PR #80, opened 2022-05-03
added a loading bar to the upload

ARADDCC002:


a3957273:
Screenshot 2022-05-04 at 12 19 11

Percentage looks off center for me.  Also, clicking 'twice' on the Submit button appears to start two requests that fight with one another.  Could we disable the submit button once a submit is starting?

a3957273 (approved):

PR #81, opened 2022-05-05
Feature/tslint

ARADDCC002:


pages/deployment/[uuid].tsx
```diff
@@ -128,18 +134,18 @@ export default function Deployment() {
   if (error) return error
 
   if (isDeploymentLoading || isUiConfigLoading || isCurrentUserLoading) {
-    return <Wrapper title={'Loading...'} page={'deployment'} />
+    return <Wrapper title='Loading...' page='deployment' />
   }
 
-  const deploymentTag = `${uiConfig?.registry.host}/${currentUser!.id}/${tag}`
+  const deploymentTag = `${uiConfig?.registry.host}/${currentUser?.id}/${tag}`
```
Comment (a3957273) [inline pages/deployment/[uuid].tsx:140]: This change isn't valid.  If `currentUser` is undefined we should error, instead of returning `undefined` here.

pages/deployment/[uuid].tsx
```diff
@@ -128,18 +134,18 @@ export default function Deployment() {
...
 
   const requestApprovalReset = async () => {
     await postEndpoint(`/api/v1/deployment/${deployment?.uuid}/reset-approvals`, {}).then((res) => res.json())
   }
 
   return (
     <>
-      <Wrapper title={`Deployment: ${deployment!.metadata.highLevelDetails.name}`} page={'deployment'}>
+      <Wrapper title={`Deployment: ${deployment?.metadata.highLevelDetails.name}`} page='deployment'>
```
Comment (a3957273) [inline pages/deployment/[uuid].tsx:148]: If deployment is undefined, `highLevelDetails` on `deployment.metadata` will also be undefined, so this will break.  We should inform TypeScript that this is defined in other ways.

Comment (a3957273) [reply pages/deployment/[uuid].tsx:148]: This can be done using a Type Guard.

pages/deployment/[uuid].tsx
```diff
@@ -181,11 +187,11 @@ export default function Deployment() {
           </Box>
           <Box sx={{ marginBottom: 3 }} />
 
-          {tab === 'overview' && <DeploymentOverview version={deployment} use={'DEPLOYMENT'} />}
+          {tab === 'overview' && <DeploymentOverview version={deployment} use='DEPLOYMENT' />}
 
           {tab === 'compliance' && <ComplianceFlow initialElements={complianceFlow} />}
 
-          {tab === 'build' && <TerminalLog logs={deployment!.logs} title='Deployment Build Logs' />}
+          {tab === 'build' && <TerminalLog logs={deployment?.logs} title='Deployment Build Logs' />}
```
Comment (a3957273) [inline pages/deployment/[uuid].tsx:194]: This isn't a valid transformation

pages/deployment/[uuid].tsx
```diff
@@ -49,10 +49,16 @@ function CodeLine({ line }) {
         style={{
           cursor: 'pointer',
         }}
+        role='button'
+        tabIndex={index}
```
Comment (a3957273) [inline pages/deployment/[uuid].tsx:53]: Tab index should basically never be greater than zero.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

> Warning: Avoid using tabindex values greater than 0. Doing so makes it difficult for people who rely on assistive technology to navigate and operate page content. Instead, write the document with the elements in a logical sequence.

pages/deployment/[uuid].tsx
```diff
@@ -196,20 +202,20 @@ export default function Deployment() {
...
 
               <div># Run Docker image</div>
-              <CodeLine line={`docker run -p 9999:9000 ${deploymentTag}`} />
+              <CodeLine index={3} line={`docker run -p 9999:9000 ${deploymentTag}`} />
               <div># (the container exposes on port 9000, available on the host as port 9999)</div>
               <br />
 
               <div># Check that the Docker container is running</div>
-              <CodeLine line={`docker ps`} />
+              <CodeLine index={4} line='docker ps' />
```
Comment (a3957273) [inline pages/deployment/[uuid].tsx:218]: See previous comments about how `index` should be zero.  Doesn't massively need to be an attribute.

a3957273 (changes requested):

pages/deployment/[uuid].tsx
Comment (a3957273) [reply pages/deployment/[uuid].tsx:140]: I note that at this stage `currentUser` cannot be undefined.  So, instead we should use a `Type Guard` to tell TypeScript that it's not undefined.

pages/deployment/[uuid].tsx
Comment (a3957273) [reply pages/deployment/[uuid].tsx:194]: I think again this only renders if deployment exists, so a Type Guard can be used to ignore the `!` or `?` entirely.

pages/deployment/[uuid].tsx
```diff
@@ -49,10 +49,16 @@ function CodeLine({ line }) {
...
           cursor: 'pointer',
         }}
+        role='button'
+        tabIndex={index}
         onClick={() => {
           navigator.clipboard.writeText(line)
           setOpenSnackbar(true)
         }}
+        onKeyDown={() => {
+          navigator.clipboard.writeText(line)
```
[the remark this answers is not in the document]
Comment (ARADDCC002) [reply pages/deployment/[uuid].tsx:59]: Changed it so keydown only applies when enter key is pressed, moved the repeated lines into a function.

pages/deployment/[uuid].tsx
Comment (ARADDCC002) [reply pages/deployment/[uuid].tsx:53]: Resolved

pages/deployment/[uuid].tsx
Comment (ARADDCC002) [reply pages/deployment/[uuid].tsx:218]: I think linting was complaining about it, but  can make it 0 for now if that still works

PR #82, opened 2022-05-06
added a duplicate version name check for models

ARADDCC002:


ARADDCC002:
test

server/routes/v1/upload.ts
```diff
@@ -107,7 +107,7 @@ export const postUpload = [
           metadata: metadata,
         })
       } catch (err: any) {
-        if (err.toString().includes('duplicate key error')) {
+        if (err.toString().includes('Duplicate version name for model')) {
```
a3957273, line 110: We should be testing the error status code, instead of the text.

server/services/version.ts
```diff
@@ -93,6 +93,14 @@ export async function createVersion(user: UserDoc, data: CreateVersion) {
     throw Forbidden({ data }, 'Unable to create version, failed permissions check.')
   }
 
+  let allVersionsForModel = VersionModel.find({ model: version.model })
```
a3957273, line 96: Should be const.

server/services/version.ts
```diff
@@ -93,6 +93,14 @@ export async function createVersion(user: UserDoc, data: CreateVersion) {
...
+  if (
+    (await allVersionsForModel).filter((versionDuplicateCheck) => versionDuplicateCheck.version === version.version)
```
a3957273, line 98: You shouldn't be doing this logic on the server, you should be getting Mongo to only find models with the specific model _and_ the specified version.

a3957273 (changes requested):

server/services/version.ts
ARADDCC002: done

server/routes/v1/upload.ts
ARADDCC002: Resolved

server/services/version.ts
ARADDCC002: Resolved

server/routes/v1/upload.ts
```diff
@@ -40,7 +40,8 @@ export const postUpload = [
     const session = await mongoose.startSession()
     return session.withTransaction(async () => {
       const files = req.files as unknown as MinioFile
-      const mode = req.query.mode
+      const mode: string = req.query.mode as string
```
a3957273, line 43: I don't think you need `: string` here, as TypeScript should already know it's a string.

server/routes/v1/upload.ts
```diff
@@ -102,12 +103,16 @@ export const postUpload = [
...
+          req.user!,
+          {
+            version: metadata.highLevelDetails.modelCardVersion,
+            metadata: metadata,
+          },
+          modelUuid
+        )
       } catch (err: any) {
-        if (err.toString().includes('duplicate key error')) {
+        if (err.code === 400) {
```
a3957273, line 115: Sorry, clearly didn't mean code.  I was vaguely expecting the code `E11000` we use elsewhere (and is documented https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml).  Which is the unique error code so we only catch the duplicate error here.  It should be somewhere on the response error we get?

Also, if it isn't this error we want to rethrow the error, e.g.:

```
} catch(err: any) {
  if (err.some.code === 'E11000') {
    return res.status(409.json({ ... })
  }

  throw err
}
```

server/services/version.ts
```diff
@@ -86,13 +87,22 @@ interface CreateVersion {
   metadata: any
 }
 
-export async function createVersion(user: UserDoc, data: CreateVersion) {
+export async function createVersion(user: UserDoc, data: CreateVersion, parentUuid: string) {
```
a3957273, line 90: Services shouldn't interact with other domains.  E.g. here, the `Version` service interacts with the `Model` domain.  Instead, you should use the `Model` service (or expect the calling process to pass in a full Model instance).

a3957273 (changes requested):

server/routes/v1/upload.ts
ARADDCC002: Resvoled

a3957273 (approved):

PR #85, opened 2022-05-15
Add some example unit tests for UI / server

a3957273:


server/services/schema.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+  const deploymentSchema: any = {
+    name: 'deployment-schema',
+    reference: 'deployment',
+    use: 'DEPLOYMENT',
+    schema: {}
+  }
+
+  beforeEach(async () => {
+    const uploadDoc = new SchemaModel(uploadSchema)
```
a3957273, line 27: `await SchemaModel.create(uploadSchema)` is the easier way of doing this.

server/services/schema.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  })
+
+  test('we can fetch schema by reference', async () => {
+    const fetchedSchema: any = await findSchemaByRef('upload')
+    expect(fetchedSchema).not.toEqual(null)
+    expect(fetchedSchema.name).toBe(uploadSchema.name)
+  })
+
+  test('we can find upload schemas with a limit of 1', async () => {
+    const fetchedSchema: any = await findSchemasByUse('UPLOAD', 1)
```
a3957273, line 47: No need for any here, code works without it.

server/services/schema.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    expect(fetchedSchema).not.toEqual(null)
+    expect(fetchedSchema.length).toBe(1)
+  })
+
+  test('that we can overwrite an existing schema with some new metadata', async () => {
+    const updatedSchema = uploadSchema
+    updatedSchema.schema.testProperty = 'test field'
```
a3957273, line 54: Hah.  I don't think this works, does it?  It still updates `uploadSchema`.  You'll need to use `lodash` `_.cloneDeep()` or equivalent.

server/services/user.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  })
+
+  test('that the serializer returns the correct properties', () => {
+    const properties: any = serializedUserFields()
+    expect(properties.mandatory).toStrictEqual(['_id', 'id', 'email'])
+  })
+
+  test('that user is cached', async () => {
+    const user: any = await findUserCached(testUser1)
+    expect(user).not.toEqual(null)
```
a3957273, line 48: Never use `.not.toEqual(null)`.  Always use `toBeTruthy()`.  Jest recommends this because JavaScript equality testing is awful.

server/services/user.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    expect(allUsers.length).toBe(2)
+  })
+
+  test('that the serializer returns the correct properties', () => {
+    const properties: any = serializedUserFields()
+    expect(properties.mandatory).toStrictEqual(['_id', 'id', 'email'])
+  })
+
+  test('that user is cached', async () => {
+    const user: any = await findUserCached(testUser1)
```
a3957273, line 47: I would expect a test to make sure the user is cached requests the user once.  Updates the user.  Requests the user again.  Then checks to make sure the change isn't there (aka the user is being cached).  At the moment this just appears to test we can find a user?

server/services/version.spec.ts
```diff
@@ -0,0 +1,87 @@
+import { ObjectId } from 'mongodb'
+import { ApprovalStates } from '../models/Deployment'
+import '../utils/mockMongo'
+import VersionModel from '../models/Version'
+import { createVersion, findModelVersions, findVersionById, findVersionByName, serializedVersionFields } from './version'
+import UserModel from '../models/User'
+
+const modelId = new ObjectId
```
a3957273, line 8: This is bad / archaic syntax.  You should use `new ObjectId()`.

server/services/version.spec.ts
```diff
@@ -0,0 +1,87 @@
+import { ObjectId } from 'mongodb'
+import { ApprovalStates } from '../models/Deployment'
+import '../utils/mockMongo'
```
a3957273, line 3: Mocking should always go first in imports.

server/services/version.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  userId: 'user1',
+  email: 'user1@email.com',
+  data: { some: 'value' },
+}
+const userDoc = new UserModel(testUser)
+
+describe('test version service', () => {
+
+  beforeEach(async () => {
+    const versionDoc = new VersionModel(testVersion)
```
a3957273, line 47: Again, `await VersionModel.create(testVersion)` is shorter.

server/services/version.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    await versionDoc.save()
+    const versionDoc2 = new VersionModel(testVersion2)
+    await versionDoc2.save()
+    testVersion._id = versionDoc._id
+  })
+
+  test('that the serializer returns the correct properties', () => {
+    const properties: any = serializedVersionFields()
+    expect(properties.mandatory).toStrictEqual(['_id', 'version', 'metadata.highLevelDetails.name'])
```
a3957273, line 56: I'm not a massive fan of any of these tests because, to me, I don't think they test anything.  I'd rather a unit test for the serializedVersionFields to try to serialize a `version` and make sure only those fields are available on it.

server/services/schema.spec.ts
ARADDCC002: Done

server/services/version.spec.ts
ARADDCC002: done

server/services/version.spec.ts
ARADDCC002: As said this was a typo and was resolved :)

server/services/schema.spec.ts
ARADDCC002: Oh duh I think you're right! I'll whip something up.

server/services/schema.spec.ts
ARADDCC002: removed

PR #92, opened 2022-06-17
added checkbox warning to the upload existing page

ARADDCC002:


src/Form/FormUpload.tsx
```diff
@@ -19,6 +23,12 @@ export default function FormUpload({
   const dataSteps = splitSchema.steps.filter((step) => step.type === 'Data')
   const [metadata, setMetadata] = useState(JSON.stringify(getStepsData(splitSchema), null, 4))
   const [validationErrorText, setValidationErrorText] = useState<string>('')
+  const { uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
+  const [warningCheckboxVal, setWarningCheckboxVal] = useState<boolean>(false)
```
a3957273, line 27: The `formUpload` component is independent of page, so it'll display when updating / on deployments / any other forms.  Because of this, I feel this should instead be on a tab?  If you create a 'renderBasic' tab with this same data on it, it should render in the same place?

a3957273 (changes requested):

a3957273 (approved):

PR #101, opened 2022-08-05
added tab url param to model and deployment pages

ARADDCC002:


pages/deployment/[uuid].tsx
```diff
@@ -39,7 +39,19 @@ import { postEndpoint } from '../../data/api'
 
 const ComplianceFlow = dynamic(() => import('../../src/ComplianceFlow'))
 
-type TabOptions = 'overview' | 'compliance' | 'build'
+type TabOptions = 'overview' | 'compliance' | 'build' | 'settings'
```
Comment (ARADDCC012) [inline pages/deployment/[uuid].tsx:42]: Instead of having this type and type guard below (`isTabOption`) defined in both files, could we just export from one and use in both?

pages/deployment/[uuid].tsx
```diff
@@ -84,9 +96,9 @@ function CodeLine({ line }) {
 
 export default function Deployment() {
   const router = useRouter()
-  const { uuid }: { uuid?: string } = router.query
+  const { uuid, tab }: { uuid?: string; tab?: TabOptions } = router.query
 
-  const [tab, setTab] = useState<TabOptions>('overview')
+  const [group, setGroup] = useState<TabOptions>('overview')
```
Comment (ARADDCC012) [inline pages/deployment/[uuid].tsx:101]: Personal preference, but instead of having to set state to update it each time we change tab, we could do the following:
```
  const group: TabOptions = useMemo(() => (tab && isTabOption(tab) ? tab : 'overview'), [tab])
```
Same for the other file.

ARADDCC012 (approved):

pages/deployment/[uuid].tsx
Comment (ARADDCC002) [reply pages/deployment/[uuid].tsx:99]: Because this is how we do it across the app I'd rather keep things consistent. However, if we think that this change is worth making across the whole app I'm happy to raise a ticket to get it all changed in the future.

PR #103, opened 2022-08-11
Feature/bai 192 export raw model

ARADDCC002:


src/RawModelExportList.tsx
```diff
@@ -0,0 +1,60 @@
...
+    setDisplayText(!displayText)
+  }
+
+  return (
+    <Stack direction='row'>
+      <Button sx={{ m: 1 }} variant='text' onClick={toggleText} startIcon={<VisibilityIcon />}>
+        {displayText ? 'Hide' : 'Show'} {type} path
+      </Button>
+      <Box component={Stack} direction='column' justifyContent='center'>
+        <Typography variant='body1'>{displayText ? text : 'xxxxxxxx/xxxxxxxx/xxxxxxxx'}</Typography>
```
a3957273, line 26: I think it'd be better to hide the implementational detail of their location all together and just have two buttons for "code" and "binary".  They're not sensitive, they're entirely predictable based on deployment ID.  People don't really need to know where they are, just have the ability to copy them / visit them.

a3957273: Screenshot 2022-08-15 at 00 01 16

What about this type of screen with an input box + copy button?

types/interfaces.ts
```diff
@@ -2,6 +2,7 @@ import { Date, Types } from 'mongoose'
 import Logger from 'bunyan'
 import { UserDoc } from '../server/models/User'
 import { XYPosition } from 'react-flow-renderer'
+import { boolean } from 'yargs'
```
a3957273, line 5: This shouldn't be an import.

a3957273: `boolean` is a basic type that _should_ exist by default.

a3957273 (changes requested):

src/Form/RenderBuildOptionsTab.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <Stack direction='row'>
+            <FormControlLabel
+              control={<Checkbox checked={rawModelExport} />}
+              onChange={handleRawModelExportChange}
+              label='Enable raw model export'
+            />
+            <Divider orientation='vertical' flexItem />
+            <Box component={Stack} direction='column' justifyContent='center' sx={{ pl: 2 }}>
+              <Typography variant='body1'>
+                Model code and binary files can be store/accessed directly from Minio
```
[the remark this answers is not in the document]
ARADDCC002: Good point - I like your wording so I've changed it to that!

src/Form/RenderBuildOptionsTab.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <Stack direction='row'>
+            <FormControlLabel
+              control={<Checkbox checked={allowGuestDeployments} />}
+              onChange={handleGuestDeploymentChange}
+              label='Allow guest deployment'
+            />
+            <Divider orientation='vertical' flexItem />
+            <Box component={Stack} direction='column' justifyContent='center' sx={{ pl: 2 }}>
+              <Typography variant='body1'>
+                Allow for deployments to be requested without the need of creating metadata
```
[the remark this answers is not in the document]
ARADDCC002: Yeah I think there's a few places where we have similar wording for different things. I do like this wording though so happy to change it!

src/RawModelExportList.tsx
ARADDCC002: Yeah so the reason I've hidden them isn't really out of any sort of sensitivity, it was more to make the page look less jarring because of all the text on it.

I can try the above - I'll see how it looks!

types/interfaces.ts
ARADDCC002: This looks like an automatic import from my IDE! Will remove.

server/utils/build.ts
```diff
@@ -158,9 +158,9 @@ export async function buildPython(version: VersionDoc, builderFiles: BuilderFile
   rm('-rf', buildDir)
   rm('-rf', s2iDir)
 
-  await Promise.all([deleteMinioFile(builderFiles.binary), deleteMinioFile(builderFiles.code)])
-  const removeImageCmd = `img rm ${tag}`
-  await logCommand(removeImageCmd, (level: string, message: string) => logger[level](message))
+  //await Promise.all([deleteMinioFile(builderFiles.binary), deleteMinioFile(builderFiles.code)])
+  //const removeImageCmd = `img rm ${tag}`
+  //await logCommand(removeImageCmd, (level: string, message: string) => logger[level](message))
```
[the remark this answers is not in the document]
ARADDCC002: Done

src/Form/RenderBasicBuildOptionsTab.tsx
```diff
@@ -0,0 +1,44 @@
+import { SplitSchema, Step } from '../../types/interfaces'
```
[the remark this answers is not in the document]
ARADDCC002: I like this idea. I've done the same with the file tab too to keep things consistent.

server/routes/v1/upload.ts
```diff
@@ -201,6 +218,32 @@ export const postUpload = [
       res.json({
         uuid: model.uuid,
       })
+
+      if (buildOptions.rawModelExport) {
```
a3957273, line 222: I think we should keep a copy of the raw model / code here regardless of whether `rawModelExport` is enabled.  So, just run this code regardless.

server/routes/v1/upload.ts
```diff
@@ -201,6 +218,32 @@ export const postUpload = [
...
+          await client.copyObject(
+            files.code[0].bucket,
+            codeLocation,
+            `${files.code[0].bucket}/${files.code[0].path}`,
+            new Minio.CopyConditions()
+          )
+
+          version.rawBinaryPath = binaryLocation
+          version.rawCodePath = codeLocation
+          version.save()
```
a3957273, line 242: At the end here you should make sure to clean up the old copies of this data.

src/Form/RenderBuildOptionsTab.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const handleRawModelExportChange = (event: any) => {
+    setStepState(splitSchema, setSplitSchema, step, { ...state, rawModelExport: event.target.checked })
+  }
+
+  const handleGuestDeploymentChange = (event: any) => {
+    setStepState(splitSchema, setSplitSchema, step, { ...state, allowGuestDeployments: event.target.checked })
+  }
+
+  return (
+    <Stack direction='column' spacing={2} sx={{ p: 3 }}>
```
a3957273, line 36: This should be within the jsonschema and use the existing form instead of creating a new one.  Custom tabs should only be used for things that _can't_ be done within the schema, like the submission tab (no data) and the file upload tab.

a3957273 (changes requested):

pages/model/[uuid]/new-version.tsx
```diff
@@ -14,10 +14,11 @@ import { useGetModelVersions } from '../../../data/model'
...
 import SubmissionError from '../../../src/Form/SubmissionError'
 import Form from '../../../src/Form/Form'
-import RenderFileTab, { FileTabComplete } from '../../../src/Form/RenderFileTab'
+import { RenderFileTab, RenderBasicFileTab, FileTabComplete } from '../../../src/Form/RenderFileTab'
 import useCacheVariable from '../../../utils/useCacheVariable'
 import LoadingBar from '../../../src/common/LoadingBar'
 import ModelExportAndSubmission from '../../../src/Form/ModelExportAndSubmission'
+import { RenderBasicBuildOptionsTab, RenderBuildOptionsTab } from '../../../src/Form/RenderBuildOptionsTab'
 
 function renderSubmissionTab(
```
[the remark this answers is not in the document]
Comment (ARADDCC002) [reply pages/model/[uuid]/new-version.tsx:23]: Any reason why? The new version page needs a submission tab too

server/routes/v1/upload.ts
ARADDCC002: Sure - I made this condition because I thought it would save on storage etc if the user doesn't want the files stored. Happy to have them stored anyway, though. 

Maybe not for this ticket but if we're always storing the files the user could perhaps edit the version and allow for uses to export the model code even if the original version upload didn't tick the box.

pages/model/[uuid]/new-version.tsx
```diff
@@ -141,6 +142,9 @@ function Upload() {
 
     data.schemaRef = model?.schemaRef
 
+    form.append('buildOptions', JSON.stringify(data.buildOptions))
```
Comment (a3957273) [inline pages/model/[uuid]/new-version.tsx:145]: This is presumably no longer necessary, now that buildOptions is handled by RJSF.

pages/model/[uuid]/new-version.tsx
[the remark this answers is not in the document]
Comment (a3957273) [reply pages/model/[uuid]/new-version.tsx]: My bad, got confused.

src/RawModelExportList.tsx
```diff
@@ -0,0 +1,78 @@
...
+      </Tooltip>
+      <CopiedSnackbar {...{ openSnackbar, setOpenSnackbar }} />
+    </Stack>
+  )
+}
+
+const RawModelExportList = ({ deployment }: { deployment: Deployment }) => {
+  const { model } = useGetModelById(deployment.model.toString())
+  const { versions, isVersionsLoading, isVersionsError } = useGetModelVersions(model?.uuid)
+  console.log(versions)
```
a3957273, line 49: No reason for log here.

a3957273 (changes requested):

pages/model/[uuid]/new-version.tsx
Comment (ARADDCC002) [reply pages/model/[uuid]/new-version.tsx:145]: oops missed this! deleted

src/RawModelExportList.tsx
ARADDCC002: was left over when debugging! removed

src/RawModelExportList.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <>
+      {!isVersionsLoading &&
```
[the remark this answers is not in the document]
ARADDCC002: Done

server/routes/v1/upload.ts
```diff
@@ -201,6 +202,30 @@ export const postUpload = [
...
+        await client.copyObject(
+          files.code[0].bucket,
+          codeLocation,
+          `${files.code[0].bucket}/${files.code[0].path}`,
+          new Minio.CopyConditions()
+        )
+
+        version.rawBinaryPath = binaryLocation
+        version.rawCodePath = codeLocation
+        version.save()
```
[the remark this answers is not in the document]
ARADDCC002: done

server/routes/v1/deployment.ts
```diff
@@ -161,3 +165,59 @@ export const resetDeploymentApprovals = [
...
+
+export const fetchRawModelFiles = [
+  ensureUserRole('user'),
+  bodyParser.json(),
+  async (req: Request, res: Response) => {
+    const { uuid, version, fileType } = req.params
+    const deployment: any = await findDeploymentByUuid(req.user!, uuid)
+
+    if (!req.user._id.equals(deployment.owner)) {
+      throw Unauthorised(
```
a3957273, line 177: You should include things like the deployment owner in the first argument here.

server/routes/v1/deployment.ts
```diff
@@ -161,3 +165,59 @@ export const resetDeploymentApprovals = [
...
+
+    if (fileType !== 'code' && fileType !== 'binary') {
+      throw NotFound({ fileType }, 'Unknown file type specificed')
+    }
+
+    if (deployment === null) {
+      throw NotFound({ deploymentUuid: uuid }, `Unable to find deployment for uuid ${uuid}`)
+    }
+
+    const versionDocument: any = await findVersionByName(req.user!, deployment.model, version)
```
a3957273, line 191: Don't set `versionDocument` to any, use the type from `findVersionByName`.  It will give a warning about possibly being undefined, which is a valid check you need to do.

server/routes/v1/deployment.ts
```diff
@@ -161,3 +165,59 @@ export const resetDeploymentApprovals = [
     return res.json(deployment)
   },
 ]
+
+export const fetchRawModelFiles = [
+  ensureUserRole('user'),
+  bodyParser.json(),
+  async (req: Request, res: Response) => {
+    const { uuid, version, fileType } = req.params
+    const deployment: any = await findDeploymentByUuid(req.user!, uuid)
```
a3957273, line 174: Don't set `deployment` to any, use the type from `findDeploymentByUuid`.  It will give a warning about possibly being undefined, which is a valid check you need to do.

server/routes/v1/deployment.ts
```diff
@@ -161,3 +165,59 @@ export const resetDeploymentApprovals = [
...
+        {},
+        `User is not authorised to download this file. Requester: ${req.user._id}, owner: ${deployment.owner}`
+      )
+    }
+
+    if (fileType !== 'code' && fileType !== 'binary') {
+      throw NotFound({ fileType }, 'Unknown file type specificed')
+    }
+
+    if (deployment === null) {
```
a3957273, line 187: This check is too late.

server/routes/v1/deployment.ts
```diff
@@ -161,3 +165,59 @@ export const resetDeploymentApprovals = [
...
+      filePath = versionDocument.rawCodePath
+    }
+    if (fileType === 'binary') {
+      filePath = versionDocument.rawBinaryPath
+    }
+
+    // Stat object to get size so browser can determine progress
+    const { size } = await client.statObject(bucketName, filePath)
+
+    res.set('Content-Disposition', contentDisposition(fileType, { type: 'inline' }))
```
a3957273, line 207: This is not the filetype but the file name

server/routes/v1/deployment.ts
```diff
@@ -161,3 +165,59 @@ export const resetDeploymentApprovals = [
...
+    res.set('Content-Type', 'application/zip')
+    res.set('Cache-Control', 'private, max-age=604800, immutable')
+    res.set('Content-Length', size.toString())
+    res.writeHead(200)
+
+    const stream: any = await client.getObject(bucketName, filePath)
```
a3957273, line 213: Stop setting things to any!

a3957273: This one again eliminates a null check that is _REQUIRED_.

server/routes/v1/deployment.ts
```diff
@@ -161,3 +165,59 @@ export const resetDeploymentApprovals = [
...
+    await stream.on('data', async (_obj, error) => {
```
a3957273, line 214: I don't think you can await it like this.

server/routes/v1/upload.ts
```diff
@@ -212,13 +212,16 @@ export const postUpload = [
           `${files.binary[0].bucket}/${files.binary[0].path}`,
           new Minio.CopyConditions()
         )
+        await client.removeObject(files.binary[0].bucket, `${files.binary[0].bucket}/${files.binary[0].path}`)
```
a3957273, line 215: This doesn't seem to delete the files.

src/RawModelExportList.tsx
```diff
@@ -4,73 +4,90 @@ import Divider from '@mui/material/Divider'
...
-      <Tooltip title='Copy' arrow placement='bottom'>
-        <IconButton aria-label='delete' onClick={handleButtonClick}>
-          <CopyIcon />
-        </IconButton>
-      </Tooltip>
-      <CopiedSnackbar {...{ openSnackbar, setOpenSnackbar }} />
-    </Stack>
-  )
-}
+  const downloadFile = async (version: string, fileType: string) => {
```
a3957273, line 27: We don't want to download files like this.  We want to use an `a href` so that the user can either click it to have it download, or right click -> copy URL to get the URL.

server/routes/v1/deployment.ts
a3957273: You shouldn't actually need this `stream.on` at all, the `pipe` is equivalent, mostly.

server/routes/v1/deployment.ts
```diff
@@ -161,3 +165,59 @@ export const resetDeploymentApprovals = [
...
+      if (error) {
+        req.log.error('Error fetching model exports from storage')
+      }
+    })
+    stream.pipe(res)
+
+    return res
```
a3957273, line 221: You don't need to return `res` here.

a3957273 (changes requested):

server/routes/v1/deployment.ts
ARADDCC002: added

server/routes/v1/deployment.ts
ARADDCC002: Done

server/routes/v1/deployment.ts
ARADDCC002: Done

server/routes/v1/deployment.ts
ARADDCC002: Fixed

server/routes/v1/deployment.ts
ARADDCC002: I was using the file type as the name, so when it's download it'll be "code.zip" or "binary.zip". I assumed they'd be better than an objectId but happy to change

server/routes/v1/deployment.ts
ARADDCC002: Done

server/routes/v1/deployment.ts
ARADDCC002: Removed

server/routes/v1/deployment.ts
ARADDCC002: Removed

src/RawModelExportList.tsx
ARADDCC002: Done

server/routes/v1/upload.ts
```diff
@@ -192,15 +193,43 @@ export const postUpload = [
...
       req.log.info({ code: 'created_upload_job', jobId }, 'Successfully created job in upload queue')
 
       // then return reference to user
       res.json({
         uuid: model.uuid,
       })
+
+      try {
+        const binaryLocation = `model/${model._id}/version/${version._id}/raw/binary/${files.binary[0].path}`
+        const client = new Minio.Client(config.get('minio'))
```
a3957273, line 209: There's a `getClient()` function to get a Minio instance.  It would be good turn this into a `copyFile(from, to)` function.

server/routes/v1/upload.ts
```diff
@@ -192,15 +193,43 @@ export const postUpload = [
...
+          files.code[0].bucket,
+          codeLocation,
+          `${files.code[0].bucket}/${files.code[0].path}`,
+          new Minio.CopyConditions()
+        )
+        await client.removeObject(files.code[0].bucket, files.code[0].path)
+
+        version.rawBinaryPath = binaryLocation
+        version.rawCodePath = codeLocation
+        version.save()
```
a3957273, line 229: `.save()` is an async function and needs to be run with `await`.  I wonder if that led to some of the random issues you were seeing before.

src/Form/RenderFileTab.tsx
```diff
@@ -70,3 +71,29 @@ export default function RenderFileTab({ step: step, splitSchema, setSplitSchema
 export function FileTabComplete(step: Step) {
   return step.state.binary && step.state.code
 }
+
+export function RenderBasicFileTab({ step: step, splitSchema, setSplitSchema }: RenderInterface) {
```
a3957273, line 75: The `step: step` is pointless, you're renaming a variable to itself.

a3957273 (changes requested):

server/routes/v1/deployment.ts
```diff
@@ -161,3 +166,59 @@ export const resetDeploymentApprovals = [
...
+
+    //res.set('Content-Disposition', contentDisposition(fileType, { type: 'inline' }))
+    res.set('Content-disposition', `attachment; filename=${fileType}.zip`)
+    res.set('Content-Type', 'application/zip')
+    res.set('Cache-Control', 'private, max-age=604800, immutable')
+    res.set('Content-Length', size.toString())
+    res.writeHead(200)
+
+    const stream: Readable = await client.getObject(bucketName, filePath)
+    if (stream !== null) {
```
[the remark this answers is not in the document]
ARADDCC002: Done

server/routes/v1/deployment.ts
```diff
@@ -161,3 +166,59 @@ export const resetDeploymentApprovals = [
     return res.json(deployment)
   },
 ]
+
+export const fetchRawModelFiles = [
+  ensureUserRole('user'),
+  bodyParser.json(),
+  async (req: Request, res: Response) => {
+    const { uuid, version, fileType } = req.params
+    const deployment: DeploymentDoc | null = await findDeploymentByUuid(req.user!, uuid)
```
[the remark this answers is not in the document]
ARADDCC002: Removed - was added because I thought we were typing those find() calls now

src/Form/RenderFileTab.tsx
ARADDCC002: Whoops! Fixed

server/routes/v1/upload.ts
ARADDCC002: Done

server/models/Version.ts
```diff
@@ -13,6 +13,9 @@ export interface Version {
   managerApproved: ApprovalStates
   reviewerApproved: ApprovalStates
 
+  rawBinaryPath?: string
```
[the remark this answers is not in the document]
ARADDCC002: done

server/routes/v1/upload.ts
```diff
@@ -192,15 +193,43 @@ export const postUpload = [
       ).add({
         versionId: version._id,
         userId: req.user?._id,
-        binary: normalizeMulterFile(files.binary[0]),
-        code: normalizeMulterFile(files.code[0]),
+        binary: normalizeMulterFile(files.binary[0], 'binary', version),
```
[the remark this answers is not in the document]
a3957273: Internally it has the type `FileRef`, so maybe `createFileRef`?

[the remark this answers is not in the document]
ARADDCC002: done

server/routes/v1/deployment.ts
```diff
@@ -161,3 +166,67 @@ export const resetDeploymentApprovals = [
...
+        { approvalStatus: deployment.managerApproved },
+        'User is not authorised to download this file as it has not been approved.'
+      )
+    }
+
+    if (fileType !== 'code' && fileType !== 'binary') {
+      throw NotFound({ fileType }, 'Unknown file type specificed')
+    }
+
+    const versionDocument: VersionDoc | null = await findVersionByName(req.user!, deployment.model, version)
```
a3957273, line 199: Don't add the `VersionDoc | null` typing here

server/routes/v1/deployment.ts
```diff
@@ -161,3 +166,67 @@ export const resetDeploymentApprovals = [
...
+    const { size } = await client.statObject(bucketName, filePath)
+
+    //res.set('Content-Disposition', contentDisposition(fileType, { type: 'inline' }))
+    res.set('Content-disposition', `attachment; filename=${fileType}.zip`)
+    res.set('Content-Type', 'application/zip')
+    res.set('Cache-Control', 'private, max-age=604800, immutable')
+    res.set('Content-Length', size.toString())
+    res.writeHead(200)
+
+    const stream: Readable = await client.getObject(bucketName, filePath)
```
a3957273, line 226: This is an invalid type, it should be `Readable | null`, but again you should be able to omit it entirely:

```javascript
const stream = await client.getObject(...)
```

a3957273 (changes requested):

a3957273 (approved):

PR #104, opened 2022-08-12
Feature/bai 229 finish unit tests for client

jo77419:
Finished unit testing for client.py and separated some of the client methods for better code/easier testing

### DIFF
--- a/lib/python/bailoclient/client.py
+++ b/lib/python/bailoclient/client.py
@@ -169,6 +169,17 @@ class Client:
         logger.warning("User with display name %s not found", name)
         return None
 
+    def __model(self, model: dict) -> Model:
+        """Create Model with schema
+
+        Args:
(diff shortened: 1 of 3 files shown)


lib/python/bailoclient/client.py
```diff
@@ -396,9 +448,17 @@ def __increment_version(self, model_uuid: str):
         """
 
         model_versions = self.api.get(f"model/{model_uuid}/versions")
-        model_versions = [
-            int(model_version["version"]) for model_version in model_versions
-        ]
+        try:
+            model_versions = [
+                int(model_version["version"]) for model_version in model_versions
```
a3957273, line 453: This is a holdover from the old system where we enforced linear versioning from 1, 2, 3, 4, etc.

a3957273 (approved):

PR #108, opened 2022-08-17
App upgraded from CJS to ESM

a3957273:


a3957273:
I think the `config` files can trivially be migrated to ESM by replacing `module.exports` with `export default`.

server/index.ts
```diff
@@ -106,14 +106,12 @@ export async function startServer() {
...
-    return handle(req, res)
-  })
+  server.use((req, res) => handle(req, res))
 
   http.createServer(server).listen(port)
   logger.info({ port }, `Listening on port ${port}`)
 }
 
-if (require.main === module) {
+if (import.meta.url.startsWith('file:') && import.meta.url === pathToFileURL(process.argv[1]).href) {
```
a3957273, line 115: Oh my...

ARADDCC012: Yep it's horrible isn't it. Feel free to suggest a better alternative if you can find one.

From the Node docs: `When the entry point is not a CommonJS module, require.main is undefined, and the main module is out of reach.` - https://nodejs.org/api/modules.html#accessing-the-main-module

ARADDCC012:
> I think the `config` files can trivially be migrated to ESM by replacing `module.exports` with `export default`.

node-config doesn't support ESM - https://github.com/node-config/node-config/wiki/Special-features-for-JavaScript-configuration-files

I have noticed that I've missed changing the production configs to `.cjs` so I'll sort that out.

PR #109, opened 2022-08-18
Feature/bai 231 auth tests

jo77419:
Only unit testing the cognito auth since the others don't really do anything. PKI auth could probably use an e2e test.

### DIFF
--- a/lib/python/tests/test_auth.py
+++ b/lib/python/tests/test_auth.py
@@ -1,82 +1,55 @@
-import os
+from unittest.mock import MagicMock
 
 import pytest
-from dotenv import load_dotenv
 
 from ..bailoclient.auth import CognitoSRPAuthenticator
 from ..bailoclient.config import APIConfig, BailoConfig, CognitoConfig
 from ..bailoclient.utils.exceptions import UnauthorizedException
 
-load_dotenv()
-
 
 @pytest.fixture()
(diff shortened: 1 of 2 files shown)


lib/python/tests/test_auth.py
```diff
@@ -0,0 +1,82 @@
...
+def cognito_authenticator():
+    cognito_config = CognitoConfig(
+        user_pool_id=os.environ["BAILO_USERPOOL"],
+        client_id=os.environ["BAILO_CLIENT_ID"],
+        client_secret=os.environ["BAILO_CLIENT_SECRET"],
+        region=os.environ["BAILO_REGION"],
+    )
+
+    config = BailoConfig(
+        cognito=cognito_config, api=APIConfig(url="http://bailo.com", ca_verify=True)
```
a3957273, line 27: Can you find & replace all instances of `bailo.com`, which might be an address that exists, with either `localhost:8080` or `example.com`.

a3957273 (changes requested):
Are these tests like, 'config validation' tests?  With unit tests I'd expect to ignore the config set in environment variables.  With integration tests I'd expect to target a real Bailo instance.  These are an interesting middle ground.  Not necessarily saying the tests are bad.

a3957273 (approved):

PR #110, opened 2022-08-21
Fix linting issues in utils

a3957273:
Mostly Typescript issues, a few changes of functions -> React elements.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,6 +25,8 @@ module.exports = {
     'prettier',
   ],
   rules: {
+    'import/no-extraneous-dependencies': 'off',
+    'no-restricted-syntax': 'off',
     '@typescript-eslint/no-var-requires': 'off',
     'no-underscore-dangle': 'off',
     'react/jsx-filename-extension': [1, { extensions: ['.ts', '.tsx'] }],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,7 +26,7 @@ module.exports = {
   ],
   rules: {
     // allow dot notation for styles (e.g. .A200)
-    'dot-notation': ['error', { allowPattern: "^[a-zA-Z][0-9]{3}$" }],
+    'dot-notation': ['error', { allowPattern: '^[a-zA-Z][0-9]{3}$' }],
     'import/no-extraneous-dependencies': 'off',
     'no-restricted-syntax': 'off',
     '@typescript-eslint/no-var-requires': 'off',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,6 +25,8 @@ module.exports = {
     'prettier',
   ],
   rules: {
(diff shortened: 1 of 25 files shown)


.eslintrc.js
```diff
@@ -25,6 +25,8 @@ module.exports = {
     'prettier',
   ],
   rules: {
+    'import/no-extraneous-dependencies': 'off',
+    'no-restricted-syntax': 'off',
```
a3957273, line 29: Sometimes `for await` loops are useful.

.eslintrc.js
```diff
@@ -25,6 +25,8 @@ module.exports = {
     'prettier',
   ],
   rules: {
+    'import/no-extraneous-dependencies': 'off',
```
a3957273, line 28: This breaks our test harness linting.

src/Form/ModelEditSubmission.tsx
```diff
@@ -20,7 +20,7 @@ const ModelEditSubmission = ({
   onSubmit: any
   activeStep: number
   setActiveStep: Function
-  modelUploading: boolean
+  modelUploading: boolean | undefined
```
ARADDCC002, line 23: How are we handling an undefined modelUploading value here? Hard to see in this big PR!

utils/complianceFlow.ts
```diff
@@ -4,7 +4,7 @@ import { Deployment, Version } from 'types/interfaces'
 
 function approvalColour(state: string) {
   if (state === 'No Response') return yellow['200']
-  if (state === 'Accepted') return green['A200']
+  if (state === 'Accepted') return green.A200
   if (state === 'Declined') return red['200']
```
ARADDCC002, line 8: Is there a reason we're now using a different format for green and red values here?

utils/complianceFlow.ts
```diff
@@ -13,7 +13,7 @@ function approvalColour(state: string) {
 export default function createComplianceFlow(version: Version) {
   const position: XYPosition = { x: 0, y: 0 }
 
-  const success = green['A200']
+  const success = green.A200
```
ARADDCC002, line 16: Same as above

utils/fetcher.ts
```diff
@@ -20,7 +20,9 @@ export const getErrorMessage = async (res: Response) => {
   let messageError = res.statusText
   try {
     messageError = `${res.statusText}: ${(await res.json()).message}`
-  } catch (e) {}
+  } catch (e) {
+    // unable to identify error message, possibly a network failure
```
ARADDCC002, line 24: Should we not log or return something here instead of just a comment

a3957273: Nah, it will then use the default message `res.statusText` which is fine for handling that error.

ARADDCC002 (approved):

PR #121, opened 2022-08-28
Fix `eslint` issues in `src`.

a3957273:
Now we have 750 less eslint issues!


a3957273:
This adds two rules to be ignored:

- `import/export` complains that a lot of our utility functions have one non-default export.
- `no-restricted-exports` incorrectly refuses the `es6` default export format we use for our `rjsf` themes.

It also adds three or four eslint exceptions for specific instances it messes up.

src/ComplianceFlow.tsx
```diff
@@ -43,9 +43,13 @@ const ComplianceFlow = ({ initialElements }: { initialElements: Elements }) => {
     const layoutedElements = elements.map((el) => {
       if (isNode(el)) {
         const nodeWithPosition = dagreGraph.node(el.id)
+
+        // ESLint confuses `el` for a function.
```
ARADDCC012, line 47: This is just saying that we shouldn't reassign a function param. You can do:
```
    const layoutedElements = elements.map((el) => {
      if (isNode(el)) {
        const updatedEl = { ...el }
        const nodeWithPosition = dagreGraph.node(el.id)

        updatedEl.targetPosition = isHorizontal ? Position.Left : Position.Top
        updatedEl.sourcePosition = isHorizontal ? Position.Right : Position.Bottom
        updatedEl.position = { x: nodeWithPosition.x + Math.random() / 1000, y: nodeWithPosition.y }
        return updatedEl
      }

      return el
    })
```

src/Form/RenderForm.tsx
```diff
@@ -40,6 +40,7 @@ export default function RenderForm({
       omitExtraData
       liveOmit
     >
+      {/* eslint-disable-next-line react/jsx-no-useless-fragment */}
```
ARADDCC012, line 43: Can do:
```
    <SchemaForm
      schema={step.schema}
      formData={step.state}
      onChange={onFormChange}
      widgets={{
        userSelector: UserSelector,
        textArea: TextareaWidget,
        nothing: Nothing,
      }}
      uiSchema={step.uiSchema}
      liveValidate={step.shouldValidate}
      omitExtraData
      liveOmit
    />
```

src/Wrapper.tsx
```diff
@@ -361,8 +352,8 @@ export default function Wrapper({ title, page, children }: { title: any; page: s
         <Box
           component='main'
           sx={{
-            backgroundColor: (theme) =>
-              theme.palette.mode === 'light' ? theme.palette.grey[100] : theme.palette.grey[900],
+            backgroundColor: (curTheme) =>
```
ARADDCC012, line 355: You can just remove the function param and use the `theme` var that exists above

PR #126, opened 2022-08-30
Feature/bai 205 support in app documentation

ARADDCC012:
Some of the changes in this PR have already been reviewed before being merged into this branch. To cut down on some of the noise you can filter by the following file types in Github:
Screenshot 2022-08-30 at 14 38 53

Tests to follow.


pages/help.tsx
```diff
@@ -73,13 +74,11 @@ export default function Help() {
...
                 <CardActions>
-                  <Button
-                    variant='contained'
-                    href={uiConfig.help?.documentationUrl}
-                    sx={{ margin: 'auto', mb: 1.5, width: 200 }}
-                  >
-                    View documentation
-                  </Button>
+                  <Link passHref href='/docs'>
+                    <Button variant='contained' sx={{ margin: 'auto', mb: 1.5, width: 200 }}>
```
ARADDCC002, line 78: Really minor, but "margin" can just be "m"

src/Wrapper.tsx
```diff
@@ -42,7 +42,7 @@ import UserAvatar from './common/UserAvatar'
 import Copyright from './Copyright'
 import { lightTheme } from './theme'
 
-const drawerWidth: number = 240
+const drawerWidth = 240
```
ARADDCC002, line 45: Is there a reason the type's been removed?

server/services/docs.ts
```diff
@@ -0,0 +1,99 @@
...
+  childPriority: number | undefined
+): number | undefined => {
+  if (childPriority && childPriority > 0 && (!headingPriority || childPriority < headingPriority)) {
+    return childPriority
+  }
+  return headingPriority
+}
+
+const extractFrontMatterFromFile = (fileOrDirPath: string): DocsFrontMatter => {
+  const fileContents = readFileSync(fileOrDirPath).toString()
```
a3957273, line 40: Never use `readFileSync`, it will block all other threads from running until this file is read.  This will prevent other requests from being handled.  Instead you should be doing something similar to:

```
import { readFile } from 'fs/promises';

const fileContents = await readFile(fileOrDirPath, { encoding: 'utf8' })
```

server/services/docs.ts
```diff
@@ -0,0 +1,99 @@
...
+          children: headingChildren,
+          ...(highestChildPriority && { priority: highestChildPriority }),
+        })
+      }
+
+      return docsMetadata
+    }, [])
+    .sort(sortByPriority)
+
+  return [docsMenuContent, headingHasIndex, headingPriority]
```
a3957273, line 90: Don't return unnamed tuples, instead return them as named objects using `{}` (e.g. `{ docsMenuContent, headingHasIndex, headingPriority }`).  It has basically the same syntax as using an unnamed tuple, but it will also vehemently complain if you get the wrong names for the tuple entries.

types/interfaces.ts
```diff
@@ -145,3 +141,23 @@ export interface RenderInterface {
   splitSchema: SplitSchema
   setSplitSchema: Dispatch<SetStateAction<SplitSchema>>
 }
+
+export type DocHeading = {
+  id: string
```
a3957273, line 146: This `id` seems to only be used as a React key, why not instead use the `slug` and remove the `id` entirely?

server/services/docs.ts
```diff
@@ -0,0 +1,99 @@
+import { readdirSync, readFileSync, statSync } from 'fs'
+import matter from 'gray-matter'
+import { nanoid } from 'nanoid'
+import { DocFileOrHeading, DocsMenuContent } from '../../types/interfaces'
+
+type DocsFrontMatter = {
+  priority?: number
+}
+
+const sortByPriority = (a: DocFileOrHeading, b: DocFileOrHeading): number => {
```
a3957273, line 10: I think this can be more trivially implemented as `return (a.priority || Infinity) - (b.priority || Infinity)`, unless I'm missing something.

server/services/docs.ts
```diff
@@ -0,0 +1,99 @@
...
+  return 0
+}
+
+const convertFileOrDirNameToTitle = (str: string): string =>
+  str
+    .split('-')
+    .map((word) => `${word[0].toUpperCase()}${word.slice(1)}`)
+    .join(' ')
+
+const removeFileExtension = (fileOrDirName: string): string => {
```
a3957273, line 24: Prefer standard library functions, in this case `Path.parse()` provides the required functionality. https://nodejs.org/api/path.html#pathparsepath

```
const { name } = parse(fileOrDirName)
```

server/services/docs.ts
a3957273: I would usually suggest making `priority` a guaranteed value on Doc with a value of `Infinity`, but if I remember correctly it doesn't have a JSON representation beyond `null` which is pretty annoying.

server/services/docs.ts
```diff
@@ -0,0 +1,99 @@
...
+  let headingPriority: number | undefined
+  const docsDirectoryContents = readdirSync(dirPath)
+
+  const docsMenuContent: DocsMenuContent = docsDirectoryContents
+    .reduce<DocsMenuContent>((docsMetadata, fileOrDirName) => {
+      const nameWithoutExtension = removeFileExtension(fileOrDirName)
+      const title = convertFileOrDirNameToTitle(nameWithoutExtension)
+      const currentSlug = partialSlug ? `${partialSlug}/${nameWithoutExtension}` : nameWithoutExtension
+      const fileOrDirPath = `${dirPath}/${fileOrDirName}`
+      const stats = statSync(fileOrDirPath)
```
a3957273, line 58: Don't use `sync` functions, see above fix using `fs/promises`.

server/services/docs.ts
```diff
@@ -0,0 +1,99 @@
...
+
+      if (stats.isFile() && fileOrDirName.slice(-4) === '.mdx') {
```
a3957273, line 60: Prefer `fileOrDirName.endsWith('.mdx')` over end of slice checking.

server/services/docs.ts
```diff
@@ -0,0 +1,99 @@
...
+        const frontMatter = extractFrontMatterFromFile(fileOrDirPath)
+        headingPriority = updateHeadingPriority(headingPriority, frontMatter.priority)
+        if (fileOrDirName === 'index.mdx') {
+          headingHasIndex = true
+        } else {
+          docsMetadata.push({
+            id: nanoid(),
```
a3957273, line 67: See above comment on whether `id` is necessary.

src/Wrapper.tsx
```diff
@@ -90,45 +90,49 @@ const Drawer = styled(MuiDrawer, { shouldForwardProp: (prop) => prop !== 'open'
...
-  const [open, setOpen] = React.useState(false)
-  const toggleDrawer = () => {
+type WrapperProps = {
+  title: string
+  page: string
+  children?: ReactNode
+}
+
+export default function Wrapper({ title, page, children }: WrapperProps): ReactElement {
+  const isDocsPage = useMemo(() => page.slice(0, 4) === 'docs', [page])
```
a3957273, line 100: Prefer `page.startsWith('docs')`.

src/Wrapper.tsx
```diff
@@ -171,7 +170,7 @@ export default function Wrapper({ title, page, children }: { title: any; page: s
       <Banner />
       <Box sx={{ display: 'flex' }}>
         <CssBaseline />
-        {!isUiConfigLoading && uiConfig!.banner.enable && <Box sx={{ mt: 20 }} />}
+        {!isUiConfigLoading && uiConfig?.banner.enable && <Box sx={{ mt: 20 }} />}
```
a3957273, line 173: Replacing this with `{uiConfig && uiConfig.banner.enable...` prevents TypeScript getting confused and requiring a bunch of `?` everywhere.

src/Wrapper.tsx
```diff
@@ -361,19 +359,24 @@ export default function Wrapper({ title, page, children }: { title: any; page: s
         <Box
           component='main'
           sx={{
-            backgroundColor: (theme) =>
-              theme.palette.mode === 'light' ? theme.palette.grey[100] : theme.palette.grey[900],
+            backgroundColor: () => (theme.palette.mode === 'light' ? theme.palette.grey[100] : theme.palette.grey[900]),
```
a3957273, line 362: Function definition unnecessary, just put the returned value directly in here.

utils/useDocsMenu.ts
```diff
@@ -0,0 +1,32 @@
+import { useCallback, useEffect, useState } from 'react'
+import axios, { AxiosResponse } from 'axios'
+import { DocsMenuContent } from '../types/interfaces'
+
+export type DocsMenuHook = {
+  docsMenuContent: DocsMenuContent
+  errorMessage?: string
+}
+
+export default function useDocsMenu(): DocsMenuHook {
```
a3957273, line 10: Fine to leave it using Axios for now, but in the future if you're doing data fetching on the client we prefer using `swr` in the `/data` directory.

a3957273 (changes requested):

src/Wrapper.tsx
ARADDCC012: We're instantiating the var with a number, so it's already typed as a number. Giving it the type of `number` explicitly does nothing. Our eslint rules state that we shouldn't be doing this, so it's been removed by the auto-fixing.

pages/help.tsx
ARADDCC012: Done

src/Wrapper.tsx
ARADDCC012: Done

src/Wrapper.tsx
ARADDCC012: Done

src/Wrapper.tsx
ARADDCC012: Done

types/interfaces.ts
ARADDCC012: Done

server/services/docs.ts
ARADDCC012: Removed

server/services/docs.ts
ARADDCC012: Done

server/services/docs.ts
ARADDCC012: Done

server/services/docs.ts
ARADDCC012: Done

server/services/docs.ts
ARADDCC012: Done

server/services/docs.ts
ARADDCC012: Done

server/services/docs.ts
ARADDCC012: I've changed priority to be a required property of `DocHeading` and `DocFile` that defaults to `Number.MAX_SAFE_INTEGER` if not in the front matter, so this has been simplified.

ARADDCC012:
> Looks really nice! Only two suggestions for dark mode styling:
> 
> * The "learn more" link buttons are too dark, it should either use the theme so that it changes automatically when swapping modes, or check which mode the app's in and pick colours
> * The anchor links are also too hard to read - will need to change both normal anchors (blue), and previously visited (purple)
> 
> Other than that it looks ready to me (Obviously once the conflicts are resolved!)

I'll spin these out into separate tickets

a3957273 (approved):

PR #128, opened 2022-09-01
Implement stroom logger in Bailo

a3957273:


server/utils/logger.ts
```diff
@@ -166,8 +169,109 @@ if (config.get('logging.file.enabled')) {
...
     path: logPath,
   })
+}
+
+async function processStroomFiles() {
+  const stroomFolder = resolve(appRoot, config.get('logging.stroom.folder'))
+  const processingFolder = resolve(stroomFolder, 'processing')
+
+  const files = (
+    await fsPromise.readdir(stroomFolder, {
```
ARADDCC012, line 179: This logic is repeated below for `processingFolder`, could spin out into a reusable function

server/utils/logger.ts
```diff
@@ -166,8 +169,109 @@ if (config.get('logging.file.enabled')) {
...
+
+  await fsPromise.unlink(file)
+}
+
+if (config.get('logging.stroom.enabled')) {
+  const stroomFolder = resolve(appRoot, config.get('logging.stroom.folder'))
+  const processingFolder = resolve(stroomFolder, 'processing')
+  const processedFolder = resolve(stroomFolder, 'processed')
+
+  let date = new Date()
```
ARADDCC012, line 240: change to const?

server/utils/logger.ts
```diff
@@ -166,8 +169,109 @@ if (config.get('logging.file.enabled')) {
...
-  console.log(logPath, dirname(logPath))
+  if (!fs.existsSync(processedFolder)) {
+    fs.mkdirSync(processedFolder, { recursive: true })
+  }
+
+  const stroomStream = new WritableStream({
+    async write(message) {
+      if (message.code) {
+        const path = resolve(stroomFolder, `logs.${+date}.txt`)
+        await fsPromise.appendFile(path, JSON.stringify(message) + '\n')
```
ARADDCC012, line 258: Our eslint rules will want this to be a template string instead of using `+`
```
await fsPromise.appendFile(path, `${JSON.stringify(message)}\n`)
```

server/utils/logger.ts
ARADDCC012, line 240: Use const?

ARADDCC012 (approved):

PR #129, opened 2022-09-02
Feature/bai 246 add name of approving user to UI

jo77419:
* Unsure whether I should have added the Menu section to the ApprovalsChip or to the [uuid].tsx
* Added in functionality to remove undefined if a managerResponse isn't provided - does this ever actually happen? I added it because there was a test for it but unsure if it's actually needed
* Generally does it look OK and should I add this info anywhere else?


src/common/ApprovalsChip.tsx
```diff
@@ -1,14 +1,31 @@
...
+import DownArrow from '@mui/icons-material/KeyboardArrowDown'
+import UpArrow from '@mui/icons-material/KeyboardArrowUp'
+import Stack from '@mui/material/Stack'
+
 import useTheme from '@mui/styles/useTheme'
-import React from 'react'
-import { Theme } from '../theme'
+import { Theme } from '../../src/theme'
 
 export default function ApprovalsChip({ approvals }: { approvals: any }) {
```
ARADDCC012, line 20: Having props typed as `any` makes things harder than they need to be. Does it make sense to swap out `approvals` for `managerResponse` and `reviewerResponse` as two strongly typed objects? As it stands there's a lot of properties blindly being accessed.

src/common/ApprovalsChip.tsx
```diff
@@ -1,14 +1,31 @@
...
-  const numApprovals = approvals.filter((e: string) => e === 'Accepted').length
-  const totalApprovals = approvals.length
+  Object.keys(approvals).forEach((key) => (approvals[key] === undefined ? delete approvals[key] : {}))
+
+  const approvalOutcomes = [approvals?.managerResponse, approvals?.reviewerResponse].filter(Boolean)
```
ARADDCC012, line 23: I feel like this shouldn't be necessary. Surely we know how many approvers we have, so why are we providing and then having to filter out falsy values?

src/common/ApprovalsChip.tsx
```diff
@@ -1,14 +1,31 @@
...
+  const numApprovals = approvalOutcomes.filter((e: string) => e === 'Accepted').length
```
ARADDCC012, line 24: All of these in-line calculations (`numApprovals`, `totalApprovals`, `open`) will be calculated on every single render. I tend to prefer to memoize them so that they're only recalculated when they need to be (when one of the values in the dependency array changes). E.g:
```
const numApprovals = useMemo(() => {
  approvalOutcomes.filter((e: string) => e === 'Accepted').length
}, [approvalOutcomes])
```

There is a slight overhead to using `useMemo`, but I find it's usually worth playing it safe and using it for anything mildly complex.

src/common/ApprovalsChip.tsx
```diff
@@ -1,14 +1,31 @@
...
+  const totalApprovals = approvalOutcomes.length
+
+  const [anchorEl, setAnchorEl] = useState<HTMLDivElement | null>(null)
+  const open = Boolean(anchorEl)
```
ARADDCC012, line 28: Instead of using the `Boolean` global you can just do 
```
const open = !!anchorEl
```

src/common/ApprovalsChip.tsx
```diff
@@ -25,11 +42,75 @@ export default function ApprovalsChip({ approvals }: { approvals: any }) {
...
+    return [Icon, secondaryText]
+  }
+
+  const [ReviewerIcon, reviewerSecondaryText] = getRequestResponses(approvals?.reviewerResponse)
+  const reviewerPrimaryText = `Technical reviewer (${approvals?.reviewer})`
+
+  const [ManagerIcon, managerSecondaryText] = getRequestResponses(approvals?.managerResponse)
+  const managerPrimaryText = `Model manager (${approvals?.manager})`
+
+  const approvalsClicked = (event: MouseEvent) => {
```
ARADDCC012, line 69: Convention for handling events is to name the handler as `handle<event description>` so in this case `handleApprovalsClicked`

src/common/ApprovalsChip.tsx
```diff
@@ -25,11 +42,75 @@ export default function ApprovalsChip({ approvals }: { approvals: any }) {
...
+    setAnchorEl(null)
+  }
+
   return (
-    <Chip
-      icon={<Icon sx={{ color: 'white !important', pl: 1 }} />}
-      sx={{ borderRadius: 1, color: 'white', height: 'auto', backgroundColor }}
-      label={`Approvals ${numApprovals}/${totalApprovals}`}
-    />
+    <Stack direction='row' spacing={0}>
```
ARADDCC012, line 78: Stack's default spacing is 0 so you can omit that prop

src/common/ApprovalsChip.tsx
```diff
@@ -25,11 +42,75 @@ export default function ApprovalsChip({ approvals }: { approvals: any }) {
...
+          ) : (
+            <DownArrow sx={{ color: 'white !important', pl: 1 }} />
+          )
+        }
+        aria-controls='model-approvals-menu'
+        aria-haspopup='true'
+        aria-expanded={open ? 'true' : undefined}
+      />
+      <Menu id='model-approvals-menu' anchorEl={anchorEl} open={open} onClose={handleClose}>
+        <List dense={true}>
```
ARADDCC012, line 95: For boolean props you don't have to pass `true` or `false`. Instead you should do `<List dense>` or just `<List>` respectively.

src/common/ApprovalsChip.tsx
```diff
@@ -25,11 +42,75 @@ export default function ApprovalsChip({ approvals }: { approvals: any }) {
...
+          {approvals?.reviewerResponse !== undefined && (
+            <ListItem>
+              <ListItemIcon>
+                <ReviewerIcon aria-hidden={true} />
+              </ListItemIcon>
+              <ListItemText primary={reviewerPrimaryText} secondary={reviewerSecondaryText} />
+            </ListItem>
+          )}
+          {approvals?.managerResponse !== undefined && (
```
ARADDCC012, line 104: The `!== undefined` isn't needed as `approvals?.managerResponse` will either be truthy or falsy which is enough in this case. Also `'' !== undefined` returns `true` and so do a lot of other things in JS, so you might get some unintended content being rendered with this check.

src/common/ApprovalsChip.tsx
```diff
@@ -25,11 +42,75 @@ export default function ApprovalsChip({ approvals }: { approvals: any }) {
     backgroundColor = '#4c8a4c'
   }
 
+  const getRequestResponses = (response) => {
+    let Icon
+    let secondaryText
+
+    if (response === 'Accepted') {
```
ARADDCC012, line 49: Instead of using "magic" strings we should use an enum so that we only have to manage the strings in one place. Something like:
```
enum ApprovalStates {
  ACCEPTED = 'Accepted',
  DECLINED = 'Declined',
  'NO_RESPONSE' = 'No Response',
}

if (response === ApprovalStates.ACCEPTED) {
  // ...
}
```

src/common/ApprovalsChip.tsx
```diff
@@ -1,35 +1,107 @@
...
+import Stack from '@mui/material/Stack'
 import useTheme from '@mui/styles/useTheme'
-import React from 'react'
+
+import { ApprovalStates } from '../../types/interfaces'
 import { Theme } from '../theme'
 
-export default function ApprovalsChip({ approvals }: { approvals: any }) {
-  const numApprovals = approvals.filter((e: string) => e === 'Accepted').length
+export default function ApprovalsChip({
```
ARADDCC012, line 20:
```
import React, { useState, useMemo, useCallback, MouseEvent, ReactElement } from 'react'

import Close from '@mui/icons-material/CloseTwoTone'
import Done from '@mui/icons-material/DoneTwoTone'
import Chip from '@mui/material/Chip'
import Menu from '@mui/material/Menu'
import List from '@mui/material/List'
import ListItem from '@mui/material/ListItem'
import ListItemIcon from '@mui/material/ListItemIcon'
import ListItemText from '@mui/material/ListItemText'
import AccessTime from '@mui/icons-material/AccessTime'
import DownArrow from '@mui/icons-material/KeyboardArrowDown'
import UpArrow from '@mui/icons-material/KeyboardArrowUp'
import Stack from '@mui/material/Stack'
import useTheme from '@mui/styles/useTheme'

import { ApprovalStates } from '../../types/interfaces'
import { Theme } from '../theme'

type Approval = {
  reviewer: string
  status: ApprovalStates
}

type ApprovalsChipProps = {
  approvals: Approval[]
}

export default function ApprovalsChip({ approvals }: ApprovalsChipProps): ReactElement {
  const theme = useTheme<Theme>()
  const [anchorEl, setAnchorEl] = useState<HTMLDivElement | null>(null)

  const open = useMemo(() => !!anchorEl, [anchorEl])
  const numApprovals = useMemo(() => approvals.filter((approval) => approval.status === 'Accepted').length, [approvals])
  const totalApprovals = useMemo(() => approvals.length, [approvals])

  const backgroundColor = useMemo(() => {
    if (numApprovals === 0) {
      return theme.palette.error.main
    }
    if (numApprovals < totalApprovals) {
      return '#dc851b'
    }
    return '#4c8a4c'
  }, [numApprovals, totalApprovals, theme])

  const getRequestResponses = useCallback((approval: Approval, index: number) => {
    let Icon
    let secondaryText = ''
    const primaryText = `${approval.reviewer}`

    if (approval.status === ApprovalStates.Accepted) {
      Icon = Done
      secondaryText = 'Approved'
    } else if (approval.status === ApprovalStates.Declined) {
      Icon = Close
      secondaryText = 'Declined'
    } else if (approval.status === ApprovalStates.NoResponse) {
      Icon = AccessTime
      secondaryText = 'Awaiting response'
    }

    return (
      <ListItem key={index}>
        <ListItemIcon>
          <Icon aria-hidden />
        </ListItemIcon>
        <ListItemText primary={primaryText} secondary={secondaryText} />
      </ListItem>
    )
  }, [])

  const requestResponseListItems = useMemo(
    () => approvals.map((approval, index) => getRequestResponses(approval, index)),
    [approvals, getRequestResponses]
  )

  const handleApprovalsClicked = (event: MouseEvent<HTMLDivElement>) => {
    setAnchorEl(event.currentTarget)
  }

  const handleClose = () => {
    setAnchorEl(null)
  }

  return (
    <Stack direction='row'>
      <Chip
        sx={{ borderRadius: 1, color: 'white', height: 'auto', backgroundColor }}
        label={`Approvals ${numApprovals}/${totalApprovals}`}
        onClick={handleApprovalsClicked}
        icon={
          open ? (
            <UpArrow sx={{ color: 'white !important', pl: 1 }} />
          ) : (
            <DownArrow sx={{ color: 'white !important', pl: 1 }} />
          )
        }
        aria-controls='model-approvals-menu'
        aria-haspopup='true'
        aria-expanded={open ? 'true' : undefined}
      />
      <Menu id='model-approvals-menu' anchorEl={anchorEl} open={open} onClose={handleClose}>
        <List dense>{requestResponseListItems}</List>
      </Menu>
    </Stack>
  )
}
```

server/services/request.ts
```diff
@@ -1,6 +1,6 @@
 import { Types } from 'mongoose'
-import { Request } from '../../types/interfaces'
-import { ApprovalStates, DeploymentDoc } from '../models/Deployment'
+import { ApprovalStates, Request } from '../../types/interfaces'
+import { DeploymentDoc } from '../../models/Deployment'
```
ARADDCC002, line 3: Is this path correct? Other paths to the model directory look like '../models/'

src/common/ApprovalsChip.spec.tsx
```diff
@@ -35,36 +36,12 @@ describe('ApprovalsChip', () => {
...
-        <ApprovalsChip approvals={['Accepted', 'Accepted']} />
+        <ApprovalsChip approvals={[{ reviewer: 'Alice', status: ApprovalStates.Accepted},  { reviewer: 'Bob', status: ApprovalStates.Accepted }]} />
       </ThemeProvider>
     )
 
     await waitFor(async () => {
       expect(await screen.findByText('Approvals 2/2')).not.toBeUndefined()
     })
   })
-
```
ARADDCC002, line 46: Can we re-add the deleted tests with the new props? Because the component has different states it would be nice to make sure other cases are displaying correctly.

ARADDCC002 (approved):

PR #130, opened 2022-09-05
deletes old requests whenever a version is edited

ARADDCC002:
For some reason RequestModel.deleteMany() isn't working, so I've had to do it a different way which is a bit longer.


a3957273 (changes requested):
Can we only delete requests that are going to be updated?  Specifically:

- Delete requests that have a response that is not 'No Response'
- Delete requests where the requesting user has changed.

The only issue I have with this method is, during my tests, it spams emails quite a lot for people who have not answered the previous review request either.

a3957273 (approved):

PR #131, opened 2022-09-05
updated mdx-mermaid

ARADDCC002:


a3957273 (changes requested):
Screenshot 2022-09-06 at 02 00 12

This adds broken code to the top of each of our docs.

ARADDCC002:
> Screenshot 2022-09-06 at 02 00 12
> 
> This adds broken code to the top of each of our docs.

That's weird I didn't notice that when I tested it? Let me take another look.

ARADDCC002:
Latest version has a bug that displays the module import at the top of each document page

PR #132, opened 2022-09-05
increased max size of upload existing text field so it is more readab…

ARADDCC002:
…le with more content

### DIFF
--- a/src/Form/FormUpload.tsx
+++ b/src/Form/FormUpload.tsx
@@ -67,7 +67,7 @@ export default function FormUpload({
       <TextField
         fullWidth
         multiline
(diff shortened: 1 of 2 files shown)


a3957273 (changes requested):
Can we keep `minRows={4}`?  It looks weird when you remove all the data from the upload existing field!

ARADDCC002:
> Can we keep `minRows={4}`? It looks weird when you remove all the data from the upload existing field!

Good point - I left it out because I thought it will always have some data, even with an empty form. Easy enough to add back in.

a3957273 (approved):

PR #139, opened 2022-09-13
added an option for different styles of model uploads. if user upload…

ARADDCC002:
…s just a model card this will be displayed to users on the corresponding model/deployment pages

I've had to add ammend the linting config so that we can export the new ModelUploadType enum from the interfaces file.


pages/deployment/[uuid].tsx
```diff
@@ -162,6 +165,10 @@ export default function Deployment() {
     await postEndpoint(`/api/v1/deployment/${deployment?.uuid}/reset-approvals`, {}).then((res) => res.json())
   }
 
+  const initialVersionRequested: ObjectId | VersionDoc = deployment.versions.filter(
+    (version: any) => version.version === deployment.metadata.highLevelDetails.initialVersionRequested
```
Comment (ARADDCC002) [inline pages/deployment/[uuid].tsx:169]: I'm not happy using "any" here, but it seems TS is complaining if I try anything else. Might revisit this when I get time

Comment (ARADDCC012) [reply pages/deployment/[uuid].tsx:170]: TS is complaining because `deployment.versions` is typed as `(Types.ObjectId | VersionDoc)[]`, so  you'd need to use a type guard in your filter to ensure you're working with `VersionDoc`s. 
```
// type guard
const isVersionDoc = (value: unknown): value is VersionDoc => value && value._id && value.version

// using type guard
// I've made the assumption that we only care about `VersionDoc`s as the UI should have the full object 
const initialVersionRequested = deployment.versions.filter(
  version => isVersionDoc(version) && version.version === deployment.metadata.highLevelDetails.initialVersionRequested
)[0]
```

.eslintrc.js
```diff
@@ -33,7 +33,7 @@ module.exports = {
     'import/no-cycle': [0, { ignoreExternal: true }],
     '@typescript-eslint/no-use-before-define': ['error', { functions: false, classes: false, variables: true }],
     '@typescript-eslint/no-namespace': ['error', { allowDeclarations: true }],
-    '@typescript-eslint/no-shadow': 'error',
+    '@typescript-eslint/no-shadow': ['error'],
```
ARADDCC012, line 36: You should only need to provide an array if you're setting custom options. I imagine it'll work like this too, but 'error' is clear that we are using the default options

pages/deployment/[uuid].tsx
Comment (ARADDCC012) [reply pages/deployment/[uuid].tsx:170]: To add to the above, for completeness and ease of handling the optional `initialVersionRequested`, you could return a simple error page if `initialVersionRequested` is `undefined`. That way, for the rest of the JSX, you can work with it being a guaranteed value.

pages/deployment/[uuid].tsx
```diff
@@ -162,14 +166,40 @@ export default function Deployment() {
...
+
   return (
     <>
       <Wrapper title={`Deployment: ${deployment.metadata.highLevelDetails.name}`} page='deployment'>
-        <Box sx={{ textAlign: 'right', pb: 3 }}>
-          <Button variant='outlined' color='primary' startIcon={<Info />} onClick={handleClickOpen}>
-            Show download commands
-          </Button>
-        </Box>
+        {initialVersionRequested.metadata.buildOptions.uploadType === undefined ||
```
Comment (ARADDCC012) [inline pages/deployment/[uuid].tsx:176]: Instead of doing this long check multiple times, you could use the following:
```
const hasUploadType = useMemo(() => !!initialVersionRequested.metadata.buildOptions.uploadType, [initialVersionRequested])
```
Then in place of `initialVersionRequested.metadata.buildOptions.uploadType === undefined` you can do `!hasUploadType`

.eslintrc.js
ARADDCC002: I believe this is a due to a merge conflict, but I can change it back

pages/model/[uuid].tsx
```diff
@@ -170,6 +170,14 @@ function Model() {
 
   return (
     <Wrapper title={`Model: ${version.metadata.highLevelDetails.name}`} page='model'>
+      {version.metadata.buildOptions.uploadType !== undefined &&
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:173]: See https://github.com/gchq/Bailo/pull/139/files#r975295397 - applies here too

src/Form/RenderFileTab.tsx
```diff
@@ -1,11 +1,16 @@
 import Box from '@mui/material/Box'
 import Button from '@mui/material/Button'
 import Divider from '@mui/material/Divider'
+import FormControl from '@mui/material/FormControl'
```
ARADDCC012, line 4: A bunch of these imports are unused. If you set up the vscode eslint plugin it'll help with catching these as you dev

src/Form/RenderFileTab.tsx
```diff
@@ -20,6 +25,10 @@ export default function RenderFileTab({ step, splitSchema, setSplitSchema }: Ren
     display: 'none',
   })
 
+  const buildOptionsStep: Step = splitSchema.steps.filter(
```
ARADDCC012, line 28: This could be replaced with a `find`. Also it's worth memoing this as we shouldn't be performing array iterations unless it's needed:
```
  const buildOptionsStep = useMemo(
    () => splitSchema.steps.find((buildOptionSchemaStep) => buildOptionSchemaStep.section === 'buildOptions'),
    [splitSchema.steps]
  )
```
Again, this will now be typed as `Steps | undefined`, which is the same as doing `splitSchema.steps.filter(...)[0]` as the filter could return an empty array. You'll need to add in conditional logic to handle the optional `buildOptionsStep`, but as with before you can just conditionally return an informative message/error instead of the entire component.

src/Form/RenderFileTab.tsx
```diff
@@ -37,43 +46,61 @@ export default function RenderFileTab({ step, splitSchema, setSplitSchema }: Ren
...
+      {buildOptionsStep.state.uploadType === ModelUploadType.ModelCard && (
+        <Typography sx={{ p: 2 }}>Uploading a model card without any code or binary files</Typography>
+      )}
     </Grid>
   )
 }
 
 export function FileTabComplete(step: Step) {
-  return step.state.binary && step.state.code
+  const buildOptionsStep: Step = step.state.steps.filter(
```
ARADDCC012, line 84: Can be replaced with a `find`:
```
  const buildOptionsStep: Step = step.state.steps.find(
    (buildOptionSchemaStep) => buildOptionSchemaStep.section === 'buildOptions'
  )
```

src/Form/RenderFileTab.tsx
```diff
@@ -37,43 +46,61 @@ export default function RenderFileTab({ step, splitSchema, setSplitSchema }: Ren
...
+    (buildOptionsStep.state.uploadType === ModelUploadType.Zip && step.state.binary && step.state.code) ||
+    buildOptionsStep.state.uploadType === ModelUploadType.ModelCard
+  )
 }
 
 export function RenderBasicFileTab({ step, splitSchema, setSplitSchema }: RenderInterface) {
   const { state } = step
   const { binary, code } = state
 
+  const buildOptionsStep: Step = splitSchema.steps.filter(
```
ARADDCC012, line 100: Can be replaced with a `find`:
```
  const buildOptionsStep = splitSchema.steps.find(
    (buildOptionSchemaStep) => buildOptionSchemaStep.section === 'buildOptions'
  )
```

.eslintrc.js
ARADDCC012: Bit cheeky but it looks like `'no-shadow': 'off',` has been duplicated at some point, would you mind cleaning that up too? Probs also due to a merge conflict.

src/Form/RenderFileTab.tsx
```diff
@@ -84,14 +111,27 @@ export function RenderBasicFileTab({ step, splitSchema, setSplitSchema }: Render
 
   return (
     <Box sx={{ pb: 4, pt: 4 }}>
-      <Stack direction='row' spacing={2} alignItems='center'>
-        <FileInput label='Select Code' file={code} onChange={handleCodeChange} accepts='.zip' />
-        <FileInput label='Select Binary' file={binary} onChange={handleBinaryChange} accepts='.zip' />
-      </Stack>
+      {(buildOptionsStep.state.uploadType === undefined ||
```
ARADDCC012, line 114: See https://github.com/gchq/Bailo/pull/139#discussion_r975295397 - same concept could be applied here

src/Form/RenderFileTab.tsx
```diff
@@ -84,14 +111,27 @@ export function RenderBasicFileTab({ step, splitSchema, setSplitSchema }: Render
...
+        buildOptionsStep.state.uploadType === ModelUploadType.ModelCard && (
+          <Typography sx={{ pt: 2, pb: 2 }}>Uploading a model card without any code or binary files</Typography>
+        )}
     </Box>
   )
 }
 
 export function BasicFileTabComplete(step: Step) {
-  return step.state.binary && step.state.code
+  const buildOptionsStep: Step = step.state.steps.filter(
```
ARADDCC012, line 130: Can use `find` and `useMemo`:
```
  const buildOptionsStep = useMemo(
    () => step.state.steps.find((buildOptionSchemaStep) => buildOptionSchemaStep.section === 'buildOptions'),
    [step]
  )
```

src/Form/RenderFileTab.tsx
```diff
@@ -37,43 +43,66 @@ export default function RenderFileTab({ step, splitSchema, setSplitSchema }: Ren
...
+      {buildOptionsStep !== undefined && buildOptionsStep.state.uploadType === ModelUploadType.ModelCard && (
+        <Typography sx={{ p: 2 }}>Uploading a model card without any code or binary files</Typography>
+      )}
     </Grid>
   )
 }
 
 export function FileTabComplete(step: Step) {
-  return step.state.binary && step.state.code
+  const buildOptionsStep = step.state.steps.find(
```
ARADDCC012, line 81: Should be able to wrap this with `useMemo` so that we only search through the array when it's changed

src/Form/RenderFileTab.tsx
```diff
@@ -37,43 +43,66 @@ export default function RenderFileTab({ step, splitSchema, setSplitSchema }: Ren
...
+        </Stack>
+      )}
+      {buildOptionsStep !== undefined && buildOptionsStep.state.uploadType === ModelUploadType.ModelCard && (
+        <Typography sx={{ p: 2 }}>Uploading a model card without any code or binary files</Typography>
+      )}
     </Grid>
   )
 }
 
 export function FileTabComplete(step: Step) {
```
ARADDCC012, line 80: I keep thinking this is returning a component, consider using camel case instead -> `fileTabComplete`

src/Form/RenderFileTab.tsx
```diff
@@ -37,43 +43,66 @@ export default function RenderFileTab({ step, splitSchema, setSplitSchema }: Ren
...
-  return step.state.binary && step.state.code
+  const buildOptionsStep = step.state.steps.find(
+    (buildOptionSchemaStep) => buildOptionSchemaStep.section === 'buildOptions'
+  )
+  const hasUploadType = useMemo(() => !!buildOptionsStep.state.uploadType, [buildOptionsStep])
```
ARADDCC012, line 84: Sorry if I've suggested to add the `useMemo` here, but it feels strange to be using hooks inside something that isn't a react component or hook itself. Also, this is already being used inside of a `useEffect` so I reckon we should keep this simple and avoid using hooks.

src/Form/RenderFileTab.tsx
```diff
@@ -82,16 +111,34 @@ export function RenderBasicFileTab({ step, splitSchema, setSplitSchema }: Render
...
+        </Stack>
+      )}
+      {hasUploadType && buildOptionsStep.state.uploadType === ModelUploadType.ModelCard && (
+        <Typography sx={{ pt: 2, pb: 2 }}>Uploading a model card without any code or binary files</Typography>
+      )}
     </Box>
   )
 }
 
 export function BasicFileTabComplete(step: Step) {
```
ARADDCC012, line 132: This returns a boolean, not a component. Consider renaming to `basicFileTabComplete`

src/Form/RenderFileTab.tsx
```diff
@@ -82,16 +111,34 @@ export function RenderBasicFileTab({ step, splitSchema, setSplitSchema }: Render
...
-  return step.state.binary && step.state.code
+  const buildOptionsStep = useMemo(
```
ARADDCC012, line 133: See https://github.com/gchq/Bailo/pull/139/files?diff=unified&w=1#r976362869

ARADDCC012 (approved):

PR #140, opened 2022-09-13
updated post model to check that upload mode is valid 

jo77419:


server/routes/v1/upload.ts
```diff
@@ -173,6 +180,7 @@ export const postUpload = [
         })
       }
 
+      console.log('should not be here')
```
ARADDCC002, line 183: Remove log

server/routes/v1/upload.ts
```diff
@@ -68,6 +68,13 @@ export const postUpload = [
         })
       }
 
+      if (mode !== 'newVersion' && mode !== 'newModel') {
+        req.log.warn({ code: 'upload_mode_invalid', mode: { mode } }, `'${mode}' is not a valid upload mode.`)
+        return res.status(400).json({
```
ARADDCC002, line 73: This can be replaced with the "BadReq" method we use

server/routes/v1/upload.ts
```diff
@@ -68,6 +68,13 @@ export const postUpload = [
         })
       }
 
+      if (mode !== 'newVersion' && mode !== 'newModel') {
```
ARADDCC002, line 71: I appreciate I made this modes originally, but I wonder if we can make them into a type as we reuse them in various places

server/routes/v1/upload.ts
```diff
@@ -40,7 +41,7 @@ export const postUpload = [
     const session = await mongoose.startSession()
     return session.withTransaction(async () => {
       const files = req.files as unknown as MinioFile
-      const mode = req.query.mode as string
+      const mode = (req.query.mode as string) || 'newModel'
```
ARADDCC002, line 44: const mode = (req.query.mode as string) || UploadModes.NewModel

ARADDCC002 (approved):

ARADDCC002 (approved):

PR #141, opened 2022-09-14
Feature/bai 228 delete model

ARADDCC011:
Now has a model delete button, which softs deletes the model
Works, and passes npm tests


pages/model/[uuid].tsx
```diff
@@ -168,6 +179,25 @@ function Model() {
...
+    setDeleteConfirmOpen(true)
+  }
+
+  const handleDeleteCancel = () => {
+    setDeleteConfirmOpen(false)
+  }
+
+  const handleDeleteConfirm = async () => {
+    await deleteEndpoint(`/api/v1/model/${uuid}`, {}).then((res) => {
+      if (res.status < 400) {
```
Comment (ARADDCC002) [inline pages/model/[uuid].tsx:193]: Should be 200

server/models/Deployment.ts
```diff
@@ -1,11 +1,14 @@
...
+import { Schema, model, Document, Types } from 'mongoose'
+import MongooseDelete from 'mongoose-delete'
 import logger from '../utils/logger'
 import { ModelDoc } from './Model'
 import { UserDoc } from './User'
 import { VersionDoc } from './Version'
 
 export const approvalStateOptions = ['Accepted', 'Declined', 'No Response']
 
+// eslint-disable-next-line no-shadow
```
ARADDCC002, line 11: This is now being handled by the eslint config and can be removed

server/models/Deployment.ts
```diff
@@ -1,11 +1,14 @@
-import { Document, model, Schema, Types } from 'mongoose'
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
ARADDCC002, line 1: Not a fan of this - we should try and work out what type we're using

server/models/Deployment.ts
```diff
@@ -41,7 +44,7 @@ export interface Deployment {
 
 export type DeploymentDoc = Deployment & Document<any, any, Deployment>
 
-const DeploymentSchema = new Schema<Deployment>(
+const DeploymentSchema: any = new Schema<Deployment>(
```
ARADDCC002, line 47: try const DeploymentSchema: Schema<Deployment> = new Schema<Deployment>

server/models/Request.ts
```diff
@@ -1,17 +1,21 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { Document, model, Schema, Types } from 'mongoose'
+import MongooseDelete from 'mongoose-delete'
 import { approvalStateOptions, ApprovalStates, DeploymentDoc } from './Deployment'
 import { UserDoc } from './User'
 import { VersionDoc } from './Version'
 
 export const approvalTypeOptions = ['Manager', 'Reviewer']
 
+// eslint-disable-next-line no-shadow
```
ARADDCC002, line 10: remove

server/models/Request.ts
```diff
@@ -1,17 +1,21 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
ARADDCC002, line 1: Remove as long as we fix typing

server/models/Request.ts
```diff
@@ -33,7 +37,7 @@ export interface Request {
 
 export type RequestDoc = Request & Document<any, any, Request>
 
-const RequestSchema = new Schema<Request>(
+const RequestSchema: any = new Schema<Request>(
```
ARADDCC002, line 40: try const RequestSchema: Schema<Request> = new Schema<Request>

server/models/Version.ts
```diff
@@ -34,7 +35,7 @@ export interface Version {
 
 export type VersionDoc = Version & Document<any, any, Version>
 
-const VersionSchema = new Schema<Version>(
+const VersionSchema: any = new Schema<Version>(
```
ARADDCC002, line 38: try VersionSchema: Schema<Version> = new Schema<Version>

server/routes/v1/model.spec.ts
```diff
@@ -73,6 +73,14 @@ describe('test model routes', () => {
     expect(res.body.version).toBe('1')
   })
 
+  test('that we can delete a specific version for a model', async () => {
+    const resDelete = await authenticatedDeleteRequest(`/api/v1/model/${modelUuid}`)
+    validateTestRequest(resDelete)
```
ARADDCC002, line 78: The delete API should return the model UUID, so I would also check for that in the response

server/routes/v1/model.ts
```diff
@@ -1,4 +1,6 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
ARADDCC002, line 1: We should fix what's causing this first, and if we really can't resolve then use a "next-line" one instead of ignoring the rule for the full file

src/common/ConfirmationDialogue.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+      expect(await screen.findByText(testTitle)).not.toBeUndefined()
+      expect(await screen.findByText(testText)).not.toBeUndefined()
+    })
+  })
+
+  it('does not render a ConfirmationDialogue component', async () => {
+    render(
+      <ConfirmationDialogue
+        showConfirmationDialogue={false}
+        onCancel={handleCancel}
```
ARADDCC002, line 41: We should be able to replace these function calls with just:

onCancel={() => {}}

pages/model/[uuid].tsx
Comment (ARADDCC002) [reply pages/model/[uuid].tsx:193]: As discussed if @a3957273 is happy with it being < 400 we can leave it - although it might have issues in the future if we change the API call to have forms of 2XX responses!

Comment (ARADDCC011) [reply pages/model/[uuid].tsx:193]: Leaving as <400, may possibly change in future pending discussion

server/models/Deployment.ts
ARADDCC011: Removed

server/models/Deployment.ts
ARADDCC011: Leaving as any as Schema causes errors

> Type 'Schema<Deployment, Model<Deployment, any, any>, undefined, {}>' is not assignable to type 'Schema<Document<any, any, any>, Model<Document<any, any, any>, any, any>, undefined, {}>'.
  Types of property 'methods' are incompatible.
    Type '{ [name: string]: (this: Deployment & Document<any, any, Deployment>, ...args: any[]) => any; }' is not assignable to type '{ [name: string]: (this: Document<any, any, any>, ...args: any[]) => any; }'.
      'string' index signatures are incompatible.
        Type '(this: Deployment & Document<any, any, Deployment>, ...args: any[]) => any' is not assignable to type '(this: Document<any, any, any>, ...args: any[]) => any'.
          The 'this' types of each signature are incompatible.
            Type 'Document<any, any, any>' is not assignable to type 'Deployment & Document<any, any, Deployment>'.
              Type 'Document<any, any, any>' is missing the following properties from type 'Deployment': schemaRef, uuid, versions, metadata, and 7 more.ts(2322)

ARADDCC012: I'd argue that typing as `any` is a backwards step. That being said, we do this already for a bunch of other schemas, so maybe worth writing up a ticket for and fixing them all at once.

package.json
```diff
@@ -31,6 +31,7 @@
     "@next/bundle-analyzer": "^12.0.8",
     "@next/mdx": "^12.2.3",
     "@rjsf/core": "^4.1.1",
+    "@types/mongoose-delete": "^0.5.9",
```
ARADDCC012, line 34: This should be move into `devDependencies`

data/api.ts
```diff
@@ -15,3 +15,7 @@ export async function postEndpoint(url: string, data: any) {
 export async function putEndpoint(url: string, data: any) {
   return fetchEndpoint(url, 'PUT', data)
 }
+
+export async function deleteEndpoint(url: string, data: any) {
+  return fetchEndpoint(url, 'DELETE', data)
```
ARADDCC012, line 20: We shouldn't be including a `body` in a DELETE request. You'll need to modify `fetchEndpoint` to make `data` optional and only include the `body` in the request if `data` is provided

pages/model/[uuid].tsx
```diff
@@ -168,6 +179,25 @@ function Model() {
...
+    setDeleteConfirmContent(<Box />)
+    setDeleteConfirmOpen(true)
+  }
+
+  const handleDeleteCancel = () => {
+    setDeleteConfirmOpen(false)
+  }
+
+  const handleDeleteConfirm = async () => {
+    await deleteEndpoint(`/api/v1/model/${uuid}`, {}).then((res) => {
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:192]: We shouldn't be using promise chains if we can avoid it. Instead we can use async/await like so:
```
  const handleDeleteConfirm = async () => {
    const response = await deleteEndpoint(`/api/v1/model/${uuid}`, {})

    if (response.status < 400) {
      router.push('/')
    } else {
      setDeleteConfirmContent(<ConfirmationFeedbackText status={response.status} statusText={response.statusText} />)
    }
  }
```

pages/model/[uuid].tsx
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:193]: Should we not be checking that `res.status >= 200 && res.status < 300`? Also, as this is the sort of logic that we'll want to reuse, is it worth creating a reusable util function for this?

pages/model/[uuid].tsx
```diff
@@ -81,6 +90,8 @@ function Model() {
   const { versions, isVersionsLoading, isVersionsError } = useGetModelVersions(uuid)
   const { version, isVersionLoading, isVersionError, mutateVersion } = useGetModelVersion(uuid, selectedVersion)
   const { deployments, isDeploymentsLoading, isDeploymentsError } = useGetModelDeployments(uuid)
+  const [deleteConfirmOpen, setDeleteConfirmOpen] = useState<boolean>(false)
+  const [deleteConfirmContent, setDeleteConfirmContent] = useState<JSX.Element>(<Box />)
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:94]: Instead of defaulting to an empty `Box` you should use `null`. The difference is that a `Box` will be rendered on the page as an empty `div` which, however minor, is work that the browser has to do. We should always aim to minimise the work that the browser has to do as that is a limited resource. `null` on the other hand will simply not be rendered at all.

src/common/ConfirmationDialogue.tsx
```diff
@@ -0,0 +1,40 @@
...
+  onCancel: () => void
+  onConfirm: () => void
+}) {
+  return (
+    <Dialog open={showConfirmationDialogue} onClose={onCancel}>
+      <DialogTitle id='alert-dialog-title'>{confirmationModalTitle}</DialogTitle>
+      <DialogContent>
+        <Typography>Are you sure you want to perform this action?</Typography>
+      </DialogContent>
+      {confirmationContent}
```
ARADDCC012, line 29: Should this not be inside `DialogContent`? You can wrap it in a `Box` to allow you to put some padding/margin between it and the other content

src/common/ConfirmationDialogue.tsx
```diff
@@ -0,0 +1,40 @@
...
+  showConfirmationDialogue: boolean
+  confirmationModalTitle: string
+  confirmationContent: JSX.Element
+
+  onCancel: () => void
+  onConfirm: () => void
+}) {
+  return (
+    <Dialog open={showConfirmationDialogue} onClose={onCancel}>
+      <DialogTitle id='alert-dialog-title'>{confirmationModalTitle}</DialogTitle>
```
ARADDCC012, line 25: Is this `id` needed?

src/common/ConfirmationDialogue.tsx
```diff
@@ -0,0 +1,40 @@
...
+  showConfirmationDialogue,
+  confirmationModalTitle,
+  confirmationContent,
+  onCancel,
+  onConfirm,
+}: {
+  showConfirmationDialogue: boolean
+  confirmationModalTitle: string
+  confirmationContent: JSX.Element
+
```
ARADDCC012, line 19: Extra space in your type definition

src/common/ConfirmationDialogue.tsx
```diff
@@ -0,0 +1,40 @@
...
+export default function ConfirmationDialogue({
+  showConfirmationDialogue,
+  confirmationModalTitle,
+  confirmationContent,
+  onCancel,
+  onConfirm,
+}: {
+  showConfirmationDialogue: boolean
+  confirmationModalTitle: string
+  confirmationContent: JSX.Element
```
ARADDCC012, line 18: As this isn't actually the entire content of the dialog, I'd argue that `errorMessage` or something similar is a clearer name

pages/model/[uuid].tsx
```diff
@@ -404,10 +434,18 @@ function Model() {
 
             <Box sx={{ mb: 4 }} />
 
+            <ConfirmationDialogue
+              showConfirmationDialogue={deleteConfirmOpen}
+              onCancel={handleDeleteCancel}
+              onConfirm={handleDeleteConfirm}
+              confirmationModalTitle='Delete the model?'
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:441]: I think a better title would be simply `Delete Model`. The dialog content is already being used to ask the user a question, we don't need to ask them twice :)

server/routes/v1/model.ts
ARADDCC012: We really shouldn't be ignoring our eslint rules at all, unless absolutely necessary. The whole point of having them is to set a standard that we want our code to reach.

pages/model/[uuid].tsx
```diff
@@ -62,6 +63,14 @@ const Alert = React.forwardRef<HTMLDivElement, AlertProps>((props, ref) => (
   <MuiAlert elevation={6} ref={ref} variant='filled' {...props} />
 ))
 
+function ConfirmationFeedbackText({ status, statusText }: { status: number; statusText: string }) {
+  return (
+    <Alert severity='error' sx={{ m: 2, width: 400 }}>
```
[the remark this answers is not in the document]
Comment (ARADDCC012) [reply pages/model/[uuid].tsx:68]: You also should steer clear of using fixed widths/heights where possible as this won't work on every screen size

src/common/ConfirmationDialogue.tsx
ARADDCC012: If this is changed to be a string as recommended in https://github.com/gchq/Bailo/pull/141/files#r975493106 then I'd suggest the following changes here:

```
      <DialogContent>
        <Typography>Are you sure you want to perform this action?</Typography>
        <Alert severity='error' sx={{ m: 2, width: 400 }}>
          {errorMessage}
        </Alert>
      </DialogContent>
```

pages/model/[uuid].tsx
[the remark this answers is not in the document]
Comment (ARADDCC002) [reply pages/model/[uuid].tsx:68]: I think this was me - I added a width because by default the window looked very small, perhaps a min-width would be better, but I'm not against removing it altogether.

server/models/Deployment.ts
ARADDCC011: Covered in BAI-290

server/models/Version.ts
ARADDCC011: Covered in BAI-290

package.json
ARADDCC011: done

src/common/ConfirmationDialogue.tsx
ARADDCC011: No

pages/model/[uuid].tsx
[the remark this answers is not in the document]
Comment (ARADDCC012) [reply pages/model/[uuid].tsx]: By default the Alert is using the full width, so we have removed the `width` property

ARADDCC012 (approved):

PR #142, opened 2022-09-15
added a tooltip wrapper for disabled buttons explaining why they're d…

ARADDCC002:
…isabled


src/common/DisabledButtonInfo.tsx
```diff
@@ -0,0 +1,34 @@
...
+          updatedTitleText = `${updatedTitleText} ${condition}`
+        }
+      })
+      setTitleText(updatedTitleText)
+    }
+  }, [conditions])
+
+  return (
+    <Tooltip title={titleText} arrow placement={placement !== undefined ? placement : 'right'}>
+      <span>{children}</span>
```
ARADDCC012, line 29: I don't think it actually makes a difference in this case, but I prefer to keep `span` for when we specifically want something to be displayed inline. As this is a generic button wrapper I'd consider using a `div` instead

src/common/DisabledButtonInfo.tsx
```diff
@@ -0,0 +1,34 @@
...
+        if (condition !== '') {
+          updatedTitleText = `${updatedTitleText} ${condition}`
+        }
+      })
+      setTitleText(updatedTitleText)
+    }
+  }, [conditions])
+
+  return (
+    <Tooltip title={titleText} arrow placement={placement !== undefined ? placement : 'right'}>
```
ARADDCC012, line 28: Should be able to do:
```
placement={placement || 'right'}
```

src/common/DisabledButtonInfo.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { ReactNode, useEffect } from 'react'
+import Tooltip, { TooltipProps } from '@mui/material/Tooltip'
+
+function DisabledButtonInfo({
+  conditions,
+  placement,
+  children,
+}: {
+  conditions: string[]
+  placement?: TooltipProps['placement']
```
ARADDCC012, line 10: Personal preference is to place optional properties at the bottom (with any type definition). I find it helps when quickly scanning code, but not something worth changing unless it's a convention we want to enforce

src/common/DisabledButtonInfo.tsx
```diff
@@ -0,0 +1,34 @@
...
+  children: ReactNode
+}) {
+  const [titleText, setTitleText] = React.useState<string>('')
+
+  useEffect(() => {
+    if (conditions !== undefined) {
```
ARADDCC012, line 16: By using this logic we're effectively changing the API of MUI's `Tooltip`. Would it not be better to stick with simply extending `TooltipProps` and managing the `title` outside of this component as we would with a regular tooltip?

src/common/DisabledButtonInfo.tsx
ARADDCC012, line 28: Another convention I've found useful is to always provide true boolean props first, so in this case:
```
<Tooltip arrow title={titleText} placement={placement || 'right'}>
```
Again, not something worth changing unless it's a convention we want to enforce.

src/common/DisabledButtonInfo.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { ReactNode, useEffect } from 'react'
+import Tooltip, { TooltipProps } from '@mui/material/Tooltip'
+
+function DisabledButtonInfo({
```
ARADDCC012, line 4: This sort of wrapper component could be used for any disabled element. Is it worth renaming to reflect that? Something like `DisabledElementTooltip`?

ARADDCC002: This was something I actually considered but never got round - but yeah I agree a name change would be worth doing!

src/common/DisabledButtonInfo.tsx
ARADDCC002: I agree, just used <div /> as that's what people suggested using for this workaround. Changed it!

src/common/DisabledButtonInfo.tsx
ARADDCC002: I like this to be honest, perhaps we should do this from now on.

PR #149, opened 2022-09-28
added an alert that apears for reviewers if the model they're viewing…

ARADDCC002:
… has been recently updated


pages/model/[uuid].tsx
```diff
@@ -78,6 +78,9 @@ function Model() {
   const open = Boolean(anchorEl)
   const [copyModelCardSnackbarOpen, setCopyModelCardSnackbarOpen] = useState(false)
   const [complianceFlow, setComplianceFlow] = useState<Elements>([])
+  const [showLastViewedWarning, setShowLastViewedWarning] = useState<boolean>(false)
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:81]: Super minor but the `<boolean>` type isn't needed as `useState` will infer the type from the initial value you've given it

data/api.ts
```diff
@@ -12,6 +12,6 @@ export async function postEndpoint(url: string, data: any) {
   return fetchEndpoint(url, 'POST', data)
 }
 
-export async function putEndpoint(url: string, data: any) {
+export async function putEndpoint(url: string, data?: any) {
```
ARADDCC012, line 15: As a PUT request is usually expected to have a body, we should probably include a `Content-Length: 0` header to clarify that the body is expected to be empty.

data/api.ts
```diff
@@ -12,6 +12,6 @@ export async function postEndpoint(url: string, data: any) {
...
   return fetchEndpoint(url, 'PUT', data)
```
ARADDCC012, line 16: After a recent change that has gone into `main` you should be able to omit the `data` param if it's not been provided. `fetchEndpoint` will then omit the `body` from the request if not provided.

pages/model/[uuid].tsx
```diff
@@ -108,12 +111,47 @@ function Model() {
     setCopyModelCardSnackbarOpen(false)
   }
 
+  useEffect(() => {
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:114]: Could be shortened to
```
  useEffect(() => {
    if (currentUser && version) {
      if (currentUser.id === version.metadata.contacts.manager) {
        updateLastViewed('manager')
      }
      if (currentUser.id === version.metadata.contacts.reviewer) {
        updateLastViewed('reviewer')
      }
    }
  }, [managerLastViewed, reviewerLastViewed, currentUser])
```

It looks like you're missing `version` and `updateLastViewed` as a dependencies though. I'd recommend wrapping `updateLastViewed` with `useCallback` and moving it above this `useEffect` to avoid triggering this effect on every render and to avoid the no-use-before-define warnings.

pages/model/[uuid].tsx
```diff
@@ -108,12 +111,47 @@ function Model() {
...
+      if (currentUser !== undefined && currentUser.id === version?.metadata.contacts.manager) {
+        updateLastViewed('manager')
+      }
+      if (currentUser !== undefined && currentUser.id === version?.metadata.contacts.reviewer) {
+        updateLastViewed('reviewer')
+      }
+    }
+  }, [managerLastViewed, reviewerLastViewed, currentUser])
+
+  const updateLastViewed = (role: string) => {
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:125]: We should steer clear of type casting and null-null assertions as much as we can. They both nullify the type-safety that TS gives us. In line with https://github.com/gchq/Bailo/pull/149/files?diff=unified&w=1#r982569243 I recommend changing this to:
```
  const updateLastViewed = useCallback(
    (role: string) => {
      if (version && currentUser?.id === version.metadata.contacts.manager) {
        const updatedAtDate = new Date(version.updatedAt)
        if (
          (managerLastViewed && managerLastViewed.getTime() < updatedAtDate.getTime()) ||
          (reviewerLastViewed && reviewerLastViewed.getTime() < updatedAtDate.getTime())
        ) {
          setShowLastViewedWarning(true)
        }
        putEndpoint(`/api/v1/version/${version._id}/lastViewed/${role}`)
      }
    },
    [currentUser, version, managerLastViewed, reviewerLastViewed]
  )
```

pages/model/[uuid].tsx
```diff
@@ -108,12 +111,47 @@ function Model() {
...
+    }
+  }
+
   useEffect(() => {
     if (tab !== undefined && isTabOption(tab)) {
       setGroup(tab)
     }
   }, [tab])
 
+  useEffect(() => {
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:146]: Using non-null assertion and missing dependencies, can also be simplified slightly:
```
  useEffect(() => {
    if (version && !managerLastViewed) {
      setManagerLastViewed(version.managerLastViewed)
    }
    if (version && !reviewerLastViewed) {
      setReviewerLastViewed(version.reviewerLastViewed)
    }
  }, [version, managerLastViewed, reviewerLastViewed])
```

server/routes/v1/version.ts
```diff
@@ -99,3 +99,37 @@ export const resetVersionApprovals = [
     return res.json(version)
   },
 ]
+
+export const updateLastViewed = [
+  ensureUserRole('user'),
+  async (req: Request, res: Response) => {
+    const { id, role } = req.params
+    const { user } = req
+    const version = await findVersionById(req.user!, id, { populate: true })
```
ARADDCC012, line 108: Null-assertion being used here `req.user!`, our eslint rules forbid this. If user is optional we should handle it by using some sort of conditional logic

pages/model/[uuid].tsx
Comment (ARADDCC012) [reply pages/model/[uuid].tsx:125]: I've realised after spinning up the app that `managerLastViewed` and `reviewerLastViewed` are ISO strings, so the second `if` statement should instead be:
```
        if (
          (managerLastViewed && new Date(managerLastViewed).getTime() < updatedAtDate.getTime()) ||
          (reviewerLastViewed && new Date(reviewerLastViewed).getTime() < updatedAtDate.getTime())
        )
```

data/api.ts
```diff
@@ -2,6 +2,7 @@ export async function fetchEndpoint(url: string, method: string, data?: any) {
   return fetch(url, {
     method,
     headers: {
+      ...((!data || data !== 0) && { 'Content-Length': '0' }),
```
ARADDCC012, line 5: Looks like a small typo here, should be `&&` instead of `||`
`...(!data && data !== 0 && { 'Content-Length': '0' }),`

pages/model/[uuid].tsx
```diff
@@ -111,17 +116,60 @@ function Model() {
...
+    if (currentUser && version) {
+      setIsManager(currentUser.id === version.metadata.contacts.manager)
+      setIsReviewer(currentUser.id === version.metadata.contacts.reviewer)
     }
-  }, [tab])
+  }, [currentUser, version])
+
+  const updateLastViewed = useCallback(
+    (role: string) => {
+      if (isManager && version?.updatedAt) {
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:127]: If you change this to `if (isManager && version && version.updatedAt)` then you can remove the null access operators `?` and non-null assertions `!` within the code block below

pages/model/[uuid].tsx
```diff
@@ -111,17 +116,60 @@ function Model() {
...
+        if (
+          (managerLastViewed && new Date(managerLastViewed).getTime() < versionLastUpdatedAt!.getTime()) ||
+          (reviewerLastViewed && new Date(reviewerLastViewed).getTime() < versionLastUpdatedAt!.getTime())
+        ) {
+          setShowLastViewedWarning(true)
+        }
+        putEndpoint(`/api/v1/version/${version?._id}/lastViewed/${role}`)
+      }
+    },
+    [managerLastViewed, reviewerLastViewed, version?._id, version?.updatedAt, isManager]
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:138]: Could replace the following dependencies `version?._id, version?.updatedAt` with simply `version`

pages/model/[uuid].tsx
```diff
@@ -111,17 +116,60 @@ function Model() {
...
+      }
+      if (isReviewer) {
+        updateLastViewed('reviewer')
+      }
+    }
+  }, [currentUser, updateLastViewed, isManager, isReviewer])
+
+  useEffect(() => {
+    if (version && managerLastViewed === undefined) {
+      setManagerLastViewed(version!.managerLastViewed)
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:154]: Can remove the non-null assertion `!`

server/models/Version.ts
```diff
@@ -43,6 +45,8 @@ const VersionSchema: any = new Schema<Version>(
     built: { type: Boolean, default: false },
     managerApproved: { type: String, required: true, enum: approvalStateOptions, default: 'No Response' },
     reviewerApproved: { type: String, required: true, enum: approvalStateOptions, default: 'No Response' },
+    managerLastViewed: { type: Schema.Types.Mixed },
```
ARADDCC012, line 48: I'm not super familiar with mongoose schema types, but is an ISO string considered a `Schema.Types.Date`?

server/routes/v1/version.ts
```diff
@@ -99,3 +99,40 @@ export const resetVersionApprovals = [
...
+    const { id, role } = req.params
+    const { user } = req
+    if (!user) {
+      throw Forbidden({ code: 'user_unauthorised' }, 'Not user details found in request.')
+    }
+    const version = await findVersionById(req.user!, id, { populate: true })
+    if (!version) {
+      throw BadReq({ code: 'version_not_found' }, 'Unabled to find requested version')
+    }
+    if (user?.id !== version?.metadata.contacts[role]) {
```
ARADDCC012, line 115: `user` should be guaranteed at this point as we've guarded against `if (!user)` above. Should be able to remove the null access `?`

server/routes/v1/version.ts
```diff
@@ -99,3 +99,40 @@ export const resetVersionApprovals = [
...
+      throw Forbidden({ code: 'user_unauthorised' }, 'User is not authorised to do this operation.')
+    }
+    if (role === 'manager') {
+      updateManagerLastViewed(id)
+      req.log.info(
+        { code: 'version_last_viewed_updated', version: id, role },
```
ARADDCC012, line 121: Do we need to update the logging translations with the new logs you've added in this PR?

server/services/version.ts
```diff
@@ -96,3 +96,19 @@ export async function createVersion(user: UserDoc, data: CreateVersion) {
 
   return version
 }
+
+export async function updateManagerLastViewed(id: ModelId) {
+  return VersionModel.findOneAndUpdate(
+    { _id: id },
+    { $set: { managerLastViewed: new Date() as DateString } },
```
ARADDCC012, line 103: Ah this is where that new type doesn't work. I believe this is similar to `ObjectId`s in that they change when sent via a REST API.

Are we at the point where we'd benefit from having augmented types for the UI, taking into account `Date` and `ObjectId`s changing when being sent via REST.

If we do that then we can correctly type both server and UI.

If this is the way we want to go I'd vote for that being a separate ticket to the work you've done here.

types/interfaces.ts
```diff
@@ -182,3 +182,6 @@ export enum UploadModes {
   NewModel = 'newModel',
   NewVersion = 'newVersion',
 }
+
+export enum DateStringBrand {}
+export type DateString = string & DateStringBrand
```
ARADDCC012, line 187: Is it worth adding a comment explaining that `DateString` is used to represent date strings in ISO 8601 format?

types/interfaces.ts
```diff
@@ -182,3 +182,6 @@ export enum UploadModes {
   NewModel = 'newModel',
   NewVersion = 'newVersion',
 }
+
+export enum DateStringBrand {}
```
ARADDCC012, line 186: We shouldn't export this enum in order to prevent counterfeit `DateString`s from being created elsewhere. The point of the "brand" is to make `DateString` unique.

server/models/Version.ts
ARADDCC002: I believe because we're using the DateString it doesn't recognise it as a Date object and is failing to use the schema. I can double check but I believe that's why I changed it to this.

pages/model/[uuid].tsx
Comment (ARADDCC002) [reply pages/model/[uuid].tsx:138]: As far as I've tested, if you replace with just "version" it will continue to call because version is constantly updating.

pages/model/[uuid].tsx
Comment (ARADDCC002) [reply pages/model/[uuid].tsx:127]: This works for the content of the useEffect, but not the dependencies

server/routes/v1/version.ts
ARADDCC002: I would say so, yes. I think we need to make a note of any new log codes somewhere so that we can update logging, but not sure where's best.

server/services/version.ts
ARADDCC002: I'm happy to leave it as it is for now and explore this as some sort of future tech debt task yes.

pages/model/[uuid].tsx
Comment (ARADDCC012) [reply pages/model/[uuid].tsx]: You're entirely right, go ahead with your solution

server/routes/v1/version.ts
ARADDCC012: Do we create a confluence page to track them? Or we could create a ticket for each code that needs a new translation?

ARADDCC012 (approved):

PR #150, opened 2022-09-29
Feature/bai 248 display archived approvals

ARADDCC011:


server/routes/v1/requests.ts
```diff
@@ -43,6 +44,7 @@ export const getRequests = [
     const requests = await readRequests({
       type: type as RequestTypes,
       filter: filter === 'all' ? undefined : req.user._id,
+      archived: !!archived && archived === 'true',
```
ARADDCC012, line 47: I think I wrote this bit, but it can be simplified to `archived: archived === 'true',`

ARADDCC011: fixed

ARADDCC002 (approved):

PR #151, opened 2022-09-30
Feature/bai 290 remove usage of any type in code base

ARADDCC012:


__tests__/selenium_tests/test_upload.ts
```diff
@@ -85,35 +88,41 @@ describe('End to end test', () => {
...
+
+            logger.info('Model not built, retrying in 2 seconds.')
+            await pause(2000)
           }
 
-          logger.info('Model not built, retrying in 2 seconds.')
-          await pause(2000)
+          logger.info({ modelInfo }, 'Received model information')
+        } else {
+          logger.error(`No matches found for URL: ${modelUrl}`)
```
a3957273, line 124: Probably should crash here.

pages/model/[uuid].tsx
```diff
@@ -27,11 +27,10 @@ import Stack from '@mui/material/Stack'
 import Tab from '@mui/material/Tab'
 import Tabs from '@mui/material/Tabs'
 import Typography from '@mui/material/Typography'
-import { useTheme } from '@mui/material'
+import { SelectChangeEvent, useTheme } from '@mui/material'
```
Comment (a3957273) [inline pages/model/[uuid].tsx:30]: We should try to rework this so we don't include `@mui/material`.  This alone increases our build times by 100% (well, at least when doing `import { Tab } from '@mui/material` instead of `import Tab from '@mui/material/Tab'`, I assume the same applies here).

a3957273 (changes requested):

pages/model/[uuid].tsx
Comment (ARADDCC012) [reply pages/model/[uuid].tsx:30]: Good point, it should only affect dev build times, but I'll start changing any named mui imports to defaults.

pages/deployment/[uuid].tsx
```diff
@@ -53,6 +51,9 @@ function isTabOption(value: string): value is TabOptions {
   return ['overview', 'compliance', 'build', 'exports', 'settings'].includes(value)
 }
 
+const isVersionDoc = (value: unknown): value is VersionDoc =>
```
Comment (a3957273) [inline pages/deployment/[uuid].tsx:54]: This will be used in a bunch of places, so should be extracted to a utility file.

Comment (a3957273) [reply pages/deployment/[uuid].tsx:54]: However, in all the cases I can think of, `value` isn't `unknown`, it's either `ObjectId | VersionDoc` and we want to distinguish between the two.

This is because each of our `interface Model{}` stuff either populate the version field (so they're the full version doc) or just leave it as the identifier.

So I think this function should be changed to something like.

```
const isVersionPopulated = (value: Types.ObjectId | VersionDoc): value is VersionDoc => {
  return value.createdAt !== undefined
}
```

a3957273 (changes requested):

a3957273 (approved):

PR #153, opened 2022-10-03
Feature/BAI-297 deployment to model button

ARADDCC011:
Adds a button to the deployment page to return to the model page


pages/deployment/[uuid].tsx
```diff
@@ -188,11 +190,25 @@ export default function Deployment() {
     <>
       <Wrapper title={`Deployment: ${deployment.metadata.highLevelDetails.name}`} page='deployment'>
         {hasUploadType && initialVersionRequested?.metadata.buildOptions.uploadType === ModelUploadType.Zip && (
-          <Box sx={{ textAlign: 'right', pb: 3 }}>
-            <Button variant='outlined' color='primary' startIcon={<Info />} onClick={handleClickOpen}>
-              Show download commands
-            </Button>
-          </Box>
+          <Grid container spacing={1}>
```
Comment (ARADDCC002) [inline pages/deployment/[uuid].tsx:193]: I would change this to use a <Stack /> component instead https://mui.com/components/stack/

There's nothing wrong with using <Grid /> (as we've used it in various places in the app), but I'd say that for "simple" in-line layouts, Stack is better. Grid is better when dealing with larger layouts where you're sorting out layouts both vertically _and_ horizontally.

ARADDCC002 (changes requested):

pages/deployment/[uuid].tsx
```diff
@@ -28,6 +27,8 @@ import Link from 'next/link'
 import { useRouter } from 'next/router'
 import React, { MouseEvent, useEffect, useMemo, useState } from 'react'
 import { Elements } from 'react-flow-renderer'
+import { ModelDoc } from 'server/models/Model'
```
Comment (ARADDCC002) [inline pages/deployment/[uuid].tsx:30]: should be "import { ModelDoc } from '../../server/models/Model'"

pages/deployment/[uuid].tsx
```diff
@@ -187,30 +188,50 @@ export default function Deployment() {
...
-                },
-              }}
+        {deployment && (
+          <Stack direction='row' alignItems='center' justifyContent='space-between'>
+            <Button
+              variant='text'
+              color='primary'
+              sx={{ pb: 3 }}
+              startIcon={<ArrowBackIosNewIcon />}
+              onClick={() => router.push(`/model/${(deployment.model as ModelDoc).uuid}`)}
```
Comment (ARADDCC012) [inline pages/deployment/[uuid].tsx:198]: We should be type-guarding instead of using a type assertion (similar to casting in other languages) in situations like this.

That being said, I've created a new ticket (BAI-308) to address the root cause of this issue, so I'd argue this is fine for now so long as we address this in that ticket.

pages/deployment/[uuid].tsx
Comment (ARADDCC011) [reply pages/deployment/[uuid].tsx]: fixed

pages/deployment/[uuid].tsx
```diff
@@ -187,30 +188,50 @@ export default function Deployment() {
...
-                '& .MuiAlert-icon': {
-                  color: '#fff',
-                },
-              }}
+        {deployment && (
+          <Stack direction='row' alignItems='center' justifyContent='space-between'>
+            <Button
+              variant='text'
+              color='primary'
+              sx={{ pb: 3 }}
```
Comment (ARADDCC012) [inline pages/deployment/[uuid].tsx:196]: Instead of doing this you should use the `spacing` prop on `Stack`. `spacing={3}` should do what you want, but the benefit is that you won't have to keep manually adding `sx={{ pb: 3 }}` to every button bar the last one.

Comment (ARADDCC011) [reply pages/deployment/[uuid].tsx:196]: The pb is the padding below the row of buttons, not between the buttons

ARADDCC002 (approved):

PR #154, opened 2022-10-04
Feature/bai 125 split logs

ARADDCC002:


server/routes/v1/model.ts
```diff
@@ -121,7 +121,7 @@ export const getModelVersions = [
       throw NotFound({ code: 'model_not_found', uuid }, `Unable to find model '${uuid}'`)
     }
 
-    const versions = await findModelVersions(req.user, model._id, { thin: true })
+    const versions = await findModelVersions(req.user, model._id, { thin: true, showLogs: true })
```
ARADDCC012, line 124: Is this intentionally hard-coded to `true` or should we do something like `const showLogs = logs === 'true'`?

server/services/version.ts
```diff
@@ -34,6 +35,7 @@ export async function filterVersion<T>(user: UserDoc, unfiltered: T): Promise<T>
 export async function findVersionById(user: UserDoc, id: ModelId, opts?: GetVersionOptions) {
   let version = VersionModel.findById(id)
   if (opts?.thin) version = version.select({ state: 0, logs: 0, metadata: 0 })
+  if (!opts?.showLogs) version = version.select({ logs: 0 })
```
ARADDCC012, line 38: Would there be an issue if `opts.thin === true` and `opts.showLogs === false`? Would line 38 overwrite the data fetched on line 37?

ARADDCC002: I did think about this, and I actually think there is somewhere this could happen, so I'm going to change it back.

server/routes/v1/model.ts
ARADDCC002: Oh no this one slipped under the radar! Good spot

a3957273 (changes requested):
Looks all good to me, you just need to update the Swagger docs (sorry!).

a3957273 (approved):
Looks good to me! :)

PR #158, opened 2022-10-05
Fix extra spacing issue on forms

a3957273:


src/MuiForms/DescriptionField/DescriptionField.tsx
```diff
@@ -5,7 +5,7 @@ import React from 'react'
 function DescriptionField({ description }: FieldProps) {
   if (description) {
     return (
-      <Typography variant='subtitle2' sx={{ mt: 5 }}>
+      <Typography variant='subtitle2' sx={{ marginTop: '5px' }}>
```
ARADDCC002, line 8: Why change from mt? Can we not use mt: '5px' if you want to use exact?

src/MuiForms/ObjectFieldTemplate/ObjectFieldTemplate.tsx
```diff
@@ -24,7 +24,7 @@ function ObjectFieldTemplate({
     <>
       {(uiSchema['ui:title'] || title) && <TitleField id={`${idSchema.$id}-title`} title={title} required={required} />}
       {description && <DescriptionField id={`${idSchema.$id}-description`} description={description} />}
-      <Grid container spacing={2} sx={{ mt: 10 }}>
+      <Grid container spacing={2} sx={{ marginTop: '10px' }}>
```
ARADDCC002, line 27: Same as above

ARADDCC002 (approved):

PR #163, opened 2022-10-08
In-app Log Monitoring

a3957273:


ARADDCC002:
I think the search button on the admin log filters should be moved to be inside the menu

pages/admin.tsx
```diff
@@ -0,0 +1,54 @@
...
+    search: '',
+    isRegex: false,
+  })
+
+  const handleGetLogs = (filters: LogFilters): void => {
+    setLogFilters(filters)
+  }
+
+  return (
+    <Wrapper title='Admin' page='admin'>
```
ARADDCC002, line 25: We should check the user role here as well and return an error message. Right now if you view the page as a non-admin you just get the log box with a "loading..." text.

server/routes/v1/admin.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export const getApplicationLogs = [
+  ensureUserRole(['user', 'admin']),
+  async (req: Request, res: Response) => {
+    const after = parseDateQuery(req.query.after, new Date(0))
+    const before = parseDateQuery(req.query.before, new Date(Date.now()))
+    const level = parseInt(parseString('level', req.query.level) || '0', 10)
+
+    if (Number.isNaN(level)) {
+      throw BadReq({ level: req.query.level }, 'Level cannot be passed as number')
```
ARADDCC002, line 67: Would parsed make more sense here?

ARADDCC012: I think that's what the intention was. I'll update under that assumption.

pages/admin.tsx
ARADDCC012: Good catch, I'll put some generic unauthorised message in the event that someone manually navigates to the page.

server/models/Version.ts
```diff
@@ -1,8 +1,14 @@
 import { Document, IndexOptions, model, Schema, Types } from 'mongoose'
 import { LogStatement } from './Deployment'
-import logger from '../utils/logger'
 import { approvalStateOptions, ApprovalStates, DateString } from '../../types/interfaces'
 import { ModelDoc } from './Model'
+import logger from '../utils/logger'
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+function doNotUse() {
```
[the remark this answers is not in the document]
ARADDCC012: I've reimplemented the use of `logger` in `VersionSchema.log`. If that's not something we want then we might need to see why it's required for Version.

PR #165, opened 2022-10-09
Fix all eslint warnings

a3957273:


__tests__/selenium_tests/e2e_tests.ts
```diff
@@ -90,7 +90,10 @@ describe('End to end test', () => {
         logger.info('waiting until url contains model')
         await driver.wait(until.urlContains('/model/'))
         const modelUrl = await driver.getCurrentUrl()
-        const mName = modelUrl.match('/.*/model/(?<name>[^/]*)')!.groups!.name
+
+        const match = modelUrl.match('/.*/model/(?<name>[^/]*)')
+        if (!match || !match.groups) throw new Error('Could not parse model UUID from URL')
```
ARADDCC012, line 95: This will have conflicts with my branch that has finally been merged in. I prefer your solution so feel free to override my changes

src/Link.spec.tsx
```diff
@@ -7,7 +7,6 @@ import * as router from 'next/router'
 
 describe('Link', () => {
   it('renders a Link component', async () => {
-    const url = 'example_url'
```
ARADDCC012, line 10: Again, this will have conflicts with my changes. I removed most of the test as it wasn't actually using the `mockRouter` or `Link` at all.

I've written a test in `DocsWrapper.spec.tsx` that does use a mocked router, so if we want to update this down the line I'd use that as a reference.

src/MuiForms/IconButton/IconButton.tsx
```diff
@@ -19,9 +20,18 @@ type IconButtonProps = MuiIconButtonProps & {
 }
 
 function IconButton(props: IconButtonProps) {
-  const { icon, className, iconProps, ...otherProps } = props
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
ARADDCC012, line 23: Don't think this comment is needed anymore

ARADDCC012 (approved):

PR #166, opened 2022-10-10
Feature/update python client to include model download

jo77419:
Added functionality to download the code/binary files for a deployment. 

Added a few different ways of getting info about deployments

Updated examples

### DIFF
--- a/lib/python/examples/cognito_client.py
+++ b/lib/python/examples/cognito_client.py
@@ -13,7 +13,7 @@ load_dotenv()
 
 logging.basicConfig(stream=sys.stdout, level=logging.INFO)
 
-### Configure client based on local secrets
+## Configure client based on local secrets
 client = create_cognito_client(
     user_pool_id=os.getenv("BAILO_USERPOOL"),
     client_id=os.getenv("BAILO_CLIENT_ID"),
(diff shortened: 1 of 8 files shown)


lib/python/examples/cognito_client.py
```diff
@@ -103,3 +107,19 @@
 
 # You can also grab and inspect the model schema used for validation:
 schema = client.get_model_schema(model_uuid)
+
+
+## You can download the code and binary for a model if you have a deployment
+
+user_deployments = client.get_my_deployments()
+deployment = user_deployments[0]
```
a3957273, line 115: Handle if a user has no deployments

lib/python/examples/pki_client.py
```diff
@@ -102,3 +104,19 @@
 
 # You can also grab and inspect the model schema used for validation:
 schema = client.get_model_schema(model_uuid)
+
+
+## You can download the code and binary for a model if you have a deployment
+
+user_deployments = client.get_my_deployments()
+deployment = user_deployments[0]
```
a3957273, line 112: Handle if no deployments

a3957273 (changes requested):
Just need to handle the edge case where raw model exports have been rejected.

a3957273 (approved):

PR #167, opened 2022-10-10
Dev/fix delete model

ARADDCC002:


data/api.ts
```diff
@@ -16,3 +16,7 @@ export async function postEndpoint(url: string, data: unknown) {
 export async function putEndpoint(url: string, data?: unknown) {
   return fetchEndpoint(url, 'PUT', data)
 }
+
+export async function deleteEndpoint(url: string, data?: any) {
```
ARADDCC012, line 20: `data?: unknown`?

server/routes/v1/model.ts
```diff
@@ -163,3 +163,11 @@ export const getModelVersion = [
     return res.json(version)
   },
 ]
+
+export const deleteModel = [
+  ensureUserRole('user'),
+  async (_req: Request, _res: Response) => {
+    // This API is temporarily locked down
+    throw Unauthorised({}, 'This API call is temporarily unavailable')
```
ARADDCC012, line 171: Reminder that you suggested changing this to Not Implemented

src/common/ConfirmationDialogue.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+        open
+        onCancel={doNothing}
+        onConfirm={doNothing}
+        title={testTitle}
+        errorMessage={testError}
+      />
+    )
+
+    await waitFor(async () => {
+      expect(await screen.findByText(testTitle)).not.toBeUndefined()
```
ARADDCC012, line 25: Replace `.not.toBeUndefined()` with `.toBeDefined()`?

src/common/ConfirmationDialogue.tsx
```diff
@@ -0,0 +1,45 @@
...
+  onConfirm: () => void
+  errorMessage?: string
+}
+
+export default function ConfirmationDialogue({
+  open,
+  title,
+  onCancel,
+  onConfirm,
+  errorMessage,
```
ARADDCC012, line 22: As this is optional we could do `errorMessage = '',` so we don't have to handle `undefined`

server/routes/v1/deployment.ts
```diff
@@ -194,6 +194,16 @@ export const fetchRawModelFiles = [
       throw NotFound({ deploymentUuid: uuid }, `Unable to find deployment for uuid ${uuid}`)
     }
 
+    const versionDocument = await findVersionByName(req.user!, deployment.model, version)
```
ARADDCC012, line 197: `req.user!` -> `req.user`?

ARADDCC002:
* Delete items in registry 
* Warn users that when deleting a version it will also delete any associated deployments

server/routes/v1/version.ts
```diff
@@ -138,3 +140,56 @@ export const updateLastViewed = [
...
+      throw Forbidden({ code: 'user_unauthorised' }, 'User is not authorised to do this operation.')
+    }
+
+    const versionRequests = await RequestModel.find({ version: version._id })
+    if (versionRequests.length > 0) {
+      await Promise.all(versionRequests.map((versionRequest) => versionRequest.delete()))
+    }
+
+    const deployments = await DeploymentModel.find({ versions: { $in: [id] } })
+    if (deployments.length > 0) {
```
a3957273, line 166: There's no reason for this, you can just `forEach` over an empty array to do nothing.

server/routes/v1/version.ts
```diff
@@ -138,3 +140,56 @@ export const updateLastViewed = [
...
+      deployments.forEach(async (deployment) => {
```
a3957273, line 167: This is still not async safe, you need to do similar to above with a `.map()` and a `Promise.all()` call

server/routes/v1/version.ts
```diff
@@ -138,3 +140,56 @@ export const updateLastViewed = [
...
+        deployment.versions.remove(id)
+        if (deployment.versions.length === 0) {
+          const deploymentRequests = await RequestModel.find({ deployment: deployment._id })
+          if (deploymentRequests.length > 0) {
```
a3957273, line 171: Similar unnecessary `.length` check.

server/routes/v1/version.ts
```diff
@@ -138,3 +140,56 @@ export const updateLastViewed = [
...
+    const version = await findVersionById(user, id)
+
+    if (!version) {
+      throw NotFound({ code: 'version_not_found', id }, 'Unable to find version')
+    }
+
+    if (req.user.id !== version.metadata.contacts.uploader) {
+      throw Forbidden({ code: 'user_unauthorised' }, 'User is not authorised to do this operation.')
+    }
+
```
a3957273, line 159: Almost all of this logic and below should be in `services/` functions.  Routes handle formatting the request, business and response logic.  Services handle altering / searching the database.  This idea has several merits:

- Less duplication of logic (e.g. we'll want to delete versions when deleting models as well, so good to put that code somewhere accessible to both)
- More secure (most services require `req.user` which can be used for access control, etc).

a3957273 (changes requested):

server/routes/v1/version.ts
ARADDCC002: This was actually my plan to refactor this method once it's finished (still working on the registry bit). I was going to split it up into separate functions if I could, but it might all go in one service function, think it depends how large it's going to be.

pages/model/[uuid].tsx
```diff
@@ -228,6 +231,25 @@ function Model() {
...
+    setDeleteConfirmOpen(false)
+  }
+
+  const handleDeleteConfirm = async () => {
+    const response = await deleteEndpoint(`/api/v1/version/${version._id}`)
+
+    if (response.ok) {
+      router.push('/')
+    } else {
+      setDeleteModelErrorMessage(`Error ${response.status}: ${response.statusText}`)
```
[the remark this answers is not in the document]
Comment (ARADDCC002) [reply pages/model/[uuid].tsx:249]: Was going to sort out the UI side of things once the rest of it was finished, but thanks for the reminder. :)

ARADDCC002:
- When all versions deleted for a model, also delete docker registry name space (currently only the tags/containers are removed)

server/utils/modelCleanUp.ts
```diff
@@ -0,0 +1,51 @@
...
+        await deployment.save()
+      }
+    })
+  )
+}
+
+export async function updateModelByVersion(version: VersionDoc) {
+  const model = await ModelModel.findById(version.model)
+  if (model) {
+    model.versions.remove(version._id)
```
a3957273, line 44: I think this is actually a promise, so needs to be awaited?  I think this also means you don't need the `await model.save()` but testing is required.

server/routes/v1/version.ts
```diff
@@ -138,3 +144,30 @@ export const updateLastViewed = [
...
+    await deleteVersionRequests(version)
+    await updateDeploymentsByVersion(version, user)
+    await deleteDeploymentsByInitialVersion(version)
+    await updateModelByVersion(version)
+
+    await version.delete()
+
+    return res.json(id)
```
a3957273, line 171: Let's return `{ id }` here instead, I'd like to try to always return objects in the future.

server/services/deployment.ts
```diff
@@ -104,3 +111,36 @@ export async function updateDeploymentVersions(user: UserDoc, modelId: ModelId,
     })
   )
 }
+
+export async function deleteRegistryObjects(user: UserDoc, deployment: DeploymentDoc, namespace: string) {
+  let digest = ''
+  const { modelID, initialVersionRequested } = deployment.metadata.highLevelDetails
```
a3957273, line 117: What if we're removing a version that isn't the initialVersionRequested?

server/utils/modelCleanUp.ts
```diff
@@ -0,0 +1,51 @@
...
+  await Promise.all(
+    intialVersionRequestedDeployments.map((initialVersionDeloyment) => initialVersionDeloyment.delete())
+  )
+}
+
+export async function updateDeploymentsByVersion(version: VersionDoc, user: UserDoc) {
+  const deployments = await DeploymentModel.find({ versions: { $in: [version._id] } })
+  await Promise.all(
+    deployments.map(async (deployment) => {
+      deleteRegistryObjects(user, deployment, 'internal')
```
a3957273, line 27: needs an await.

server/utils/modelCleanUp.ts
```diff
@@ -0,0 +1,51 @@
...
+      deleteRegistryObjects(user, deployment, user.id)
```
a3957273, line 28: needs an await.

server/utils/modelCleanUp.ts
```diff
@@ -0,0 +1,51 @@
...
+      deployment.versions.remove(version._id)
```
a3957273, line 29: As mentioned before, possibly needs an await because it returns a promise?

a3957273 (changes requested):

PR #168, opened 2022-10-11
Docs/bai 265 update documentation

ARADDCC002:


docs/user-guide.md
```diff
@@ -44,8 +44,8 @@ laid out on this page. This includes making use of a template model class, which
 consistency with respect to the code structure, but avoids imposing unnecessary restrictions. At present, this class
 must be present in your codebase but the intention is to make it available for import as an external library.
 
-The model code is uploaded as a <b>zip file</b> on the upload files section on the user interface upload form. The
-uploaded code must contain:
+The model code is uploaded as a <b>zip file</b> (however, other file types may be added as options at a later date) on
```
a3957273, line 47: Don't include information about _future_ features in documentation, it defines the _current_ state of the project.  When other file types are added, we can add it to the documentation.  Otherwise we'll be adding a lot of 'in the futures' everywhere!

pages/docs/users/approvals/model-approvals.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-priority: 11000
+priority: 300
```
a3957273, line 2: I think we agreed priorities in a previous meeting, but the idea was mostly around having increasing priorities as you go down in layers.  So, everything at this layer is a 5 digit number.

pages/docs/users/upload-a-model/why-upload-a-model.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-priority: 10010
+priority: 100
```
a3957273, line 2: Similar comment about priorities.

pages/docs/users/deployments/index.mdx
```diff
@@ -0,0 +1,9 @@
+---
+priority: 200
```
a3957273, line 2: Similar comment about priorities.

pages/docs/users/automation/index.mdx
```diff
@@ -0,0 +1,9 @@
+---
+priority: 400
```
a3957273, line 2: Similar comment about priorities.

a3957273 (changes requested):

docs/user-guide.md
ARADDCC002: I've updated it to add docker container as we're now supporting that.

PR #171, opened 2022-10-12
Feature/bai 273 fix table styling markdown

ARADDCC002:


public/css/table.css
```diff
@@ -0,0 +1,13 @@
+table {
```
ARADDCC012, line 1: Reminder to check that the following props are unaffected by this global CSS:
- padding
- size
- stickyHeader

ARADDCC012 (approved):

PR #173, opened 2022-10-16
Allow uploading a docker tar directly

a3957273:


pages/deployment/[uuid].tsx
```diff
@@ -220,6 +220,24 @@ export default function Deployment() {
                 </Alert>
               </Box>
             )}
+            {hasUploadType && initialVersionRequested?.metadata.buildOptions.uploadType === ModelUploadType.Docker && (
+              <Box>
```
Comment (ARADDCC002) [inline pages/deployment/[uuid].tsx:224]: Not for this ticket, but I wonder if we can change how the alerts on these pages work. We could either use a  re-usable component, but I don't know if there might be cases where there should be multiple alerts?

server/utils/build/PushDockerTar.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    if (bytesRead < size) {
+      const { headers: responseHeaders } = await axios.patch(followUploadUrl, chunk, { headers })
+      followUploadUrl = `${registry}${responseHeaders.location}`
+    }
+  }
+
+  // last chunk
+  const digest = `sha256:${sha256.digest('hex')}`
+  await axios.put(`${followUploadUrl}&digest=${digest}`, chunk, { headers })
```
ARADDCC002, line 81: Is there a reason we switch from using fetch to axios in different places? Is it worth being consist and using one, or does axios need to be used here?

a3957273: Yes, I agree we should aim to be consistent.  It seems that over time we've switched from `node-fetch` to `axios` so have defaulted to using that.  No technical reason beyond it having better support for progress, cancellation, etc.

ARADDCC002: Happy to use axios in future work, I'll make a ticket for going through and making previous uses consistent.

PR #177, opened 2022-10-20
Made it so that you can click the Bailo text in the app bar to navigate back to marketplace

ARADDCC002:


a3957273 (approved):

src/Wrapper.tsx
```diff
@@ -194,14 +194,16 @@ export default function Wrapper({ title, page, children }: WrapperProps): ReactE
...
-              sx={{ flexGrow: 1, ml: 2, display: { xs: 'none', md: 'flex' } }}
-            >
-              Bailo
-            </Typography>
+            <Link href='/' passHref>
+              <Typography
+                variant='h6'
+                noWrap
+                component='div'
+                sx={{ flexGrow: 1, ml: 2, display: { xs: 'none', md: 'flex', cursor: 'pointer' } }}
```
a3957273, line 202: We shouldn't be defining `cursor` on our typography, it should be caused by our `Link` element.  Do we need to add it here?

a3957273 (changes requested):

src/Wrapper.tsx
ARADDCC002: Yes - because I'm using the Typography component underneath. Perhaps I can remove it and just use the text without Typography?

a3957273: Screenshot 2022-10-21 at 15 03 33

It seems to be using a div instead of a link.  We'll need to fix this to conform with accessibility requirements.

ARADDCC002: I think this might because of the typography thing. I think the best thing to do is use an anchor tag and change the styling of it?

a3957273 (approved):

PR #180, opened 2022-10-21
Fix/bai 338 deployments arrange by dark

ARADDCC011:
Deployments page arrange by drop down menu in light mode:
arrange-by-light

Deployments page arrange by drop down menu in dark mode:
arrange-by-dark


a3957273 (approved):
Looks great to me! :)

ARADDCC002:
I've very quickly removed the white background that was being used for each deployment if you listed it by model as well.

PR #182, opened 2022-10-21
Feature/bai 311 add nav buttons to docs

ARADDCC002:


src/docs/DocsWrapper.tsx
```diff
@@ -22,14 +27,30 @@ const paddingIncrement = 2
...
+  const router = useRouter()
   const { docsMenuContent, errorMessage } = useContext(DocsMenuContext)
+  const [flattenedPages, setFlattenedPages] = useState<DocsMenuContent>([])
 
   const linkColour = useMemo(
     () => (theme.palette.mode === 'light' ? theme.palette.primary.main : theme.palette.secondary.main),
     [theme]
   )
 
+  const flattenPages = useCallback((array) => {
```
ARADDCC012, line 39: Could update this to TS:
```
  const flattenPages = useCallback((array: DocsMenuContent) => {
    let result: DocsMenuContent = []
    array.forEach((item) => {
      result.push(item)
      if (isDocHeading(item)) {
        result = result.concat(flattenPages(item.children))
      }
    })
    return result.filter((item) => !isDocHeading(item) || item.hasIndex)
  }, [])
```

src/docs/DocsWrapper.tsx
```diff
@@ -22,14 +27,30 @@ const paddingIncrement = 2
...
+    array.forEach((a) => {
+      result.push(a)
+      if (Array.isArray(a.children)) {
+        result = result.concat(flattenPages(a.children))
+      }
+    })
+    return result.filter((item: any) => item.hasIndex || item.hasIndex === undefined)
+  }, [])
+
+  useEffect(() => {
```
ARADDCC012, line 50: You could condense this effect hook and the above state hook into a single memo hook:
```
  const flattenedPages = useMemo(() => flattenPages(docsMenuContent), [docsMenuContent, flattenPages])
```

src/docs/DocsWrapper.tsx
```diff
@@ -67,20 +88,32 @@ export default function DocsWrapper({ children }: DocsWrapperProps): ReactElemen
...
-    [pathname]
+    [router.pathname]
   )
 
   const docsMenu = useMemo(
     () => docsMenuContent.map((doc) => createDocElement(doc)),
     [docsMenuContent, createDocElement]
   )
 
+  const reducedPath = router.pathname.replace(/^(\/docs\/)/, '')
```
ARADDCC012, line 105: Could wrap both of these in a memo hook:
```
  const reducedPath = useMemo(() => router.pathname.replace(/^(\/docs\/)/, ''), [router.pathname])

  const currentIndex = useMemo(
    () => flattenedPages.findIndex((item) => item.slug === reducedPath),
    [flattenedPages, reducedPath]
  )
```

src/docs/DocsWrapper.tsx
```diff
@@ -116,7 +149,31 @@ export default function DocsWrapper({ children }: DocsWrapperProps): ReactElemen
...
+                            Home
+                          </Button>
+                        )}
+                        {currentIndex > 0 && (
+                          <Button startIcon={<ArrowBack />} onClick={() => changePage(currentIndex - 1)}>
+                            {flattenedPages[currentIndex - 1].title}
+                          </Button>
+                        )}
+                        {currentIndex < flattenedPages.length - 1 && (
+                          <Button endIcon={<ArrowForward />} onClick={() => changePage(currentIndex + 1)} sx={{}}>
```
ARADDCC012, line 168: Empty `sx` prop being provided here

src/docs/DocsWrapper.tsx
```diff
@@ -22,14 +27,30 @@ const paddingIncrement = 2
 
 export default function DocsWrapper({ children }: DocsWrapperProps): ReactElement {
   const theme = useTheme()
-  const { pathname } = useRouter()
+  const router = useRouter()
```
ARADDCC012, line 30: Is there a reason for reverting from destructuring? I tend to prefer destructuring as it reduces the amount of dot notation being used throughout the code.

ARADDCC012 (changes requested):

src/docs/DocsWrapper.tsx
ARADDCC002: We need to use router itself to change page, L#114 for example.

ARADDCC012 (approved):

PR #187, opened 2022-10-26
removed export raw model build option

ARADDCC002:


src/RawModelExportList.tsx
```diff
@@ -17,43 +16,37 @@ function RawModelExportList({ deployment }: { deployment: Deployment }) {
...
+        versions.map((version: any) => (
+          <Box key={version.version}>
+            <Box sx={{ p: 1 }}>
+              <Box sx={{ p: 2 }}>
+                <Typography variant='h4'>Version: {version.version}</Typography>
               </Box>
-            )
-        )}
-      {versions && versions.filter((e) => e.metadata.buildOptions?.exportRawModel).length === 0 && (
-        <EmptyBlob text='No exportable versions' />
```
a3957273, line 54: You still need the empty blob check.

a3957273 (changes requested):

src/RawModelExportList.tsx
ARADDCC002: I didn't think it would be possible for a deployment to _not_ have any versions but can add it back in.

lib/python/bailoclient/client.py
```diff
@@ -207,9 +207,6 @@ def download_model_files(
             str: Response status code
         """
 
-        if not self.__allow_exports(deployment_uuid):
-            raise ModelFileExportNotAllowed("Files are not exportable for this model")
```
jo77419, line 211: can delete this exception now - it's in utils/exceptions.py. 

will also need to remove the import in this file and the reference to the exception in the docstring

jo77419:
several of the python tests are broken - should be an easy fix but let me know if you need a hand

a3957273 (approved):

PR #189, opened 2022-10-27
added the ui components for seldon selector

ARADDCC002:


config/default.js
```diff
@@ -71,6 +71,9 @@ module.exports = {
       showWarning: true,
       checkboxText: 'By checking here you confirm that the information is correct',
     },
+    seldon: {
+      versions: ['v1', 'v2'],
```
a3957273, line 75: These should point to the Docker Seldon versions we use in the application.  They should have both an 'image' and a 'name', as the image may not be user readable on some registries.

pages/model/[uuid]/new-version.tsx
```diff
@@ -142,7 +143,8 @@ function Upload() {
 
     form.append('code', data.files.code)
     form.append('binary', data.files.binary)
-
+    form.append('seldonVersion', data.files.seldonVersion)
```
Comment (a3957273) [inline pages/model/[uuid]/new-version.tsx:146]: Seldon version is _not_ a file and so must not be added as an extra field, it should remain in the model card like every other item.

src/Form/RenderFileTab.tsx
```diff
@@ -29,23 +43,42 @@ export default function RenderFileTab({ step, splitSchema, setSplitSchema }: Ren
...
           <Divider orientation='vertical' flexItem />
           <Box sx={{ textAlign: 'center' }}>
-            <Typography sx={{ p: 1 }} variant='h5'>
-              Upload a binary file (.zip)
-            </Typography>
+            <Typography variant='h5'>Upload a binary file (.zip)</Typography>
             <FileInput label='Select Binary' onChange={handleBinaryChange} file={binary} accepts='.zip' />
           </Box>
+          <Divider orientation='vertical' flexItem />
+          <FormControl>
```
a3957273, line 65: The selector should be implemented in the JSON Schema.  Maybe a custom selector?

src/Form/RenderFileTab.tsx
```diff
@@ -29,23 +43,42 @@ export default function RenderFileTab({ step, splitSchema, setSplitSchema }: Ren
...
+            <InputLabel id='demo-simple-select-label'>Seldon version</InputLabel>
+            <Select
+              labelId='demo-simple-select-label'
```
a3957273, line 68: This label is wrong.

src/Form/RenderFileTab.tsx
```diff
@@ -29,23 +43,42 @@ export default function RenderFileTab({ step, splitSchema, setSplitSchema }: Ren
...
           <Box sx={{ textAlign: 'center' }}>
-            <Typography sx={{ p: 1 }} variant='h5'>
-              Upload a binary file (.zip)
-            </Typography>
+            <Typography variant='h5'>Upload a binary file (.zip)</Typography>
             <FileInput label='Select Binary' onChange={handleBinaryChange} file={binary} accepts='.zip' />
           </Box>
+          <Divider orientation='vertical' flexItem />
+          <FormControl>
+            <InputLabel id='demo-simple-select-label'>Seldon version</InputLabel>
```
a3957273, line 66: This id is wrong.

a3957273 (changes requested):

src/MuiForms/SeldonVersionSelector.tsx
```diff
@@ -0,0 +1,42 @@
+import { SeldonVersion } from '@/types/interfaces'
+import MenuItem from '@mui/material/MenuItem'
+import Select from '@mui/material/Select'
+import Box from '@mui/system/Box'
+import * as React from 'react'
+import { useGetUiConfig } from '../../data/uiConfig'
+
+export default function SeldonVersionSelector(props: any) {
```
ARADDCC012, line 8: Can `props` be typed?

src/MuiForms/SeldonVersionSelector.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  React.useEffect(() => {
+    if (uiConfig !== undefined) {
+      setSeldonVersions(uiConfig.seldonVersions)
+    }
+  }, [uiConfig])
+
+  const { onChange, value: currentValue, required, label, readonly } = props
+
+  const _onChange = (_event: any, newValue: any) => {
```
ARADDCC012, line 21: Personal preference for naming convention would be to call this `handleChange`

src/MuiForms/SeldonVersionSelector.tsx
```diff
@@ -0,0 +1,42 @@
...
+    onChange(newValue?.props.value)
+  }
+
+  return readonly ? (
+    <Box />
+  ) : (
+    <Select
+      labelId='seldon-version-label'
```
ARADDCC012, line 29: Where are these `labelId` and `id` props being used?

src/MuiForms/SeldonVersionSelector.tsx
```diff
@@ -0,0 +1,42 @@
...
+  }, [uiConfig])
+
+  const { onChange, value: currentValue, required, label, readonly } = props
+
+  const _onChange = (_event: any, newValue: any) => {
+    onChange(newValue?.props.value)
+  }
+
+  return readonly ? (
+    <Box />
```
ARADDCC012, line 26: Should return `null` if we don't want to render anything

src/MuiForms/SeldonVersionSelector.tsx
```diff
@@ -0,0 +1,42 @@
...
+  ) : (
+    <Select
+      labelId='seldon-version-label'
+      id='seldon-version-selector'
+      value={currentValue || ''}
+      label={label + (required ? ' *' : '')}
+      onChange={_onChange}
+    >
+      {seldonVersions.map((version: any) => (
```
ARADDCC012, line 35: can `version` be typed as `SeldonVersion` or something? Yet another `any` that we'll have to remove down the line otherwise

src/MuiForms/SeldonVersionSelector.tsx
ARADDCC002: Removed

ARADDCC012 (approved):

PR #190, opened 2022-10-31
Support multiple users within Bailo

a3957273:
Known issues:
- Users come up as 'datasets' on the model view page.
- No migration script for MongoDB
- No migration script for existing Docker deployments


__tests__/example_models/minimal_model/deployment.json
```diff
@@ -6,7 +6,6 @@
     "name": "Test Deployment"
   },
   "contacts": {
-    "requester": "user",
-    "secondPOC": "user"
+    "owner": [{ "kind": "user", "id": "user" }]
```
ARADDCC002, line 9: Is there a reason for "kind" over "type"? Does type have too many connotations to language types?

a3957273: It's a keyword in both TypeScript and Mongoose, which makes it a real pain to use.  I went with `kind` as an alternative, although happy to take other suggestions.

ARADDCC002: Thought as much. The only other name I can think of is category, but I'm happy to leave it as kind.

a3957273:
Now just need a Docker migration script.

pages/model/[uuid].tsx
```diff
@@ -86,10 +87,15 @@ function Model() {
...
   const { version, isVersionLoading, isVersionError, mutateVersion } = useGetModelVersion(uuid, selectedVersion, true)
   const { deployments, isDeploymentsLoading, isDeploymentsError } = useGetModelDeployments(uuid)
+  const { versionAccess } = useGetVersionAccess(version?._id)
 
   const hasUploadType = useMemo(() => version !== undefined && !!version.metadata.buildOptions?.uploadType, [version])
   const sendNotification = useNotification()
 
+  // possiblyUploader stores whether an uploader could plausibly have access to privileged functions.
+  // It defaults to true, until it hears false from the network access check.
+  const possiblyUploader = useMemo(() => versionAccess?.uploader !== false, [versionAccess])
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:97]: Would this be easier to read if `possiblyUploader` was renamed to something like `hasUploadAccess` or `isUploader`?

pages/upload.tsx
```diff
@@ -71,7 +71,7 @@ function Upload() {
 
     const { reference } = currentSchema
     const defaultState = {
-      contacts: { uploader: user.id },
+      contacts: { uploader: [{ kind: 'user', id: user.id }] },
```
ARADDCC012, line 74: Could use an enum for `kind` so that there's less chance of typos. Would be nice if we could use the same enum in our mongoose schema definition, so that we could manage both in one place.

Would something like this work?
```
export enum Kind {
  USER = 'user',
}

// example usage

const defaultState = contacts: { uploader: [{ kind: 'user', id: user.id }] },

// schema definition

kind: {
  type: String,
  enum: Object.values(Kind),
  required: true,
  },
```

src/MuiForms/EntitySelector.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { EntityKind } from '../../types/interfaces'
+
+interface Entity {
+  kind: string
+  id: string
+  data: any
+}
+
+export default function EntitySelector(props: any) {
+  const { users, isUsersLoading } = useListUsers()
```
ARADDCC012, line 15: Instead of renaming below you could do `const { users, isUsersLoading: isLoading } = useListUsers()`

src/MuiForms/EntitySelector.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { useListUsers } from '../../data/user'
+import { EntityKind } from '../../types/interfaces'
+
+interface Entity {
+  kind: string
+  id: string
+  data: any
+}
+
+export default function EntitySelector(props: any) {
```
ARADDCC012, line 14: Can `props` have a proper type?

src/MuiForms/EntitySelector.tsx
```diff
@@ -0,0 +1,74 @@
...
+          data: user,
+        }))
+      )
+
+    return tempEntities
+  }, [users])
+
+  const { onChange, value: currentValue, required, label } = props
+
+  const _onChange = (_event: any, newValue: any) => {
```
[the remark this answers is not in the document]
a3957273: Intend to keep it as `onChange` to stay consistent with the React Json Schema Forms library, which requires the parameter to be named that.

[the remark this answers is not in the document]
ARADDCC012: I'm talking about the function name, not the prop. The prop should be `onChange` to signify it's an event to handle. But the handler being called `_onChange` is confusing.

We only have to adhere to their API, which we are with our props. Internal functions can be named as we please.

src/MuiForms/EntitySelector.tsx
```diff
@@ -0,0 +1,80 @@
...
+      multiple
+      open={open}
+      onOpen={() => {
+        setOpen(true)
+      }}
+      onClose={() => {
+        setOpen(false)
+      }}
+      // we might get a string or an object back
+      isOptionEqualToValue={(option: any, value: any) => option.id === value.id}
```
ARADDCC012, line 57: Can `option` and `value` be typed? Looks like we know what they can be

src/MuiForms/EntitySelector.tsx
```diff
@@ -0,0 +1,80 @@
...
+import CircularProgress from '@mui/material/CircularProgress'
+import TextField from '@mui/material/TextField'
+import * as React from 'react'
+import { useListUsers } from '../../data/user'
+import { EntityKind } from '../../types/interfaces'
+
+interface Entity {
+  kind: string
+  id: string
+  data: any
```
ARADDCC012, line 11: Can `data` have a type? I'm guessing `unknown` would be suitable here with the expectation that we test the type as and when we need to manipulate the value.

ARADDCC002 (approved):

PR #192, opened 2022-11-01
Feature/helm uniformity

ARADDCC013:
HELM Uniformity

### DIFF
--- a/helm/bailo/templates/mail/mail.service.yaml
+++ b/helm/bailo/templates/mail/mail.service.yaml
@@ -7,12 +7,12 @@ metadata:
     name: mail
 spec:
   ports:
-    - port: {{ .Values.config.smtp.port }}
-      targetPort: {{ .Values.config.smtp.port }}
+    - port: 1080
+      targetPort: 1080
       protocol: TCP
       name: http
-    - port: 1025
-      targetPort: 1025
+    - port: {{ .Values.config.smtp.port }}
+      targetPort: {{ .Values.config.smtp.port }}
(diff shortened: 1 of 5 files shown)


helm/bailo/templates/mail/mail.service.yaml
```diff
@@ -7,13 +8,14 @@ metadata:
...
-    - port: {{ .Values.config.smtp.port }}
-      targetPort: {{ .Values.config.smtp.port }}
+    - port: 1080
+      targetPort: 1080
       protocol: TCP
       name: http
-    - port: 1025
-      targetPort: 1025
+    - port: {{ .Values.config.smtp.port }}
+      targetPort: {{ .Values.config.smtp.port }}
```
a3957273, line 16: This target port will always be 1025 right?  It's encoded into the container we use.  I think it makes sense to configure the externally visible port (under .Values.config.smtp.port), but not the targetPort.

a3957273 (changes requested):

helm/bailo/templates/mail/mail.service.yaml
ARADDCC013: I have configured .Values.config.smtp.port on mail.deployment.yaml

ARADDCC013: I had configured mail.deployment.yaml to listen to {{ .Values.config.smtp.port }}, hence the mail.service.yaml reflects that

a3957273 (approved):

PR #194, opened 2022-11-02
Feature/bai 363 model card tests

ARADDCC002:


cypress/fixtures/minimal_metadata_edit.json
```diff
@@ -0,0 +1,17 @@
+{
```
a3957273, line 1: With these new fixtures we now store identical test schemas in four different places:

```
server/scripts/example_schemas
lib/python/tests/resources
__tests__/example_models/minimal_model
cypress/fixtures
```

We should use existing sources, instead of duplicating them yet again.

a3957273 (changes requested):

cypress/fixtures/minimal_metadata_edit.json
ARADDCC002: The test schemas do have slight differences, but I'm wondering if we can fetch the content of the test schemas in the cypress test and edit them in there, therefore only having one overall?

Either way it's common practice to keep any test material inside the fixtures directory, so not sure how comfortable I am moving it. We could make a directory for any shared example schemas for tests (python and cypress - we'll be removing __tests__ soon). If we want to use them all in the same place I'd say we should make a new directory at the route level

cypress/e2e/model-card.cy.ts
```diff
@@ -0,0 +1,78 @@
...
+      metadata.buildOptions.uploadType = 'Model card only'
+      cy.get('[data-test=metadataTextarea]')
+        .clear()
+        .type(JSON.stringify(metadata), { parseSpecialCharSequences: false, delay: 0 })
+    })
+
+    cy.log('Submitting model')
+    cy.get('[data-test=warningCheckbox]').click()
+    cy.get('[data-test=submitButton]').click()
+    cy.url().should('contain', '/model/', { timeout: 10000 })
```
ARADDCC012, line 23: The timeout should be on the command, not the assertion:
`cy.url({ timeout: 10000 }).should('contain', '/model/')`

cypress/fixtures/minimal_metadata_edit.json
ARADDCC002: I've removed them from the python test folder (they now use python fixtures directory).
The __tests__ directory will be deleted once Cypress tests are finished. That just leaves them in two places, the /server/scripts dir, and /cypress/fixtures. I've ammended the tests to only use one metadata file though and edit the metadata for any test specific changes.

ARADDCC012 (approved):

PR #198, opened 2022-11-04
added favouriting test

ARADDCC002:


cypress/e2e/favouriting.cy.ts
```diff
@@ -0,0 +1,52 @@
...
+        .type(JSON.stringify(metadata), { parseSpecialCharSequences: false, delay: 0 })
+    })
+
+    cy.log('Submitting model')
+    cy.get('[data-test=warningCheckbox]').click()
+    cy.get('[data-test=submitButton]').click()
+    cy.url({ timeout: 10000 }).as('modelUrl').should('contain', '/model/')
+  })
+
+  it('Is able to favourite and unfavourite a model', function favouriteModelTest() {
```
ARADDCC012, line 31: Cypress has a different eslint config that means we dont have to name these callbacks (as we already name it in the first argument to `it()`, so can just be ` it('Is able to favourite and unfavourite a model', function () { ...`

cypress/e2e/favouriting.cy.ts
```diff
@@ -0,0 +1,52 @@
...
+    cy.visit(this.modelUrl)
```
ARADDCC012, line 32: I think it's useful to check that the URL is as we expect it to be before continuing with the rest of the test as this acts like a "wait for the page to load" sort of thing. In my tests I've done the following:
```
  // import util function
  import convertNameToUrlFormat from '../utils/convertNameToUrlFormat'

  // then in your test you can do
  cy.visit(this.modelUrl)
    cy.fixture('minimal_metadata.json').then((modelMetadata) => {
      cy.url().should('contain', `/model/${convertNameToUrlFormat(modelMetadata.highLevelDetails.name)}`)
    })
```

cypress/e2e/favouriting.cy.ts
```diff
@@ -0,0 +1,52 @@
...
+      updatedMetadata.buildOptions.uploadType = 'Model card only'
+      cy.get('[data-test=metadataTextarea]')
+        .clear()
+        .type(JSON.stringify(metadata), { parseSpecialCharSequences: false, delay: 0 })
+    })
+
+    cy.log('Submitting model')
+    cy.get('[data-test=warningCheckbox]').click()
+    cy.get('[data-test=submitButton]').click()
+    cy.url({ timeout: 10000 }).as('modelUrl').should('contain', '/model/')
```
ARADDCC012, line 28: I've done the following in my tests to make these url checks a bit stronger:
```
      import convertNameToUrlFormat from '../utils/convertNameToUrlFormat'

      cy.url({ timeout: 10000 })
        .as('modelUrl')
        .should('contain', `/model/${convertNameToUrlFormat(modelMetadata.highLevelDetails.name)}`)
        
```

In this case, if you wanted to do this, you'd have to move lines 25-28 inside your `.then()` callback on line 16 to have access to the fixture.

cypress/e2e/favouriting.cy.ts
```diff
@@ -8,6 +9,8 @@ describe('Model favouriting', () => {
     cy.visit('/upload')
     cy.get('[data-test=uploadJsonTab]').click({ force: true })
 
+    cy.log(modelName)
```
ARADDCC012, line 12: I don't think this log is needed

a3957273 (approved):

a3957273:
Before merging, can you check why it's only checking the CLA and not the integration / unit tests?

PR #202, opened 2022-11-10
Docs/bai 376 administration docs

ARADDCC011:


pages/docs/administration/getting-started/configuration/making-a-schema.mdx
```diff
@@ -0,0 +1,19 @@
...
+priority: 200
+---
+
+import DocsWrapper from '@/src/docs/DocsWrapper'
+
+# Making a Schema
+
+export default ({ children }) => <DocsWrapper>{children}</DocsWrapper>
+
+- The schemas use a JSON schema (existing examples are in server/scripts/example_schemas)
```
ARADDCC002, line 11: This needs to be expanded on much more. 

We should be detail why each property is used (like we are with widget) and also give other examples. Examples of how to make multiple choice, nested pages, dependencies, etc. 

We don't need to go into HUGE amounts of detail as we can forward users to the library's documentation, but as a user I'd expect a quick guide on how to make a very small schema from scratch.

pages/docs/administration/getting-started/helm-deployments.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+export default ({ children }) => <DocsWrapper>{children}</DocsWrapper>
+
+## Note
+
+Below deployment is configured for OpenShift. You will have to change route configuraiton to ingress for k8s.
+
+## Prerequisite
+
+- OpenShift CLI: https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz
```
a3957273, line 17: Use the Markdown format for links, `[name](http://example.com)`

pages/docs/administration/getting-started/building-docker-image.mdx
```diff
@@ -0,0 +1,32 @@
...
+Below lists options to change application configuration during dockerbuild.
+
+- default application configuration is applied from config/default.js.
+- production.js overwrites mongo configuration.
+- production.js overwrites minio configuration.
+- production.js overwrites registry configurration.
+- build.environment: openshift //must be set for OpenShift
+- openshift.[ namespace | appPublicRoute ] //must be set for OpenShift
+- smtp.enabled: true //must be set, else Bailo does not start the s2i build
+- logging.file.enabled: false //set true to debug Bailo
```
a3957273, line 26: None of these actually affect the built Docker image.  These should all be removed.

a3957273: These are all configured at runtime!

pages/docs/administration/getting-started/configuration/app-configuration.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+# App Configuration
+
+export default ({ children }) => <DocsWrapper>{children}</DocsWrapper>
+
+## Note
+
+Below listed configurations can be changed during dockerbuild.
+
+- default application configuration is applied from config/default.js.
```
a3957273, line 15: These are many more, it's odd to pick out just these few options.

a3957273: I'd argue these are either obvious or should be documented within the config themselves.  Not here.

pages/docs/administration/getting-started/configuration/app-configuration.mdx
```diff
@@ -0,0 +1,138 @@
...
+- smtp.enabled: true //must be set, else Bailo does not start the s2i build
+- logging.file.enabled: false //set true to debug Bailo
+
+## production.js
+
+production.js can be overwritten during HELM deployment. See helm/bailo/templates/bailo/bailo.configmap.yaml
+
+## Example **default.js**
+
+```
```
a3957273, line 30: This isn't an example, this _is_ the default configuration.  It shouldn't be included here, as it changes often.  Instead, you should link / direct the user to the location.

You should instead talk about `node-config` and how it overrides files (including the order of those overrides).

a3957273 (changes requested):

pages/docs/administration/getting-started/building-docker-image.mdx
ARADDCC013: amended - removed as per comment.

pages/docs/administration/getting-started/configuration/app-configuration.mdx
ARADDCC013: added link to node config docs page and reference to helm making changes to application during runtime.

pages/docs/administration/getting-started/helm-deployments.mdx
ARADDCC013: complete.

a3957273 (approved):

PR #211, opened 2022-11-15
updated node version for dev dockerfile

ARADDCC002:


ARADDCC013 (approved):

ARADDCC013:
In future, might have to look into using newer alpine with same node version.

PR #217, opened 2022-11-24
fixed models not correctl making nw versions

ARADDCC002:


server/utils/general.ts
```diff
@@ -8,9 +8,10 @@ export function getPropertyFromEnumValue<Enum extends { [key: string]: string |
 ): string | undefined {
   const entries = Object.entries(t)
 
-  for (const [index, value] of entries) {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  for (const [_index, value] of entries) {
```
ARADDCC012, line 12: Another way of doing this is
```
export function getPropertyFromEnumValue<Enum extends { [key: string]: string | number }>(
  t: Enum,
  key: string
): string | undefined {
  const entries = Object.values(t)

  for (const value of entries) {
    if (value === key) {
      return value
    }
  }

  return undefined
}
```

ARADDCC012 (approved):

PR #219, opened 2022-12-08
Feature/bai 380 add create model page

ARADDCC002:


ARADDCC002:
MinLength and maxLength for string are breaking

pages/schemas.tsx
```diff
@@ -0,0 +1,55 @@
...
+import Box from '@mui/material/Box'
+import Typography from '@mui/material/Typography'
+import SchemaDesigner from '@/src/SchemaDesign/SchemaDesigner'
+import { useRouter } from 'next/router'
+import React from 'react'
+import Tab from '@mui/material/Tab'
+import Tabs from '@mui/material/Tabs'
+import SchemaList from '@/src/SchemaDesign/SchemaList'
+
+type TabOptions = 'schemas' | 'designer'
```
ARADDCC012, line 12: Could use an enum to make everything a bit more type-safe:
```
enum TabOption {
  SCHEMAS = 'schemas',
  DESIGNER = 'designer'
} 
```

pages/schemas.tsx
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }, [tab])
+
+  return (
+    <Wrapper title='Design Schema' page='schemas'>
+      <Box>
+        {currentUser && currentUser.roles.includes('admin') ? (
+          <Box>
+            <Tabs value={group} onChange={handleGroupChange} aria-label='basic tabs example'>
+              <Tab label='Schemas' value='schemas' />
```
ARADDCC012, line 41: You could then use the enum for the Tab values `value={TabOption.SCHEMAS}`

pages/schemas.tsx
```diff
@@ -0,0 +1,55 @@
...
+              <Tab label='Designer' value='designer' />
+            </Tabs>
+            {group === 'schemas' && <SchemaList />}
```
ARADDCC012, line 44: You could also use the enum here `group === TabOption.SCHEMAS`

pages/schemas.tsx
```diff
@@ -0,0 +1,55 @@
...
+      setGroup(tab)
+    }
+  }, [tab])
+
+  return (
+    <Wrapper title='Design Schema' page='schemas'>
+      <Box>
+        {currentUser && currentUser.roles.includes('admin') ? (
+          <Box>
+            <Tabs value={group} onChange={handleGroupChange} aria-label='basic tabs example'>
```
ARADDCC012, line 40: Looks like the `aria-label`  can be removed

pages/schemas.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  React.useEffect(() => {
+    if (tab !== undefined && isTabOption(tab)) {
+      setGroup(tab)
+    }
+  }, [tab])
+
+  return (
+    <Wrapper title='Design Schema' page='schemas'>
+      <Box>
```
ARADDCC012, line 37: This can also be a fragment

pages/schemas.tsx
```diff
@@ -0,0 +1,55 @@
...
+import Typography from '@mui/material/Typography'
+import SchemaDesigner from '@/src/SchemaDesign/SchemaDesigner'
+import { useRouter } from 'next/router'
+import React from 'react'
+import Tab from '@mui/material/Tab'
+import Tabs from '@mui/material/Tabs'
+import SchemaList from '@/src/SchemaDesign/SchemaList'
+
+type TabOptions = 'schemas' | 'designer'
+function isTabOption(value: string): value is TabOptions {
```
ARADDCC012, line 13: If you go with the enum approach this can be
```
const isTabOption = (value: string): value is TabOption => {
  return !!TabOption[value]
}
``

File: src/SchemaDesign/QuestionPicker.tsx [hunk 7]
```diff
@@ -0,0 +1,84 @@
...
+  handleClose,
+  questionPickerOpen,
+}: {
+  onSubmit: (data: SchemaQuestion) => void
+  handleClose: () => void
+  questionPickerOpen: boolean
+}) {
+  const [expanded, setExpanded] = React.useState('')
+
+  const handleAccordianChange = (panel: string) => {
```
ARADDCC012, line 30: Typo, change to `handleAccordionChange`?

src/SchemaDesign/QuestionPicker.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <DialogContent>
+        <DialogContentText sx={{ mb: 2 }}>
+          The question reference must be unique, if it is left blank it will be generated automatically using the
+          question title. The question title is what is displayed on the form itself.
+        </DialogContentText>
+        {questionList.map((question) => (
+          <Accordion
+            key={question.title}
+            expanded={expanded === question.title}
+            onChange={() => handleAccordianChange(question.title)}
```
ARADDCC012, line 68: Same typo, change to `handleAccordionChange`?

src/SchemaDesign/QuestionPicker.tsx
```diff
@@ -0,0 +1,84 @@
...
+  handleClose: () => void
+  questionPickerOpen: boolean
+}) {
+  const [expanded, setExpanded] = React.useState('')
+
+  const handleAccordianChange = (panel: string) => {
+    setExpanded(panel)
+  }
+
+  const questionList = [
```
ARADDCC012, line 34: This can be moved outside of the component. It doesn't need to be recreated on every render.

src/SchemaDesign/QuestionPicker.tsx
```diff
@@ -0,0 +1,84 @@
...
+          The question reference must be unique, if it is left blank it will be generated automatically using the
+          question title. The question title is what is displayed on the form itself.
+        </DialogContentText>
+        {questionList.map((question) => (
+          <Accordion
+            key={question.title}
+            expanded={expanded === question.title}
+            onChange={() => handleAccordianChange(question.title)}
+          >
+            <AccordionSummary expandIcon={<ExpandMoreIcon />} aria-controls='panel1a-content' id='panel1a-header'>
```
ARADDCC012, line 70: What are the `aria-controls` and `id` props being used for?

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,418 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
ARADDCC012, line 1: I'm not a fan of disabling this rule. The docs give some good reasons as to why it's useful to follow:

"Enforces that there is no spreading for any JSX attribute. This enhances readability of code by being more explicit about what props are received by the component. It is also good for maintainability by avoiding passing unintentional extra props and allowing react to emit warnings when invalid HTML props are passed to HTML elements."

Can we instead destructure the object and explicitly provide the props we need?

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,418 @@
...
+    setSteps(steps.map((step) => (step.reference === selectedStep ? { ...step, questions: updatedQuestions } : step)))
+  }
+
+  const submitSchema = () => {
+    const reference = schemaReference === '' ? _.camelCase(schemaName) : _.camelCase(schemaReference)
+    if (schema) {
+      const userSchema = { ...schema, name: schemaName, reference, use: schemaUse }
+      console.log(userSchema)
+    }
+    // Submit schema
```
ARADDCC012, line 171: Make as a TODO with a reference to the ticket covering this?

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,418 @@
...
+    const stepToAmmend = updatedSteps.find((step) => step.reference === selectedStep)
+    const updatedQuestions = stepToAmmend.questions.filter((question) => question.reference !== questionReference)
+    setSteps(steps.map((step) => (step.reference === selectedStep ? { ...step, questions: updatedQuestions } : step)))
+  }
+
+  const submitSchema = () => {
+    const reference = schemaReference === '' ? _.camelCase(schemaName) : _.camelCase(schemaReference)
+    if (schema) {
+      const userSchema = { ...schema, name: schemaName, reference, use: schemaUse }
+      console.log(userSchema)
```
ARADDCC012, line 169: Log can be removed

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,418 @@
...
+              </Stack>
+
+              {steps.map((step) => (
+                <Box key={step.reference}>
+                  {step.reference === selectedStep && (
+                    <Box sx={{ pt: 2 }}>
+                      <Typography variant='caption'>Questions can be reordered by drag and drop</Typography>
+                      <DragDropContext onDragEnd={onQuestionDragEnd}>
+                        <Droppable droppableId='questionList'>
+                          {(droppableProvided) => (
```
ARADDCC012, line 269: This is just typed as `any`, looks like we're missing the type definitions for this package. From what I can see `npm i -D @types/react-beautiful-dnd` should fix this.

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,418 @@
...
+import Box from '@mui/material/Box'
+import Button from '@mui/material/Button'
+import Divider from '@mui/material/Divider'
+import Paper from '@mui/material/Paper'
+import Stack from '@mui/material/Stack'
+import Tab from '@mui/material/Tab'
+import Tabs from '@mui/material/Tabs'
+import TextField from '@mui/material/TextField'
+import Typography from '@mui/material/Typography'
+import React, { useEffect } from 'react'
```
ARADDCC012, line 13: Can we be consistent by destructuring the package and only pulling out the functionality we want to use? Saves us having to do `React.` all over the place.

This also applies to some of your other `tsx` files in this PR.

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,418 @@
...
+    } else {
+      setStepErrorText(`Value ${stepName} already exists`)
+    }
+  }
+
+  const addQuestion = (data: SchemaQuestion) => {
+    const question = data
+    question.reference = data.reference === '' ? _.camelCase(data.title) : _.camelCase(data.reference)
+    const updatedSteps = steps
+    const stepToAmmend = updatedSteps.find((step) => step.reference === selectedStep)
```
ARADDCC012, line 152: Spelling, amend on has one 'm' :) `stepToAmend`

src/SchemaDesign/SchemaDesignerQuestion.tsx
```diff
@@ -0,0 +1,81 @@
...
+import TextField from '@mui/material/TextField'
+import React from 'react'
+
+export default function SchemaDesignerQuestion({
+  onSubmit,
+  question,
+  closeQuestionPicker,
+}: {
+  onSubmit: (data: SchemaQuestion) => void
+  question: any // type this
```
ARADDCC012, line 15: Comment to be addressed

src/SchemaDesign/SchemaDesignerQuestion.tsx
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable dot-notation */
```
ARADDCC012, line 1: We don't need to disable this rule, I've commented on the line below with more info.

ARADDCC012, line 1: We shouldn't be disabling eslint to get around typing issues

src/SchemaDesign/SchemaDesignerQuestion.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [formInput, setFormInput] = React.useState<SchemaQuestion>({
+    title: '',
+    description: '',
+    type: question.type,
+    reference: '',
+    format: question.format || '',
+  })
+
+  const handleChange = (event: React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) => {
+    const { value, name }: { value: string | number; name: string } = event.target
```
ARADDCC012, line 27: By doing this we are overwriting the types with incorrect types. The default types are correct, both `value` and `name` will be of type `string` as TextFields will always emit change events with the `value` being a `string`, even if you provide a `number` as the `value` prop.

I'd suggest removing this line and just accessing the value and name via `event.target.value` and `event.target.name`.

src/SchemaDesign/SchemaDesignerQuestion.tsx
```diff
@@ -0,0 +1,81 @@
...
+    if (value || (!value && event.target.attributes['required'])) {
```
ARADDCC012, line 28: Something's not right here, we shouldn't be disabling eslint to get around this type issue. It looks to me like you've typed `event` correctly. Are you sure `event.target.attributes.required` exists?

Happy to look into this further on Monday, but do not have time this evening.

src/SchemaDesign/SchemaDesignerQuestion.tsx
```diff
@@ -0,0 +1,81 @@
...
+          name='description'
+          value={formInput.description}
+          onChange={handleChange}
+        />
+        {question.additionalQuestions.includes('minLength') && (
+          <TextField
+            label='Minimum length'
+            type='number'
+            name='minLength'
+            value={formInput.minLength || undefined}
```
ARADDCC012, line 64: Providing `undefined`/`null` as a value will make this an uncontrolled input. As we want it to be a controlled input we should instead use an empty string as a fallback value={formInput.minLength || ''}

src/SchemaDesign/SchemaDesignerQuestion.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <TextField
+          label='Question description'
+          name='description'
+          value={formInput.description}
+          onChange={handleChange}
+        />
+        {question.additionalQuestions.includes('minLength') && (
+          <TextField
+            label='Minimum length'
+            type='number'
```
ARADDCC012, line 62: `number` inputs are very old and do not conform to modern accessibility standards for what we're using it for here. Instead you should use `type="text" inputmode="numeric" pattern="[0-9]*"` for the most comprehensive number input field.

If we want to limit the input to only accept certain characters, it's much better if we use form validation instead of blocking values outside of a given set of values. For example, if you use `type='number'` and try to put in alphabetic characters nothing will appear in the input field, but the user isn't given any feedback as to why their keystrokes have been ignored. We can instead validate the value when they try to submit the form and provide a useful error/helper text message telling them what values the field accepts.

You can easily fall into a bit of a rabbit hole when looking into this so here's a few useful sources that help to describe the problem:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#using_number_inputs
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#accessibility

src/SchemaDesign/SchemaDesignerQuestion.tsx
```diff
@@ -0,0 +1,81 @@
...
+            name='minLength'
+            value={formInput.minLength || undefined}
+            onChange={handleChange}
+          />
+        )}
+        {question.additionalQuestions.includes('maxLength') && (
+          <TextField
+            label='Maximum length'
+            type='number'
```
ARADDCC012, line 71: See https://github.com/gchq/Bailo/pull/219/files#r1044751463

pages/schemas.tsx
a3957273: It should be kept, but given a useful description!

src/SchemaDesign/SchemaDesigner.tsx
ARADDCC002: I agree, I'm happy to move it to be global. Some rules like this can be really nit-picky when sometimes there's no issue spreading props.

src/SchemaDesign/SchemaDesignerQuestion.tsx
ARADDCC002: According to the MUI docs, <TextArea ..> should use "type: number". Looking at the generated HTML, it creates an <input /> element with "type: number". Not sure how best to approach this - should we use MUI TextArea, or should we use base <input />? The attribute of "pattern" isn't valid for MUI's TextArea

package.json
```diff
@@ -118,6 +119,8 @@
     "@types/node": "^17.0.30",
     "@types/nodemailer": "^6.4.4",
     "@types/react": "^17.0.44",
+    "@types/react-beautiful-dnd": "^13.1.3",
+    "@types/selenium-webdriver": "^4.0.16",
```
ARADDCC012, line 123: We don't use the `selenium-webdriver` package anymore so don't think this is needed

src/SchemaDesign/SchemaDesignerQuestion.tsx
ARADDCC012: You're using `TextField` here, not `TextArea`.

Could you link me to the documentation you're referring to? I can see that MUI provides a couple of `TextArea`s via the `@mui/joy` and `@mui/base` packages, but we use `@mui/material` which doesn't offer a `TextArea`, so a bit confused as to what you're referring to.

types/interfaces.ts
```diff
@@ -242,3 +242,24 @@ export enum LogType {
...
+  maxLength?: number
+  widget?: string
+  readOnly?: boolean
+}
+
+export type SchemaPage = {
+  title: string
+  description: string
+  type: string
+  properties: SchemaQuestion
```
ARADDCC012, line 262: This doesn't look quite right to me. I thought `properties` was use to store multiple questions. Can't remember if it was a map of key/value pairs or an array, but as this is typed you can only have one `SchemaQuestion` per page.

src/SchemaDesign/SchemaDesignerQuestion.tsx
```diff
@@ -0,0 +1,87 @@
...
+  additionalQuestions: string[]
+}
+
+export default function SchemaDesignerQuestion({
+  onSubmit,
+  question,
+  closeQuestionPicker,
+}: {
+  onSubmit: (data: SchemaQuestion) => void
+  question: DesignerQuestion // type this
```
ARADDCC012, line 22: Comment can be removed :)

src/SchemaDesign/SchemaList.tsx
```diff
@@ -0,0 +1,62 @@
...
+import Grid from '@mui/material/Grid'
+import Link from '@mui/material/Link'
+import { useGetSchemas } from '../../data/schema'
+
+export default function SchemaList() {
+  const { schemas: uploadSchemas, isSchemasLoading: isUploadSchemasLoading } = useGetSchemas('UPLOAD')
+  const { schemas: deploymentSchemas, isSchemasLoading: isDeploymentSchemasLoading } = useGetSchemas('DEPLOYMENT')
+
+  return (
+    <Box>
```
ARADDCC012, line 15: Can be swapped out for a fragment

src/SchemaDesign/SchemaList.tsx
```diff
@@ -0,0 +1,62 @@
...
+      {!isUploadSchemasLoading && !isDeploymentSchemasLoading && (
+        <Grid container spacing={2}>
+          <Grid item sm={12} md={6} sx={{ width: '100%' }}>
+            <Paper sx={{ mt: 2, mb: 2, p: 2 }}>
```
ARADDCC012, line 19: Fine as is, but could be shortened to `sx={{ my: 2, p: 2 }}`

src/SchemaDesign/SchemaList.tsx
```diff
@@ -0,0 +1,62 @@
...
+                  ))}
+              </Box>
+            </Paper>
+          </Grid>
+          <Grid item sm={12} md={6} sx={{ width: '100%' }}>
+            <Paper sx={{ mt: 2, mb: 2, p: 2 }}>
+              <Stack spacing={2} alignItems='center' sx={{ textAlign: 'center' }}>
+                <Typography>Have an existing schema? Upload it directly below! </Typography>
+                <Typography variant='caption'>
+                  For more information on creating a schema yourself, please see{' '}
```
ARADDCC012, line 46: Not necessarily asking for a change here, but just throwing out an idea.

This looks like it's formatted correctly, but I've had issues with prettier adding/removing spaces when writing text like this. I find the following is safer/clearer to write like this:
```
                  {`For more information on creating a schema yourself, please see ${(
                    <Link target='_blank' href='/docs/administration/getting-started/configuration/making-a-schema'>
                      the documentation
                    </Link>
                  )}.`}
                </Typography>
```

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,416 @@
...
+import IconButton from '@mui/material/IconButton'
+import DeleteIcon from '@mui/icons-material/Clear'
+import { Theme as MaterialUITheme } from '../MuiForms'
+import QuestionPicker from './QuestionPicker'
+
+const SchemaForm = withTheme(MaterialUITheme)
+
+export default function SchemaDesigner() {
+  const [questionPickerOpen, setQuestionPickerOpen] = useState(false)
+  const [steps, setSteps] = useState<any[]>([])
```
ARADDCC012, line 40: I seem to remember `steps` being awkward to type. Is it worth us creating a `Step` type that simply extends `any`, but comes with an internal comment explaining why we have to type it as such? That way we could at least be consistent with using the `Step` type and reduce the number of `any`s we use. I know it makes littler difference functionally, but at least a new developer could understand why we use `any` and what the value is meant to represent by reading the comment.

As this would be a change that affects code out of the scope of this ticket I'd suggest making this change in a separate ticket, if you agree with my reasoning of course.

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,416 @@
...
+  }
+
+  return (
+    <Box>
+      <Alert sx={{ mt: 2, mb: 2 }} severity='info'>
+        This page is currently not fully implemented. More advanced schema options will be added in the future.
+      </Alert>
+      <Grid container spacing={2}>
+        <Grid item xs={12} sm={12} md={6}>
+          <Box>
```
ARADDCC012, line 215: Could be a fragment

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,416 @@
...
+    return <TextFieldsIcon />
+  }
+
+  return (
+    <Box>
+      <Alert sx={{ mt: 2, mb: 2 }} severity='info'>
+        This page is currently not fully implemented. More advanced schema options will be added in the future.
+      </Alert>
+      <Grid container spacing={2}>
+        <Grid item xs={12} sm={12} md={6}>
```
ARADDCC012, line 214: `sm={12}` can be removed and it would work the same

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,416 @@
...
+                  variant='text'
+                  startIcon={<AddIcon />}
+                  onClick={() => setNewStepDialogOpen(true)}
+                >
+                  Add step
+                </Button>
+              </Stack>
+
+              {steps.map((step) => (
+                <Box key={step.reference}>
```
ARADDCC012, line 261: Can be swapped out for a fragment. You can attach a `key` to a fragment like so:
```
import { Fragment } from 'react'

<Fragment key={step.reference}>
  ...
</Fragment>
```

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,416 @@
...
+                      <Button onClick={handleClickOpen}>Add new question</Button>
+                    </Box>
+                  )}
+                </Box>
+              ))}
+              {selectedStep === '' && <Typography>Add a new step to begin designing your schema</Typography>}
+            </Paper>
+          </Box>
+        </Grid>
+        <Grid item xs={12} sm={12} md={6}>
```
ARADDCC012, line 307: `sm={12}` can be removed

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,416 @@
...
+            {showForm && splitSchema?.steps[stepIndex]?.schema === undefined && (
+              <Typography>Nothing to preview!</Typography>
+            )}
+            {showForm && splitSchema?.steps[stepIndex]?.schema !== undefined && (
+              <SchemaForm
+                schema={splitSchema.steps[stepIndex].schema}
+                formData={splitSchema.steps[stepIndex].state}
+                uiSchema={splitSchema.steps[stepIndex].uiSchema}
+              >
+                {/* eslint-disable-next-line react/jsx-no-useless-fragment */}
```
ARADDCC012, line 326: As we've now had to disable this rule in multiple places, is it worth creating a simple wrapper component around a fragment that explains why we are using it like this? That way we on't have to keep disabling `eslint`.

Also, as someone who's not worked closely on the `SchemaForm` before, it's not clear as to why we're doing this. Personally I'd find it useful to have this explained in code.

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,416 @@
...
+                value={stepReference}
+              />
+              <Stack>
+                <TextField
+                  required
+                  label='Step name'
+                  onChange={(event): void => setStepName(event.target.value)}
+                  value={stepName}
+                />
+                <Typography variant='caption' sx={{ color: '#e35a5a' }}>
```
ARADDCC012, line 365: Instead of having a random colour controlled at the component level, can we define this in our theme and give it a descriptive name?

As we've experienced recently, these are the sort of things that can be a pain to resolve when supporting both light and dark modes if they're not controlled at the theme level.

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,416 @@
...
+              onChange={(event): void => setSchemaReference(event.target.value)}
+              value={schemaReference}
+            />
+            <TextField
+              required
+              label='Schema name'
+              onChange={(event): void => setSchemaName(event.target.value)}
+              value={schemaName}
+            />
+            <InputLabel id='schema-use-label'>Schema Type</InputLabel>
```
ARADDCC012, line 396: Is this `id` needed?

src/SchemaDesign/SchemaDesigner.tsx
```diff
@@ -0,0 +1,416 @@
...
+            <Select
+              labelId='schema-use-label'
```
ARADDCC012, line 398: I don't think these `labelId` and `id` props are needed

src/SchemaDesign/SchemaDesignerQuestion.tsx
ARADDCC002: apologies I meant <TextField /> :)

types/interfaces.ts
ARADDCC002: Deleting as it looks like I'm not using it anyway, which is good because typing the format that the library needs is a bit of a pain!

src/SchemaDesign/SchemaList.tsx
ARADDCC002: Didn't even know `my` existed! Very useful to know

src/SchemaDesign/SchemaDesigner.tsx
ARADDCC002: We already have` theme.palette.error.main`, so should be fine to use that

src/SchemaDesign/SchemaDesigner.tsx
ARADDCC002: AFAIK it's used to associate with the `<Select />` component below, it might be needed for accessibility

src/SchemaDesign/SchemaDesigner.tsx
ARADDCC002: I'm under the impression it's needed to tie the label and select component together for accessibility reasons, but happy to remove if this is not the case

src/SchemaDesign/SchemaDesignerQuestion.tsx
ARADDCC002: There's still a bug with the minLengh and maxLength fields so for now I might disable them anyway, as I want to draw a line in the scope of this ticket.

src/SchemaDesign/SchemaDesigner.tsx
ARADDCC002: I've removed this altogether and it seems to be working. I think I'll leave it in the other part of the code for now as I don't want to touch the upload functionality in this ticket, just in case something goes wrong!

src/SchemaDesign/SchemaDesigner.tsx
ARADDCC002: I think that you're right that this should be moved to a new ticket. Created `BAI-388` to look into this!

src/SchemaDesign/SchemaDesigner.tsx
ARADDCC012: You're right, we do need it. Good shout.

src/SchemaDesign/SchemaDesigner.tsx
ARADDCC012: We do need the `labelid` on the `Select` but we don't need the `id='demo-simple-select'` below

a3957273 (approved):

PR #221, opened 2022-12-09
bugfix/BAI-298-deployments-made-of-the-same-model-by-the-same-user-overwrite-each-other

ARADDCC012:


ARADDCC002 (approved):

a3957273:
Technically, ticket BAI-298 refers to the _registry_ deployments overwriting each other.  This will be resolved in the `groups` PR anyway, so the ticket is void.

This still seems like a useful feature.  Could we have some user facing information that this is the selected minimum version though?  I don't think we should hide it from the schema, we could include it with every other question.  To take this further, we could even make a custom component that allows you to change the selected minimum version.

ARADDCC002:
Ah looks like there was some missing information in the ticket - we thought it was to do with the application. We'll have a dicussion about whether to keep this in with some quick UI changes, or to leave it as-is.

ARADDCC012:
> Technically, ticket BAI-298 refers to the _registry_ deployments overwriting each other. This will be resolved in the `groups` PR anyway, so the ticket is void.
> 
> This still seems like a useful feature. Could we have some user facing information that this is the selected minimum version though? I don't think we should hide it from the schema, we could include it with every other question. To take this further, we could even make a custom component that allows you to change the selected minimum version.

Done

a3957273 (approved):

PR #222, opened 2022-12-11
Convert seldon to use the base select widget.

a3957273:
Converts the Seldon widget to use the basic MUI select, instead of redesigning the wheel.  This also converts the term `SeldonVersion` to `ImageBuilder` throughout our code, requested by another corporation.


ARADDCC002 (approved):

ARADDCC002:
When uploading a form, the build options tab doesn't validate as is the builder image question is required (even though it has a value)

image

src/MuiForms/BuilderImageSelector.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react'
+import { WidgetProps } from '@rjsf/core'
+import { Typography } from '@mui/material'
+import { useGetUiConfig } from '../../data/uiConfig'
+import SelectWidget from './SelectWidget'
+
+export default function BuilderImageSelector(props: WidgetProps): any {
```
ARADDCC012, line 7: Return type should be `ReactElement`

src/MuiForms/BuilderImageSelector.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const [selectProps, setSelectProps] = React.useState(props)
+
+  React.useEffect(() => {
+    if (!uiConfig) return
+
+    const { options, value } = props
+    options.enumOptions = uiConfig.builderImages.map((version) => ({ value: version.image, label: version.name }))
+
+    setSelectProps({ ...props, value: value || options.enumOptions[0].value })
+  }, [uiConfig, props])
```
ARADDCC012, line 18: If you destructure the props on line 7 then you won't have to depend on the whole object here and can instead depend on only `options`, `value` and `uiConfig`

a3957273:
Superseded by the deployment initial version PR.

PR #224, opened 2022-12-12
[Fun] Add snow that falls on each page

a3957273:
The snow is animated using CSS transitions for efficiency.  Concept taken from https://embed.im/snow/.

Screenshot 2022-12-12 at 15 54 20

This is just to add a little bit of Christmas cheer.  This likely shouldn't be merged in.


ARADDCC002:
I think this should be a permanent feature...

PR #226, opened 2022-12-14
feature/add-action-minikube

ARADDCC013:


.github/workflows/minikube.yml
```diff
@@ -0,0 +1,172 @@
...
+      # Install dependencies
+      - run: npm ci
+
+      # Create certs
+      - run: sudo chown -R runner:docker certs
+      - run: npm run certs
+
+      # Remove route set for openshift
+      - name: remove route config
+        run: sudo rm helm/bailo/templates/route.yaml
```
a3957273, line 40: Don't remove the `route.yaml` file, instead add a `value.yaml` entry to disable it.

.github/workflows/minikube.yml
```diff
@@ -0,0 +1,172 @@
...
+
+      # Add bailo NodePort
+      - name: Create NodePort Config
```
a3957273, line 43: Change this to be in the helm chart, possibly using an `ingress`.

.github/workflows/minikube.yml
```diff
@@ -0,0 +1,172 @@
...
+        run: |
+          cat << EOT >> helm/bailo/templates/nodePort.yaml
+          ---
+          apiVersion: v1
+          kind: Service
+          metadata:
+            name: bailo-node-port
```
a3957273, line 50: Don't hardcode bailo, instead use the variable.

.github/workflows/minikube.yml
```diff
@@ -0,0 +1,172 @@
...
+          openshift:
+            appPublicRoute: need.to.set.ingress
+
+          nginxcert:
+            cert: cert.pem
+            key: key.pem
+          EOT
+
+      - name: Set Bailo image pull policy to Never
+        run: sed -i 's/{{ .Values.image.pullPolicy }}/Never/g' helm/bailo/templates/bailo/bailo.deployment.yaml
```
a3957273, line 126: Should be an altered to be changed in the override yaml file.

.github/workflows/minikube.yml
```diff
@@ -0,0 +1,172 @@
...
+      - name: Start minikube
+        uses: medyagh/setup-minikube@master
+
+      - name: List Cluster
+        run: kubectl get pods -A
+
+      - name: Build Image
+        run: |
+          export SHELL=/bin/bash
+          eval $(minikube -p minikube docker-env)
```
a3957273, line 137: Probably not needed.

.github/workflows/minikube.yml
```diff
@@ -0,0 +1,172 @@
...
+          docker images
+
+      - name: Load docker image to minikube
+        run: |
+          export SHELL=/bin/bash
+          eval $(minikube -p minikube docker-env)
+          docker save local/bailo | (eval $(minikube docker-env) && docker load)
+
+      - name: HELM update dependency update
+        run: current_dir=$PWD;cd helm/bailo/;helm dependency update;cd $current_dir;
```
a3957273, line 149: Use multiple lines!

.github/workflows/minikube.yml
```diff
@@ -0,0 +1,172 @@
...
+
+      - name: List Pods and Services
+        run: |
+          echo "------------------listing services------------------"
+          kubectl get services
+          echo "------------------listing pods----------------------"
+          kubectl get pods
+
+      - name: HELM test Bailo
+        run: current_dir=$PWD;cd helm/bailo/;helm test bailo ;cd $current_dir;
```
a3957273, line 166: multiple lines!

a3957273 (changes requested):

.github/workflows/minikube.yml
ARADDCC013: route is not optional

.github/workflows/minikube.yml
ARADDCC013: ref to bailo now removed. This is temp file created only with action is run.

.github/workflows/minikube.yml
ARADDCC013: Unfortunately it fails without it.

.github/workflows/minikube.yml
ARADDCC013: updated with working-directory.

.github/workflows/minikube.yml
ARADDCC013: updated with working-directory.

.github/workflows/minikube.yml
ARADDCC013: For ingress to work (minikube specific), we need;
- Additional Ingress Controller Pod,  simply done by  "$ minikube addons enable ingress".
- Then create an ingress rule, and for rule to work we need dns mapping to IP; this can be done via /etc/hosts.
- **Issue**: However for the mapping we need IP which is dynamic and takes more than few seconds to be assigned. Hence, not able to create mapping in /etc/hosts on fly
- Once /etc/hosts is amended, we have all request coming to minikube cluster handed over to ingress controller, which validates the rules and forwards that request to service.

a3957273:
I'm going to close this as it's included in the other PR, but 👍, this looks amazing.  Look forward to improved testing.

PR #227, opened 2022-12-16
Improve reliability of tests

a3957273:


a3957273:
Reduces image build times from 504 seconds to 175 seconds.  Reduces image size from 1.6GB to 765MB.  Still installs `docker` despite it not being technically required to fix a certificate issue.  Removing this saves another 25 seconds of build times and another 250MB of image space.

a3957273:
Should be noted that 175 seconds is the _worst case scenario_, where no layer was previously cached.  As can be seen from the last image build on this PR, with cache the image build can be under 60 seconds.

ARADDCC002 (approved):

PR #228, opened 2022-12-16
Feature/bai 381 bailo home page

ARADDCC002:


a3957273 (changes requested):

a3957273:
I think I've fixed most of the issues with it.  Primarily around the next image, which I swapped out to be a cool slideshow!  You were totally write, there were some truly traumatizing issues with styling the image component.

Only thing left is that it doesn't work well on mobile devices.  Can you adapt it so that it works regardless of device?

a3957273 (approved):

PR #236, opened 2023-01-05
Feature/bai 244 view model files

ARADDCC011:


ARADDCC002:
You have a lot of empty files that need to be deleted

server/routes/v1/deployment.ts
```diff
@@ -257,6 +260,131 @@ export const fetchRawModelFiles = [
...
+  directories: Map<string, DirectoryMetadata>
+  files: string[]
+}
+
+const addToTree = (tree: DirectoryMetadata | undefined, path: string): void => {
+  if (tree === undefined) {
+    throw new Error('Error: Tree does not exist')
+  }
+  if (path.includes('/')) {
+    const i = path.indexOf('/')
```
ARADDCC002, line 274: Rename to index

server/routes/v1/deployment.ts
```diff
@@ -257,6 +260,131 @@ export const fetchRawModelFiles = [
...
+  }
+}
+
+const createTree = (rootName: string, files: string[]): DirectoryMetadata => {
+  const tree: DirectoryMetadata = { name: `${rootName}/`, directories: new Map(), files: [] }
+  files.forEach((file) => addToTree(tree, file))
+  return tree
+}
+
+// const treeToJSON = (tree: DirectoryMetadata): JSON => {
```
ARADDCC002, line 293: Commented out code should be removed if it's not needed

server/routes/v1/deployment.ts
```diff
@@ -257,6 +260,131 @@ export const fetchRawModelFiles = [
...
+
+    if (!(await isUserInEntityList(req.user, deployment.metadata.contacts.owner))) {
+      const owners = deployment.metadata.contacts.owner.map((owner) => owner.id).join(', ')
+      throw Unauthorised(
+        { deploymentOwner: deployment.metadata.contacts.owner },
+        `User is not authorised to download this file. Requester: ${req.user.id}, owners: ${owners}`
+      )
+    }
+
+    if (deployment.managerApproved !== 'Accepted') {
```
ARADDCC002, line 340: use ApprovalStates.Accepted instead of the string

server/routes/v1/deployment.ts
```diff
@@ -257,6 +260,131 @@ export const fetchRawModelFiles = [
...
+    stream.on('data', (data) => {
+      if (Buffer.isBuffer(data)) buffers.push(data)
+    })
+    stream.on('end', async () => {
+      const zip = JSZip()
+      const zipBuffer = Buffer.concat(buffers)
+      const zipFile = await zip.loadAsync(zipBuffer)
+      const fileNames = Object.keys(zipFile.files)
+      const tree = createTree(fileType, fileNames)
+      // tree.directories.forEach((item) => console.log(item))
```
ARADDCC002, line 382: Remove this

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+import Button from '@mui/material/Button'
+
+import { useState } from 'react'
+import { useRouter } from 'next/router'
+import DirectoryTreeView from './common/DirectoryTreeView'
+import SplitButton from './common/SplitButton'
+
+function RawModelExportItem({ deploymentUuid, version }: { deploymentUuid: string; version: any }) {
+  const router = useRouter()
+  const [displayTree, setDisplayTree] = useState<boolean>(false)
```
ARADDCC002, line 14: Typing isn't really necessary for strings and booleans (applies to line below too!)

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+import Typography from '@mui/material/Typography'
+import Box from '@mui/system/Box'
+import Button from '@mui/material/Button'
+
+import { useState } from 'react'
+import { useRouter } from 'next/router'
+import DirectoryTreeView from './common/DirectoryTreeView'
+import SplitButton from './common/SplitButton'
+
+function RawModelExportItem({ deploymentUuid, version }: { deploymentUuid: string; version: any }) {
```
ARADDCC002, line 12: Is there a reason why version is typed as any? You're using it as a key below so I'm assuming it's a string?

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const router = useRouter()
+  const [displayTree, setDisplayTree] = useState<boolean>(false)
+  const [codeMenuItems, setCodeMenuItems] = useState<string[]>(['Show Code File Tree'])
+
+  const handleCodeMenuItemClicked = (item: string) => {
+    if (item === 'Show Code File Tree') {
+      setDisplayTree(!displayTree)
```
ARADDCC002, line 19: This should be "true". Line #23 should be "false". I'm not 100% convinced of this implementation though and think it's something we could revisit

src/common/ApprovalsChip.tsx
```diff
@@ -45,6 +45,7 @@ export default function ApprovalsChip({ approvals }: ApprovalsChipProps): ReactE
   const getRequestResponses = useCallback((approval: Approval, index: number) => {
     let Icon
     let secondaryText = ''
+    console.log(approval.reviewers)
```
ARADDCC002, line 48: Remove

server/routes/v1/deployment.ts
ARADDCC011: That's a javascript function, not a variable name

src/RawModelExportItem.tsx
ARADDCC011: That was a holdover from before there was an if statement

src/RawModelExportItem.tsx
ARADDCC011: I believe this was a holdover from when it was using the parent variable (confusingly also named version), and I missed that when I decided to use the child variable rather than clutter the code with version.version everywhere

server/routes/v1/deployment.ts
ARADDCC011: For this the line should read as follows?
if (deployment.managerApproved !== ApprovalStates.Accepted) {

ARADDCC002: Yup :)

server/routes/v1/deployment.ts
ARADDCC012: I think he meant changing the `i` to `index`

server/routes/v1/deployment.ts
```diff
@@ -257,6 +258,120 @@ export const fetchRawModelFiles = [
...
+}
+
+export const fetchModelFileList = [
+  ensureUserRole('user'),
+  bodyParser.json(),
+  async (req: Request, res: Response<DirectoryArrayMetadata>) => {
+    const { uuid, version, fileType } = req.params
+    const deployment = await findDeploymentByUuid(req.user, uuid)
+
+    if (deployment === null) {
```
ARADDCC012, line 310: Change to `if (!deployment)`?

src/common/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,79 @@
+import TreeView from '@mui/lab/TreeView'
```
ARADDCC012, line 1: I think we need to be careful about which components end up in `/common`. This is a component designed for a specific purpose. It is not a generic, reusable component and therefore shouldn't be in this directory imo.

src/common/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,79 @@
...
+import TreeItem from '@mui/lab/TreeItem'
+import { useState, useCallback, useEffect } from 'react'
+import { DirectoryArrayMetadata } from 'server/routes/v1/deployment'
+import ChevronRightIcon from '@mui/icons-material/ChevronRight'
+import ExpandLessIcon from '@mui/icons-material/ExpandLess'
+import useNotification from './Snackbar'
+import { getEndpoint } from '../../data/api'
+
+export default function DirectoryTreeView({
```
ARADDCC012, line 10: I think we need to be careful about which components end up in `/common`. This is a component designed for a specific purpose. It is not a generic, reusable component and therefore shouldn't be in this directory imo.

src/common/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }, [version, getTreeResponse])
+
+  function directoryTreeItem(tree: DirectoryArrayMetadata | undefined, parentId: string) {
+    let id = ''
+    if (tree) {
+      id = `${parentId}${tree.name}`
+    }
+    return (
+      tree && (
```
ARADDCC012, line 51: Instead of doing this consider adding the following condition above
```
if (!tree) {
  return null
}
```

src/common/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <TreeItem key={`${id}`} nodeId={`${id}`} label={`${tree.name}`}>
```
ARADDCC012, line 52: Can be simplified to `<TreeItem key={id} nodeId={id} label={tree.name}>`

src/common/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,79 @@
...
+          {tree.directories.map((child) => directoryTreeItem(child, id))}
+          {tree.files.map((file) => (
+            <TreeItem key={`${id}${file}`} nodeId={`${id}${file}`} label={`${file}`} />
```
ARADDCC012, line 55: `label={`${file}`}` can be simplified to `label={file}`

src/common/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return null
+  }
+
+  if (treeLoading) {
+    return <div>Loading files...</div>
+  }
+
+  return (
+    <TreeView
+      aria-label={`${fileType} file list`}
```
ARADDCC012, line 72: Does this actually get read by a screen reader or has this been put here for no reason?

server/services/deployment.ts
```diff
@@ -161,3 +161,27 @@ export async function deleteDeploymentsByVersion(user: UserDoc, version: Version
...
+          return directoryMetadata.children
+        }, accumulator)
+      }
+      return accumulator
+    }, [])
+  }
+  return {
+    id: 'bailotreecodefilesroot',
+    name: 'Code files',
+    children: [...result],
```
ARADDCC012, line 185: Can simplify to `children: result,`

server/services/deployment.ts
```diff
@@ -161,3 +161,27 @@ export async function deleteDeploymentsByVersion(user: UserDoc, version: Version
     })
   )
 }
+
+export const createTree = (files: string[]) => {
+  let result = []
```
ARADDCC012, line 166: Maybe rename `result` to `fileList` or something else more descriptive?

src/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,65 @@
...
+export default function DirectoryTreeView({
+  uuid,
+  version,
+  displayTree,
+}: {
+  uuid: string
+  version: string
+  displayTree: boolean
+}) {
+  const [treeResponse, setTreeResponse] = useState<DirectoryMetadata>()
```
ARADDCC012, line 20: Rename to `fileList`?

src/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [treeLoading, setTreeLoading] = useState(false)
+  const sendNotification = useNotification()
+
+  const getTreeResponse = useCallback(async () => {
```
ARADDCC012, line 24: Rename to `getFileList`?

src/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,65 @@
...
+    if (displayTree) {
+      setTreeLoading(true)
+      const response = await getEndpoint(`/api/v1/deployment/${uuid}/version/${version}/file-list`)
+      if (response.status === 200) {
+        const directoryArrayMetadata: DirectoryMetadata = await response.json()
```
ARADDCC012, line 29: Rename to `directoryMetadata`?

src/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }, [uuid, version, sendNotification, displayTree])
+
+  useEffect(() => {
+    if (version) {
+      getTreeResponse()
+    }
+  }, [version, getTreeResponse])
+
+  if (treeLoading) {
```
ARADDCC012, line 44: In order to prevent an empty tree from being rendered you can change this to `if (treeLoading || !treeResponse)` and change the line below from `{treeResponse && TreeRender(treeResponse)}` to `{TreeRender(treeResponse)}`

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+import DisabledElementTooltip from './common/DisabledElementTooltip'
+
+function RawModelExportItem({
+  deploymentUuid,
+  version,
+  uploadType,
+}: {
+  deploymentUuid: string
+  version: string
+  uploadType: string
```
ARADDCC012, line 21: Should this be typed as `ModelUploadType`?

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+}) {
+  const [displayTree, setDisplayTree] = useState(false)
+
+  let primaryButtonDisabledReason: string | undefined = 'No files to download'
```
ARADDCC012, line 25: Instead of making this `string | undefined` it would be cleaner to always treat it as a `string` and set it to `''` instead of `undefined`

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+  if (uploadType === ModelUploadType.Zip || uploadType === undefined) {
+    primaryButtonDisabledReason = undefined
```
ARADDCC012, line 27: `primaryButtonDisabledReason = ''`

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+  uploadType,
+}: {
+  deploymentUuid: string
+  version: string
+  uploadType: string
+}) {
+  const [displayTree, setDisplayTree] = useState(false)
+
+  let primaryButtonDisabledReason: string | undefined = 'No files to download'
+  if (uploadType === ModelUploadType.Zip || uploadType === undefined) {
```
ARADDCC012, line 26: `uploadType === undefined` is either an unnecessary check or we should update the prop definition to specify that this is an optional field.

If we go with the latter, we should always treat it as a `string` and default it to `''`. That would make then make this check `if (!uploadType || uploadType === ModelUploadType.Zip)`

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+    primaryButtonDisabledReason = undefined
+  }
+
+  const hideTree = () => {
+    setDisplayTree(false)
+  }
+
+  return (
+    <Box key={version}>
```
ARADDCC012, line 35: We shouldn't be providing a key here, it does nothing. I've double checked the consumers of this component to see if we should be providing keys there instead, but we do not. Therefore this `Box` can be swapped out for a fragment `<>`.

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <Stack spacing={2} direction='row' sx={{ p: 1 }}>
+          <DisabledElementTooltip
+            conditions={[primaryButtonDisabledReason === undefined ? '' : primaryButtonDisabledReason]}
+          >
+            <Button
+              variant='contained'
+              href={`/api/v1/deployment/${deploymentUuid}/version/${version}/raw/code`}
+              target='_blank'
+              data-test='downloadCodeFile'
+              disabled={!(primaryButtonDisabledReason === undefined)}
```
ARADDCC012, line 49: Can be simplified to `disabled={!!primaryButtonDisabledReason}`

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+            >
+              Download code files
+            </Button>
+          </DisabledElementTooltip>
+          <DisabledElementTooltip
+            conditions={[primaryButtonDisabledReason === undefined ? '' : primaryButtonDisabledReason]}
```
ARADDCC012, line 55: Can be simplified to `conditions={[primaryButtonDisabledReason]}`

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+              target='_blank'
+              data-test='downloadBinaryFile'
+              disabled={!(primaryButtonDisabledReason === undefined)}
+            >
+              Download binary file
+            </Button>
+          </DisabledElementTooltip>
+          <Divider flexItem orientation='vertical' />
+          <DisabledElementTooltip
+            conditions={[primaryButtonDisabledReason === undefined ? '' : primaryButtonDisabledReason]}
```
ARADDCC012, line 69: Can be simplified to `conditions={[primaryButtonDisabledReason]}`

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+          >
+            <Button
+              variant='outlined'
+              disabled={!(primaryButtonDisabledReason === undefined)}
```
ARADDCC012, line 73: Can be simplified to `disabled={!!primaryButtonDisabledReason}`

src/RawModelExportItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+              onClick={() => setDisplayTree(true)}
+            >
+              Display files
+            </Button>
+          </DisabledElementTooltip>
+        </Stack>
+      </Box>
+      <Dialog open={displayTree} onClose={hideTree}>
+        <DialogTitle>Showing files for {deploymentUuid}</DialogTitle>
+        <DialogContent sx={{ maxHeight: '500px', overflowX: 'auto', py: 2 }}>
```
ARADDCC012, line 83: Instead of limiting the height like this, we could instead let the dialog use as much height as it wants. This would help with viewing large code files.

types/interfaces.ts
```diff
@@ -254,3 +254,9 @@ export enum LogType {
   Request = 'request',
   Misc = 'misc',
 }
+
+export type DirectoryMetadata = {
```
ARADDCC012, line 258: Consider changing to `FileOrDirectoryMetadata` or similar, as this is now used for both files and directories.

src/RawModelExportItem.tsx
ARADDCC002: I've gone with making it optional. I believe I put this in (or did this somewhere else?) with the thought that existing / older models might not have this functionality as BuildOptions is relatively new.

server/services/deployment.ts
```diff
@@ -161,3 +161,25 @@ export async function deleteDeploymentsByVersion(user: UserDoc, version: Version
...
+
+export const createTree = (files: string[]) => {
+  let fileList = []
+  if (files) {
+    fileList = files.reduce((accumulator, path) => {
+      const names = path.split('/')
+      if (names) {
+        names.reduce((nestedAccumulator: FileOrDirectoryMetadata[], name, index) => {
+          let directoryMetadata = nestedAccumulator.find((item: any) => item.name === name)
+          if (!directoryMetadata) nestedAccumulator.push((directoryMetadata = { name, children: [], id: name + index }))
```
ARADDCC012, line 173: Name + index isnt guaranteed to be unique

src/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [fileList, setFileList] = useState<FileOrDirectoryMetadata>()
+  const [treeLoading, setTreeLoading] = useState(false)
+  const sendNotification = useNotification()
+
+  const getFileList = useCallback(async () => {
+    if (displayTree) {
+      setTreeLoading(true)
+      const response = await getEndpoint(`/api/v1/deployment/${uuid}/version/${version}/file-list`)
+      if (response.status === 200) {
+        const directoryMetadata: FileOrDirectoryMetadata = await response.json()
```
ARADDCC012, line 29: Update name to match the new type?

src/DirectoryTreeView.tsx
```diff
@@ -0,0 +1,65 @@
...
+    </TreeView>
+  )
+}
+
+function TreeRender(data: FileOrDirectoryMetadata) {
+  const { name, children, id } = data
+  if (Array.isArray(children)) {
+    return (
+      <TreeItem key={id} nodeId={id} label={name}>
+        {children.map((treeNode, _idx) => TreeRender(treeNode))}
```
ARADDCC012, line 60: I think `_idx` is unused and can probs be removed

src/RawModelExportItem.tsx
ARADDCC012: Good thinking 👍

PR #237, opened 2023-01-05
added ungoverned deployments

ARADDCC002:


pages/model/[uuid].tsx
```diff
@@ -252,6 +259,33 @@ function Model() {
...
+      name: ungovernedDeploymentName,
+      modelUuid: uuid,
+      initialVersionRequested: version.version,
+    })
+    if (response.status >= 400) {
+      let responseError = response.statusText
+      try {
+        responseError = `${response.statusText}: ${(await response.json()).message}`
+      } catch (e) {
+        setErrorMessage('No response from server')
```
Comment (a3957273) [inline pages/model/[uuid].tsx:278]: This `setErrorMessage` is pointless.  It instantly gets overrode on line 281.

pages/model/[uuid].tsx
```diff
@@ -320,6 +354,18 @@ function Model() {
...
+                <DisabledElementTooltip conditions={[!version.built ? 'Version needs to build.' : '']}>
+                  <MenuItem
+                    onClick={() => setUngovernedDialogOpen(true)}
+                    disabled={!version.built}
+                    data-test='submitUngovernedDeployment'
+                  >
+                    <ListItemIcon>
+                      <UploadIcon fontSize='small' />
+                    </ListItemIcon>
+                    <ListItemText>Request Ungoverned deployment</ListItemText>
```
Comment (a3957273) [inline pages/model/[uuid].tsx:366]: Not consistent capitalisation

pages/model/[uuid].tsx
```diff
@@ -554,6 +600,31 @@ function Model() {
...
       </Paper>
+      <Dialog open={ungovernedDialogOpen} onClose={handleUngovernedDialogClose}>
+        <DialogContent>
+          <Stack spacing={1}>
+            <Stack direction='row' spacing={2}>
+              <TextField
+                label='Deployment name'
+                variant='outlined'
+                value={ungovernedDeploymentName}
+                onChange={(event): void => setUngovernedDeploymentName(event.target.value)}
```
Comment (a3957273) [inline pages/model/[uuid].tsx:611]: I don't think you need the `: void` here.

server/routes/v1/deployment.ts
```diff
@@ -151,6 +152,98 @@ export const postDeployment = [
...
+
+    const versionArray = [version._id]
+
+    const owner = {
+      kind: 'user',
+      id: req.user.id,
+    }
+
+    const deployment = await createDeployment(req.user, {
+      schemaRef: 'n/a',
```
a3957273, line 200: schemaRef should not be `n/a`.  Maybe null?

a3957273 (changes requested):

pages/model/[uuid].tsx
```diff
@@ -554,6 +600,31 @@ function Model() {
...
+              <Button
+                disabled={ungovernedDeploymentName === ''}
+                variant='contained'
+                onClick={requestUngovernedDeployment}
+                autoFocus
+              >
+                Request
+              </Button>
+            </Stack>
+            <Typography sx={{ color: 'red' }} variant='caption'>
```
Comment (ARADDCC012) [inline pages/model/[uuid].tsx:622]: Consider using `theme.error.main` instead of `red`

pages/deployment/[uuid].tsx
```diff
@@ -225,6 +225,24 @@ export default function Deployment() {
...
               </Box>
             )}
+            {deployment.ungoverned && (
+              <Box>
+                <Alert
+                  severity='info'
+                  sx={{
+                    width: 'fit-content',
+                    m: 'auto',
+                    backgroundColor: '#0288d1',
```
Comment (ARADDCC012) [inline pages/deployment/[uuid].tsx:235]: Are you changing the colours here because the standard info colours don't work for us? I feel like we should be able to rely on our theme here. If we want to use another alert with the same style somewhere else we'll have to copy/paste these colours. If we use the theme then we don't have to do anything aside from providing the severity.

pages/model/[uuid].tsx
Comment (ARADDCC002) [reply pages/model/[uuid].tsx:622]: For some reason even though theme seems to be correctly imported, it doesn't recognise "error". It definitely exists in the theme file, though...

Comment (ARADDCC002) [reply pages/model/[uuid].tsx:622]: Ignore above!

server/routes/v1/deployment.ts
```diff
@@ -151,6 +152,98 @@ export const postDeployment = [
...
+        {
+          code: 'version_not_found',
+          modelId: body.modelID,
+          version: body.initialVersionRequested,
+        },
+        `Unable to find version with name: '${body.initialVersionRequested}'`
+      )
+    }
+
+    const versionArray = [version._id]
```
a3957273, line 192: This is inaccurate.  There could be multiple versions above the initial version requested.

server/routes/v1/deployment.ts
```diff
@@ -151,6 +152,98 @@ export const postDeployment = [
...
+
+    const owner = {
+      kind: 'user',
```
a3957273, line 195: There's an enum somewhere around here for a kind of 'user'.

server/routes/v1/deployment.ts
```diff
@@ -151,6 +152,98 @@ export const postDeployment = [
...
+        contacts: {
+          owner: [owner],
+        },
+      },
+
+      owner,
+    })
+
+    deployment.managerApproved = ApprovalStates.Accepted
+    deployment.ungoverned = true
```
a3957273, line 220: Why not set these in the `createDeployment` function above?  Alter it to make them optional args.

a3957273 (changes requested):

a3957273 (changes requested):

a3957273 (approved):

ARADDCC012 (approved):

PR #317, opened 2023-01-19
Feature/helm update

ARADDCC013:
helm update
- added stroom
- added app
- updated ingress  (tested with AWS and k8)
- updates logger.ts
- updated values.yaml

### DIFF
--- a/helm/bailo/templates/bailo/bailo.configmap.yaml
+++ b/helm/bailo/templates/bailo/bailo.configmap.yaml
@@ -17,6 +17,11 @@ data:
         registryBucket: '{{ .Values.minio.registryBucket }}',
         createBuckets: {{ .Values.minio.createBuckets }},
       },
+      app: {
+        protocol: '{{ .Values.config.app.protocol }}',
+        host: '{{ .Values.config.app.host }}',
+        port: {{ .Values.config.app.port }},
+      },
       smtp: {
         host: '{{ .Values.config.smtp.host }}',
         port: {{ .Values.config.smtp.port }},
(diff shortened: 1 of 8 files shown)


ARADDCC013:
Also updated Chart.yaml as old charts do not exist. Also, tested this config on AWS and Openshift.

helm/bailo/templates/bailo/bailo.configmap.yaml
```diff
@@ -53,6 +67,9 @@ data:
           supportHref: 'mailto:{{ .Values.config.supportEmail }}?subject=Bailo%20Support',
           contactHref: 'mailto:{{ .Values.config.supportEmail }}?subject=Bailo%20Contact',
         },
+        help: {
+          documentationUrl: '{{ .Values.config.documentationLink }}',
```
a3957273, line 71: This value doesn't appear to be used in the application anymore.

helm/bailo/values.yaml
```diff
@@ -174,15 +176,16 @@ config:
   s2i:
     image: seldonio/seldon-core-s2i-python37:1.10.0
   supportEmail: "support@example.com"
+  documentationLink: "https://example.com/support"
```
a3957273, line 179: This value is not used anymore.

helm/bailo/values.yaml
```diff
@@ -174,15 +176,16 @@ config:
...
 
   ui:
     banner:
       enabled: "true"
```
a3957273, line 183: Let's default the banner to disabled?

a3957273 (changes requested):

helm/bailo/templates/bailo/bailo.configmap.yaml
ARADDCC013: removed

helm/bailo/values.yaml
ARADDCC013: removed

helm/bailo/values.yaml
ARADDCC013: set to disabled

a3957273 (approved):

PR #318, opened 2023-01-23
Feature/bai 342 replace all references of request with approval

ARADDCC012:
Refactored the following:
Request -> Approval
RequestType -> ApprovalCategory

Some of the logging has been updated as a result. This is likely to mean we will need to update our translations HS.

Also fixed some typos I noticed while CTRL+F'ing the codebase. These were all in .md files so won't have any knock-on effects.


data/approvals.ts
```diff
@@ -0,0 +1,38 @@
+import qs from 'qs'
+import useSWR from 'swr'
+import { Approval } from '../types/interfaces'
+import { fetcher } from '../utils/fetcher'
+
+export type ApprovalCategory = 'Upload' | 'Deployment'
+export type ReviewFilterType = 'user' | 'archived'
```
a3957273, line 7: Should this be `ApprovalFilterType`?

server/migrations/002-user-array-support.ts
```diff
@@ -52,27 +52,27 @@ export async function up() {
     await model.save()
   }
 
-  const requests = await RequestModel.find({})
+  const approvals = await ApprovalModel.find({})
```
a3957273, line 55: This is a migration, so shouldn't be altered in this way.

a3957273 (changes requested):

data/approvals.ts
ARADDCC012: I've not refactored "review"s, only "request"s. Do you also want to change all reviews to approvals?

server/migrations/002-user-array-support.ts
ARADDCC012: Yeah I wondered if we wanted to change it or not. It'll complain about `RequestModel` not existing anymore, but will revert any changes to this file.

a3957273 (approved):

PR #322, opened 2023-01-24
Add code explorer tab

a3957273:


src/model/CodeExplorer.tsx
```diff
@@ -0,0 +1,93 @@
+import Editor from '@monaco-editor/react'
```
ARADDCC002, line 1: Can we add a unit test for this component

src/model/CodeExplorer.tsx
```diff
@@ -0,0 +1,93 @@
...
+          changePath(nodes.id)
+        }
+      }}
+    >
+      {Array.isArray(nodes.children) ? nodes.children.map((node) => renderTree(node)) : null}
+    </TreeItem>
+  )
+
+  return (
+    <Box>
```
ARADDCC002, line 68: As we're not using any properties, this can be a fragment `<></>`

src/model/CodeExplorer.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <Grid container spacing={2} sx={{ maxHeight: '100%' }}>
```
ARADDCC002, line 69: I'd personally move the "height: 80vh" property from the code viewer, and add it to the parent <Grid />, this means that the tree list _and_ the code viewer will be the same height, and the scroll property you added to tree list will appear. Right now there's no scroll bar because the tree list doesn't have or inherit any height property.

I'd probably also look into replacing <Grid /> with <Stack /> as it tends to be better for two dimensional arrangement, whereas <Grid /> is used for three dimensional arrangements.

src/model/CodeExplorer.tsx
ARADDCC002: Actually, can you just be removed altogether, as you've only got one top level component

src/model/CodeExplorer.tsx
ARADDCC002: I've done some tinkering, let me know if this works for you:

`return (
    <Grid container spacing={2} sx={{ maxHeight: '700px' }}>
      <Grid item xs={12} md={3} sx={{ maxHeight: 'inherit' }}>
        <TreeView
          aria-label='code navigator'
          defaultCollapseIcon={<ExpandMoreIcon />}
          defaultExpandIcon={<ChevronRightIcon />}
          defaultExpanded={['code/']}
          sx={{ flexGrow: 1, overflowY: 'auto', maxHeight: 'inherit' }}
        >
          {renderTree(tree)}
        </TreeView>
      </Grid>
      <Grid item xs={12} md={9} sx={{ scroll: 'auto' }}>
        <Editor
          height='700px'
          path={path}
          value={file}
          theme={theme.palette.mode === 'dark' ? 'vs-dark' : 'light'}
          options={{ fontSize: theme.typography.fontSize }}
        />
      </Grid>
    </Grid>
  )`

ARADDCC002: Apologies for the format..... your IDE should auto format that!

ARADDCC002 (approved):

ARADDCC002 (approved):

PR #325, opened 2023-01-25
untied versions from deployments

ARADDCC002:


pages/deployment/[uuid].tsx
```diff
@@ -96,37 +97,28 @@ export default function Deployment() {
...
+  const [imageName, setImageName] = useState('')
+  const [selectedImageTag, setSelectedImageTag] = useState('')
+
   const actionOpen = anchorEl !== null
 
   const { currentUser, isCurrentUserLoading, isCurrentUserError } = useGetCurrentUser()
   const { deployment, isDeploymentLoading, isDeploymentError, mutateDeployment } = useGetDeployment(uuid, true)
   const { uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
+  const model: ModelDoc = deployment?.model as ModelDoc
+  const { versions } = useGetModelVersions(model?.uuid)
```
Comment (ARADDCC012) [inline pages/deployment/[uuid].tsx:111]: If we can guarantee this is a ModelDoc in the UI and this is just a type issue then we can simplify it slightly `const { versions } = useGetModelVersions((deployment?.model as ModelDoc).uuid)`

pages/deployment/[uuid].tsx
```diff
@@ -163,6 +155,10 @@ export default function Deployment() {
     setAnchorEl(null)
   }
 
+  const onSelectedTagChange = (event: SelectChangeEvent) => {
+    setSelectedImageTag(event.target.value as string)
```
Comment (ARADDCC012) [inline pages/deployment/[uuid].tsx:159]: The type assertion here isn't needed as `event.target.value` is already considered a `string`

pages/deployment/[uuid].tsx
```diff
@@ -364,35 +297,55 @@ export default function Deployment() {
...
-              <p style={{ margin: 0 }}># The model is accessible at localhost:9000</p>
-            </Box>
-          </DialogContentText>
-        </DialogContent>
+        {versions && (
+          <DialogContent>
+            <Stack justifyContent='flex-start' alignItems='center' direction='row' spacing={2} sx={{ m: 2 }}>
+              <Typography>Select a version</Typography>
+              <Select value={selectedImageTag} label='Selected version' onChange={onSelectedTagChange}>
+                {versions
```
Comment (ARADDCC012) [inline pages/deployment/[uuid].tsx:305]: Unlikely to be a problem, but if we had a lot of versions then doing multiple array manipulations on ever render is more expensive than needed. Consider pulling this filter and map into a `useMemo`

pages/deployment/[uuid].tsx
Comment (ARADDCC002) [reply pages/deployment/[uuid].tsx:111]: Unfortunately returns undefined!

ARADDCC012 (approved):

PR #332, opened 2023-01-31
Feature/bai 407 change current metadata properties

ARADDCC002:


server/migrations/001_6-remove-current-metadata-from-model.ts
```diff
@@ -0,0 +1,22 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
ARADDCC012, line 1: This comment can now be removed too

ARADDCC012 (dismissed):

ARADDCC012 (approved):

PR #353, opened 2023-02-16
added a latest version tag to the version list on the deployment comm…

ARADDCC002:
…ands drop down menu


a3957273 (approved):

a3957273 (changes requested):
Couple of comments

a3957273 (approved):

PR #360, opened 2023-02-25
Test bun install speeds

a3957273:


Jarred-Sumner:
Are the unit tests failing due to a bug in bun? Or is it an incompatible version update of a dependency as a result of the lockfile changing (since bun doesn't read other package manager lockfiles)?

a3957273:
Hey @Jarred-Sumner.  I think these issues are caused by `bun` not using `package-lock.json` to identify Node dependency versions.  We were relying on it to lock some of our dependencies to non-esm versions.

We have a separate pull request in this project to enable esm-only dependencies (e.g. to get the latest modules created by sindresorhus) and are hoping that it should fix this pull request too.

That is to say, the errors here are caused by our mistakes alone and not the fault of the `bun` tool, which during testing shows some remarkable improvements (reducing cold dependency install times from 300 seconds down to 12 seconds).

PR #364, opened 2023-03-06
Feature/bai 464 email deployment owner on model version deletion

JR40159:


.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
+  // Use IntelliSense to learn about possible attributes.
```
ARADDCC012, line 2: You can't put comments in JSON

server/services/model.ts
```diff
@@ -100,11 +101,11 @@ export async function deleteModel(user: UserDoc, modelId: ModelId) {
 export async function removeVersionFromModel(user: UserDoc, version: VersionDoc) {
   // Deletes model if no versions left
   const model = await ModelModel.findById(version.model)
-
   if (!model) {
     throw NotFound({ modelId: version.model }, 'Unable to find model to remove.')
   }
 
+  // Remove model record
```
ARADDCC012, line 108: I don't think this comment is correct. It's removing the version of a model, not the model itself.

I'd also argue that the comment isn't needed as the function name describes this clearly, but I don't feel too strongly about that.

server/services/deployment.ts
```diff
@@ -72,15 +77,6 @@ export async function findDeploymentsByModel(user: UserDoc, model: ModelDoc, opt
   return filterDeployment(user, await deployments)
 }
 
-export async function findDeploymentsByVersion(user: UserDoc, version: VersionDoc, opts?: GetDeploymentOptions) {
```
ARADDCC012, line 75: Was this removed for being dead/unused code?

JR40159: Yes, I couldn't find any references to it and I don't think we include version info in deployment records anymore

.vscode/launch.json
JR40159: Those comments were auto-generated but I can remove them

server/services/deployment.spec.ts
```diff
@@ -51,4 +61,42 @@ describe('test deployment service', () => {
...
+
+    await emailDeploymentOwnersOnVersionDeletion(deployments, testVersion)
+
+    expect(emailClient.sendEmail).toHaveBeenCalledTimes(deployments.length * userList.length)
+  })
+
+  test('that we can email deployment owners about changes for a single deployment', async () => {
+    const deployments: DeploymentDoc[] = [testDeployment]
+    const spy: any = jest.spyOn(entityUtils, 'getUserListFromEntityList')
+    const userList: Array<any> = [
```
ARADDCC012, line 82: Instead of making new test data here you could construct the array of users by using the existing test data in `utils/test/testModels`. E.g. `testUser` and `testUser2`.

server/services/deployment.spec.ts
```diff
@@ -51,4 +61,42 @@ describe('test deployment service', () => {
...
+    ]
+    spy.mockReturnValue(userList)
+
+    await emailDeploymentOwnersOnVersionDeletion(deployments, testVersion)
+
+    expect(emailClient.sendEmail).toHaveBeenCalledTimes(deployments.length * userList.length)
+  })
+
+  test('that we can email deployment owners about changes for a single deployment', async () => {
+    const deployments: DeploymentDoc[] = [testDeployment]
```
ARADDCC012, line 80: Unfortunately our test data isn't fully representative of the full DeploymentDoc and won't have all the properties of a full document. It's probably best to leave it to infer the type itself at the moment, which will just be `any[]` as we haven't typed our test data.

server/services/deployment.spec.ts
```diff
@@ -51,4 +61,42 @@ describe('test deployment service', () => {
...
+    const spy: any = jest.spyOn(entityUtils, 'getUserListFromEntityList')
+    const userList: Array<any> = [
+      { name: 'Alice', email: 'alice@email.com' },
+      { name: 'Bob', email: 'bob@email.com' },
+    ]
+    spy.mockReturnValue(userList)
+
+    await emailDeploymentOwnersOnVersionDeletion(deployments, testVersion)
+
+    // Assert that for each deployment, each owner has been send an email about that deployment
```
ARADDCC012, line 90: Typo `send` -> `sent`

server/routes/v1/version.ts
```diff
@@ -337,6 +341,17 @@ export const deleteVersion = [
 
     await version.delete(user._id)
 
+    // Send email to owners of affected deployments
+    const model = await ModelModel.findById(version.model)
+    if (model) {
```
ARADDCC012, line 346: I would think that we should expect to find the model 100% of the time so it would be a valid reason to throw an error if we don't:
```
if (!model) {
  throw NotFound({ code: 'model_not_found', version.model }, `Unable to find model '${version.model}'`)
}

const deployments = await findDeploymentsByModel(user, model)
if (deployments.length > 0) {
  emailDeploymentOwnersOnVersionDeletion(deployments, version)
}
```

server/routes/v1/version.spec.ts
```diff
@@ -1,22 +1,46 @@
...
 import {
+  authenticatedDeleteRequest,
   authenticatedGetRequest,
   authenticatedPostRequest,
   authenticatedPutRequest,
   validateTestRequest,
 } from '../../utils/test/testUtils'
 
+jest.mock('../../services/model', () => ({
+  removeVersionFromModel: jest.fn(() => Promise.resolve()),
```
ARADDCC012, line 21: Could be written as `removeVersionFromModel: jest.fn(async () => undefined)`

server/routes/v1/version.spec.ts
```diff
@@ -1,22 +1,46 @@
...
+  serializedModelFields: () => ({
+    mandatory: ['_id', 'uuid', 'latestVersion.metadata.highLevelDetails.name', 'schemaRef'],
+  }),
+}))
+
+jest.mock('../../services/deployment', () => ({
+  findDeploymentsByModel: jest.fn(() => Promise.resolve({ irrelevant: 'content' })),
```
ARADDCC012, line 28: You can just make this a jest function for now as you're setting the return values in your tests (which is the right way to do it anyway) so these values will be overwritten at that point. `findDeploymentsByModel: jest.fn()`

The same applies to the other functions you're mocking in this file. We only tend to set a specific return value at this level is we want the same return value for all tests in this file. Otherwise just setting `jest.fn()` here and using `mockReturnValueOnce` or `mockImplementationOnce` in each test that uses them is the preferred approach.

server/routes/v1/version.spec.ts
```diff
@@ -27,6 +51,65 @@ describe('test version routes', () => {
     expect(res.body.version).toBe('1')
   })
 
+  test('that when we delete a version, we send an email to the relvant deployment owners', async () => {
+    await ModelModel.create(testModel)
+    const spy: any = jest.spyOn(approvalService, 'deleteApprovalsByVersion')
+    spy.mockReturnValue({ irrelevant: 'content' })
+    const updateImplementation = (deploymentService.findDeploymentsByModel as jest.Mock).mockImplementationOnce(() => [
```
ARADDCC012, line 58: Doesn't really matter but you could instead use `mockReturnValueOnce(['deployment'])`. `mockImplementation` is more useful if you want to change the functionality of something in a test (which is pretty uncommon)

server/services/deployment.spec.ts
```diff
@@ -14,8 +19,13 @@ import {
 
 const userDoc = new UserModel(testUser)
 
+jest.mock('../utils/smtp', () => ({
+  sendEmail: jest.fn(() => Promise.resolve()),
```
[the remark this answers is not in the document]
ARADDCC012: My mistake, I did check this but got it wrong

.vscode/launch.json
a3957273: VS: Code always using 'JSONC', which allows comments and trailing commas.  Can keep the comments in if they're useful.

server/routes/v1/version.spec.ts
JR40159: After discussion, decided to leave as is to avoid repeated code in multiple tests

ARADDCC012 (approved):
Approved but needs to wait for the ESM work to be merged in before this can be merged

PR #367, opened 2023-03-13
Bugfix/bai 529 latest version attribute of a model is not being updated when deleting the latest

JR40159:


ARADDCC012 (approved):
Approved but will need to wait for ESM changes to be merged

PR #378, opened 2023-04-06
Convert to using 'vitest' for all unit testing

a3957273:


ARADDCC002:
Running "npm run test" locally is causing it to hang at the end.

a3957273:
> Running "npm run test" locally is causing it to hang at the end.

This is intentional.  In development mode it will watch all files and rerun only the related tests when they change.  E.g. if you edit the 'createEmotionCache' file it will understand that it only needs to run the `createEmotionCache.spec.ts` file.

In CI it will end as expected.

This can be circumvented by running `vitest run` to force it to run only once.

backend/src/routes/v1/version.spec.ts
```diff
@@ -105,8 +105,7 @@ describe('test version routes', () => {
   //   expect(deploymentService.emailDeploymentOwnersOnVersionDeletion).not.toBeCalled()
```
JR40159, line 105: Have you checked whether this tests can now work with vitest?

a3957273: I have not.  I would rather get this vitest change merged in before trying to re-enable things that were similarly broken in Jest.

backend/src/routes/v1/approvals.spec.ts
```diff
@@ -1,25 +1,20 @@
...
-    const approval = await import('../../services/approval.js')
-    jest.unstable_mockModule('../../services/approval.js', () => {
-      return {
-        ...approval,
-        readNumApprovals: jest.fn(() => 1),
-      }
-    })
-
-    const { authenticatedGetRequest } = await import('../../utils/test/testUtils.js')
+    vi.spyOn(approval, 'readNumApprovals').mockImplementation(async () => 1)
```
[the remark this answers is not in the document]
a3957273: Resolved.

JR40159:
> Running "npm run test" locally is causing it to hang at the end.

I don't mind this when running locally but not sure about the best way to handle this in the automated pipeline. Can a flag be passed into npm test so that it doesn't hang?

backend/src/routes/v1/version.spec.ts
JR40159: I had them working with Jest with in my original PR, we couldn't get them working with ESM-Jest

backend/src/routes/v1/approvals.spec.ts
```diff
@@ -1,25 +1,20 @@
...
-    const approval = await import('../../services/approval.js')
-    jest.unstable_mockModule('../../services/approval.js', () => {
-      return {
-        ...approval,
-        readNumApprovals: jest.fn(() => 1),
-      }
-    })
-
-    const { authenticatedGetRequest } = await import('../../utils/test/testUtils.js')
+    vi.spyOn(approval, 'readNumApprovals').mockReturnValue(Promise.resolve(1))
```
JR40159, line 11: mockresolvedvalue()?

JR40159 (approved):

PR #385, opened 2023-04-09
Add 'entity display' component generic

a3957273:


backend/src/scripts/addFakeData.ts
```diff
@@ -1,5 +1,5 @@
 import { findAndUpdateUser } from '../services/user.js'
-import { connectToMongoose, disconnectFromMongoose } from '../utils/database.js' assert { type: 'json' }
```
ARADDCC002, line 2: This has always been needed before, is there a reason we don't need it anymore?

a3957273: We do need it on JSON files, but `../utils/database.js` is not a JSON file.

ARADDCC002 (approved):

PR #419, opened 2023-04-11
Deep clone deployment data so deployment can be submitted twice

a3957273:
Before it would add misc data to the highLevelDetails field due to it being a reference clone.  With clone deep we prevent that from happening.


a3957273:
The same bug actually occurs in upload.  I've changed the fix so that it makes 'getStepsData' a deep clone.  It's not used in any performance important locations (only when rendering the submission page / submitting a form)

JR40159 (approved):

PR #410, opened 2023-04-10
Fixes next import type, fixes all eslint issues.

a3957273:

PR #432, opened 2023-04-15
Fix all backend linting issues, move serializers to serializers.ts

a3957273:


a3957273:
Approved under https://github.com/gchq/Bailo/pull/410

PR #436, opened 2023-04-15
Fix sorting by model on deployments page

a3957273:


backend/src/services/model.ts
```diff
@@ -37,7 +37,7 @@ export async function findModelByUuid(user: UserDoc, uuid: string, opts?: GetMod
 
 export async function findModelById(user: UserDoc, id: string | Types.ObjectId | ModelDoc, opts?: GetModelOptions) {
   let model = ModelModel.findById(id)
-  if (opts?.populate) model = model.populate('latestVersion', 'metadata')
```
JR40159, line 40: Does the same need to be done in ```findModels``` and ```findModelByUuid```?

a3957273: Resolved.

PR #468, opened 2023-05-04
bugfix/backend-node-version

ARADDCC012:


a3957273 (approved):

JR40159:
For consistency, is it worth also downgrading the node version for the frontend images?

ARADDCC012:
> For consistency, is it worth also downgrading the node version for the frontend images?

If we want to do this we can do so in a separate PR

PR #470, opened 2023-05-05
Bugfix/fix 500 response when getting images

JR40159:


ARADDCC013 (approved):
Tests pass in github and OS. Looks good.

PR #489, opened 2023-05-12
Add documentation for useful commands

a3957273:


frontend/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@bailo/frontend",
   "private": true,
   "scripts": {
-    "dev": "next dev -- --turbo",
+    "dev": "next dev",
```
m19671, line 5: Should this change be in this PR?

m19671 (approved):

frontend/package.json
a3957273: Yeah, it doesn't seem to work with several aspects of our application.

PR #507, opened 2023-05-23
latestVersion now always populates on marketplace models

ARADDCC002:


a3957273:
Instead of defaulting to populating everything always, we should instead enable it specifically in the instances we want it.

PR #515, opened 2023-05-25
Add export functionality to Bailo

a3957273:


backend/src/scripts/generateAccessToken.ts
```diff
@@ -2,7 +2,7 @@ import { getAccessToken } from '../routes/v1/registryAuth.js'
 import { consoleLog } from '../utils/logger.js'
 
 async function main() {
-  const model = 'nginx'
+  const model = 'test_deployment/test_yolo'
```
m19671, line 5: Not sure this should be this?

backend/src/routes/v1/export.ts
```diff
@@ -0,0 +1,63 @@
...
+  ensureUserRole('user'),
+  bodyParser.json(),
+  async (req: Request, res: Response) => {
+    // Get model params
+    const { uuid, version, deploymentId } = req.params
+
+    // Set .zip extension to request header
+    res.set('Content-disposition', `attachment; filename=${uuid}.zip`)
+    res.set('Content-Type', 'application/zip')
+    // res.set('Cache-Control', 'private, max-age=604800, immutable')
```
m19671, line 26: remove unused comments?

backend/src/types/types.ts
```diff
@@ -341,3 +343,16 @@ export interface Version {
 }
 
 export type VersionDoc = Version & Document<any, any, Version>
+
+export interface fsLayer {
+  blobSum: string
+}
+
+export interface ImageManifest {
```
m19671, line 351: Can probably get rid of this now as it won't be accurate or used

backend/src/routes.ts
```diff
@@ -100,4 +101,6 @@ server.get('/api/v1/admin/logs', ...getApplicationLogs)
 server.get('/api/v1/admin/logs/build/:buildId', ...getItemLogs)
 server.get('/api/v1/admin/logs/approval/:approvalId', ...getItemLogs)
 
+server.get('/api/v1/export/:uuid/:deploymentId/version/:version', ...exportModel)
```
a3957273, line 104: We can get the uuid from the deploymentId, so we don't actually need the uuid value here.

PR #552, opened 2023-06-12
Action K8s

ARADDCC013:
configured use Secuirty context
removed host config for Mongo, now managed my _helper.tpl


ARADDCC013:
updated cypress config to use correct base_url. Cypress does rum but fails due to timeout during model upload test. I have increased timeout to 60 mins but the job is still failing.

ARADDCC002 (approved):

PR #556, opened 2023-06-14
added condition to stop approving from API

TT38665:
A condition was added to the backend, in deployment.ts to make sure that any approved versions of a model can be seen, when requesting a deployment and to also make sure that any unapproved models are not approved for deployment from the API.


backend/src/routes/v1/deployment.ts
```diff
@@ -94,6 +94,24 @@ export const postDeployment = [
...
+    const version = await findVersionById(req.user, (model as ModelDoc).latestVersion)
+
+    if (!version) {
+      throw NotFound(
+        { code: 'version_not_found', versionId: (model as ModelDoc).latestVersion },
+        'Unable to find version'
+      )
+    }
+
+    if (version.managerApproved === ApprovalStates.Accepted && version.reviewerApproved === ApprovalStates.Accepted) {
```
ARADDCC012, line 108: Could this be simplified to the following?
```
    if (version.managerApproved !== ApprovalStates.Accepted || version.reviewerApproved !== ApprovalStates.Accepted) {
      throw BadReq({}, 'Latest version of this model must be fully approved before making a deployment')
    }
```

ARADDCC002 (approved):

PR #558, opened 2023-06-16
Feature/add tag filter component and reorganise marketplace

ARADDCC002:


frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+export default function ExploreModels() {
+  const [group, setGroup] = useState<ListModelType>('all')
+  const [filter, setFilter] = useState('')
+  const debouncedFilter = useDebounce(filter, 250)
+
+  const { models, isModelsError, mutateModels } = useListModels(group, debouncedFilter)
+
+  const theme = useTheme()
+
+  const error = MultipleErrorWrapper(`Unable to load marketplace page`, {
```
ARADDCC012, line 34: This is what causes multiple Wrapper components to be rendered on the page when we display errors. It's probably worth fixing this all in one go in a separate ticket, but just an FYI that any further uses of `MultipleErrorWrapper` like this is adding more work to that ticket.

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const handleFilterChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setFilter(e.target.value)
+  }
+
+  const onFilterSubmit = (e: React.FormEvent) => {
+    e.preventDefault()
+  }
+
+  const updateSelectedTags = (selectedTags: string[]) => {
+    console.log(selectedTags)
```
ARADDCC012, line 48: Is it worth adding TODOs to any placeholder code like this?

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  const updateSelectedType = (selectedType: string) => {
+    if (selectedType) {
+      switch (selectedType) {
+        case 'My models':
```
ARADDCC012, line 54: Would be better to use enums rather than "magic" strings. That way we only have to update the enum if we change these in future and those changes will propagate across the app.

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+          setGroup('user')
```
ARADDCC012, line 55: Same here, would be better to use an enum for groups

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+  return (
+    <Wrapper title='Explore Models' page='marketplace'>
+      <Paper
+        component='form'
+        onSubmit={onFilterSubmit}
+        sx={{
+          p: '2px 4px',
+          display: 'flex',
+          alignItems: 'center',
+          width: '70%',
```
ARADDCC012, line 76: setting `width` like this seems a bit pointless when we're setting a `maxWidth` below

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+        default:
+          setGroup('all')
+      }
+      mutateModels()
+    }
+  }
+
+  return (
+    <Wrapper title='Explore Models' page='marketplace'>
+      <Paper
```
ARADDCC012, line 69: Is there a reason you're making a custom textfield here instead of just using the pre-built mui `TextField`?

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+      </Paper>
+      <Grid container>
+        <Grid item sm={8} xs={12}>
+          <Paper sx={{ py: 2, px: 4 }}>
+            <Box sx={{ mb: 2, borderBottom: 1, borderColor: 'divider' }} data-test='indexPageTabs'>
+              <Tabs value={'bailo'}>
+                <Tab label={`Models ${models ? `(${models.length})` : ''}`} value='bailo' />
+              </Tabs>
+            </Box>
+            <Box data-test='modelListBox'>
```
ARADDCC012, line 94: Can just be a `div` if you're not using mui utilities

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+        <InputBase sx={{ ml: 1, flex: 1 }} placeholder='Filter Models' value={filter} onChange={handleFilterChange} />
+        <IconButton color='primary' type='submit' sx={{ p: '10px' }} aria-label='filter'>
+          <SearchIcon />
+        </IconButton>
+      </Paper>
+      <Grid container>
+        <Grid item sm={8} xs={12}>
+          <Paper sx={{ py: 2, px: 4 }}>
+            <Box sx={{ mb: 2, borderBottom: 1, borderColor: 'divider' }} data-test='indexPageTabs'>
+              <Tabs value={'bailo'}>
```
ARADDCC012, line 90: This looks like WIP code. Worth adding a TODO to highlight it?

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+                <Tab label={`Models ${models ? `(${models.length})` : ''}`} value='bailo' />
+              </Tabs>
+            </Box>
+            <Box data-test='modelListBox'>
+              {models &&
+                models.map((model: Model, index: number) => {
+                  const latestVersion = model.latestVersion as Version
+                  return (
+                    <Box key={model.uuid}>
```
ARADDCC012, line 99: This could be a `<Fragment key={model.uuid}>` as you're not using any mui utilities

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+                        ))}
+                      </Stack>
+                      {index !== models.length - 1 && (
+                        <Box sx={{ borderBottom: 1, borderColor: 'divider', marginBottom: 2 }} />
+                      )}
+                    </Box>
+                  )
+                })}
+
+              {models?.length === 0 && <EmptyBlob data-test='emptyModelListBlob' text='No models here' />}
```
ARADDCC012, line 123: Instead of doing `models?` here it would be better to fallback to an error message before getting to this point. That way we can treat `models` as a guaranteed array from that point onwards.

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+            </Box>
+          </Paper>
+        </Grid>
+        <Grid sm={4} xs={12}>
+          <Stack>
+            <Box sx={{ px: 2 }}>
+              <TagSelector
+                label='Tags'
+                multiple
+                tags={[
```
ARADDCC012, line 133: TODO?

frontend/src/Wrapper.tsx
```diff
@@ -175,6 +178,10 @@ export default function Wrapper({ title, page, children }: WrapperProps): ReactE
     },
   })
 
+  function displayBetaTag() {
```
ARADDCC012, line 181: Is it better if you do:
```
const betaAdornment = <Box component="span" sx={{ marginLeft: 1, color: '#cecece', fontSize: 15 }}>beta</span>
```
Then you could use it below like:
```
              <Link
                href='/'
                color='inherit'
                underline='none'
                style={{ color: 'inherit', textDecoration: 'inherit', fontSize: '1.25rem', fontWeight: 500 }}
              >
                Bailo
                {router.asPath.includes('/beta') && betaAdornment}
              </Link>
```
I've renamed to betaAdornment because tags are a different thing within Bailo and it's worth avoiding any confusion.

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <Button onClick={toggleExpansion}>{expanded ? 'Show less' : 'Show more...'}</Button>
+      )}
+    </>
+  )
+}
+
+type TagProps = {
+  tag: string
+  handleChange: (value: string) => void
+  size?: 'small' | 'medium'
```
ARADDCC012, line 77: Super duper minor, but as this is always going to be the same type as the prop of the same name in `TagSelectorProps` you could change this to be:
```
size?: TagSelectorProps['size']
```
That way we only need to change the type in once place if we ever decide we want to.

frontend/src/theme.ts
```diff
@@ -10,7 +10,28 @@ declare module '@mui/material/styles/createPalette' {
   }
 }
 
+const removeElevation = {
```
ARADDCC012, line 13: 🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

frontend/pages/beta/index.tsx
ARADDCC002: This is actually left from the original marketplace, so I don't know why `width `and `maxWith `are set. I've actually gone and removed `width`.

frontend/pages/beta/index.tsx
ARADDCC002: Again left over from the original marketplace. I'm not 100% sure why it's like that. I'm probably going to leave as-is for now just for continuity with the previous page. If you're set on having it changed though I'm sure we can update it.

frontend/pages/beta/index.tsx
ARADDCC002: I think according to our design mock-up it should be clear as-is

frontend/pages/beta/index.tsx
ARADDCC002: I've added an error before but TS is still complaining unfortunately

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type TagProps = {
+  tag: string
+  handleChange: (value: string) => void
+  size?: TagSelectorProps['size']
+  active: boolean
+}
+
+function Tag({ tag, handleChange, size, active }: TagProps) {
```
ARADDCC012, line 85: Same here, it looks like this is likely to represent more than just tags. Maybe rename to `SelectableChip` or `ActiveChip`?

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,161 @@
+import SearchIcon from '@mui/icons-material/Search'
```
a3957273, line 1: Similar comment to before about maybe combining these into one import.  And the below imports being made absolute by using '@' instead of '../..'

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+        default:
+          setGroup(MarketPlaceModelGroup.ALL)
+      }
+      mutateModels()
+    }
+  }
+
+  return (
+    <Wrapper title='Explore Models' page='marketplace'>
+      <TextField
```
a3957273, line 77: I feel like the original flat white search bar looks better than this.

frontend/pages/beta/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+              </Tabs>
+            </Box>
+            <div data-test='modelListBox'>
+              {(!models || models.length === 0) && <EmptyBlob data-test='emptyModelListBlob' text='No models here' />}
+              {models &&
+                models.map((model: Model, index: number) => {
+                  const latestVersion = model.latestVersion as Version
+                  return (
+                    <Fragment key={model.uuid}>
+                      <Link href={`/model/${model.uuid}`} passHref legacyBehavior>
```
a3957273, line 110: We shouldn't be adding legacy behaviour to new pages.

a3957273 (changes requested):

frontend/pages/beta/index.tsx
ARADDCC002: Left over from the original page, removed.

frontend/pages/beta/index.tsx
ARADDCC002: @ Isn't working for me!

frontend/pages/model/[uuid].tsx
```diff
@@ -87,7 +86,6 @@ function Model() {
   const { versionAccess } = useGetVersionAccess(version?._id)
 
   const hasUploadType = useMemo(() => version !== undefined && !!version.metadata.buildOptions?.uploadType, [version])
-  const sendNotification = useNotification()
```
Comment (m19671) [inline frontend/pages/model/[uuid].tsx:90]: Why has this been taken out - seems out of scope for this ticket?

Comment (ARADDCC002) [reply frontend/pages/model/[uuid].tsx:90]: I don't remember doing this... I'll revert!

PR #562, opened 2023-06-16
Feature/bai 677 create initial new model page

ARADDCC002:


frontend/pages/beta/create-new-model.tsx
```diff
@@ -0,0 +1,97 @@
+import Box from '@mui/material/Box'
```
a3957273, line 1: Should we start joining up these import statements?  Maybe switch this up to:

```
import { Box, Button, Card, Divider, ... } from '@mui/material'
```

frontend/pages/beta/create-new-model.tsx
```diff
@@ -0,0 +1,97 @@
...
+import Divider from '@mui/material/Divider'
+import FormControlLabel from '@mui/material/FormControlLabel'
+import Radio from '@mui/material/Radio'
+import RadioGroup from '@mui/material/RadioGroup'
+import Stack from '@mui/material/Stack'
+import TextField from '@mui/material/TextField'
+import Typography from '@mui/material/Typography'
+import { useState } from 'react'
+
+import Wrapper from '../../src/Wrapper'
```
a3957273, line 13: Should we use absolute imports for this, so that we can keep our templates the same no matter what level the page is at?  An import like...

```
import Wrapper from '@src/Wrapper'
```

Might be better.

frontend/pages/beta/create-new-model.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+type NewModelData = {
+  teamName: string
+  modelName: string
+  description: string
+  accessibility: 'public' | 'private'
+}
+
+export default function CreateNewModelWizard() {
```
a3957273, line 22: The URL for this page should probably be something simple like `/model/new` or `/model/create`.  The name might want to match with that.

a3957273 (changes requested):

frontend/pages/beta/create-new-model.tsx
ARADDCC002: Isn't working unfortunately, think we need to update the config somewhere but I agree this change would be nice

frontend/pages/beta/create-new-model.tsx
```diff
@@ -0,0 +1,97 @@
...
+import Radio from '@mui/material/Radio'
+import RadioGroup from '@mui/material/RadioGroup'
+import Stack from '@mui/material/Stack'
+import TextField from '@mui/material/TextField'
+import Typography from '@mui/material/Typography'
+import { useState } from 'react'
+
+import Wrapper from '../../src/Wrapper'
+
+type NewModelData = {
```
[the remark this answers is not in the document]
ARADDCC002: Moved NewModelData to types, but I'm unsure what you mean about the API call? For now I wasn't going to bother adding any, and come back to these pages once the API is either fully mocked or finished.

frontend/pages/beta/create-new-model.tsx
```diff
@@ -0,0 +1,97 @@
...
+      teamName,
+      modelName,
+      description,
+      accessibility,
+    }
+    // TODO - after new model page is implemented, forward this data
+    console.log(formData)
+  }
+  return (
+    <Wrapper title='Create a new Model' page='marketplace'>
```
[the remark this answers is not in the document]
ARADDCC002: Do you mean the overall form? It's based on the wireframe designs, I think it just looks odd because the wireframes are much more compact width wise whereas you're viewing it on a wider resolution. I've discussed this with people before but we actually agreed having it wider looks better. 

My reservation about reducing the width is that it might look odd if some pages have a smaller width than others; this way we can keep it consistent. Thoughts?

PR #624, opened 2023-07-14
Bai 559 create UI element to enable the import of the model

k25831:
Model description now extends further across the page, reducing the dead space. New spacing dimensions approved by Alex.


m19671:
This work has been completed on the wrong branch.

PR #627, opened 2023-07-18
Implement POST /v2/models and GET /v2/model

a3957273:


backend/src/middleware/validate.ts
```diff
@@ -0,0 +1,36 @@
+import { NextFunction, Request, Response } from 'express'
+import z, { AnyZodObject, ZodError } from 'zod'
+
+import { BadReq } from '../utils/v2/error.js'
+
+export function validate(schema: AnyZodObject) {
```
JR40159, line 6: Unit Tested?

backend/src/middleware/validate.ts
```diff
@@ -0,0 +1,36 @@
...
+      })
+      return next()
+    } catch (err: unknown) {
+      const error = err as ZodError
+      throw BadReq(error.issues[0].message, { errors: error.issues })
+    }
+  }
+}
+
+export function parse<T extends AnyZodObject>(req: Request, schema: T): z.infer<T> {
```
JR40159, line 22: Unit Tested?

backend/src/routes/v2/model/getModel.spec.ts
```diff
@@ -0,0 +1,21 @@
+// Some bad parts
+import '../../../utils/test/testUtils.js'
+
+import { describe, expect, test, vi } from 'vitest'
+
+import { createFixture, testGet } from '../../../utils/v2/test/routes.js'
+import { getModelSchema } from './getModel.js'
+
+describe('routes > model > getModel', () => {
```
JR40159, line 9: Tests for when parse (BadReq) and getModelByID (Forbidden, NotFound) throw errors?

backend/src/routes/v2/model/getModel.ts
```diff
@@ -0,0 +1,36 @@
+import bodyParser from 'body-parser'
+import { Request, Response } from 'express'
+import { z } from 'zod'
+
+import { parse } from '../../../middleware/validate.js'
+import { ModelInterface } from '../../../models/v2/ModelModel.js'
+import { getModelById } from '../../../services/v2/model.js'
+
+export const getModelSchema = z.object({
+  params: z.object({
```
JR40159, line 10: When importing 'z' could we give it useful name? 'z.object' and the like isn't the nicest to read

backend/src/services/v2/model.ts
```diff
@@ -0,0 +1,39 @@
+import authorisation, { ModelAction } from '../../external/v2/authorisation/index.js'
```
JR40159, line 1: Unit Tests?

backend/src/services/v2/model.ts
JR40159, line 1: Unit Tests?

backend/src/external/v2/authorisation/index.ts
JR40159, line 1: Test that the different (mocked) config returns the auth class you expect it to and an unrecognised 'kind' throws the expected error.

backend/src/utils/v2/config.ts
```diff
@@ -0,0 +1,21 @@
+import bunyan from 'bunyan'
+import _config from 'config'
+
+export interface Config {
+  connectors: {
+    user: {
+      kind: 'silly' | string
```
JR40159, line 7: Maybe 'name' or 'type' rather than 'kind'?

backend/src/middleware/expressLogger.ts
```diff
@@ -0,0 +1,85 @@
...
+    return false
+  }
+
+  // Internal and context are both unknown, so they could be any value.
+  // Thus, we don't check them here.
+
+  return true
+}
+
+export async function expressErrorHandler(err: unknown, req: Request, res: Response, _next: NextFunction) {
```
JR40159, line 67: Having this as part of the 'expressLogger.ts' file seems a bit random to me, could this not just be in it's own errorHandler file?

JR40159: Unit test logic around internal- logger gets called with the internal content but that content is not returned by the function

ARADDCC002 (approved):

ARADDCC002 (dismissed):
Unapproved

backend/src/routes/v2/model/getModel.ts
ARADDCC002: Whilst I'm generally a proponent of using useful names for things, this is how it's exported by default and also how it's used in the official documentation, so I think it's acceptable to keep it as 'z' (although I'm happy to have it changed to 'zod' if I'm overruled by the majority).

ARADDCC002:
I see that we've moved all of the tests to a new directory. Given that number of individual files (especially under routes) I think this is a good change. However, should we keep this consistent with the frontend? Or are we OK with having frontend unit tests in the same directories as the code they're testing? Not overly bothered, just thought I'd raise it. :)

backend/src/routes/v2/model/getModel.ts
JR40159: I'm happy to yield on this one- might just be a nice to have for me personally.

backend/src/routes.ts
```diff
@@ -136,4 +145,65 @@ server.get('/api/v1/admin/logs', ...getApplicationLogs)
 server.get('/api/v1/admin/logs/build/:buildId', ...getItemLogs)
 server.get('/api/v1/admin/logs/approval/:approvalId', ...getItemLogs)
 
-server.use('/api', expressErrorHandler)
+// V2 APIs
+
+server.post('/api/v2/models', ...postModel)
```
JR40159, line 150: We're only allowing users to POST a single model? Should this endpoint be '/api/v2/model'?

backend/test/services/model.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { describe, expect, test, vi } from 'vitest'
+
+import { createModel, getModelById } from '../../src/services/v2/model.js'
+
+const mocks = vi.hoisted(() => {
```
JR40159, line 5: For readability and organisation could we use vi.hoisted to define a factory mock for each of our mocks rather than putting methods for different mocks in a single object called 'mocks'?

backend/test/services/model.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+describe('services > model', () => {
+  test('createModel > good', async () => {
+    await createModel({} as any, {} as any)
+
+    expect(mocks.save).toBeCalled()
+    expect(mocks.Model).toBeCalled()
+  })
+
+  test('getModelById > good', async () => {
```
JR40159, line 42: Tests for model not found and user not authorized?

backend/test/routes/model/getModel.spec.ts
```diff
@@ -0,0 +1,21 @@
+// Some bad parts
+import '../../../src/utils/test/testUtils.js'
+
+import { describe, expect, test, vi } from 'vitest'
+
+import { getModelSchema } from '../../../src/routes/v2/model/getModel.js'
+import { createFixture, testGet } from '../../../src/utils/v2/test/routes.js'
+
+describe('routes > model > getModel', () => {
```
JR40159, line 9: Still no mocked parse function? I think we agreed to mock it for unit tests and write system integration tests that would include the parse function (zod functionality) and the middleware

backend/src/external/v2/authorisation/index.ts
```diff
@@ -0,0 +1,34 @@
...
+} as const
+
+export type ModelActionKeys = (typeof ModelAction)[keyof typeof ModelAction]
+
+export abstract class BaseAuthorisationConnector {
+  abstract userModelAction(user: UserDoc, model: ModelDoc, action: ModelActionKeys): Promise<boolean>
+}
+
+let authConnector: undefined | BaseAuthorisationConnector = undefined
+export function getAuthorisationConnector() {
```
JR40159, line 18: Unit test this function?

backend/config/default.cjs
```diff
@@ -218,4 +218,18 @@ module.exports = {
   experimental: {
     v2: false,
   },
+
+  connectors: {
+    user: {
+      kind: 'silly',
+    },
+
+    authorisation: {
```
JR40159, line 227: Could this be combined with the config that relates to the chosen auth provider. Using oauth as an example, to use oauth you have to both add the oauth section of config and also have to set the oath connector in a different place in the config. We have oauth.enabled, if that's set to true then I shouldn't also need to set the connectors.authorisation to be oauth.

backend/src/external/v2/authorisation/index.ts
JR40159, line 1: Could we rename the 'external' folder? If we're having the section of config called 'connectors' perhaps we use that to be consistent?

backend/src/models/v2/User.ts
```diff
@@ -0,0 +1,32 @@
+import { Document, model, Schema } from 'mongoose'
+
+// This interface stores information about the properties on the base object.
+// It should be used for plain object representations, e.g. for sending to the
+// client.
+export interface UserInterface {
+  // Do not store user role information on this object.  This information
+  // should be stored in an external corporate store.
+  dn: string
```
JR40159, line 9: Are we expecting every auth provider to include a 'dn' in users' information?

backend/src/routes.ts
```diff
@@ -86,7 +88,9 @@ if (config.oauth.enabled) {
 }
 
 server.use(getUser)
+
 server.use(expressLogger)
+server.use('/api/v2', expressLoggerV2)
```
JR40159, line 93: Worth including this in the if statement with the V2 routes?

backend/test/routes/model/getModel.spec.ts
JR40159, line 9: No unhappy test cases?

backend/test/routes/model/getModel.spec.ts
```diff
@@ -0,0 +1,21 @@
+// Some bad parts
+import '../../../src/utils/test/testUtils.js'
+
+import { describe, expect, test, vi } from 'vitest'
+
+import { getModelSchema } from '../../../src/routes/v2/model/getModel.js'
+import { createFixture, testGet } from '../../../src/utils/v2/test/routes.js'
```
JR40159, line 7: FYI now on main we have a testUtils directory in the test directory

backend/src/models/v2/User.ts
a3957273: We can't use the word `id` as it's restricted by Mongoose.  Thus, I figure `dn` is a common enough phrase to get the point across of having a single globally unique identifier per person?  I'm happy to change it if a better name is around that doesn't relate to LDAP servers.

backend/src/utils/test/testUtils.ts
```diff
@@ -45,7 +45,7 @@ vi.mock('../../utils/config.js', () => {
         },
       },
       experimental: {
-        v2: false,
+        v2: true,
```
JR40159, line 48: This config should only be being used by V1 tests?

backend/test/routes/model/getModel.spec.ts
```diff
@@ -0,0 +1,21 @@
+// Some bad parts
+import '../../../src/utils/test/testUtils.js'
```
JR40159, line 2: You shouldn't need this for V2 route unit tests as we have testUtils/routes.ts. To mock the config and user implementation you just need to include the mock in the route test files without any factory and it will use the mocked implementation in the __mocks__ folder.

backend/test/connectors/authorisation/authorisation.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+import { getAuthorisationConnector } from '../../../src/connectors/v2/authorisation/index.js'
+
+const configMock = vi.hoisted(() => ({
+  connectors: {
+    authorisation: {
+      kind: 'silly',
+    },
+  },
+}))
+vi.mock('../../../src/utils/v2/config.js', () => ({
```
JR40159, line 12: Use the mock in __mocks__ rather than defining another one here?

backend/test/connectors/authorisation/authorisation.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  __esModule: true,
+  default: configMock,
+}))
+
+describe('connectors > authorisation', () => {
```
JR40159, line 17: Connectors tests - can you add an unhappy test for each connector for the default error case?

JR40159 (approved):

PR #632, opened 2023-07-19
Bai 650 add warning message for max model upload size

P20179:
Added functionality so that if the user tries to upload a combination of code, binary and docker files (or any one of these) that exceeds the maximum upload size we allow, they are presented with an error message which tells them the issue, what the upper limit is, and they cannot proceed to select upload.
The amount set is configurable and at the time of this PR is set within backend>config>default.cjs to 50 GB.
This PR also includes an improved Loading state which displays a loading spinner rather than a Wrapper, which was confusing to the user and not visually attractive.

### DIFF
--- a/backend/src/utils/config.ts
+++ b/backend/src/utils/config.ts
@@ -147,6 +147,8 @@ export interface Config {
       name: string
       image: string
     }>
+    //max model size is calculated in gigabytes
+    maxModelSizeGB: number
   }
 }
 
--- a/frontend/src/Form/RenderFileTab.tsx
+++ b/frontend/src/Form/RenderFileTab.tsx
@@ -34,7 +34,7 @@ export default function RenderFileTab({ step, splitSchema, setSplitSchema }: Ren
     if (isUiConfigError) setError(isUiConfigError.message)
     else if (!uiConfig) setError('Failed to load ui config')
     else if (totalFileSize > convertGigabytesToBytes(uiConfig.maxModelSizeGB))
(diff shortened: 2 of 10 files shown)


backend/src/utils/config.ts
```diff
@@ -147,6 +147,8 @@ export interface Config {
       name: string
       image: string
     }>
+    //max model size is calculated in gigabytes
+    maxModelSizeGB: number
```
ARADDCC012, line 151: You should also update the `UiConfig` type in `frontend/types/types.ts` to match this.

frontend/src/Form/RenderFileTab.tsx
```diff
@@ -1,69 +1,112 @@
...
+  useEffect(() => {
+    const codeSize = state.code ? state.code.size : 0
+    const binarySize = state.binary ? state.binary.size : 0
+    const dockerSize = state.docker ? state.docker.size : 0
+    setTotalFileSize(codeSize + binarySize + dockerSize)
+  }, [state.binary, state.code, state.docker])
+
+  useEffect(() => {
+    if (isUiConfigError) setError(isUiConfigError.message)
+    else if (!uiConfig) setError('Failed to load ui config')
```
ARADDCC012, line 35: Super minor, but change to `'Failed to load UI config'`?

ARADDCC012 (approved):

P20179:
Need to pull over most recent changes

PR #42, opened 2022-03-08
Bump Express to v5 to fix error handling

a3957273:
To fix asynchronous error handling, we need to use Express v5.  It has already gone through 8 alpha versions to improve stability and has been stabilizing for 7 years.  I'm not sure why it hasn't been released yet.

PR #644, opened 2023-07-31
Feature/bai 820 integrate beta mocked api

ARADDCC002:


frontend/src/common/ChipSelector.tsx
```diff
@@ -23,22 +27,15 @@ type TagSelectorProps = {
...
   tags,
-  onChange,
+  setSelectedTags,
+  selectedTags,
   multiple,
   size = 'medium',
   expandThreshold = 5,
 }: TagSelectorProps): ReactElement {
-  const [selectedTags, setSelectedTags] = useState<string[]>([])
+  //const [selectedTags, setSelectedTags] = useState<string[]>([])
```
ARADDCC012, line 36: Commented out code

frontend/src/common/MultiFileInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export default function MultiFileInput({
+  label,
+  setFiles,
+  files,
+  accepts,
+  disabled,
+  fullWidth = false,
+}: {
```
ARADDCC012, line 24: Can this inline type be pulled out into a reusable type? I find this generally helps with readability and also allows us to easily reference the type if needed.
```
type MultiFileInputProps = {
  label: string
  setFiles: Dispatch<SetStateAction<File[] | undefined>>
  files?: File[]
  accepts?: string
  disabled?: boolean
  fullWidth?: boolean
}
```

frontend/src/common/MultiFileInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+      setFiles(updatedFileList)
+    }
+  }
+
+  function handleFileChange(event: ChangeEvent<HTMLInputElement>) {
+    const newFiles = event.target.files ? Array.from(event.target.files) : []
+    if (newFiles) {
+      if (files) {
+        const updatedFiles = newFiles.concat(
+          files.filter((file) => newFiles.every((newFile) => newFile.name != file.name))
```
ARADDCC012, line 46: Should be using strict equality checks `newFile.name !== file.name`

ARADDCC012, line 46: Slight performance improvement could be made here. Instead of always checking `every` new file we can exit early using `some`:
```
newFiles.concat(
          files.filter((file) => !newFiles.some((newFile) => newFile.name === file.name))
)
```

frontend/src/common/MultiFileInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+        )
+        setFiles(updatedFiles)
+      } else {
+        setFiles(newFiles)
+      }
+    }
+  }
+
+  return (
+    <div style={{ width: fullWidth ? '100%' : 'unset' }}>
```
ARADDCC012, line 56: For consistency can we use a `Box` here with `sx`?

frontend/src/common/MultiFileInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+          multiple
+          accept={accepts}
+          disabled={disabled}
+        />
+        <Button sx={{ width: fullWidth ? '100%' : 'unset' }} variant='outlined' component='span' disabled={disabled}>
+          {label}
+        </Button>
+      </label>
+      <Box sx={{ mt: 2 }}>
+        <Box>
```
ARADDCC012, line 72: This `Box` isn't needed

frontend/src/model/DraftNewReleaseDialog.tsx
```diff
@@ -13,35 +13,56 @@ import {
...
+import useNotification from '../../src/common/Snackbar'
+import { ReleaseInterface } from '../../types/types'
+import MultiFileInput from '../common/MultiFileInput'
 
-export default function DraftNewReleaseDialog({ open, handleClose }: { open: boolean; handleClose: () => void }) {
+export default function DraftNewReleaseDialog({
+  open,
+  handleClose,
+  modelId,
+}: {
```
ARADDCC012, line 28: Can this inline type be pulled out into a reusable type?
```
type DraftNewReleaseDialogProps = {
  open: boolean
  handleClose: () => void
  modelId: string
}
```

frontend/src/model/DraftNewReleaseDialog.tsx
```diff
@@ -13,35 +13,56 @@ import {
...
+  open: boolean
+  handleClose: () => void
+  modelId: string
+}) {
   const [releaseName, setReleaseName] = useState('')
   const [semanticVersion, setSemanticVersion] = useState('')
   const [releaseNotes, setReleaseNotes] = useState('')
   const [isMinorRelease, setIsMinorRelease] = useState(false)
-  const [artefact, setArtefact] = useState<Artefact | undefined>()
+  const [artefacts, setArtefacts] = useState<File[]>()
```
ARADDCC012, line 37: As we've typed this as an array we should default to an empty array
```
const [artefacts, setArtefacts] = useState<File[]>([])
```

frontend/src/model/DraftNewReleaseDialog.tsx
```diff
@@ -13,35 +13,56 @@ import {
...
 
-  function onSubmit(event: FormEvent<HTMLFormElement>) {
+  const sendNotification = useNotification()
```
ARADDCC012, line 39: I believe we're trying to move away from snackbars, see my other comment about this for more info.

frontend/src/model/ModelReleaseDisplay.tsx
```diff
@@ -110,7 +110,7 @@ export default function ModelReleaseDisplay({
                   spacing={2}
                 >
                   <Link href='/beta'>{file}</Link>
-                  <Typography variant='caption'>123GB</Typography>
+                  {/* <Typography variant='caption'>123GB</Typography> */}
```
ARADDCC012, line 113: Commented out code

frontend/src/model/ModelReleaseDisplay.tsx
```diff
@@ -122,7 +122,7 @@ export default function ModelReleaseDisplay({
                   spacing={2}
                 >
                   <Link href='/beta'>{image}</Link>
-                  <Typography variant='caption'>123GB</Typography>
+                  {/* <Typography variant='caption'>123GB</Typography> */}
```
ARADDCC012, line 125: Commented out code

frontend/src/model/beta/Releases.tsx
```diff
@@ -1,18 +1,24 @@
...
   const [latestRelease, setLatestRelease] = useState<string>('')
   const [openDraftNewRelease, setOpenDraftNewRelease] = useState(false)
 
+  const { releases, isReleasesLoading } = useGetReleasesForModelId(model.id)
+
   useEffect(() => {
-    if (model && model.releases) {
-      setLatestRelease(model.releases.sort((a, b) => a.semver < b.semver)[0].semver)
+    if (model && releases) {
+      const sortedReleases = releases.sort((a, b) => (a.semver < b.semver ? 1 : b.semver > a.semver ? -1 : 0))
```
ARADDCC012, line 18: With complex sorting logic like this I find it's usually pretty unreadable. Therefore I prefer to pull the logic out into a function and give it a descriptive name so that it's easy to read at a glance. I believe this is sorting by version descending, so have assumed that for the following example, but please change that if I'm incorrect.

Also, nested ternaries are always worth avoiding again for readability.
```
// Probably worth pulling this out into a utils file
const sortByReleaseVersionDescending = (a: ReleaseInterface, b: ReleaseInterface) => {
  if (a.semver < b.semver) return 1
  if (b.semver > a.semver) return -1
  return 0
}

const sortedReleases = releases.sort(sortByReleaseVersionDescending)
```

frontend/src/model/beta/Releases.tsx
```diff
@@ -26,21 +32,27 @@ export default function Releases({ model }: { model: any }) {
             Draft new Release
           </Button>
         </Box>
+        {isReleasesLoading && <Typography sx={{ m: 4 }}>{`Fetching releases for model ${model.name}`}</Typography>}
+
+        {releases && releases.length === 0 && <EmptyBlob text={`No releases found for model ${model.name}`} />}
```
ARADDCC012, line 37: If you make releases default to an empty array you can simplify this logic
```
{releases.length === 0 && <EmptyBlob text={`No releases found for model ${model.name}`} />}
```

frontend/src/model/beta/Releases.tsx
```diff
@@ -26,21 +32,27 @@ export default function Releases({ model }: { model: any }) {
...
-              <ModelReleaseDisplay
-                key={release.semver}
-                modelId={model.id}
-                release={release}
-                latestRelease={latestRelease}
-              />
-            )
-          )
-        })}
+        {releases &&
```
ARADDCC012, line 39: If you make releases default to an empty array you can simplify this logic by removing the initial `releases &&`

frontend/src/model/beta/Releases.tsx
```diff
@@ -26,21 +32,27 @@ export default function Releases({ model }: { model: any }) {
...
+          releases
+            .sort((a, b) => (a.semver < b.semver ? 1 : b.semver > a.semver ? -1 : 0))
```
ARADDCC012, line 41: This can be written in a more readable way and is also duplicating the same sorting logic. See my other comments for more info.

frontend/src/model/beta/Releases.tsx
ARADDCC012, line 18: This sort is being performed multiple times in this component. Consider using `useMemo` to ensure we're only doing it when needed.
```
const sortedReleases = useMemo(() => releases.sort(sortByReleaseVersionDescending), [releases])
```

frontend/types/types.ts
```diff
@@ -358,13 +358,26 @@ export enum MarketPlaceModelGroup {
...
+export const ModelVisibility = {
+  Private: 'private',
+  Public: 'public',
+} as const
+
+export type ModelVisibilityKeys = (typeof ModelVisibility)[keyof typeof ModelVisibility]
+
+export interface ModelInterface {
+  id: string
+
```
ARADDCC012, line 375: Super minor, but as type definitions are already very vertical, could we save some scrolling by removing the empty lines in type definitions?

ARADDCC012 (changes requested):

frontend/pages/beta/index.tsx
```diff
@@ -137,25 +138,24 @@ export default function ExploreModels() {
             <div data-test='modelListBox'>
               {(!models || models.length === 0) && <EmptyBlob data-test='emptyModelListBlob' text='No models here' />}
               {models &&
```
ARADDCC012, line 140: `models &&` can be removed if you default to an array as suggested in another comment

frontend/pages/beta/index.tsx
```diff
@@ -137,25 +138,24 @@ export default function ExploreModels() {
...
+                          {model.name}
                         </MuiLink>
                       </Link>
                       <Typography variant='body1' sx={{ marginBottom: 2 }}>
-                        {latestVersion.metadata.highLevelDetails.modelInASentence}
+                        {model.description}
                       </Typography>
                       <Stack direction='row' spacing={1} sx={{ marginBottom: 2 }}>
-                        {latestVersion.metadata.highLevelDetails.tags.map((tag: string) => (
+                        {/* {model.tags.map((tag: string) => (
```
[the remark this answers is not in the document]
ARADDCC002: This is there because we don't know how model tags are going to be implemented yet. I've added a TODO comment above.

frontend/src/model/ModelReleaseDisplay.tsx
ARADDCC002: This is being kept in until we know how we're going to handle sizes for artefacts (they're just returned as strings), I'll make a todo comment.

frontend/pages/beta/index.tsx
```diff
@@ -15,27 +14,47 @@ import {
...
+import { MarketPlaceModelGroup, MarketPlaceModelSelectType, ModelInterface } from '../../types/types'
 import useDebounce from '../../utils/hooks/useDebounce'
 
 export default function ExploreModels() {
   const [group, setGroup] = useState<ListModelType>('all')
   // TODO - fetch model tags from API
   const [filter, setFilter] = useState('')
+  const [selectedLibrary, setSelectedLibrary] = useState('')
+  const [selectedTask, setSelectedTask] = useState('')
+  const [selectedType, setSelectedType] = useState('')
```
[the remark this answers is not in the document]
ARADDCC002: This uses the reusable ChipSelector component which just returns strings - we then use the `useEffect` on line #42 to over the options to check if it matches. 

For the `useState` can we type as `const [selectedType, setSelectedType] = useState<MarketPlaceModelSelectType | string>('')`?

backend/config/default.cjs
```diff
@@ -212,10 +212,10 @@ module.exports = {
         image: 'seldonio/seldon-core-s2i-python37:1.10.0',
       },
     ],
-    maxModelSizeGB: 50
+    maxModelSizeGB: 50,
   },
 
   experimental: {
-    v2: false,
+    v2: true,
```
a3957273, line 219: We don't want to enable v2 by default, consider adding this in a `local.cjs` file that isn't committed to git.

a3957273 (changes requested):

frontend/actions/release.ts
```diff
@@ -0,0 +1,36 @@
+import axios from 'axios'
+import useSWR from 'swr'
+
+import { ReleaseInterface } from '../types/types'
+import { handleAxiosError } from '../utils/axios'
```
ARADDCC012, line 5: Am I being daft or does this file not exist?

frontend/pages/beta/model/[id].tsx
```diff
@@ -0,0 +1,37 @@
...
+import Overview from '../../../src/model/beta/Overview'
+import Releases from '../../../src/model/beta/Releases'
+import Wrapper from '../../../src/Wrapper.beta'
+
+export default function Model() {
+  const router = useRouter()
+  const { id }: { id?: string } = router.query
+  const { model, isModelLoading, isModelError } = useGetModel(id)
+
+  function myFunction() {
```
Comment (ARADDCC012) [inline frontend/pages/beta/model/[id].tsx:16]: TODO?

frontend/src/model/DraftNewReleaseDialog.tsx
```diff
@@ -117,10 +130,13 @@ export default function DraftNewReleaseDialog({ open, handleClose }: { open: boo
...
             variant='contained'
             type='submit'
-            disabled={!semanticVersion || !artefact || !releaseNotes || !releaseName || !isValidSemver(semanticVersion)}
+            disabled={
+              !semanticVersion || !artefacts || !releaseNotes || !releaseName || !isValidSemver(semanticVersion)
+            }
           >
             Create Release
           </Button>
+          <MessageAlert message={errorMessage} severity='error' linkText='More info' />
```
ARADDCC012, line 139: `linkText` is only required if `href` is also being provided. This should be enforce via our MessageAlert props type.
```
type PartialMessageAlertProps =
  | {
      linkText: string
      href: string
    }
  | {
      linkText: never
      href: never
    }

type MessageAlertProps = {
  message?: string
  severity?: AlertProps['severity']
} & PartialMessageAlertProps
```

frontend/src/model/beta/Releases.tsx
```diff
@@ -1,18 +1,32 @@
...
 
+  const { releases, isReleasesLoading } = useGetReleasesForModelId(model.id)
+
+  const sortByReleaseVersionDescending = (a: ReleaseInterface, b: ReleaseInterface) => {
+    if (a.semver < b.semver) return 1
+    if (b.semver > a.semver) return -1
+    return 0
+  }
+
+  const sortedReleases = releases.sort(sortByReleaseVersionDescending)
```
ARADDCC012, line 23: Sorry another thing I should've spotted the first time around. Can we wrap this with a `useMemo` so we're not sorting on every render?
```
const sortedReleases = useMemo(() => releases.sort(sortByReleaseVersionDescending), [releases])
```

frontend/src/model/beta/Releases.tsx
```diff
@@ -1,18 +1,32 @@
...
 import ModelReleaseDisplay from '../ModelReleaseDisplay'
 
-export default function Releases({ model }: { model: any }) {
+export default function Releases({ model }: { model: ModelInterface }) {
   const [latestRelease, setLatestRelease] = useState<string>('')
   const [openDraftNewRelease, setOpenDraftNewRelease] = useState(false)
 
+  const { releases, isReleasesLoading } = useGetReleasesForModelId(model.id)
+
+  const sortByReleaseVersionDescending = (a: ReleaseInterface, b: ReleaseInterface) => {
```
ARADDCC012, line 17: This sort of thing is nice to have available as a reusable piece of logic. Consider pulling out into a sorting utils file.

frontend/src/model/beta/Releases.tsx
```diff
@@ -26,21 +40,25 @@ export default function Releases({ model }: { model: any }) {
...
-        {model.releases.map((release) => {
-          return (
-            release.name && (
-              <ModelReleaseDisplay
-                key={release.semver}
-                modelId={model.id}
-                release={release}
-                latestRelease={latestRelease}
-              />
+        {sortedReleases &&
```
ARADDCC012, line 47: `sortedReleases &&` isn't needed as it'll always be an array

frontend/src/model/beta/Releases.tsx
ARADDCC012: Unresolved, still outstanding

frontend/src/model/beta/Releases.tsx
```diff
@@ -26,21 +40,25 @@ export default function Releases({ model }: { model: any }) {
...
+          sortedReleases.map((release) => {
```
ARADDCC012, line 48: Again sorry for not noticing this the first time around, but this logic is effectively a `reduce` written as a `map`. I prefer to avoid this as if `release.name` was a falsy value that React will render (e.g. `0`) then you can end up with unintended elements being rendered on the page. That might not have any tangible effect in this case, but I believe it's a pattern we should avoid.

 I also always prefer to put this sort of logic in a `useMemo` so that we're not iterating through the array on every render. Consider changing to:
```
  const modelReleaseDisplays = useMemo(
    () =>
      sortedReleases.reduce<ReactElement[]>((releaseDisplays, release) => {
        if (release.name) {
          releaseDisplays.push(
            <ModelReleaseDisplay
              key={release.semver}
              modelId={model.id}
              release={release}
              latestRelease={latestRelease}
            />
          )
        }
        return releaseDisplays
      }, []),
    [latestRelease, model.id, sortedReleases]
  )
```

ARADDCC012 (approved):

PR #663, opened 2023-08-03
Bugfix/bai 570 approved first by technical reviewer

TT38665:


backend/src/routes/v1/approvals.ts
```diff
@@ -132,6 +137,18 @@ export const postApprovalResponse = [
...
       await version.save()
 
+      if (approval.approvalType === ApprovalTypes.Manager && choice === ApprovalStates.Accepted) {
+        if (version.reviewerApproved !== ApprovalStates.Accepted) {
+          throw BadReq(
+            {
+              code: 'bad_request_type',
+              version: versionDoc.managerApproved,
+            },
+            'Managers cannot approve model versions until it has been accepted by a technical reviewer'
```
ARADDCC002, line 147: Can we get these role names from the schema itself? "Manager" and "Technical Reviewer" might not be the right definition if the schema changes.

frontend/cypress/e2e/cypress.config.js
```diff
@@ -0,0 +1,9 @@
+const { defineConfig } = require("cypress");
```
ARADDCC002, line 1: Is there a reason this is included?

frontend/cypress/e2e/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
ARADDCC002, line 1: Is there a reason this is included?

frontend/cypress/e2e/cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
ARADDCC002, line 1: Is there a reason this is included?

frontend/pages/review.tsx
```diff
@@ -106,136 +152,129 @@ function ApprovalList({ category, filter }: { category: ApprovalCategory; filter
...
-            </Grid>
-            <Grid item xs={12} md sx={{ display: 'flex' }}>
-              <Box ml='auto' my='auto'>
-                {approvalObj.status !== 'No Response' && (
-                  <Chip label={approvalObj.status} color={approvalObj.status === 'Accepted' ? 'success' : 'error'} />
+                    <Stack direction='row' spacing={2}>
+                      <Chip color='primary' label={approvalObj.approvalType} size='small' />
+                      <Chip color='primary' label={`Version: ${approvalObj.version?.version}`} size='small' />
+                      <Box sx={{ mt: 'auto !important', mb: 'auto !important' }}>
+                        <Typography variant='body1'>
```
ARADDCC002, line 176: Can we remove `variant='body1'` as `body1` is the default value for `variant`

frontend/pages/review.tsx
```diff
@@ -106,136 +152,129 @@ function ApprovalList({ category, filter }: { category: ApprovalCategory; filter
...
+                  <>
+                    <Link href={`/deployment/${approvalObj.deployment?.uuid}`} passHref legacyBehavior>
+                      <MuiLink
+                        variant='h5'
+                        sx={{ fontWeight: '500', textDecoration: 'none', color: theme.palette.secondary.main }}
+                      >
+                        {approvalObj.deployment?.metadata?.highLevelDetails?.name}
+                      </MuiLink>
+                    </Link>
+                    <Typography variant='body1'>
```
ARADDCC002, line 200: See previous comment about removing `variant='body1'`

frontend/pages/review.tsx
```diff
@@ -106,136 +152,129 @@ function ApprovalList({ category, filter }: { category: ApprovalCategory; filter
...
+                    color='secondary'
+                    variant='outlined'
+                    onClick={() => changeState('Declined', approvalObj)}
+                    sx={{ mr: 1 }}
+                    disabled={approvalObj.status === 'Declined'}
+                  >
+                    Reject
+                  </Button>
+                  <DisabledElementTooltip
+                    conditions={[!canApprove() ? 'technical reviewer needs to appove this model first' : '']}
```
ARADDCC002, line 239: See previous comment about getting the actual schema title for "technical reviewer" in case it differs in other schemas

ARADDCC002 (changes requested):

frontend/pages/review.tsx
```diff
@@ -35,29 +36,70 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+
+      {approvals.length === 0 && (
+        <EmptyBlob text={`All done! No ${getUploadCategory(category)} are waiting for approval.`} />
+      )}
+    </Paper>
+  )
+}
+
+type ApprovalItemProps = {
+  approvalObj: any
```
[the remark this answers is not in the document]
TT38665: It clashes with the approval useState on line 86.  Do you have any suggestions?

frontend/cypress/e2e/cypress.config.js
TT38665: Removed

frontend/cypress/e2e/cypress/fixtures/example.json
TT38665: Removed

frontend/cypress/e2e/cypress/support/commands.js
TT38665: Removed

frontend/cypress/e2e/cypress/support/e2e.js
```diff
@@ -0,0 +1,20 @@
+// ***********************************************************
```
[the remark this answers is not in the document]
TT38665: Removed

frontend/pages/review.tsx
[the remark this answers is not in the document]
TT38665: Done

backend/src/routes/v1/approvals.ts
TT38665: Done

frontend/pages/review.tsx
TT38665: Resolved

frontend/pages/review.tsx
TT38665: Resolved

backend/src/routes/v1/approvals.ts
```diff
@@ -212,3 +236,14 @@ export const postApprovalResponse = [
     })
   },
 ]
+async function getManagerAndReviewer(metadata: ModelMetadata) {
```
a3957273, line 239: This should not be a route, but should instead probably be in the schema service?

backend/src/routes/v1/approvals.ts
```diff
@@ -132,6 +139,20 @@ export const postApprovalResponse = [
       version[field] = choice as ApprovalStates
       await version.save()
 
+      const { managerTitle, reviewerTitle } = await getManagerAndReviewer(version.metadata)
```
a3957273, line 142: There's no reason to call this every single time.  Only call it when an error is made by putting it inside the two `if` statements.

backend/src/services/schema.ts
```diff
@@ -1,9 +1,9 @@
-import SchemaModel, { Schema } from '../models/Schema.js'
-import { SchemaType } from '../types/types.js'
+import SchemaModel from '../models/Schema.js'
+import { Schema, SchemaType } from '../types/types.js'
 import config from '../utils/config.js'
 import logger from '../utils/logger.js'
 
-export async function findSchemaByRef(ref: string) {
+export async function findSchemaByRef(ref: string): Promise<Schema | null> {
```
a3957273, line 6: Is this type hint you've added here necessary?  It should work just as well without it.  This return type isn't `Schema` anyway I think, it's `SchemaDoc`?

a3957273 (changes requested):

frontend/pages/review.tsx
ARADDCC012: The default value of `variant` is `body1`, so this can just be
```
<Typography>
    {approvalObj.version?.metadata?.highLevelDetails?.modelInASentence}
</Typography>
```

frontend/pages/review.tsx
```diff
@@ -35,29 +37,86 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
 }
 
 function ApprovalList({ category, filter }: { category: ApprovalCategory; filter: ApprovalFilterType }) {
+  const { approvals, isApprovalsLoading, isApprovalsError, mutateApprovals } = useListApprovals(category, filter)
+
+  console.log(approvals)
```
ARADDCC012, line 42: This can be removed

frontend/pages/review.tsx
```diff
@@ -35,29 +37,86 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+
+      {approvals.length === 0 && (
+        <EmptyBlob text={`All done! No ${getUploadCategory(category)} are waiting for approval.`} />
+      )}
+    </Paper>
+  )
+}
+
+type ApprovalItemProps = {
+  approval: any
```
ARADDCC012, line 81: This should have a proper type, we should avoid `any` unless absolutely necessary. Can we use `Approval` as a type here?

frontend/pages/review.tsx
```diff
@@ -35,29 +37,86 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+
+  if (isApprovalsLoading || !approvals) {
+    return <Paper sx={{ mt: 2, mb: 2 }} />
+  }
+  return (
+    <Paper sx={{ mt: 2, mb: 2, pb: 2 }}>
+      <Typography sx={{ p: 3 }} variant='h4'>
+        {category === 'Upload' ? 'Models' : 'Deployments'}
+      </Typography>
+      {approvals.map((approval: any) => (
```
ARADDCC012, line 63: This `any` isn't necessary, it should automatically be picking up the `Approval` type from `useListApprovals`

frontend/pages/review.tsx
```diff
@@ -35,29 +37,86 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
 }
 
 function ApprovalList({ category, filter }: { category: ApprovalCategory; filter: ApprovalFilterType }) {
+  const { approvals, isApprovalsLoading, isApprovalsError, mutateApprovals } = useListApprovals(category, filter)
```
ARADDCC012, line 40: `approvals` can be `undefined` here due to our implementation in `useListApprovals`. We should update line 25 inside that hook (see `frontend/data/approvals.ts`) from:
```
approvals: data?.approvals
```
to
```
approvals: data ? data.approvals : []
```

That way we can guarantee that we're dealing with an array, instead of having to check for `undefined`.

frontend/pages/review.tsx
```diff
@@ -106,136 +169,127 @@ function ApprovalList({ category, filter }: { category: ApprovalCategory; filter
...
-                    >
-                      {approvalObj.deployment?.metadata?.highLevelDetails?.name}
-                    </MuiLink>
-                  </Link>
-                  <Typography variant='body1'>
-                    Requesting deployment of{' '}
-                    <Link href={`/model/${approvalObj.deployment?.model?.uuid}`} passHref legacyBehavior>
-                      <MuiLink sx={{ fontWeight: '500', textDecoration: 'none', color: theme.palette.secondary.main }}>
-                        {approvalObj.deployment?.model?.latestVersion?.metadata?.highLevelDetails?.name}
+    <>
```
ARADDCC012, line 173: This fragment can be removed

frontend/pages/review.tsx
```diff
@@ -106,136 +169,127 @@ function ApprovalList({ category, filter }: { category: ApprovalCategory; filter
...
+      {approval && (
```
ARADDCC012, line 174: This check shouldn't be needed once you've addressed the following comment https://github.com/gchq/Bailo/pull/663/files?diff=unified&w=1#r1302024117

frontend/pages/review.tsx
TT38665: Resolved

backend/src/services/schema.ts
```diff
@@ -47,3 +48,15 @@ export async function addDefaultSchemas() {
...
 }
+
+export async function getManagerAndReviewer(metadata: ModelMetadata) {
+  const schema = await findSchemaByRef(metadata.schemaRef)
+
+  if (!schema) {
+    return { managerTitle: 'Managers', reviewerTitle: 'Technical Reviewer' }
+  }
+
+  const managerTitle = schema.schema.properties.contacts.properties.manager.title
```
ARADDCC012, line 59: Schema is typed as `any`, but we're accessing numerous properties unsafely here. Either:
1. If we know these properties will always exist, update the type from `any` to accurately reflect what we know about `schema`
2. If we can't guarantee that these properties will always exist, update this code to check each property exists before accessing it. Also include fallback logic.

backend/src/models/Schema.ts
```diff
@@ -5,7 +5,7 @@ import { SchemaType } from '../types/types.js'
 export interface Schema {
   name: string
   reference: string
-  schema: unknown
+  schema: any
```
ARADDCC012, line 8: `unknown` is better than `any` as it forces us to assert that properties of an unknown type exist before accessing them. `any` simply turns off TypeScript.

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+  version: VersionDoc
+
+  mutateApprovals: () => void
+}
+
+function ApprovalItem({ approval, category, mutateApprovals }: ApprovalItemProps) {
   const [open, setOpen] = useState(false)
   const [choice, setChoice] = useState('')
-  const [approval, setApproval] = useState<Approval | undefined>(undefined)
+  //const [getapproval, setGetApproval] = useState<Approval | undefined>(undefined)
```
ARADDCC012, line 90: Commented out code

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+
+type ApprovalItemProps = {
+  approval: Approval // changing this from 'any' to 'Approval' causes errors
+  category: string
+  version: VersionDoc
+
+  mutateApprovals: () => void
+}
+
+function ApprovalItem({ approval, category, mutateApprovals }: ApprovalItemProps) {
```
ARADDCC012, line 87: The is a separate component and should be pulled out into its own file

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+
+      {approvals.length === 0 && (
+        <EmptyBlob text={`All done! No ${getUploadCategory(category)} are waiting for approval.`} />
+      )}
+    </Paper>
+  )
+}
+
+type ApprovalItemProps = {
+  approval: Approval // changing this from 'any' to 'Approval' causes errors
```
ARADDCC012, line 80: This comment should not be in code, but instead you can leave a comment on your own PR. Could you provide any more info as to what these errors are?

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+      isApprovalsError,
+    },
+    ErrorWrapper
+  )
+  if (error) return error
+
+  const getUploadCategory = (approvalCategory: ApprovalCategory) =>
+    approvalCategory === 'Upload' ? 'models' : 'deployments'
+
+  if (isApprovalsLoading || !approvals) {
```
ARADDCC012, line 54: We should handle these two situations differently.

- If it's loading then we should show a loading spinner (`<Loading />`)
- If we don't have any approvals to show, we should instead show a message to the user. Something like "You don't have anything to review"

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
 }
 
 function ApprovalList({ category, filter }: { category: ApprovalCategory; filter: ApprovalFilterType }) {
+  const { approvals, isApprovalsLoading, isApprovalsError, mutateApprovals } = useListApprovals(category, filter)
+
+  const error = MultipleErrorWrapper(
```
ARADDCC012, line 42: We could memoize this so that we're not running through this logic on every render.
```
  const error = useMemo(
    () =>
      MultipleErrorWrapper(
        `Unable to load approval page`,
        {
          isApprovalsError,
        },
        ErrorWrapper
      ),
    [isApprovalsError]
  )
```

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+    `Unable to load approval page`,
+    {
+      isApprovalsError,
+    },
+    ErrorWrapper
+  )
+  if (error) return error
```
ARADDCC012, line 49: It's generally safer to put all of your return blocks for a component at the bottom. This is because:

1. It's easy to look at the bottom to see all the possible things that can be returned by the component, rather than having to read every line of code.
2. As React hooks have to run in the same order on every render, in other words they can't be used in conditional logic, having early-exit code like this can lead to someone accidentally adding a hook below it. This can easily slip by in PRs with our current setup. This would be caught by our eslint rules, but we don't currently enforce that those rules are met before being able to merge PRs.

In short, I suggest moving `if (error) return error` to just above where you're checking `isApprovalsLoading` below.

frontend/pages/review.tsx
```diff
@@ -15,16 +15,18 @@ import { useTheme } from '@mui/material/styles'
...
 import Tabs from '@mui/material/Tabs'
 import Typography from '@mui/material/Typography'
+import { useGetSchema } from 'data/schema'
 import Link from 'next/link'
 import React, { useState } from 'react'
+import DisabledElementTooltip from 'src/common/DisabledElementTooltip'
 import { getErrorMessage } from 'utils/fetcher'
 
 import { postEndpoint } from '../data/api'
 import { ApprovalCategory, ApprovalFilterType, useGetNumApprovals, useListApprovals } from '../data/approvals'
```
ARADDCC012, line 25: This isn't necessarily for you to fix, but I've noticed that we have a duplicate type here. We should be using the type in `bailo/frontend/types/types.ts`. If that type needs to change to no longer be an enum, then we should do it in that file. 

@ARADDCC002 any insight on what's happened here?

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+  const error = MultipleErrorWrapper(
+    `Unable to load approval page`,
+    {
+      isApprovalsError,
+    },
+    ErrorWrapper
+  )
+  if (error) return error
+
+  const getUploadCategory = (approvalCategory: ApprovalCategory) =>
```
ARADDCC012, line 51: I feel like `UploadCategory` should be a specified type as it's the sort of thing we're likely to want to use in future. It's nice to help future devs to be consistent with the terminology being defined here.

I suggest making a new type in `bailo/frontend/types/types.ts`:
```
export type UploadCategory = 'model' | 'deployment'
```

You can then update `getUploadCategory`. I've changed it from being a regular function to `useMemo`, so it should be moved above any conditional logic in this component. I've also updated it to use `ApprovalCategory` from the correct file (`bailo/frontend/types/types.ts`), so your import statement at the top will need updating:
```
const uploadCategory: UploadCategory = useMemo(() => approvalCategory === ApprovalCategory.Upload ? 'model' : 'deployment', [approvalCategory])
```

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
 }
 
 function ApprovalList({ category, filter }: { category: ApprovalCategory; filter: ApprovalFilterType }) {
```
ARADDCC012, line 39: As we now have upload category and approval category, we should update this prop to be clearer. Change `category` to `approvalCategory`

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+          approval={approval}
+          category={approval.approvalCategory}
+          version={approval.version as VersionDoc}
+          key={approval._id}
+          mutateApprovals={mutateApprovals}
+        />
+      ))}
+
+      {approvals.length === 0 && (
+        <EmptyBlob text={`All done! No ${getUploadCategory(category)} are waiting for approval.`} />
```
ARADDCC012, line 73: Based on https://github.com/gchq/Bailo/pull/663/files#r1308684924 this should be updated. Also, to be super picky, saying "All done" makes the assumption that the user has reviewed something. They may have never reviewed something, but will still see this message. Due to this I've made a slight change in the wording:
```
<EmptyBlob text={`No ${uploadCategory}s awaiting approval.`} />
```

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+  const getUploadCategory = (approvalCategory: ApprovalCategory) =>
+    approvalCategory === 'Upload' ? 'models' : 'deployments'
+
+  if (isApprovalsLoading || !approvals) {
+    return <Paper sx={{ mt: 2, mb: 2 }} />
+  }
+  return (
+    <Paper sx={{ mt: 2, mb: 2, pb: 2 }}>
+      <Typography sx={{ p: 3 }} variant='h4'>
+        {category === 'Upload' ? 'Models' : 'Deployments'}
```
ARADDCC012, line 60: This will need updating to the following when using the correct `ApprovalCategory` type:
`{approvalCategory === ApprovalCategory.Upload ? 'Models' : 'Deployments'}`

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+    </Paper>
+  )
+}
+
+type ApprovalItemProps = {
+  approval: Approval // changing this from 'any' to 'Approval' causes errors
+  category: string
+  version: VersionDoc
+
+  mutateApprovals: () => void
```
ARADDCC012, line 84: We shouldn't be passing this in via props. As it stands, this component would break if I passed something like `mutateApprovals={() => undefined}` to it, but that adheres to our props interface. So we should instead access `mutateApprovals` again via the same hook as before. I've updated the props to allow for this. I've also removed `version` from your props interface as it wasn't being used:
```
type ApprovalItemProps = {
  approval: Approval 
  approvalCategory: ApprovalCategory
  filter: ApprovalFilterType
}

function ApprovalItem({ approval, approvalCategory, filter }: ApprovalItemProps) {
// There's a bunch of hooks here

// Then add the following line next to the useGetNumApprovals hook 
const { mutateApprovals } = useListApprovals(category, filter)
const { mutateNumApprovals } = useGetNumApprovals()
```

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
+    return <ErrorWrapper message={isSchemaError.info.message} />
   }
-  const reviewerStyling = {
-    mb: 2,
-    borderLeft: '.3rem solid #de3c30',
-    p: 2,
-    backgroundColor: theme.palette.container.main,
+
+  if (isSchemaLoading) {
+    return null
```
ARADDCC012, line 110: We should return `<Loading />` here

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
-    mb: 2,
-    borderLeft: '.3rem solid #283593',
-    p: 2,
-    backgroundColor: theme.palette.container.main,
+  const { schema, isSchemaLoading, isSchemaError } = useGetSchema(
+    ((approval.version as VersionDoc).model as ModelDoc)?.schemaRef ||
+      ((approval.deployment as DeploymentDoc)?.model as ModelDoc)?.schemaRef
+  )
+
+  const reviewer = schema?.schema.properties.contacts.properties.reviewer.title
```
ARADDCC012, line 104: Using `?` doesn't cover you in the event that `schema` is `undefined`. It just stops to from attempting to access the property. So we should have some sort of fallback in the event that `schema` is `undefined`. In this case that means adding a `!schema` check alongside your loading check and moving this logic below. You can also useMemo this:
```
  const reviewer: string = useMemo(
    () => (schema ? schema.schema.properties.contacts.properties.reviewer.title : ''),
    [schema]
  )
```

frontend/pages/review.tsx
```diff
@@ -35,29 +37,88 @@ function ErrorWrapper({ message }: { message: string | undefined }) {
...
-    mb: 2,
-    borderLeft: '.3rem solid #de3c30',
-    p: 2,
-    backgroundColor: theme.palette.container.main,
+
+  if (isSchemaLoading) {
+    return null
+  }
+
+  function canApprove() {
```
ARADDCC012, line 113: This could be re-rwitten using `useMemo`:
```
  const canApprove = useMemo(() => {
    if (!approval.version || approval.approvalType === 'Reviewer') {
      return true
    }
    return (
      approval.approvalType === 'Manager' &&
      (approval.version as VersionDoc).reviewerApproved === ApprovalStates.Accepted
    )
  }, [approval])
```

backend/src/services/schema.ts
ARADDCC012: This work has been spun out into the following ticket: BAI-854

ARADDCC012 (approved):

a3957273 (approved):

ARADDCC002 (approved):

PR #664, opened 2023-08-03
Feature/bai 796 get api v2 schemas

JR40159:


backend/src/routes/v2/schema/getSchemas.ts
```diff
@@ -19,52 +21,13 @@ interface GetSchemaResponse {
...
+    const { query } = parse(req, getSchemasSchema)
 
-            createdAt: new Date(),
-            updatedAt: new Date(),
-          },
-          {
-            id: 'example-schema-2',
-            name: 'Example Schema 2',
-            description: '',
+    const schemas: Array<SchemaInterface> = await findSchemasByKind(query.kind)
```
a3957273, line 26: Unnecessary typing.  If TypeScript can't work it out for itself, you should add a return type to `findSchemasByKind` instead of putting it here.

backend/test/testUtils/testModels.ts
```diff
@@ -0,0 +1,45 @@
+import { SchemaKind } from '../../src/models/v2/Schema.js'
+
+export const testModelSchema = {
```
a3957273, line 3: This file should probably be 'testSchemas' instead of 'testModels'.

backend/test/services/schema.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  Schema.find = mockedMethods.find
+  Schema.deleteOne = mockedMethods.deleteOne
+
+  return {
+    ...mockedMethods,
+    Schema,
+  }
+})
+vi.mock('../../src/models/v2/Schema.js', () => ({
+  SchemaKind: {
```
a3957273, line 25: I think you should import this (via e.g. importActual), instead of mocking it.  Alternatively, we could put it in a separate `types` folder.

a3957273 (changes requested):

backend/test/testUtils/testModels.ts
JR40159: I think there's a misunderstanding of the meaning of 'models' in this context. I'm intending this file to be the V2 version of https://github.com/gchq/Bailo/blob/main/backend/src/utils/test/testModels.ts

a3957273 (approved):

PR #665, opened 2023-08-04
Bai 738 unit cypress testing

k25831:
unit tests written to cover page loading, creation of private and public model (one with team selection, one with typed team selection).


frontend/cypress/e2e/beta/beta-model.cy.ts
```diff
@@ -0,0 +1,35 @@
+const betaModelUrl = 'localhost:8080/beta/model/new/model'
```
a3957273, line 1: This should be relative, e.g. `/beta/model/new/model`.

a3957273 (changes requested):

a3957273 (approved):

PR #666, opened 2023-08-07
Bugfix/v2 crashes on reload and log formats

JR40159:


ARADDCC013:
Looks good, tested in LS bailo ops. able to see required logs and suppress unnecessary ones.

ARADDCC013 (approved):
test in LS ops. Looks good.

PR #668, opened 2023-08-07
GET /api/v2/schema/<schema_id>

JR40159:


backend/src/routes/v2/schema/getSchemas.ts
```diff
@@ -24,7 +24,7 @@ export const getSchemas = [
   async (req: Request, res: Response<GetSchemaResponse>) => {
     const { query } = parse(req, getSchemasSchema)
 
-    const schemas = await findSchemasByKind(query.kind)
+    const schemas = await getSchemasByKind(query.kind)
```
a3957273, line 27: Am tempted to say that request functions should have names like `getSchemas` and `postResult`, etc.  Services should be independent of HTTP request language and so use phrasing like `createSchema` or `findSchemas` or `readSchema`.

backend/test/services/schema.spec.ts
```diff
@@ -1,20 +1,21 @@
...
     find: vi.fn(() => ({ sort: vi.fn(() => ['schema-1', 'schema-2']) })),
+    findOne: vi.fn(),
   }
 
   const Schema: any = vi.fn(() => ({
     save: mockedMethods.save,
   }))
   Schema.find = mockedMethods.find
-  Schema.deleteOne = mockedMethods.deleteOne
+  ;(Schema.findOne = mockedMethods.findOne), (Schema.deleteOne = mockedMethods.deleteOne)
```
a3957273, line 18: What is going on here?!  Put it on multiple lines!

a3957273 (changes requested):

backend/src/routes/v2/schema/getSchemas.ts
JR40159: I made this change to follow the convention you've implemented in your PR....

backend/test/services/schema.spec.ts
JR40159: Caused by running npm run style, not sure why

a3957273 (approved):

PR #670, opened 2023-08-09
Strip colors in production logs

a3957273:


ARADDCC013:
Tested on LS Bailo
Previous log: [0mGET /api/v1/models?type=all&filter= [36m304[0m 3.651 ms - -[0m
New log: GET /api/v1/models?type=all&filter= 304 34.706 ms - -

PR #672, opened 2023-08-14
Feature/bai 836 post api v2 schemas

JR40159:


backend/src/services/v2/schema.ts
```diff
@@ -1,7 +1,21 @@
+import { MongoServerError } from 'mongodb'
+
 import { testDeploymentSchema, testModelSchema } from '../../../test/testUtils/testModels.js'
 import Schema, { SchemaInterface } from '../../models/v2/Schema.js'
 import { SchemaKindKeys } from '../../types/v2/enums.js'
-import { NotFound } from '../../utils/v2/error.js'
+import { BadReq, NotFound } from '../../utils/v2/error.js'
+
+function isMongoServerError(err: unknown): err is MongoServerError {
```
a3957273, line 8: Nice, but can we move these error guards somewhere else?  This feels like a generic function.

backend/src/utils/__mocks__/user.ts
```diff
@@ -0,0 +1,13 @@
+import { NextFunction, Request, Response } from 'express'
+
+export async function getUser(req: Request, _res: Response, next: NextFunction) {
+  console.log('auto mocked getUser')
```
a3957273, line 4: Either change these to use the `logger` or remove them?

backend/src/utils/__mocks__/user.ts
```diff
@@ -0,0 +1,13 @@
...
+  next()
+}
+
+export function ensureUserRole() {
+  return function ensureUserRoleMiddleware(req: Request, _res: Response, next: NextFunction) {
+    console.log('auto mocked ensureUserRole')
```
a3957273, line 10: Either change these to use the `logger` or remove them?

backend/test/routes/schema/__snapshots__/postSchema.spec.ts.snap
```diff
@@ -0,0 +1,27 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`routes > schema > postSchema > blah > successfully stores the schema blah 1`] = `
+{
+  "data": {},
```
a3957273, line 5: I would expect this to return some data, e.g. the schema object.

backend/test/routes/schema/__snapshots__/postSchema.spec.ts.snap
```diff
@@ -0,0 +1,27 @@
...
+{
+  "error": {
+    "message": "You do not have the 'admin' role",
+  },
+}
+`;
+
+exports[`routes > schema > postSchema > successfully stores the schema 1`] = `
+{
+  "data": {},
```
a3957273, line 19: Expect this to return a schema.

backend/test/routes/schema/__snapshots__/postSchema.spec.ts.snap
```diff
@@ -0,0 +1,27 @@
...
+}
+`;
+
+exports[`routes > schema > postSchema > successfully stores the schema 2`] = `
+{
+  "data": {},
```
a3957273, line 25: Expect this to return a schema.

a3957273 (changes requested):

backend/src/utils/__mocks__/user.ts
JR40159: Oops, added debugging while development. Good spot

PR #674, opened 2023-08-15
Add collaborator list to models

a3957273:


backend/src/models/v2/Model.ts
```diff
@@ -8,6 +8,11 @@ export const ModelVisibility = {
 
 export type ModelVisibilityKeys = (typeof ModelVisibility)[keyof typeof ModelVisibility]
 
+export interface CollaboratorEntry {
+  entity: string
+  roles: Array<'admin' | 'write' | 'read' | string>
```
ARADDCC002, line 13: In this context is "admin" for the application or is it for just the model? If it's the latter then perhaps a different name is better? If the former, then I wonder if we should be sticking application-wide roles in the same place as model-specific roles.

ARADDCC002 (approved):

PR #678, opened 2023-08-16
Feature/add beta settings tab

ARADDCC002:


frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -0,0 +1,177 @@
...
+              Save
+            </Button>
+          </Stack>
+        </Box>
+      )}
+    </>
+  )
+}
+
+function EntityItem({
```
ARADDCC012, line 120: The prevent files from becoming too large, consider pulling separate components like this into their own file. This also has to benefit of making unit testing individual components easier to organise as they can each have their own test file.

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -0,0 +1,177 @@
...
+        multiple
+        aria-label={`role selector input for entity ${entity.entity}`}
+        value={entity.roles}
+        options={roles}
+        getOptionLabel={(role) => role}
+        onChange={onRoleChange}
+        renderInput={(params) => <TextField {...params} label='Select roles' />}
+        renderTags={(value, getTagProps) =>
+          value.map((option, index) => (
+            // eslint-disable-next-line react/jsx-key
```
ARADDCC012, line 167: We shouldn't be disabling this rule. I would take a guess that `option` would make a suitable `key`

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -0,0 +1,177 @@
...
+            <Chip label={option} {...getTagProps({ index })} disabled={fixedOptions.indexOf(option) !== -1} />
+          ))
+        }
+      />
+      <IconButton onClick={removeEntity} aria-label={`remove user ${entity.entity} from model access list button`}>
```
ARADDCC012, line 172: The aria-label doesn't need to specify the type of input. Can be: 
```
aria-label={`Remove user ${entity.entity} from model access list`}
```

ARADDCC012 (changes requested):

frontend/src/model/beta/Settings.tsx
```diff
@@ -0,0 +1,39 @@
+import { Box, Button, Divider, List, ListItem, ListItemButton, Stack } from '@mui/material'
+import { useState } from 'react'
+
+import { ModelInterface } from '../../../types/v2/types'
+import ModelAccess from './settings/ModelAccess'
+
+type SettingsCategory = 'general' | 'danger'
+
+export default function Settings({ model }: { model: ModelInterface }) {
```
ARADDCC012, line 9: Reusable prop type?
```
type SettingsProps = {
  model: ModelInterface
}
```

frontend/src/model/beta/Settings.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const [selectedCategory, setSelectedCategory] = useState('general')
```
ARADDCC012, line 10: Could type this as `useState< SettingsCategory >('general')`

frontend/src/model/beta/settings/EntityItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { Dispatch, SetStateAction } from 'react'
+import Loading from 'src/common/Loading'
+
+import { useGetModelRoles } from '../../../../actions/model'
+import { CollaboratorEntry, ModelInterface } from '../../../../types/v2/types'
+
+type EntityItemProps = {
+  entity: CollaboratorEntry
+  accessList: CollaboratorEntry[]
+  setAccessList: Dispatch<SetStateAction<CollaboratorEntry[]>>
```
ARADDCC012, line 15: Personal preference would be for this to be agnostic of what the "event" is being used for `onAccessListChange: (value: CollaboratorEntry[]) => void`

frontend/src/model/beta/settings/EntityItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+  model: ModelInterface
+}
+
+export default function EntityItem({ entity, accessList, setAccessList, model }: EntityItemProps) {
+  const { modelRoles, isModelRolesLoading } = useGetModelRoles(model.id)
```
ARADDCC012, line 20: We should also handle any error from this using `isModelRolesError`

frontend/src/model/beta/settings/EntityItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+    if (!role) return { id: roleId, name: 'Unknown Role' }
+
+    return role
+  }
+
+  return (
+    <TableRow>
+      <TableCell>
+        <EntityIcon entity={entity} />
+        <Typography>{entity.entity.replace('user:', '').replace('group:', '')}</Typography>
```
ARADDCC012, line 48: Super minor, but could `useMemo` this so it's not being calculated on every render

frontend/src/model/beta/settings/EntityItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+    </TableRow>
+  )
+}
+
+type EntityIconProps = {
+  entity: CollaboratorEntry
+}
+
+function EntityIcon({ entity }: EntityIconProps) {
+  return entity.entity.startsWith('user:') ? <PersonIcon color='primary' /> : <GroupsIcon color='secondary' />
```
ARADDCC012, line 93: Could also `useMemo` here:
```
function EntityIcon({ entity }: EntityIconProps) {
  const isUser = useMemo(() => entity.entity.startsWith('user:'), [entity])

  return isUser ? <PersonIcon color='primary' /> : <GroupsIcon color='secondary' />
}
```

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -0,0 +1,136 @@
...
+import EntityItem from './EntityItem'
+
+type ModelAccessProps = {
+  model: ModelInterface
+}
+
+export default function ModelAccess({ model }: ModelAccessProps) {
+  const [open, setOpen] = useState(false)
+  const [accessList, setAccessList] = useState<CollaboratorEntry[]>(model.collaborators)
+  const { users, isUsersLoading } = useListUsers()
```
ARADDCC012, line 32: We should also be checking for any errors from this

ARADDCC012 (approved):

PR #690, opened 2023-08-22
Bai 825 create wrapper component for markdown

P20179:
Have added the reusable markdown component which offers markdown functionality for the front end. In addition, we have altered the base version so that by default the toolbar does not display all of the various options, as this keeps the ui clean while making it better for accessibility and screen readers, as otherwise they need to tab over every option. However, we have created a custom button to view the toolbar which will display all of the other advanced options to the user


package.json
```diff
@@ -14,5 +14,8 @@
   },
   "devDependencies": {
     "prettier": "^2.8.8"
+  },
+  "dependencies": {
```
ARADDCC002, line 18: This should be added to `frontend/package.json`

ARADDCC002 (changes requested):

ARADDCC002 (approved):

PR #691, opened 2023-08-25
start of live doc implementation

ARADDCC002:
This is a draft to show progress - this is not ready for review.


a3957273 (dismissed):

frontend/pages/beta/model/[modelId].tsx
```diff
@@ -30,9 +28,7 @@ export default function Model() {
             { title: 'Releases', view: <Releases model={model} /> },
             { title: 'Settings', view: <Settings model={model} /> },
           ]}
-          actionButtonOnClick={myFunction}
-          actionButtonTitle='Actions'
-          displayActionButton
+          displayActionButton={false}
```
Comment (ARADDCC012) [inline frontend/pages/beta/model/[modelId].tsx:31]: It's unusual to provide `false` as a prop. Consider not providing displayActionButton at all here. If `displayActionButton` isn't optional, it probably should be.

frontend/pages/beta/model/[modelId]/schema.tsx
```diff
@@ -1,23 +1,34 @@
...
-  function createModelUsingSchema(_schema: string) {
-    router.push(`/beta/model/${modelId}`)
+  async function createModelUsingSchema(newSchema: SchemaInterface) {
+    if (currentUser && model) {
+      await postFromSchema(model.id, newSchema.id)
+      router.push(`/beta/model/${modelId}`)
+    }
   }
 
   function schemaButton(schema: any) {
```
Comment (ARADDCC012) [inline frontend/pages/beta/model/[modelId]/schema.tsx:34]: Looks like this slipped through a previous PR, but this is essentially a component pretending to be a regular function. Up to you if you want to fix it in this PR though

frontend/pages/model/[uuid]/edit/[version].tsx
```diff
@@ -43,6 +43,8 @@ function Upload() {
   useEffect(() => {
     if (!cSchema || !cVersion || splitSchema.steps.length) return
 
+    console.log(cVersion.metadata)
```
Comment (ARADDCC012) [inline frontend/pages/model/[uuid]/edit/[version].tsx:46]: Can be removed

frontend/src/Form/beta/ModelCardForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+  setSplitSchema,
+  canEdit = false,
+}: {
+  splitSchema: SplitSchemaNoRender
+  setSplitSchema: Dispatch<SetStateAction<SplitSchemaNoRender>>
+  canEdit?: boolean
+}) {
+  const [activeStep, setActiveStep] = useState(0)
+  // TODO - add validation
+  const [_openValidateError, _setOpenValidateError] = useState(false)
```
ARADDCC012, line 26: By using `_` we are saying that our intention is to not use these vars. If this is just because it's WIP, I would argue that the linting warnings are exactly what we want so `_` could be removed.

frontend/src/Form/beta/ModelCardForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+  splitSchema,
+  setSplitSchema,
+  canEdit = false,
+}: {
+  splitSchema: SplitSchemaNoRender
+  setSplitSchema: Dispatch<SetStateAction<SplitSchemaNoRender>>
+  canEdit?: boolean
+}) {
+  const [activeStep, setActiveStep] = useState(0)
+  // TODO - add validation
```
ARADDCC012, line 25: Not something I'm going to try to force through, but I'm a big fan of linking TODOs to tickets that have been created to capture the required changes. This comment especially sounds like something we definitely don't want to miss.

frontend/src/Form/beta/ModelCardForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const theme = useTheme()
+
+  const currentStep = splitSchema.steps[activeStep]
+
+  if (!currentStep) {
+    return null
+  }
+
+  const onFormChange = (form) => {
```
ARADDCC012, line 36: This could be simplified
```
  const onFormChange = (form) => {
    if (form.schema.title === currentStep.schema.title) {
      setStepState(splitSchema, setSplitSchema, currentStep, { ...currentStep.state, ...form.formData })
    }
  }
```

ARADDCC012, line 36: This `form` param should have a type

frontend/src/Form/beta/ModelCardForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+    <Stack
+      direction={{ xs: 'column', sm: 'row' }}
+      spacing={4}
+      justifyContent='left'
+      divider={<Divider flexItem orientation='vertical' />}
+      sx={{ width: '100%' }}
+    >
+      <Box>
+        <Stepper
+          id='form-page-stepper'
```
ARADDCC012, line 58: I don't think this `id` is being used at all. Can it be removed?

frontend/src/Form/beta/ModelCardForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+          sx={{ minWidth: 'max-content' }}
+        >
+          {splitSchema.steps.map((step, index) => (
+            <MaterialStep key={step.schema.title}>
+              <StepButton sx={{ p: 0, m: 0 }} onClick={() => setActiveStep(index)} icon={<Nothing />}>
+                <StepLabel
+                  sx={{
+                    padding: 0,
+                    '& .MuiStepLabel-label': {
+                      fontSize: '16px !important',
```
ARADDCC012, line 73: Using `!important` is considered bad practice and should be avoided. It overrides other styles and can become a real pain to maintain.

frontend/src/Form/beta/ModelCardForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+                    },
+                    '& .Mui-active': {
+                      borderBottomStyle: 'solid',
+                      color: `${theme.palette.primary.main} !important`,
```
ARADDCC012, line 77: Should also avoid using `!important` here and the line below too

frontend/src/Form/beta/ModelCardForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+        formContext={{ editMode: canEdit, formSchema: currentStep.schema }}
+        templates={
+          !canEdit
+            ? {
+                DescriptionFieldTemplate,
+              }
+            : {}
+        }
+      >
+        {/* eslint-disable-next-line react/jsx-no-useless-fragment */}
```
ARADDCC012, line 115: Can this be changed to not pass in children? E.g. `<Form />`

frontend/src/MuiForms/CustomTextInput.tsx
```diff
@@ -12,40 +14,48 @@ interface EntitySelectorProps {
...
+export default function CustomTextInput(props: CustomTextInputProps) {
   const { onChange, value: currentValue, label, formContext } = props
 
-  const _onChange = (e) => {
-    onChange(e.target.value)
-  }
-
   const theme = useTheme()
 
+  const _onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
ARADDCC012, line 22: We use `_` to denote something as being intentionally unused when we have no choice but to provide the value (e.g. when we only care about the second parameter in a function). Consider removing the `_`

frontend/src/MuiForms/CustomTextInput.tsx
```diff
@@ -12,40 +14,48 @@ interface EntitySelectorProps {
...
+        size='small'
+        sx={{
+          input: {
+            color: theme.palette.mode === 'light' ? 'black' : 'white',
+          },
+          label: {
+            WebkitTextFillColor: theme.palette.mode === 'light' ? 'black' : 'white',
+          },
+          '& .MuiInputBase-input.Mui-disabled': {
+            WebkitTextFillColor: currentValue
```
ARADDCC012, line 39: This is very difficult to read. Nested ternaries are something we should avoid. Consider pulling out into a `useMemo`
```
const disabledWebkitTextFillColor = useMemo(() => {
  if (currentValue) {
    return theme.palette.mode === 'light' ? 'black' : 'white'
  } else {
    return theme.palette.mode === 'light' ? '#535353' : '#c8c8c8'
  }
}, [theme])
```

frontend/src/MuiForms/CustomTextInput.tsx
```diff
@@ -12,40 +14,48 @@ interface EntitySelectorProps {
...
-        disableUnderline: !formContext.editMode ? true : false,
-      }}
-    />
+    <Fragment key={label}>
+      <Typography sx={{ fontWeight: 'bold' }}>{label}</Typography>
+      <TextField
+        size='small'
+        sx={{
+          input: {
+            color: theme.palette.mode === 'light' ? 'black' : 'white',
```
ARADDCC012, line 33: These colours should be in our theme

frontend/src/MuiForms/CustomTextInput.tsx
```diff
@@ -12,40 +14,48 @@ interface EntitySelectorProps {
...
+          },
+          label: {
+            WebkitTextFillColor: theme.palette.mode === 'light' ? 'black' : 'white',
```
ARADDCC012, line 36: These colours should be in our theme

frontend/src/MuiForms/CustomTextInput.tsx
```diff
@@ -12,40 +14,48 @@ interface EntitySelectorProps {
...
+          },
+          '& .MuiInputBase-input.Mui-disabled': {
+            WebkitTextFillColor: currentValue
+              ? theme.palette.mode === 'light'
+                ? 'black'
```
ARADDCC012, line 41: These colours should be in our theme

frontend/src/MuiForms/CustomTextInput.tsx
```diff
@@ -12,40 +14,48 @@ interface EntitySelectorProps {
...
+                : 'white'
+              : theme.palette.mode === 'light'
+              ? '#535353'
```
ARADDCC012, line 44: These colours should be in our theme

frontend/src/MuiForms/CustomTextInput.tsx
ARADDCC012, line 36: I'm guessing that `WebkitTextFillColor` is setting the `-webkit-text-fill-color` CSS property. If that's the case then we should be able to just use `color` instead, unless MUI is setting `-webkit-text-fill-color` under the hood. 

The MDN docs on this property state:
"The -webkit-text-fill-color CSS property specifies the fill [color](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) of characters of text. If this property is not set, the value of the [color](https://developer.mozilla.org/en-US/docs/Web/CSS/color) property is used"
See: https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-fill-color

frontend/src/MuiForms/CustomTextInput.tsx
```diff
@@ -12,40 +14,48 @@ interface EntitySelectorProps {
   InputProps?: any
 }
 
-export default function EntitySelector(props: EntitySelectorProps) {
+export default function CustomTextInput(props: CustomTextInputProps) {
   const { onChange, value: currentValue, label, formContext } = props
```
ARADDCC012, line 18: Not a change that you've made, but there's no need to change `value` to `currentValue` here. It just hinders readability by renaming the prop unnecessarily. However, it's up to you if you want to change it or not in this PR.

frontend/src/MuiForms/CustomTextInput.tsx
```diff
@@ -12,40 +14,48 @@ interface EntitySelectorProps {
...
+                : 'white'
+              : theme.palette.mode === 'light'
+              ? '#535353'
+              : '#c8c8c8',
+          },
+          fontStyle: currentValue ? 'unset' : 'italic',
+        }}
+        onChange={_onChange}
+        variant={!formContext.editMode ? 'standard' : 'outlined'}
+        required={!formContext.editMode ? false : true}
```
ARADDCC012, line 51: Hard to tell what `editMode` is as `formContext` is typed as `any`. But assuming this is a boolean you can just do `required={formContext.editMode}` or if it's not a boolean you can do `required={!!formContext.editMode}`

frontend/src/MuiForms/CustomTextInput.tsx
```diff
@@ -12,40 +14,48 @@ interface EntitySelectorProps {
...
+        value={currentValue || (!formContext.editMode ? 'Unanswered' : '')}
+        disabled={!formContext.editMode}
+        InputProps={{
+          ...props.InputProps,
+          disableUnderline: !formContext.editMode ? true : false,
```
ARADDCC012, line 56: Can be simplified `disableUnderline: !formContext.editMode`

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
+import { Autocomplete, Box, Chip, Stack, TextField, Typography } from '@mui/material'
+import { useTheme } from '@mui/material/styles'
+
+export default function TagSelector(props: any) {
+  const { onChange, value: currentValue, label, formContext } = props
```
ARADDCC012, line 5: Seems unnecessary to destructure `props` like this when you can just do it on the line above as we usually do.

ARADDCC012, line 5: I'm not a fan of renaming props like this. Consider either leaving `value` and  renaming any other `value` vars below.

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const _onChange = (_event: React.SyntheticEvent<Element, Event>, newValues: any) => {
```
ARADDCC012, line 7: Rename to `handleChange`?

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+    onChange(newValues.map((value) => value))
```
ARADDCC012, line 8: This `map` seems useless. Change to `onChange(newValues)`?

If you want to create a copy of the array you can do `onChange([...newValues])`

frontend/src/MuiForms/TagSelectorBeta.tsx
ARADDCC012, line 7: Can `newValues` be given a better type? I imagine generics could come in handy here

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const theme = useTheme()
+
+  return (
+    <>
+      {formContext.editMode && (
+        <>
+          <Typography sx={{ fontWeight: 'bold' }}>{label}</Typography>
+          <Autocomplete
+            multiple
+            isOptionEqualToValue={(option: any, value: any) => option === value}
```
ARADDCC012, line 20: Can `option` and `value` be typed?

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+            value={currentValue || ''}
+            onChange={_onChange}
+            options={[]}
+            freeSolo
+            renderTags={(value: readonly string[], getTagProps) =>
```
ARADDCC012, line 25: We shouldn't be mutating parameters ever, so specifying `value` as `readonly` is unnecessary. I believe this is already covered by our eslint rules also.

frontend/pages/beta/model/[modelId].tsx
Comment (ARADDCC002) [reply frontend/pages/beta/model/[modelId].tsx:31]: It is optional already so not sure why I was adding in false here!

frontend/src/Form/beta/ModelCardForm.tsx
ARADDCC002: This is needed otherwise the style won't update

frontend/src/Form/beta/ModelCardForm.tsx
ARADDCC002: This is needed as without it, it will render the submit buttons which we no longer need. I'm not sure why this is the case, but this has been the same since we used the library in old version.

frontend/src/MuiForms/CustomTextInput.tsx
ARADDCC002: Using `color` does not work unfortunately

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+              value.map((option: string, index: number) => (
+                // eslint-disable-next-line react/jsx-key
```
ARADDCC012, line 27: This eslint rule is trying to save us from having to refactor this in later versions of react. We've discussed this before and used the following solution in `EntityItem.tsx`:
```
            renderTags={(value, getTagProps) =>
              value.map((option, index) => (
                <Chip variant='outlined' label={option} {...getTagProps({ index })} key={option} />
              ))
            }
```

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+                <Chip variant='outlined' label={option} {...getTagProps({ index })} />
+              ))
+            }
+            renderInput={(params) => (
+              <TextField
+                {...params}
+                size='small'
+                sx={{
+                  input: {
+                    color: theme.palette.mode === 'light' ? 'black' : 'white',
```
ARADDCC012, line 37: Should be using our theme colours

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+                  },
+                  label: {
+                    WebkitTextFillColor: theme.palette.mode === 'light' ? 'black' : 'white',
```
ARADDCC012, line 40: Should be using our theme

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+                  },
+                  '& .MuiInputBase-input.Mui-disabled': {
+                    WebkitTextFillColor: theme.palette.mode === 'light' ? 'black' : 'white',
```
ARADDCC012, line 43: Should be using our theme

frontend/src/MuiForms/TagSelectorBeta.tsx
ARADDCC012, line 40: This applies to all uses of `WebkitTextFillColor` in this file. See: https://github.com/gchq/Bailo/pull/691/files#r1318463150

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+                  },
+                  fontStyle: currentValue ? 'unset' : 'italic',
```
ARADDCC012, line 45: If there's no `currentValue` then surely there's no text to show, right? Why do we want to change the font style in this case?

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+                }}
+                variant='outlined'
+                required={!formContext.editMode ? false : true}
```
ARADDCC012, line 48: Can just be `required={formContext.editMode}` assuming `editMode` is a boolean

frontend/src/MuiForms/TagSelectorBeta.tsx
```diff
@@ -0,0 +1,78 @@
...
+        </>
+      )}
+      {!formContext.editMode && (
+        <>
+          <Typography sx={{ fontWeight: 'bold' }}>{label}</Typography>
+          {currentValue.length === 0 && (
+            <Typography
+              sx={{
+                fontStyle: 'italic',
+                color: theme.palette.mode === 'light' ? '#535353' : '#c8c8c8',
```
ARADDCC012, line 61: Should be using our theme

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const { mutateModel } = useGetModel(model.id)
+  const [splitSchema, setSplitSchema] = useState<SplitSchemaNoRender>({ reference: '', steps: [] })
+  const { uiConfig: _uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
+
+  async function onSubmit() {
+    if (schema) {
+      const data = getStepsData(splitSchema, true)
+      data.schemaRef = schema.id
+      const form = data
+      // TODO - submit form
```
ARADDCC012, line 31: Is this covered off by us calling `putModelCard` below?

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const { schema, isSchemaLoading, isSchemaError } = useGetSchema(model.card.schemaId)
+  const { mutateModel } = useGetModel(model.id)
+  const [splitSchema, setSplitSchema] = useState<SplitSchemaNoRender>({ reference: '', steps: [] })
+  const { uiConfig: _uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
+
+  async function onSubmit() {
+    if (schema) {
+      const data = getStepsData(splitSchema, true)
+      data.schemaRef = schema.id
+      const form = data
```
ARADDCC012, line 30: This seems unnecessary. Consider renaming `data` above or simply removing this line

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+      const res = await putModelCard(model.id, form)
+      if (res.status && res.status < 400) {
+        setIsEdit(false)
+      }
+    }
+  }
+
+  function onCancel() {
+    if (schema) {
+      mutateModel
```
ARADDCC012, line 41: Should this be `mutateModel()`?

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+      const steps = getStepsFromSchema(schema, {}, ['properties.contacts'], model.card.metadata)
+      for (const step of steps) {
+        step.steps = steps
```
ARADDCC012, line 44: I'm very confused by this logic. What's going on here?

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }
+
+  useEffect(() => {
+    if (!model || !schema) return
+    const metadata = model.card.metadata
+    const steps = getStepsFromSchema(schema, {}, ['properties.contacts'], metadata)
+
+    for (const step of steps) {
+      step.steps = steps
```
ARADDCC012, line 57: Also confused by this logic. Can you explain what we're doing here please?

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    setSplitSchema({ reference: schema.id, steps })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ARADDCC012, line 61: It goes against the rules of hooks to not include all dependencies. This rule should not be disabled ever.

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -0,0 +1,101 @@
...
+  if (isSchemaError) {
+    return <MessageAlert message={isSchemaError.info.message} severity='error' />
+  }
+
+  if (isUiConfigError) {
+    return <MessageAlert message={isUiConfigError.info.message} severity='error' />
+  }
+  return (
+    <>
+      {isSchemaLoading || (isUiConfigLoading && <Loading />)}
```
ARADDCC012, line 73: The brackets look wrong here. Should this be the following?
```
{(isSchemaLoading || isUiConfigLoading) && <Loading />}
```

frontend/src/model/beta/overview/TemplatePage.tsx
```diff
@@ -0,0 +1,80 @@
...
+              p: 4,
+              borderRadius: 2,
+              width: '300px',
+            }}
+          >
+            <Stack spacing={2}>
+              <Typography component='h3' variant='h6'>
+                Create from scratch
+              </Typography>
+              <Typography variant='body1'>Create a model from scratch using a predifined schema.</Typography>
```
ARADDCC012, line 70: Spelling: predefined

frontend/src/model/beta/overview/TemplatePage.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+type TemplatePageProps = {
+  model: ModelInterface
+}
+
+export default function TemplatePage({ model }: TemplatePageProps) {
+  const theme = useTheme()
+  const router = useRouter()
+
+  function createFromScratchOnClick() {
```
ARADDCC012, line 16: Consider renaming to `handleCreateFromScratchClick` if you like that convention

frontend/types/interfaces.ts
```diff
@@ -28,13 +34,30 @@ export interface Step {
...
-  renderBasic?: (RenderInterface) => JSX.Element | null
+  renderBasic: (RenderInterface) => JSX.Element | null
   renderButtons: (RenderButtonsInterface) => JSX.Element | null
 
   shouldValidate: boolean
   isComplete: (step: Step) => boolean
 }
 
+export interface StepNoRender {
+  schema: any
```
ARADDCC012, line 45: I know there's a ticket to type `schema`, but can we type `uiSchema` or `state` any further?

frontend/utils/beta/formUtils.ts
```diff
@@ -105,8 +90,9 @@ export function getStepsFromSchema(
   baseUiSchema: any = {},
   omitFields: Array<string> = [],
   state: any = {}
-): Array<Step> {
-  const schemaDupe = omit(schema.schema, omitFields) as any
+): Array<StepNoRender> {
+  console.log(state)
```
ARADDCC012, line 94: Console log

frontend/src/Form/beta/ModelCardForm.tsx
ARADDCC012: That's because the color is likely defined elsewhere higher up and this is not the right place to set it.

Either that or it's defined lower down with `!important` and if this is the case then we have no choice.

frontend/src/MuiForms/TagSelectorBeta.tsx
ARADDCC002: This is because we render text to say "Unanswered" if there's no value. The colour difference is just to emphasise this.

frontend/src/MuiForms/TagSelectorBeta.tsx
ARADDCC002: In this instance yes, but this is a reusable component so there's a possibility we could render a form using this under different conditions

frontend/src/model/beta/overview/FormEditPage.tsx
ARADDCC002: Genuinely not sure!

frontend/types/interfaces.ts
ARADDCC002: uischema seems to have a type but struggling for state.

frontend/src/model/beta/overview/FormEditPage.tsx
ARADDCC002: This is how it's always been

frontend/src/model/beta/overview/FormEditPage.tsx
ARADDCC002: This is how it's always been

ARADDCC012 (approved):

PR #692, opened 2023-08-30
Initial model card and schema routes

a3957273:
This adds routes for:

- Get a model card (either the latest, or a specific version).
- Initialize a blank model card from a schema.
- Update a model card.


backend/src/models/v2/Model.ts
```diff
@@ -13,6 +13,14 @@ export interface CollaboratorEntry {
   roles: Array<'owner' | 'contributor' | 'consumer' | string>
 }
 
+export interface ModelCardInterface {
+  schemaId: string
+  version: number
+  createdBy: string
```
ARADDCC002, line 19: Is this going to be just the username or is it worth adding in the full entity object?

a3957273: We should always use `users` whenever it's an action done by someone.  There's never a 'group' that creates a release, it's always a specific person.  (Also, entities are now strings of the form `${type}:${id}`).

backend/src/routes/v2/model/modelcard/getModelCard.ts
```diff
@@ -25,20 +27,25 @@ interface GetModelCardResponse {
...
-      modelCard: {
-        modelId: 'example-model',
-        schemaId: 'example-schema',
+    let modelCard: ModelCardInterface
+    if (version === GetModelCardVersionOptions.Latest) {
+      const card = (await getModelById(req.user, modelId)).card
 
-        version: 1,
-        metadata: { example: true },
+      console.log('CARD', card)
```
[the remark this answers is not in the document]
a3957273: My bad.  I had some issues mocking.  Nice catch!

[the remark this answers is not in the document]
ARADDCC002: I do it all the time :D

backend/src/models/v2/Model.ts
ARADDCC002: Yes sorry I forgot the change in entity format, yeah I agree with this in that case.

ARADDCC002 (approved):

backend/src/routes/v2/model/modelcard/getModelCard.ts
```diff
@@ -25,20 +27,23 @@ interface GetModelCardResponse {
...
-      modelCard: {
-        modelId: 'example-model',
-        schemaId: 'example-schema',
+    let modelCard: ModelCardInterface
+    if (version === GetModelCardVersionOptions.Latest) {
+      const card = (await getModelById(req.user, modelId)).card
 
-        version: 1,
-        metadata: { example: true },
+      if (!card) {
```
JR40159, line 38: Why is this logic not in the service?

PR #696, opened 2023-08-31
Add release endpoints

a3957273:
Adds create, list, get and delete endpoints and their corresponding services.

### DIFF
--- a/backend/test/services/release.spec.ts
+++ b/backend/test/services/release.spec.ts
@@ -1,0 +1,120 @@
+import { describe, expect, test, vi } from 'vitest'
+
+import { createRelease, deleteRelease, getModelReleases, getReleaseBySemver } from '../../src/services/v2/release.js'
+
+const arrayAsyncFilter = vi.hoisted(() => ({
+  asyncFilter: vi.fn(() => []),
+}))
+vi.mock('../../src/utils/v2/array.js', () => arrayAsyncFilter)
+
+const authorisationMocks = vi.hoisted(() => ({
+  userReleaseAction: vi.fn(() => true),
+}))
+vi.mock('../../src/connectors/v2/authorisation/index.js', async () => ({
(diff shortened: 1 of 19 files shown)


backend/test/services/release.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    await createRelease({} as any, {} as any)
+
+    expect(releaseModelMocks.save).toBeCalled()
+    expect(releaseModelMocks).toBeCalled()
+  })
+
+  test('createRelease > bad authorisation', async () => {
+    authorisationMocks.userReleaseAction.mockResolvedValueOnce(false)
+    expect(() => createRelease({} as any, {} as any)).rejects.toThrowError(/^You do not have permission/)
```
JR40159, line 56: For cases like this we could also assert that release.save() does not get called as well as the error getting thrown as those both seem like the two most important outcomes of someone not having the permission to do something like this?

backend/test/services/release.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+const arrayAsyncFilter = vi.hoisted(() => ({
+  asyncFilter: vi.fn(() => []),
+}))
+vi.mock('../../src/utils/v2/array.js', () => arrayAsyncFilter)
+
+const authorisationMocks = vi.hoisted(() => ({
+  userReleaseAction: vi.fn(() => true),
+}))
+vi.mock('../../src/connectors/v2/authorisation/index.js', async () => ({
+  ...((await vi.importActual('../../src/connectors/v2/authorisation/index.js')) as object),
```
JR40159, line 14: Why the need for actual implementation here?

a3957273: I don't want to mock the authorisation enums `ModelAction` and `ReleaseAction`.

PR #718, opened 2023-09-02
Update from Node 16 to Node 20.

a3957273:
This requires all our developer machines to be updated, as they currently don't suppose Node 20.


README.md
```diff
@@ -79,7 +79,7 @@ sharing.
 
 ### Requirements:
 
-- Node v16
+- Node v20
```
JR40159, line 82: Wrong Node version

JR40159 (approved):

PR #720, opened 2023-09-03
Create a service for a common http interface

a3957273:
In our default implementation it's not particularly useful, but does allow us a central place to add in deployment specific configuration for network communication.


backend/src/services/v2/http.ts
```diff
@@ -0,0 +1,7 @@
+import https from 'node:https'
+
+// This function has the same syntax as 'https.Agent', but is centralised throughout
+// the application in case it needs to be altered.
+export function getHttpsAgent(config?: https.AgentOptions) {
+  return new https.Agent({ ...config })
```
ARADDCC012, line 6: Do we want to create a new agent on every call to this function or should we be reusing a single agent?

Could do something like:
```
// Not sure what the correct type is off the top of my head
let httpsAgent: HttpsAgent | undefined

export function getHttpsAgent(config?: https.AgentOptions) {
  if (!httpsAgent) {
    httpsAgent = new https.Agent({ ...config })
  }
  return httpsAgent
}
```

ARADDCC012 (approved):

PR #722, opened 2023-09-04
Bai 827 markdown implementation

P20179:
Further changes to Markdown Editor:

Renamed to Rich Text Editor

Implemented Rich Text Editor into the Create New Model page Description box

Edited the functionality of Rich Text Editor so that the toolbar is hidden by default but activated by selecting the words "show toolbar". When toolbar is visible, Code Edit mode is shown as default. Then the user can hide the toolbar by selecting the words "hide toolbar"


frontend/cypress/e2e/beta/beta-model.cy.ts
```diff
@@ -13,7 +13,7 @@ describe('Beta create new model', () => {
     cy.get('[data-test=teamSelector]').click()
     cy.get('[role=presentation]').click()
     cy.get('[data-test=modelSelector]').type('test model')
-    cy.get('[data-test=modelDescription]').type('test description')
+    cy.get('[aria-label=modelDescription]').type('test description')
```
ARADDCC002, line 16: Is there a reason we're using the aria-label and not the data-test property?

frontend/pages/beta/model/new.tsx
```diff
@@ -9,13 +9,13 @@ import {
...
   RadioGroup,
   Stack,
-  TextField,
   Tooltip,
   Typography,
 } from '@mui/material'
 import { useTheme } from '@mui/material/styles'
 import { useRouter } from 'next/router'
 import { useState } from 'react'
+import RichTextEditor from 'src/common/RichTextEditor'
```
ARADDCC002, line 18: This should be a relative path

frontend/pages/beta/model/new.tsx
```diff
@@ -94,27 +94,29 @@ export default function NewModel() {
...
-                  </Typography>
-                  <TextField
-                    id='new-model-description'
-                    required
-                    size='small'
-                    value={description}
-                    onChange={(e) => setDescription(e.target.value)}
-                    data-test='modelDescription'
-                  />
+                  <Box id='new-model-description'>
```
ARADDCC002, line 107: If you need an ID you should be able to add it to the RichTextEditor component below and remove this Box

frontend/pages/beta/model/new.tsx
```diff
@@ -94,27 +94,29 @@ export default function NewModel() {
...
+                    <RichTextEditor
+                      dataValue={description}
+                      onDataValueChange={(value) => setDescription(value)}
+                      data-test='modelDescription'
+                      ariaLabel='modelDescription'
```
ARADDCC002, line 112: This shouldn't be camel case, rename to "Model description"

frontend/src/common/RichTextEditor.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  const handleChange = (value?: string) => {
+    onDataValueChange(value || '')
+  }
+
+  return (
+    <>
+      <Box display='flex'>
+        {label}
```
ARADDCC002, line 32: Label is an optional prop, what happens if you don't supply one?

ARADDCC002 (changes requested):

frontend/cypress/e2e/beta/beta-model.cy.ts
P20179: I have altered it now so the tag is data-test rather than aria-label to keep it consistent with the other elements on the screen

frontend/pages/beta/model/new.tsx
P20179: Cypress testing requires camel case for the data-test name as it does not accept spaces as valid characters, so we matched this for consistency with the other uses of data-test. I have removed aria-label now though

frontend/src/common/RichTextEditor.tsx
P20179: We added in label so that we could add words on the page such as "Description" on the new model page but we weren't sure if there were examples where we may have a RichTextEditor present on the page but not named above it, so kept it optional. Do you think this should be required (if perhaps we will always actually use a descriptive title above it)? Happy to change

frontend/src/common/RichTextEditor.tsx
```diff
@@ -0,0 +1,47 @@
...
+          {`${hideToolbar ? 'Show' : 'Hide'} Toolbar`}
+        </Button>
+      </Box>
+      <MDEditor
+        defaultTabEnable
+        preview='edit'
+        hideToolbar={hideToolbar}
+        value={dataValue}
+        onChange={handleChange}
+        textareaProps={{ 'aria-label': ariaLabel }}
```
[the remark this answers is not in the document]
P20179: Have now changed it for data-test

frontend/pages/beta/model/new.tsx
ARADDCC002: Sorry to be clear, aria-label should be "Model description", data-test can be whatever you want, if you're not using data-test I wold try and use the id field instead if you can.

ARADDCC002 (approved):

PR #723, opened 2023-09-06
added wip release drafts

ARADDCC002:
This is in draft until #691 has been completed. Once it's been merged in we can updated the draft release object's `modelCardVersion` property to use the latest version in the card. We will also need to only allow releases once a user has selected a schema and has a card object attached to the model.


frontend/src/model/beta/Releases.tsx
```diff
@@ -6,21 +6,21 @@ import { ModelInterface } from '../../../types/v2/types'
...
+import ModelReleaseDisplay from './releases/ModelReleaseDisplay'
 
 export default function Releases({ model }: { model: ModelInterface }) {
   const [latestRelease, setLatestRelease] = useState<string>('')
   const [openDraftNewRelease, setOpenDraftNewRelease] = useState(false)
 
-  const { releases, isReleasesLoading } = useGetReleasesForModelId(model.id)
+  const { releases, isReleasesLoading, mutateReleases } = useGetReleasesForModelId(model.id)
 
   const sortedReleases = useMemo(() => releases.sort(sortByReleaseVersionDescending), [releases])
```
a3957273, line 18: This should be done on the server-side.

frontend/src/model/beta/Releases.tsx
```diff
@@ -6,21 +6,21 @@ import { ModelInterface } from '../../../types/v2/types'
...
 
   const modelReleaseDisplays = useMemo(
     () =>
       sortedReleases.reduce<ReactElement[]>((releaseDisplays, release) => {
-        if (release.name) {
+        if (release.semver) {
```
a3957273, line 23: Releases must always have a semver, so this if statement never applies.  This `reduce` statement should be replaced with a `map`.

frontend/src/model/beta/releases/DraftNewReleaseDialog.tsx
```diff
@@ -38,28 +45,44 @@ export default function DraftNewReleaseDialog({ open, handleClose, modelId }: Dr
...
     event.preventDefault()
+    setErrorMessage('')
     if (isValidSemver(semanticVersion)) {
       const release: Partial<ReleaseInterface> = {
-        name: releaseName,
+        modelId: model.id,
         semver: semanticVersion,
         notes: releaseNotes,
-        modelCardVersion: 1,
+        modelCardVersion: model.card.version,
```
a3957273, line 54: This no longer needs to be provided and will default to 'latest'.

a3957273 (changes requested):

a3957273 (approved):

PR #726, opened 2023-09-08
added a beta settings page and added a theme selector tab

ARADDCC002:
image


frontend/pages/beta/settings.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { useGetCurrentUser } from '../../actions/user'
+import SettingsProfileTab from '../../src/settings/beta/SettingsProfileTab'
+import ThemeTab from '../../src/settings/beta/ThemeTab'
+import Wrapper from '../../src/Wrapper.beta'
+
+function TabPanel({ children, value, index, ...rest }) {
+  return (
+    <Box
+      role='tabpanel'
+      hidden={value !== index}
```
TT38665, line 15: I understand the statement, but what does it do for this attribute in his component?  If that makes sense

ARADDCC012:
The sunset and earth themes don't look like they'll pass contrast requirements. We might also have that problem with the secondary colour of the classic theme. Have they all been run through Wave to check?

frontend/pages/beta/settings.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const { currentUser, isCurrentUserLoading } = useGetCurrentUser()
+
+  const onTabChange = (_event: React.SyntheticEvent, newValue: number) => {
+    setTab(newValue)
+  }
+
+  return (
+    <Wrapper title='Settings' page='settings'>
+      <Box sx={{ bgcolor: 'background.paper' }}>
```
ARADDCC012, line 43: Could we instead use `Paper` rather than making a `Box` look like `Paper`?

frontend/pages/beta/settings.tsx
```diff
@@ -0,0 +1,62 @@
...
+import Tab from '@mui/material/Tab'
+import Tabs from '@mui/material/Tabs'
+import React, { useState } from 'react'
+
+import { useGetCurrentUser } from '../../actions/user'
+import SettingsProfileTab from '../../src/settings/beta/SettingsProfileTab'
+import ThemeTab from '../../src/settings/beta/ThemeTab'
+import Wrapper from '../../src/Wrapper.beta'
+
+function TabPanel({ children, value, index, ...rest }) {
```
ARADDCC012, line 11: Should this be pulled out into a separate file?

ARADDCC012, line 11: Can these props be typed?

frontend/pages/beta/settings.tsx
```diff
@@ -0,0 +1,62 @@
...
+      id={`simple-tabpanel-${index}`}
+      aria-labelledby={`simple-tab-${index}`}
+      {...rest}
+    >
+      {value === index && <Box sx={{ p: 3 }}>{children}</Box>}
+    </Box>
+  )
+}
+
+function a11yProps(index) {
```
ARADDCC012, line 25: This name is a bit too generic imo. Rename to `tabA11yProps`?

frontend/pages/beta/settings.tsx
ARADDCC012, line 15: Instead of hiding we could conditionally render each `TabPanel`

frontend/pages/beta/settings.tsx
```diff
@@ -0,0 +1,62 @@
...
+function TabPanel({ children, value, index, ...rest }) {
+  return (
+    <Box
+      role='tabpanel'
+      hidden={value !== index}
+      id={`simple-tabpanel-${index}`}
+      aria-labelledby={`simple-tab-${index}`}
+      {...rest}
+    >
+      {value === index && <Box sx={{ p: 3 }}>{children}</Box>}
```
ARADDCC012, line 20: We should conditionally render each `TabPanel` using this logic, rather than hiding the box and conditionally rendering the content. Also the second `Box` here is unecessary as the `p: 3` could be added to the wrapping `Box` above.

frontend/pages/beta/settings.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+function a11yProps(index) {
+  return {
+    id: `simple-tab-${index}`,
+    'aria-controls': `simple-tabpanel-${index}`,
+  }
+}
+
+export default function Settings() {
+  const [tab, setTab] = useState<number>(0)
```
ARADDCC012, line 33: Explicit `number` type isn't needed

frontend/pages/beta/settings.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  return (
+    <Wrapper title='Settings' page='settings'>
+      <Box sx={{ bgcolor: 'background.paper' }}>
+        <Tabs
+          value={tab}
+          onChange={onTabChange}
+          indicatorColor='secondary'
+          sx={{ p: 2, borderRight: 1, borderColor: 'divider' }}
```
ARADDCC012, line 48: Is there a reason we want these `Tabs` to look different to our other tabs by adding a right border?

frontend/pages/beta/settings.tsx
```diff
@@ -0,0 +1,62 @@
...
+        >
+          <Tab label='Profile' {...a11yProps(0)} />
+          <Tab label='Display' {...a11yProps(0)} />
+        </Tabs>
+        <TabPanel value={tab} index={0}>
+          {currentUser && !isCurrentUserLoading && <SettingsProfileTab user={currentUser} />}
```
ARADDCC012, line 54: If `currentUser` is truthy then `isCurrentUserLoading` will surely always be `false`. Instead, should we show a `<Loading />` component if `isCurrentUserLoading` is `true`?

frontend/pages/beta/settings.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </TabPanel>
+        <TabPanel value={tab} index={1}>
+          {currentUser && !isCurrentUserLoading && <ThemeTab user={currentUser} />}
```
ARADDCC012, line 57: See: https://github.com/gchq/Bailo/pull/726/files#r1334182621

frontend/pages/beta/settings.tsx
```diff
@@ -0,0 +1,62 @@
...
+    'aria-controls': `simple-tabpanel-${index}`,
+  }
+}
+
+export default function Settings() {
+  const [tab, setTab] = useState<number>(0)
+
+  const { currentUser, isCurrentUserLoading } = useGetCurrentUser()
+
+  const onTabChange = (_event: React.SyntheticEvent, newValue: number) => {
```
ARADDCC012, line 37: Super nit-picky, but everywhere else we would use named export for `SyntheticEvent`.
`import { SyntheticEvent } from 'react``

frontend/src/settings/beta/SettingsProfileTab.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+function SettingsProfileTab({ user }: { user: User }) {
+  const theme = useTheme()
+  const [displayToken, setDisplayToken] = useState(false)
+  const [displayedToken, setDisplayedToken] = useState('')
+
+  const regenerateToken = async () => {
+    const { token } = await fetch('/api/v1/user/token', {
+      method: 'POST',
+    }).then((res) => res.json())
```
ARADDCC012, line 22: Shouldn't be using promise chains. You're already using `await`, so could just do:
```
const response = await fetch('/api/v1/user/token', {
  method: 'POST',
})

// error handling is needed here, what if the POST request errors?

return response.json()
``

File: frontend/src/settings/beta/SettingsProfileTab.tsx [hunk 12]
```diff
@@ -0,0 +1,95 @@
...
+import { useState } from 'react'
+import { User } from 'types/types'
+
+function SettingsProfileTab({ user }: { user: User }) {
+  const theme = useTheme()
+  const [displayToken, setDisplayToken] = useState(false)
+  const [displayedToken, setDisplayedToken] = useState('')
+
+  const regenerateToken = async () => {
+    const { token } = await fetch('/api/v1/user/token', {
```
ARADDCC012, line 20: Error handling is needed here, what if the POST request fails?

frontend/src/settings/beta/SettingsProfileTab.tsx
```diff
@@ -0,0 +1,95 @@
...
+import Stack from '@mui/material/Stack'
+import { useTheme } from '@mui/material/styles'
+import Tooltip from '@mui/material/Tooltip'
+import Typography from '@mui/material/Typography'
+import { useState } from 'react'
+import { User } from 'types/types'
+
+function SettingsProfileTab({ user }: { user: User }) {
+  const theme = useTheme()
+  const [displayToken, setDisplayToken] = useState(false)
```
ARADDCC012, line 16: These state vars are confusingly similar in name imo. How about `token` and `showToken`?

frontend/src/settings/beta/SettingsProfileTab.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const showToken = async () => {
+    const token = await regenerateToken()
+    setDisplayedToken(token)
+    setDisplayToken(true)
+  }
+
+  return (
+    <>
+      <Box sx={{ pt: 0, pr: 2, pb: 2, pl: 2 }}>
```
ARADDCC012, line 35: `pt: 0` is the default, can be removed

frontend/src/settings/beta/SettingsProfileTab.tsx
```diff
@@ -0,0 +1,95 @@
...
+          Name
+        </Typography>
+        <Divider sx={{ pt: 1, mb: 1 }} />
+        <Typography variant='body1'>{user.id}</Typography>
+      </Box>
+      <Box sx={{ p: 2 }}>
+        <Typography variant='body1' sx={{ pr: 1, fontWeight: 'medium' }}>
+          Roles
+        </Typography>
+        <Divider sx={{ pt: 1, mb: 1 }} />
```
ARADDCC012, line 46: Bit picky, but can we pick either padding or margin here. It doesn't matter which one in this case, but is a bit odd to use one for the top and another for the bottom.

frontend/src/settings/beta/SettingsProfileTab.tsx
```diff
@@ -0,0 +1,95 @@
...
+    setDisplayToken(true)
+  }
+
+  return (
+    <>
+      <Box sx={{ pt: 0, pr: 2, pb: 2, pl: 2 }}>
+        <Typography variant='body1' sx={{ pr: 1, fontWeight: 'medium' }}>
+          Name
+        </Typography>
+        <Divider sx={{ pt: 1, mb: 1 }} />
```
ARADDCC012, line 39: See: https://github.com/gchq/Bailo/pull/726/files#r1334190559

frontend/src/settings/beta/SettingsProfileTab.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <Typography variant='body1'>{user.id}</Typography>
+      </Box>
+      <Box sx={{ p: 2 }}>
+        <Typography variant='body1' sx={{ pr: 1, fontWeight: 'medium' }}>
+          Roles
+        </Typography>
+        <Divider sx={{ pt: 1, mb: 1 }} />
+        <Stack direction='row' spacing={1} sx={{ p: 1 }}>
+          {user.roles.map((role) => (
+            <Chip color='primary' key={`chip-role-${role}`} label={role} />
```
ARADDCC012, line 49: `key` can simply be `key=role`

frontend/src/settings/beta/SettingsProfileTab.tsx
```diff
@@ -0,0 +1,95 @@
...
+          ))}
+        </Stack>
+      </Box>
+      <Box sx={{ p: 2 }}>
+        <Typography variant='body1' sx={{ pr: 1, fontWeight: 'medium' }}>
+          User authentication token
+        </Typography>
+        <Divider sx={{ pt: 1, mb: 2 }} />
```
ARADDCC012, line 57: See: https://github.com/gchq/Bailo/pull/726/files#r1334190559

frontend/src/settings/beta/SettingsProfileTab.tsx
```diff
@@ -0,0 +1,95 @@
...
+          <Box
+            sx={{
+              backgroundColor: theme.palette.container.main,
+              pr: 2,
+              pl: 2,
+              display: 'flex',
+              mr: 1,
+            }}
+          >
+            <Box component={Stack} direction='column' justifyContent='center'>
```
ARADDCC012, line 71: If this `Stack` is only being used to center content of a single child then we should be using another method.

If that's the case, you can simply set `display: 'flex'` on the above `Box` (already done) and `m: 'auto'` on the `Typography`:
```
          <Box
            sx={{
              backgroundColor: theme.palette.container.main,
              pr: 2,
              pl: 2,
              display: 'flex',
              mr: 1,
            }}
          >
              <Typography variant='body1' data-test='dockerPassword' sx={ m: 'auto' }>
                {displayToken ? displayedToken : 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxxx'}
              </Typography>
          </Box>
```

frontend/src/settings/beta/SettingsProfileTab.tsx
```diff
@@ -0,0 +1,95 @@
...
+              <Typography variant='body1' data-test='dockerPassword'>
+                {displayToken ? displayedToken : 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxxx'}
+              </Typography>
+            </Box>
+          </Box>
+          <Tooltip title='Regenerate & Copy to clipboard'>
```
ARADDCC012, line 77: Should the `C` from `Copy` be lower case?

frontend/src/settings/beta/SettingsProfileTab.tsx
```diff
@@ -0,0 +1,95 @@
...
+            <IconButton
+              onClick={async () => {
+                const token = await regenerateToken()
```
ARADDCC012, line 80: To avoid having too much logic inline can we move this to a `handleRegenerateAndCopyClick` function?

frontend/src/settings/beta/ThemeTab.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+interface ThemeKey {
+  name: string
+  theme: Theme
+}
+
+// TODO - determine how we are storing user themes
+
+export default function ThemeTab({ user }: ThemeTabProps) {
+  const themes: ThemeKey[] = [
```
ARADDCC012, line 19: This could be defined outside of the component, so it's not recreated on every render

frontend/src/settings/beta/ThemeTab.tsx
```diff
@@ -0,0 +1,152 @@
...
+    { name: 'Bailo Default', theme: betaLightTheme },
+    { name: 'Bailo Classic', theme: lightTheme },
+    { name: 'Bailo Dark', theme: darkTheme },
+    { name: 'Bailo Sunset', theme: sunsetTheme },
+    { name: 'Bailo Earth', theme: earthTheme },
+  ]
+
+  function userThemeOnChange(themeKey) {
+    console.log(`Setting ${themeKey.name} as default theme for ${user.id}`)
```
ARADDCC012, line 28: is this WIP?

frontend/src/settings/beta/ThemeTab.tsx
```diff
@@ -0,0 +1,152 @@
...
+      {themes.map((theme) => (
+        <Grid key={theme.name} item xs={12} sm={4} spacing={2}>
+          <ExampleThemeDisplay themeKey={theme} setUserTheme={userThemeOnChange} />
+        </Grid>
+      ))}
+    </Grid>
+  )
+}
+
+interface TestThemeDisplayProps {
```
ARADDCC012, line 41: This name doesn't match the name of the component it is used for.

frontend/src/settings/beta/ThemeTab.tsx
```diff
@@ -0,0 +1,152 @@
...
+  themeKey: ThemeKey
+  setUserTheme: (themeKey: ThemeKey) => void
+}
+
+function ExampleThemeDisplay({ themeKey, setUserTheme }: TestThemeDisplayProps) {
+  const StyledCardContent = styled(CardContent)({
```
ARADDCC012, line 47: I believe this should be defined outside of the component

frontend/src/settings/beta/ThemeTab.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <ExampleThemeDisplay themeKey={theme} setUserTheme={userThemeOnChange} />
+        </Grid>
+      ))}
+    </Grid>
+  )
+}
+
+interface TestThemeDisplayProps {
+  themeKey: ThemeKey
+  setUserTheme: (themeKey: ThemeKey) => void
```
ARADDCC012, line 43: Instead of assuming intent with what this component will be used for, can we change it to `onThemeChange` to match our event/handler naming convention?

frontend/src/settings/beta/ThemeTab.tsx
```diff
@@ -0,0 +1,152 @@
...
+}
+
+function ExampleThemeDisplay({ themeKey, setUserTheme }: TestThemeDisplayProps) {
```
ARADDCC012, line 46: Can this be pulled into a separate file?

frontend/src/theme.ts
ARADDCC012, line 1: The sunset and earth themes don't look like they'll pass contrast requirements. We might also have that problem with the secondary colour of the classic theme as it doesn't work against light backgrounds. Have they all been run through Wave to check?

If they don't pass these checks then we shouldn't make them available to users.

frontend/pages/beta/settings.tsx
ARADDCC012: `hidden` shouldn't be used. It's not a MUI or React attribute, but a default HTML attribute. It's best to avoid using `hidden` because it will be overridden if a `display` attribute is added to the element (e.g. `display: 'flex'` or `display: 'inline-block'`).

It's always better to use either:
1. Conditional rendering (preferred) - `{ value === index & <div>Hello World</div> }`
2. The display property - `<Box sx={{ display: value !== index ? 'none' : undefined }}>`

The following SO post explains in a bit more detail: https://stackoverflow.com/a/65685345

ARADDCC002:
Moved back to draft as this will not need to be merged in until V2 is complete.

ARADDCC002:
Now that v2 has been released we can come back to this work, although it will make sense to start from scratch as there are too many changes since this PR was raised.

Leaving open for now to use as a base for when we revisit this work.

PR #727, opened 2023-09-09
Allow uploads to S3 bucket for model

a3957273:


backend/src/routes/v2/model/file/getFiles.ts
```diff
@@ -28,7 +28,6 @@ export const getFiles = [
           modelId: 'example-model',
 
           name: 'example-file',
-          category: FileCategory.Other,
```
ARADDCC002, line 31: Do we need the `mime` property here too?

ARADDCC002 (approved):

PR #728, opened 2023-09-10
Initial Python library rewrite

a3957273:
Currently implements two basic functions.  Has working tests with mocking.  Does not have great documentation.


lib/python-beta/examples/create_model.py
bw27492, line 1: Do we need this if we're having example notebooks? Happy to approve for now

bw27492 (approved):

PR #732, opened 2023-09-13
add start of ui work for access requests

ARADDCC002:
Changed living doc step labels to be consistent:

image

Added a "Request access" button:

image

This loads a select schema page (this could very easily change in the future!:

image

Added a form page for making an access request:

image

This also has validation:

image

### DIFF
--- a/frontend/pages/beta/model/[modelId]/access/new.tsx
+++ b/frontend/pages/beta/model/[modelId]/access/new.tsx
@@ -1,0 +1,73 @@
+import ArrowBack from '@mui/icons-material/ArrowBack'
+import { Box, Button, Card, Stack } from '@mui/material'
+import { useRouter } from 'next/router'
+import { useEffect, useState } from 'react'
+
+import { useGetModel } from '../../../../../actions/model'
+import { useGetSchema } from '../../../../../actions/schema'
+import Loading from '../../../../../src/common/Loading'
+import ModelCardForm from '../../../../../src/Form/beta/ModelCardForm'
+import MessageAlert from '../../../../../src/MessageAlert'
+import Wrapper from '../../../../../src/Wrapper.beta'
+import { Styling } from '../../../../../types/enums'
+import { SplitSchemaNoRender } from '../../../../../types/interfaces'
+import { getStepsFromSchema } from '../../../../../utils/beta/formUtils'
+
+export default function NewAccessRequest() {
+  const router = useRouter()
+  const { modelId }: { modelId?: string } = router.query
+  const { model, isModelLoading, isModelError } = useGetModel(modelId)
+  // TODO - populate the accessRequestSettings property
+  const { schema, isSchemaLoading, isSchemaError } = useGetSchema(
+    model?.accessRequestSettings?.schemaId || 'minimal-access-request-general-v10-beta'
+  )
+
+  const [splitSchema, setSplitSchema] = useState<SplitSchemaNoRender>({ reference: '', steps: [] })
+
+  useEffect(() => {
+    if (!model || !schema) return
+    const metadata = model.card.metadata
+    const steps = getStepsFromSchema(schema, {}, ['properties.contacts'], metadata)
+
+    for (const step of steps) {
+      step.steps = steps
+    }
+
+    setSplitSchema({ reference: schema.id, steps })
+  }, [schema, model])
+
+  function onSubmit() {
+    console.log()
+  }
+
+  if (isSchemaError) {
+    return <MessageAlert message={isSchemaError.info.message} severity='error' />
+  }
+
+  if (isModelError) {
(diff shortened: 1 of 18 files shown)


frontend/pages/beta/model/[modelId]/access/new.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return <MessageAlert message={isSchemaError.info.message} severity='error' />
+  }
+
+  if (isModelError) {
+    return <MessageAlert message={isModelError.info.message} severity='error' />
+  }
+
+  return (
+    <Wrapper title='Access Request' page='Model'>
+      {(isSchemaLoading || isModelLoading) && <Loading />}
```
Comment (ARADDCC012) [inline frontend/pages/beta/model/[modelId]/access/new.tsx:54]: We should probably not render anything else other than the loading spinner if it's in a loading state. As it stands, we're showing the spinner and the `Access requests can not be requested if a schema is not set for this model.` message.

Consider wrapping the card below with the following conditional logic `!isSchemaLoading && !isModelLoading`

frontend/pages/beta/model/[modelId]/access/schema.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { useGetSchemas } from '../../../../../actions/schema'
+import EmptyBlob from '../../../../../src/common/EmptyBlob'
+import Loading from '../../../../../src/common/Loading'
+import Wrapper from '../../../../../src/Wrapper.beta'
+import { SchemaInterface, SchemaKind } from '../../../../../types/types'
+
+export default function NewSchemaSelection() {
+  const router = useRouter()
+  const { modelId }: { modelId?: string } = router.query
+  const { schemas, isSchemasLoading, isSchemasError } = useGetSchemas(SchemaKind.Deployment)
```
Comment (ARADDCC012) [inline frontend/pages/beta/model/[modelId]/access/schema.tsx:16]: We should show an error if there is one in `isSchemasError`

frontend/pages/beta/model/[modelId]/access/schema.tsx
```diff
@@ -0,0 +1,97 @@
...
+    </Wrapper>
+  )
+}
+
+interface SchemaButtonProps {
+  modelId: string
+  schema: any
+}
+
+function SchemaButton({ modelId, schema }: SchemaButtonProps) {
```
Comment (ARADDCC012) [inline frontend/pages/beta/model/[modelId]/access/schema.tsx:71]: Can this component be moved into a separate component file?

frontend/pages/beta/model/new.tsx
```diff
@@ -82,83 +83,85 @@ export default function NewModel() {
...
-                <FormControl>
-                  <Typography sx={{ fontWeight: 'bold' }}>
-                    Description <span style={{ color: theme.palette.primary.main }}>*</span>
-                  </Typography>
-                  <TextField
-                    data-test='modelDescription'
-                    onChange={(event) => setDescription(event.target.value)}
-                    value={description}
-                    size='small'
+      <Container maxWidth='sm'>
```
ARADDCC012, line 86: Consider putting this `Container` inside `Wrapper` if it's something we're always going to wrap around the content.

frontend/src/Form/beta/ModelCardForm.tsx
```diff
@@ -116,3 +103,18 @@ export default function ModelCardForm({
     </Stack>
   )
 }
+
+interface ValidationErrorIconProps {
+  displayLabelValidation: boolean
+  step: StepNoRender
+}
+
+function ValidationErrorIcon({ displayLabelValidation, step }: ValidationErrorIconProps) {
```
ARADDCC012, line 112: Can this component be moved into its own file?

frontend/src/Form/beta/ModelCardForm.tsx
```diff
@@ -116,3 +103,18 @@ export default function ModelCardForm({
     </Stack>
   )
 }
+
+interface ValidationErrorIconProps {
+  displayLabelValidation: boolean
```
ARADDCC012, line 108: What's the point of this prop? If it's `false` then we don't return anything from the component. If we dont want to use the component then it makes more sense to conditionally render the component in the parent (or not use at all), instead of hiding that behaviour inside the component itself. E.g.
```
{ displayLabelValidation && <ValidationErrorIcon step={step} /> }
```

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -131,9 +131,11 @@ export default function ModelAccess({ model }: ModelAccessProps) {
               </TableBody>
             </Table>
           </Box>
-          <Button aria-label='Save access list' onClick={updateAccessList}>
-            Save
-          </Button>
+          <Box>
```
ARADDCC012, line 134: This `Box` isn't doing anything useful. Can be removed

ARADDCC012 (approved):

PR #733, opened 2023-09-14
Bugfix/bai 862 request reviews on release creation

JR40159:


backend/src/services/v2/smtp/templates/releaseReviewRequest.ts
```diff
@@ -0,0 +1,62 @@
...
+import { BaseEmailTemplate, IEmailTemplate } from './baseEmailTemplate.js'
+
+export class ReleaseReviewRequest extends BaseEmailTemplate implements IEmailTemplate {
+  getSubject(resourceName: string): string {
+    return dedent(`
+    You have been requested to review '${resourceName}' on Bailo
+  `)
+  }
+
+  getBody(releaseName: string, reviewKind: ReviewKindKeys, modelId: string, baseUrl: string, author: string) {
```
a3957273, line 14: This should probably be called `getText`, as it's the text mime.  Both this and the HTML are technically the body.

backend/src/utils/v2/entity.ts
```diff
@@ -1,4 +1,10 @@
-export function toEntity(kind: string, value: string) {
+export const EntityKind = {
+  User: 'user',
+  Group: 'group',
```
a3957273, line 3: I don't think this can be an enum, necessarily.  Systems can generate their own entity kinds, e.g. we'll have ones specific to our internal names.

a3957273 (changes requested):

backend/package.json
```diff
@@ -106,7 +106,7 @@
     "eslint-plugin-simple-import-sort": "^10.0.0",
     "mongodb-memory-server": "^8.13.0",
     "nodemon": "^3.0.1",
-    "prettier": "3.0.3",
+    "prettier": "2.8.8",
```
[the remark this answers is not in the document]
JR40159: I'm just avoiding the extra styling while I'm still deving, I'll add it back once I'm done :)

backend/src/connectors/v2/authorisation/silly.ts
```diff
@@ -20,6 +20,23 @@ export class SillyAuthorisationConnector implements BaseAuthorisationConnector {
...
+    if (entityObject.kind !== EntityKind.User) {
+      throw new Error('Cannot get user information for a non-user entity')
+    }
+    return {
+      email: `${entityObject.value}@example.com`,
+    }
+  }
+
+  async getGroupMembers(entity: string): Promise<string[]> {
+    if (fromEntity(entity).kind !== EntityKind.User) {
```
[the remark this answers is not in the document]
JR40159: Copy and paste error, good spot

backend/src/services/v2/smtp/smtp.ts
```diff
@@ -0,0 +1,60 @@
...
+  } else {
+    throw GenericError(500, 'Error Sending email notification to unrecognised entity', { entity })
+  }
+
+  const email = new ReleaseReviewRequest()
+  const subject = email.getSubject(release.name)
+  const text = email.getBody(release.name, reviewRequest.kind, release.modelId, appBaseUrl, 'unknown')
+  const html = email.getHtml(release.name, reviewRequest.kind, release.modelId, appBaseUrl, 'unknown')
+
+  if (!config.smtp.enabled) {
```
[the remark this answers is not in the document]
JR40159: If you look in the if statement, you'll see, as with the original implementation, we log out the subject and the address. But I do think the subject and the address is pretty irrelevant if smtp has been disabled, so I'll update this

backend/src/services/v2/smtp/smtp.ts
```diff
@@ -0,0 +1,60 @@
...
+    log.info({ subject, to }, 'Not sending email due to SMTP disabled')
+    return
+  }
+
+  if (!transporter) {
```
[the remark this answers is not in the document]
JR40159: This is so we only need to create a transporter once as a pose to every time we call this function?

backend/src/connectors/v2/authorisation/silly.ts
```diff
@@ -20,23 +27,35 @@ export class SillyAuthorisationConnector implements BaseAuthorisationConnector {
...
+    if (entityObject.kind !== SillyEntityKind.User) {
+      throw new Error(`Cannot get user information for a non-user entity: ${entity}`)
     }
     return {
       email: `${entityObject.value}@example.com`,
     }
   }
 
+  async getUserInformationList(entity): Promise<Promise<{ email: string; }>[]> {
+    const entityObject = fromEntity(entity)
```
a3957273, line 44: With the suggested changes to `getEntityMembers` below, this function would becomes significantly simpler:

```typescript
{
  const entities = await this.getGroupMembers(entity)
  return entities.map(entity => this.getUserInformation(entity))
}
```

backend/src/models/v2/ReviewRequest.ts
```diff
@@ -20,12 +19,12 @@ interface Review {
...
 // client.
 export interface ReviewRequestInterface {
-  model: string
   semver?: string
   modelId?: string
   kind: ReviewKindKeys
 
-  entity: CollaboratorEntry
+  role: string
+  entities: Array<string>
```
a3957273, line 27: As mentioned in chat, we don't want `entities` to be on the review request.

backend/src/services/v2/review.ts
```diff
@@ -40,37 +41,48 @@ export async function findReviewRequestsByActive(user: UserDoc, active: boolean)
...
+  await Promise.all(createReviewRequests)
+}
 
-  // I think we care?
-  if (!requestCreated) {
-    throw BadReq('No Review Requests have been created')
-  }
+function getEntitiesForRole(collaborators: Array<CollaboratorEntry>, role: string): string[] {
+  const roleEntities: string[] = collaborators
+    .filter((collaborator) => {
```
a3957273, line 83: `.filter((collaborator) => collaborator.roles.includes(role)` is better.

backend/src/services/v2/schema.ts
```diff
@@ -33,13 +33,11 @@ export async function createSchema(schema: Partial<SchemaInterface>, overwrite =
   try {
     return await schemaDoc.save()
   } catch (error) {
-    if (isMongoServerError(error) && error.code == 11000) {
-      throw BadReq(`The following is not unique: ${JSON.stringify(error.keyValue)}`)
-    }
-    throw error
+    throw handleDuplicateKeys(error)
```
a3957273, line 36: With other suggested change to `handleDuplicateKeys` this becomes:

```
handleDuplicateKeys(error)
throw error
```

backend/src/services/v2/smtp/templates/baseEmailTemplate.ts
```diff
@@ -3,9 +3,14 @@ import dedent from 'dedent-js'
 import { ReviewKindKeys } from '../../../../types/v2/enums.js'
 
 export interface IEmailTemplate {
-  getSubject(resourceName: string): string
-  getHtml(releaseName: string, reviewKind: ReviewKindKeys, modelId: string, baseUrl: string, author: string): string
-  getBody(releaseName: string, reviewKind: ReviewKindKeys, modelId: string, baseUrl: string, author: string): string
+  to
```
a3957273, line 6: These should be typed, e.g.  as `string`.

a3957273: I don't think you need this interface for anything.  You can use the `class` to automatically generate these types.  Just add these as abstract functions to your class, and add the types in.  Then you can just use the class and not this interface.  See our v2 authorization class as an example.

backend/src/services/v2/smtp/templates/releaseReviewRequest.ts
```diff
@@ -4,17 +4,26 @@ import mjml2html from 'mjml'
 import { ReviewKindKeys } from '../../../../types/v2/enums.js'
 import { BaseEmailTemplate, IEmailTemplate } from './baseEmailTemplate.js'
 
-export class ReleaseReviewRequest extends BaseEmailTemplate implements IEmailTemplate {
-  getSubject(resourceName: string): string {
-    return dedent(`
-    You have been requested to review '${resourceName}' on Bailo
+export class ReleaseReviewRequestEmail extends BaseEmailTemplate implements IEmailTemplate {
```
a3957273, line 7: This seems over complicated.  I think I now understand why you're using both `extends` and `implements` at the same time.  

```typescript
interface Email {
  subject: string
  text: string
  html: string
  to: string
}

function emailWrapper(body: string) {
  ...
}

function sendEmail(email: Email) {
  ...
}

function sendReviewRequestNotifications(to: string, model: ModelDoc, ...) {
  const subject = `${reviewerRole.toUpperCase()}: You have been requested to review '${resourceName}' on Bailo`
  const text = dedent(`
    // some text
  `)

  const html = emailWrapper(dedent(`
    // some html
  `))

  await sendEmail({ to, subject, html, text })
}

```

Seems much more understandable, without losing any of our type safety.

a3957273 (changes requested):

backend/src/models/v2/Schema.ts
```diff
@@ -14,7 +14,7 @@ export interface SchemaInterface {
   hidden: boolean
 
   kind: SchemaKindKeys
-  jsonSchema: unknown
+  jsonSchema: object
```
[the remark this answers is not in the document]
a3957273: Because the above interface is _bad_, you can probably use some types from 'json-schema' to help out (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/json-schema/index.d.ts)

backend/src/services/v2/smtp/templates/releaseReviewRequest.ts
JR40159: I may be misunderstanding what you're suggesting here but it seems like you're wanting to include all of the static email content like the HTML inside `sendReviewRequestNotifications` which is in the `smtp` service? I think this is a bad idea and the static content specific to our different emails should not be included in the generic `smtp` service

a3957273 (approved):

PR #737, opened 2023-09-17
Add get current user route

a3957273:


backend/src/routes/v2/user/getCurrentUser.ts
```diff
@@ -0,0 +1,21 @@
+import bodyParser from 'body-parser'
+import { Request, Response } from 'express'
+import { z } from 'zod'
+
+import { UserDoc } from '../../../models/v2/User.js'
+import { parse } from '../../../utils/validate.js'
+
+export const getCurrentUserSchema = z.object({})
```
TT38665, line 8: What does z.object({}) in this case?

JR40159 (approved):

backend/src/routes/v2/user/getCurrentUser.ts
a3957273: It means the endpoint takes no input.

PR #738, opened 2023-09-17
Add get files endpoint

a3957273:
A fix was added in `deleteFiles`, which this extends, so it should be reviewed first.  The `files` service was in the wrong folder, which broke some tests.


backend/src/models/v2/File.ts
```diff
@@ -44,6 +45,8 @@ const FileSchema = new Schema<FileInterface>(
   },
 )
 
+FileSchema.plugin(MongooseDelete, { overrideMethods: 'all', deletedBy: true, deletedByType: Schema.Types.ObjectId })
```
TT38665, line 48: What does the plugin() do?

backend/src/routes/v2/model/file/deleteFile.ts
```diff
@@ -18,7 +19,11 @@ interface DeleteFileResponse {
 export const deleteFile = [
   bodyParser.json(),
   async (req: Request, res: Response<DeleteFileResponse>) => {
-    const _ = parse(req, deleteFileSchema)
```
TT38665, line 21: What does the 'const _' do and why is it not implemented anymore?

backend/src/services/v2/file.ts
```diff
@@ -0,0 +1,82 @@
+import { putObjectStream } from '../../clients/s3.js'
+import authorisation, { ModelAction } from '../../connectors/v2/authorisation/index.js'
+import FileModel from '../../models/v2/File.js'
+import { UserDoc } from '../../models/v2/User.js'
+import config from '../../utils/v2/config.js'
+import { Forbidden, NotFound } from '../../utils/v2/error.js'
+import { longId } from '../../utils/v2/id.js'
+import { getModelById } from './model.js'
+
+export async function uploadFile(user: UserDoc, modelId: string, name: string, mime: string, stream: ReadableStream) {
```
TT38665, line 10: what is 'mime'?

backend/src/models/v2/File.ts
ARADDCC002: This allows us to "delete" documentns from a Mongo collection. Normally, you can just delete a document but that means you can't undelete it. This plugin allows us to mark documents as "deleted", which means they won't be retrieved via normal GET requests, but they still exist in the collection so that we can make special requests to restore them.

backend/src/services/v2/file.ts
ARADDCC002: Mime types refer to what _type_ of file we're going to expect. There's a list of common ones [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types). Whilst there are libraries to type _common_ mime types, you can always make your own so it's best to leave as string.v

backend/src/routes/v2/model/file/deleteFile.ts
ARADDCC002: It originally wasn't being used before as the response was just mocked. It's now been been re implemented a few lines below

backend/test/services/file.spec.ts
JR40159, line 1: Are you missing a config mock for these tests as we use config in the service?

JR40159 (changes requested):

PR #739, opened 2023-09-17
Add update model endpoint

a3957273:


backend/src/services/v2/model.ts
```diff
@@ -206,6 +206,20 @@ export async function updateModelCard(
   return revision
 }
 
+export type UpdateModelParams = Pick<ModelInterface, 'name' | 'description' | 'visibility'>
+export async function updateModel(user: UserDoc, modelId: string, diff: Partial<UpdateModelParams>) {
```
TT38665, line 210: Does the 'diff' stand for different/difference??

ARADDCC002: I would assume so, a Partial is (like the name suggests) a type that contains _some_ but not all of the properties of a given type. 

I think we should probably rename this, but I am going to approve in case you're 100% set on this naming @a3957273

backend/src/services/v2/model.ts
```diff
@@ -206,6 +206,20 @@ export async function updateModelCard(
...
+  const model = await getModelById(user, modelId)
```
JR40159, line 211: Why didn't you use `findByIdAndUpdate()`?

JR40159 (changes requested):

backend/src/services/v2/model.ts
a3957273: We need to check the authorisation first.

PR #740, opened 2023-09-17
Add get docker images endpoint

a3957273:


backend/src/services/v2/registry.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    agent: httpsAgent,
+  }).then((res) => res.json() as Promise<{ repositories: Array<string> }>)
+
+  return catalog.repositories.filter((repo) => repo.startsWith(`${modelId}/`))
+}
+
+export async function listImageTags(user: UserDoc, imageRef: RepoRef) {
+  const repo = `${imageRef.namespace}/${imageRef.model}`
+  const token = await getAccessToken({ id: user.dn, _id: user.dn }, [
```
TT38665, line 44: What is the difference between 'id' and '_id' ?

backend/src/services/v2/registry.ts
```diff
@@ -0,0 +1,70 @@
...
+  }).then((res) => res.json() as Promise<{ tags: Array<string> }>)
+
+  return catalog.tags
+}
+
+export async function listModelImages(user: UserDoc, modelId: string) {
+  const repos = await listModelRepos(user, modelId)
+  const versions = await Promise.all(
+    repos.map(async (repo) => {
+      const [namespace, model] = repo.split('/')
```
TT38665, line 64: Is the .split() used to keep a division between namespace and model?

ARADDCC002 (approved):

backend/src/routes.ts
```diff
@@ -212,7 +213,7 @@ if (config.experimental.v2) {
   server.post('/api/v2/model/:modelId/files/upload/multipart/finish', ...postFinishMultipartUpload)
   server.delete('/api/v2/model/:modelId/files/:fileId', ...deleteFile)
 
-  // *server.get('/api/v2/model/:modelId/images', ...getImages)
+  server.get('/api/v2/model/:modelId/images', ...getImages)
```
JR40159, line 216: Is this obvious enough we are referring to Docker images?

JR40159 (changes requested):

backend/src/services/v2/registry.ts
a3957273: We used to have the `_id` field represent the internal MongoDB ID and the 'id' field represent the public identifier of a user.  This was confusing, so we now only use the 'dn' field to represent a user.

This `id` and `_id` is to make sure our registry continues working during the beta, as our `registryAuth` is still using the v1 authentication system.

backend/src/services/v2/registry.ts
a3957273: The repository is in the form `x/y`.  We want to separate x, the namespace with y, the model.  So we split it into two.

backend/src/services/v2/registry.ts
```diff
@@ -0,0 +1,70 @@
+import fetch from 'node-fetch'
+
+import { UserDoc } from '../../models/v2/User.js'
+import { getAccessToken } from '../../routes/v1/registryAuth.js'
```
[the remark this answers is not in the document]
a3957273: Yes but we can only run one simultaneously, :(.

backend/src/routes.ts
a3957273: We could rename it `getOCIImages`, but that looks weird!

backend/src/routes/v2/model/images/getImages.ts
```diff
@@ -14,29 +14,22 @@ export const getImagesSchema = z.object({
 })
 
 interface GetImagesResponse {
-  images: Array<ImageInterface>
+  images: Array<{
+    namespace: string
+    model: string
```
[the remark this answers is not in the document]
a3957273: No, this is the model artefact name in Docker.

JR40159 (approved):

PR #741, opened 2023-09-18
New endpoints (including unit tests and docstrings)

bw27492:
- get_model
- update_model
- get_model_card
- update_model_card
- model_card_from_schema

### DIFF
--- a/lib/python-beta/src/bailo/bailo.py
+++ b/lib/python-beta/src/bailo/bailo.py
@@ -1,7 +1,7 @@
 """Main entry point"""
 from __future__ import annotations
 import requests
-from typing import List, Optional
+from typing import List, Optional, Any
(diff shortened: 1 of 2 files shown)


lib/python-beta/src/bailo/bailo.py
```diff
@@ -50,3 +73,106 @@ def find_models(
...
+        :return: JSON response object
+        """
+        return self.agent.get(
+            f"{self.url}/v2/model/{model_id}",
+        ).json()
+
+    def update_model(
+        self,
+        model_id: str,
+        name: Optional[str] = None,
```
a3957273, line 94: Name, description and visibility could be provided as a dictionary, which would save you from a lot of the repetitive content below.

a3957273 (approved):

PR #747, opened 2023-09-19
Added delete_file method, methods to create or get teams/roles

bw27492:
Methods added:

- delete_file
- get_model_roles
- get_model_user_roles
- create_team
- get_all_teams
- get_user_teams

Also added Model, a new TypeDict class containing name, description and visibility parameters for models. create_model and update_model now use this to validate input

### DIFF
--- a/lib/python-beta/src/bailo/bailo.py
+++ b/lib/python-beta/src/bailo/bailo.py
@@ -238,8 +238,8 @@ class BailoClient():
         ).json()
     
     def get_files(
-            self,
-            model_id: str,
+        self,
+        model_id: str,
     ):
         """
         Gets files for a model.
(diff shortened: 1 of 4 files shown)


lib/python-beta/src/bailo/bailo.py
```diff
@@ -91,35 +86,18 @@ def get_model(
     def update_model(
         self,
         model_id: str,
-        name: Optional[str] = None,
-        description: Optional[str] = None,
-        visibility: Optional[ModelVisibility] = None,
+        model: Model,
```
a3957273, line 89: Name this `diff` or `patch` or similar.  It's not a full model and doesn't include the full required model fields.

lib/python-beta/src/bailo/bailo.py
```diff
@@ -19,32 +19,27 @@ def get(self, *args, **kwargs):
...
     def __init__(self, url: str, agent: Agent = Agent()):
         self.url = url.rstrip("/") + "/api"
         self.agent = agent
 
     def create_model(
             self,
-            name: str,
-            description: str,
-            visibility: ModelVisibility
+            model: Model,
```
a3957273, line 29: I _think_ the syntax for this is actually better as the old argument list, even though it will differ from the `update_model` method.  Comparison:

```python
# old
client.create_model(
  name="Example Model"
  description="Some Description"
  visibility=ModelVisibility.Public
)

# new
client.create_model(
  model={
    "name": "Example Model",
    "description": "Some Description",
    "visibility": ModelVisibility.Public
  }
)
```

This new method is also bad because the `Model` fields are all optional, which isn't accurate.  So it's weakening the typing on this as well.

lib/python-beta/src/bailo/bailo.py
```diff
@@ -272,4 +250,105 @@ def get_files(
         return self.agent.get(
             f"{self.url}/v2/model/{model_id}/files",
         ).json()
+    
+    #def simple_upload(): TBC
```
a3957273, line 254: I like these comments, just to see ones left to be done!

lib/python-beta/src/bailo/bailo.py
```diff
@@ -272,4 +250,105 @@ def get_files(
...
+        """
+        Create new team.
+
+        :param team_id: Unique team ID
+        :param name: Team name
+        :param description: Team description
+        :return: JSON response object
+        """        
+        return self.agent.post(
+            f"{self.url}/v2/teams",
```
a3957273, line 329: You seem to have missed including any data here.

lib/python-beta/src/bailo/model.py
```diff
@@ -0,0 +1,7 @@
+from typing_extensions import TypedDict, NotRequired
+from .enums import ModelVisibility
+
+class Model(TypedDict):
+    name: NotRequired[str]
```
a3957273, line 5: Fun typing!

a3957273 (changes requested):

lib/python-beta/src/bailo/bailo.py
```diff
@@ -93,19 +91,17 @@ def update_model(
         model_id: str,
         name: Optional[str] = None,
         description: Optional[str] = None,
-        visibility: Optional[ModelVisibility] = None,
+        visibility: Optional[str] = None,
```
a3957273, line 94: Isn't this still a `ModelVisibility`?

a3957273: Oh, stupid enum values.  Nevermind.

a3957273 (approved):

PR #750, opened 2023-09-22
Dev/implement post review UI

ARADDCC002:


backend/src/routes.ts
```diff
@@ -225,6 +225,7 @@ if (config.experimental.v2) {
   server.post('/api/v2/schemas', ...postSchemaV2)
 
   server.get('/api/v2/reviews', ...getReviews)
+  server.get('/api/v2/review/:reviewId', ...getReviews)
```
JR40159, line 228: I think this endpoint was added by mistake for the demo but isn't actually needed?

frontend/actions/review.ts
```diff
@@ -25,18 +25,56 @@ export function useGetReviewRequestsForUser(isActive = true) {
...
+
+  return {
+    mutateReviews: mutate,
+    reviews: data ? data.reviews : [],
+    isReviewsLoading: !error && !data,
+    isReviewsError: error,
+  }
+}
+
+// TODO - this API has been removed
```
JR40159, line 56: This reviews endpoint now returns a header with the count. Recommend changing this to a `HEAD` request and returning the value in the header.

JR40159 (changes requested):

backend/src/routes/v2/review/getReview.ts
JR40159, line 1: Don't forget to remove this!

JR40159 (dismissed):

.vscode/settings.json
```diff
@@ -1,8 +1,17 @@
 {
+  "prettier": {
```
ARADDCC012, line 2: This prettier config shouldn't be here as it'll override any changes we make to the settings in our prettierrc files.

The `editor.defaultFormatter` setting below makes sense to leave here though

backend/.prettierignore
```diff
@@ -1,2 +1,3 @@
 dist
 node_modules
+coverage
```
ARADDCC012, line 3: We don't have a `coverage` directory do we?

frontend/actions/review.ts
```diff
@@ -11,7 +11,7 @@ export function useGetReviewRequestsForUser(isActive = true) {
     },
     ErrorInfo
   >(
-    `/api/v2/reviews?${qs.stringify({
+    `/api/v2/reviews/?${qs.stringify({
```
ARADDCC012, line 14: As `active` is a query param on the `/reviews` endpoint, I wouldn't expect to see the additional `/` here.

Should it not be left as was?

frontend/actions/review.ts
```diff
@@ -25,18 +25,51 @@ export function useGetReviewRequestsForUser(isActive = true) {
   }
 }
 
-export function useGetNumReviews() {
+export function useGetReviewRequestsForModel(modelId, semver?, isActive = true) {
```
ARADDCC012, line 28: These params should be typed

frontend/src/common/MarkdownRenderer.tsx
ARADDCC012, line 1: This file has been changed in my branch, so can be removed from here

frontend/package.json
```diff
@@ -36,6 +36,7 @@
     "is-docker": "^3.0.0",
     "jsonschema": "^1.4.1",
     "lodash-es": "^4.17.21",
+    "markdown-to-jsx": "^7.3.2",
```
ARADDCC012, line 39: This can be removed as `frontend/src/common/MarkdownRenderer.tsx` has been changed in my branch

frontend/pages/_document.tsx
```diff
@@ -8,7 +8,7 @@ import { lightTheme } from '../src/theme'
 export default class MyDocument extends Document {
   render() {
     return (
-      <Html lang='en'>
+      <Html lang='en' data-color-mode='light'>
```
ARADDCC012, line 11: What is this doing?

frontend/src/common/ReviewWithComment.tsx
```diff
@@ -1,45 +1,123 @@
...
+  onSubmit,
+  release,
+}: ReviewWithCommentProps) {
+  const { reviews } = useGetReviewRequestsForModel(release.modelId, release.semver, true)
+  const { modelRoles, isModelRolesLoading, isModelRolesError } = useGetModelRoles(reviews[0].model.id)
+
   const [reviewComment, setReviewComment] = useState('')
   const [showError, setShowError] = useState(false)
+  const [selectOpen, setSelectOpen] = useState(false)
+  const [reviewKind, setReviewKind] = useState<ReviewRequestInterface>()
```
ARADDCC012, line 51: Should this not be typed as `ReviewRequestInterface | undefined`?

frontend/src/common/ReviewWithComment.tsx
```diff
@@ -1,45 +1,123 @@
...
   }
 
   function submitForm(kind: ResponseTypeKeys) {
     setShowError(false)
-    if (invalidComment() && (kind === ResponseTypes.Reject || kind === ResponseTypes.RequestChanges)) {
+    if (invalidComment() && kind === ResponseTypes.RequestChanges) {
       setShowError(true)
     } else {
-      onSubmit(kind, reviewComment)
+      if (reviewKind) {
```
ARADDCC012, line 68: If we get to this point in the logic but fail this check, we won't submit the form but also won't give the user any feedback. We should have an `else` condition that displays an informative error.

frontend/src/common/ReviewWithComment.tsx
```diff
@@ -1,45 +1,123 @@
...
+        onSubmit(kind, reviewComment, reviewKind.role)
+      }
     }
   }
 
+  function onChange(_event: React.SyntheticEvent<Element, Event>, newValue: ReviewRequestInterface | null) {
```
ARADDCC012, line 74: Super duper picky, but can we use a named import here like we do for everything else?
`import { SyntheticEvent } from 'react'`

frontend/src/common/ReviewWithComment.tsx
```diff
@@ -1,45 +1,123 @@
...
+              getOptionLabel={(option) => getRoleDisplay(option.role, modelRoles)}
+              options={reviews}
+              renderInput={(params) => (
+                <TextField
+                  {...params}
+                  label='Select your role'
+                  size='small'
+                  InputProps={{
+                    ...params.InputProps,
+                    endAdornment: <>{params.InputProps.endAdornment}</>,
```
ARADDCC012, line 116: What's the point of this? It looks like a long-winded way of writing `InputProps={params.InputProps}` to me and, as we're already spreading `params` above, this whole `InputProps` part can probably be removed.

frontend/src/model/beta/releases/ModelReleaseReviewBanner.tsx
```diff
@@ -3,17 +3,25 @@ import Button from '@mui/material/Button'
...
+import { mutate } from 'swr'
 
+import { postReviewResponse } from '../../../../actions/review'
+import { ReleaseInterface } from '../../../../types/types'
 import ReviewWithComment, { ResponseTypeKeys } from '../../../common/ReviewWithComment'
 
 type ModelReleaseReviewBannerProps = {
   label: string
+  release: ReleaseInterface
+  mutateReviews: () => void
```
ARADDCC012, line 17: Instead of passing this in as a prop, with no real safety on what's provided by the consumer of this component, could we instead do the following in this component?
```
const { mutateReviews } = useGetReviewRequestsForModel(release.modelId, release.semver, true)
const { mutateInactiveReviews } = useGetReviewRequestsForModel(release.modelId, release.semver, true)
```
We can then guarantee that we're calling the right mutate functions from this component, leaving no room for error for anyone using this component in future. It also allows us to call mutate from but active and inactive reviews.

frontend/src/model/beta/releases/ModelReleaseReviewBanner.tsx
```diff
@@ -24,8 +32,19 @@ export default function ModelReleaseReviewBanner({ label }: ModelReleaseReviewBa
     setReviewCommentOpen(false)
   }
 
-  const handleSubmit = (_kind: ResponseTypeKeys, _reviewComment: string) => {
-    //TODO some response to API endpoint- BAI-858
+  const handleSubmit = (kind: ResponseTypeKeys, reviewComment: string, reviewRole: string) => {
+    postReviewResponse(release.modelId, release.semver, reviewRole, reviewComment, kind).then(() => {
```
ARADDCC012, line 36: Should use async/await syntax instead of promise chains

frontend/src/model/beta/releases/ModelReleaseReviewDisplay.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const { modelRoles, isModelRolesLoading, isModelRolesError } = useGetModelRoles(review.model.id)
+
+  if (isModelRolesError) {
+    return <MessageAlert message={isModelRolesError.info.message} severity='error' />
+  }
+  return (
+    <>
+      {isModelRolesLoading && <Loading />}
+      <Stack direction={{ sm: 'row', xs: 'column' }}>
+        {review.responses.map((response) => {
```
ARADDCC012, line 26: This is a somewhat complex iteration. Can you pull it out into a `useMemo`?

frontend/src/model/beta/releases/ModelReleaseReviewDisplay.tsx
```diff
@@ -0,0 +1,54 @@
...
+          const [_kind, username] = response.user.split(':')
+          switch (response.decision) {
+            case ApprovalStates.Accepted:
+              return (
+                <Stack direction='row' spacing={1}>
+                  <Done color='success' fontSize='small' />
+                  <Typography variant='caption'>
+                    <span style={{ fontWeight: 'bold' }}>{username}</span>
```
ARADDCC012, line 34: Imo we should stick to using `sx` as we do for the rest of the app. Consider using a `<Box component='span' sx={{ 
fontWeight: 'bold' }}>` instead

frontend/src/model/beta/releases/ModelReleaseReviewDisplay.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    {` has approved this release (${getRoleDisplay(review.role, modelRoles)})`}
+                  </Typography>
+                </Stack>
+              )
+            case ApprovalStates.RequestChanges:
+              return (
+                <Stack direction='row' spacing={1}>
+                  <HourglassEmpty color='warning' fontSize='small' />
+                  <Typography variant='caption'>
+                    <span style={{ fontWeight: 'bold' }}>{username}</span>
```
ARADDCC012, line 44: See https://github.com/gchq/Bailo/pull/750/files#r1342658426

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -80,10 +80,18 @@ export default function SideNavigation({
...
+  const [reviewCount, setReviewCount] = useState(0)
 
   const theme = useTheme()
 
+  useEffect(() => {
+    fetchReviewCount()
+  }, [])
+
+  async function fetchReviewCount() {
+    setReviewCount((await getReviewCount()).headers.get('x-count') as unknown as number)
```
ARADDCC012, line 92: It doesn't look like there's any error handling around this. We should at least have some basic 200 status checks right?

frontend/src/common/ReviewWithComment.tsx
```diff
@@ -113,7 +120,6 @@ export default function ReviewWithComment({
                   size='small'
                   InputProps={{
```
ARADDCC012, line 121: `InputProps` doesn't need to be specified at all as it's already covered by `{...params}` above

frontend/src/model/beta/releases/ModelReleaseReviewBanner.tsx
```diff
@@ -3,26 +3,29 @@ import Button from '@mui/material/Button'
...
-  mutateReviews: () => void
 }
 
-export default function ModelReleaseReviewBanner({ label, release, mutateReviews }: ModelReleaseReviewBannerProps) {
+export default function ModelReleaseReviewBanner({ label, release }: ModelReleaseReviewBannerProps) {
   const theme = useTheme()
 
   const { mutateReleases } = useGetReleasesForModelId(release.modelId)
+  const { mutateReviews: mutateActiveReviews } = useGetReviewRequestsForModel(release.modelId, release.semver, true)
+  const { mutateReviews: mutateInactiveReviews } = useGetReviewRequestsForModel(release.modelId, release.semver, true)
```
ARADDCC012, line 25: the active param should be `false` here: `useGetReviewRequestsForModel(release.modelId, release.semver, false)`

frontend/src/wrapper/SideNavigation.tsx
ARADDCC002: TODO added to handle this (and other instances like it) in another ticket

ARADDCC012: Agreed we need accessible snackbar functionality to properly resolve this. This work will be spun out into a separate ticket to address this sort of problem across the entire app.

backend/src/services/v2/model.ts
```diff
@@ -24,7 +24,7 @@ export async function createModel(user: UserDoc, modelParams: CreateModelParams)
     collaborators: [
       {
         entity: toEntity('user', user.dn),
-        roles: ['owner'],
+        roles: ['owner', 'msro', 'mtr'],
```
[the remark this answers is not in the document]
ARADDCC002: This is hard-coded for now until the we've implemented roles and updating model card roles in the backend.

PR #754, opened 2023-09-29
Fix marketplace search

JR40159:


backend/src/models/v2/Model.ts
```diff
@@ -49,7 +49,7 @@ const ModelSchema = new Schema<ModelInterface>(
   {
     id: { type: String, required: true, unique: true, index: true },
 
-    name: { type: String, required: true },
+    name: { type: String, text: true, required: true },
```
a3957273, line 52: You likely want to run this:

```javascript
ModelSchema.index({ '$**': 'text' })
```

Which will index all fields?  We'll also want to index the model card fields.

a3957273 (changes requested):

a3957273 (approved):

PR #773, opened 2023-10-03
fixed issue with release page crashing after review

ARADDCC002:


frontend/src/common/ReviewWithComment.tsx
```diff
@@ -90,66 +90,68 @@ export default function ReviewWithComment({
...
-            >
-              <Button onClick={onClose}>Cancel</Button>
-              <Stack spacing={2} direction={{ sm: 'row', xs: 'column' }}>
-                <Button variant='outlined' onClick={() => submitForm(ResponseTypes.RequestChanges)}>
-                  Request Changes
-                </Button>
-                <Button variant='contained' onClick={() => submitForm(ResponseTypes.Approve)}>
-                  Approve
-                </Button>
+      {!modelRoles.length && isModelRolesLoading && <Loading />}
```
ARADDCC012, line 93: I don't think we need `!modelRoles.length` in this check. The loading spinner is there purely to inform the user if data if being fetched. The `models.length` check below should be enough to stop the crashing.

frontend/src/common/ReviewWithComment.tsx
```diff
@@ -90,66 +90,68 @@ export default function ReviewWithComment({
...
+      {modelRoles.length && (
```
ARADDCC012, line 94: Do we need to handle the case where the data has successfully been fetched but no modelRoles were found? I imagine not, but just confirming.

ARADDCC002: I've added a little error message, it probably won't ever crop up though.

ARADDCC012 (approved):

PR #774, opened 2023-10-03
added tabbed views as url params so they can be easily shared with users

ARADDCC002:


frontend/src/Form/beta/ModelCardForm.tsx
```diff
@@ -41,6 +54,13 @@ export default function ModelCardForm({
     return <></>
   }
 
+  function listButtonOnClick(index: number, formPageKey: string) {
```
ARADDCC012, line 57: Again, shout if you'd rather not follow this convention, but consider renaming to `handleListItemClick`

frontend/src/common/PageWithTabs.tsx
```diff
@@ -22,11 +24,23 @@ export default function PageWithTabs({
...
-    setValue(newValue)
+  const router = useRouter()
+
+  const tab = router.query.tab
+
+  useEffect(() => {
+    tab && tabs.length ? setCurrentTab(tab as string) : setCurrentTab(tabs[0].path)
+  }, [tab, setCurrentTab, tabs])
+
+  const handleChange = (_event: React.SyntheticEvent, newValue: string) => {
```
ARADDCC012, line 37: Use named import for `SyntheticEvent`?

frontend/src/common/PageWithTabs.tsx
```diff
@@ -22,11 +24,23 @@ export default function PageWithTabs({
...
+  const [currentTab, setCurrentTab] = useState('')
 
-  const handleChange = (_event: React.SyntheticEvent, newValue: number) => {
-    setValue(newValue)
+  const router = useRouter()
+
+  const tab = router.query.tab
+
+  useEffect(() => {
+    tab && tabs.length ? setCurrentTab(tab as string) : setCurrentTab(tabs[0].path)
```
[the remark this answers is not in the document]
ARADDCC002: It can also be of type `string[]` which is why the type assertion is done here

ARADDCC012 (approved):

PR #791, opened 2023-10-17
Bai 876 add history button

TT38665:


backend/src/services/v2/model.ts
```diff
@@ -147,6 +147,18 @@ export async function getModelCardRevision(user: UserDoc, modelId: string, versi
   return modelCard
 }
 
+export async function getModelCardRevisions(user: UserDoc, modelId: string) {
+  const modelCardRevisions = await ModelCardRevisionModel.find({ modelId })
+
+  if (!modelCardRevisions) {
```
ARADDCC002, line 153: I think this needs revisiting. Currently `ModelCardRevisionModel.find()` will just return an empty array, which means it shouldn't ever be undefined. 

I would probably do an additional check before this line to make sure that the `modelId` returns an actual model object. Then you can do your undefined error check on that.

backend/test/testUtils/testModels.ts
```diff
@@ -63,3 +63,9 @@ export const testReleaseReview = {
   createdAt: new Date('08/13/2023'),
   updatedAt: new Date('08/14/2023'),
 }
+
+export const testModelCardRevision = {
+  modelId: 'abc',
+  CreatedBy: 'user',
+  createdAt: new Date('08/13/2023'),
```
ARADDCC002, line 70: Unless you're checking for a specific date this can just be `new Date()`

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -69,31 +79,37 @@ export default function FormEditPage({ model }: FormEditPageProps) {
...
+          divider={<Divider orientation='vertical' flexItem />}
+          sx={{ mb: { xs: 2 } }}
+        >
+          {!isEdit && (
+            <Button variant='outlined' onClick={() => setDialogOpen(true)}>
+              View History
+            </Button>
+          )}
+          {!isEdit && (
+            <Button variant='outlined' onClick={handleEdit}>
```
ARADDCC002, line 95: Is there a reason why `setIsEdit(!isEdit)` was changed use `handleEdit()`?

frontend/src/model/beta/overview/ModelCardHistoryDialog.tsx
```diff
@@ -0,0 +1,63 @@
...
+  )
+
+  if (isModelCardRevisionsError) {
+    return <MessageAlert message={isModelCardRevisionsError.info.message} severity='error' />
+  }
+
+  return (
+    <>
+      {isModelCardRevisionsLoading && <Loading />}
+      <ThemeProvider theme={theme}>
```
ARADDCC002, line 40: We don't need to use this provider as it's already provided at the top level of the application. Calling `theme` will still work within your JSX.

frontend/src/model/beta/overview/ModelCardRevisionListDisplay.tsx
```diff
@@ -0,0 +1,31 @@
...
+type revisionProp = {
+  modelCard: ModelCardRevisionInterface
+}
+export default function ModelCardRevisionListDisplay({ modelCard }: revisionProp) {
+  const router = useRouter()
+  const { modelId, modelCardVersion }: { modelId?: string; modelCardVersion?: number } = router.query
+  const theme = useTheme()
+
+  return (
+    <ThemeProvider theme={theme}>
```
ARADDCC002, line 17: We don't need to use this provider as it's already declared at the top level of this application.

frontend/src/model/beta/overview/ModelCardRevisionListDisplay.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <TableBody>
+        <TableRow
+          onClick={() => router.push(`/beta/model/${modelId}/history/${modelCardVersion}`)}
+          // this route leads to blank page that says 'undefined'
```
ARADDCC002, line 21: Is this comment still relevant? If so it'll need to be fixed

frontend/src/model/beta/overview/ModelCardRevisionListDisplay.tsx
```diff
@@ -0,0 +1,31 @@
...
+          sx={{ '&:hover': { backgroundColor: '#f0f0f0', cursor: 'pointer' } }}
```
ARADDCC002, line 22: We shouldn't really use custom codes like this and should instead change our theme. Happy to help with this tomorrow if you need

ARADDCC002 (changes requested):

frontend/src/model/beta/overview/ModelCardRevisionListDisplay.tsx
ARADDCC002: Actually, you can remove `backgroundColor: : '#f0f0f0'` entirely, and add a prop called `hover` to your `<TableRow />`

frontend/actions/modelCard.ts
```diff
@@ -47,3 +47,19 @@ export function useModelCard(modelId?: string, modelCardVersion?: number) {
     isModelError: error,
   }
 }
+
+export function useModelCardRevisions(modelId: string) {
```
ARADDCC002, line 51: Can we rename this to `useGetModelCardRevisions`

ARADDCC002 (approved):

PR #792, opened 2023-10-17
added file uploads to post releases

ARADDCC002:
Files can now be uploaded as part of a release. The files are individually uploaded, and the `id` of each is added to the release model.

Currently we are just displaying those `id`s on the release display, but in the future we will expand on this.

image

### DIFF
--- a/frontend/actions/release.ts
+++ b/frontend/actions/release.ts
@@ -26,3 +26,11 @@ export async function postRelease(release: Partial<ReleaseInterface>, modelId: s
     body: JSON.stringify(release),
   })
 }
+
+export async function postArtefact(artefact: File, modelId: string, name: string, mime: string) {
+  return fetch(`/api/v2/model/${modelId}/files/upload/simple?name=${name}&mine=${mime}`, {
+    method: 'post',
+    headers: { 'Content-Type': 'application/json' },
+    body: artefact,
+  })
+}
--- a/frontend/actions/release.ts
+++ b/frontend/actions/release.ts
@@ -27,8 +27,8 @@ export async function postRelease(release: Partial<ReleaseInterface>, modelId: s
   })
 }
 
-export async function postArtefact(artefact: File, modelId: string, name: string, mime: string) {
-  return fetch(`/api/v2/model/${modelId}/files/upload/simple?name=${name}&mine=${mime}`, {
+export async function postFile(artefact: File, modelId: string, name: string, mime: string) {
+  return fetch(`/api/v2/model/${modelId}/files/uploadss/simple?name=${name}&mine=${mime}`, {
     method: 'post',
     headers: { 'Content-Type': 'application/json' },
     body: artefact,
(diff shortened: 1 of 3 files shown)


frontend/actions/release.ts
```diff
@@ -26,3 +26,11 @@ export async function postRelease(release: Partial<ReleaseInterface>, modelId: s
     body: JSON.stringify(release),
   })
 }
+
+export async function postArtefact(artefact: File, modelId: string, name: string, mime: string) {
```
a3957273, line 30: This should be called 'postFile'.  In general, 'artefacts' include both docker images and files.  Whereas files are just files!

frontend/src/model/beta/releases/DraftNewReleaseDialog.tsx
```diff
@@ -43,34 +44,49 @@ export default function DraftNewReleaseDialog({
...
+    setLoading(true)
     if (!model.card.version) {
+      setLoading(false)
       return setErrorMessage('Please make sure your model has a schema set before drafting a release.')
     }
     if (isValidSemver(semanticVersion)) {
+      const fileIds: string[] = []
+      for (const artefact of artefacts) {
+        const postArtefactResponse = await postArtefact(artefact, model.id, artefact.name, artefact.type)
+        if (postArtefactResponse.status === 200) {
```
a3957273, line 61: You don't provide any warnings if an error does occur.  You also shouldn't check status is equal to 200, there's the `ok` attribute to determine if a request passed.

frontend/src/model/beta/releases/DraftNewReleaseDialog.tsx
```diff
@@ -43,34 +44,49 @@ export default function DraftNewReleaseDialog({
...
+          await postArtefactResponse.json().then((res: PostSimpleUpload) => {
```
a3957273, line 62: You are using both `await` and `.then`.  You only need to use one.

```
const res = await postArtefactResponse.json()
fileIds.push(res.file._id)
```

a3957273 (changes requested):

frontend/src/model/beta/releases/DraftNewReleaseDialog.tsx
```diff
@@ -43,34 +44,53 @@ export default function DraftNewReleaseDialog({
...
     if (isValidSemver(semanticVersion)) {
+      const fileIds: string[] = []
+      for (const artefact of artefacts) {
+        const postArtefactResponse = await postFile(artefact, model.id, artefact.name, artefact.type)
+        if (postArtefactResponse.ok) {
+          const res = await postArtefactResponse.json()
+          fileIds.push(res.file._id)
+        } else {
+          setLoading(false)
+          return setErrorMessage(
```
a3957273, line 66: Use `getErrorMessage(res)` here to generate the error message.

a3957273 (changes requested):

a3957273 (approved):

PR #807, opened 2023-10-24
Bug/Remove bubble when no reviews

GB27247:


frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -147,12 +147,12 @@ export default function SideNavigation({
                   <ListItemIcon>
                     {!drawerOpen ? (
                       <Tooltip title='Review' arrow placement='right'>
-                        <Badge badgeContent={reviewCount} color='secondary'>
+                        <Badge badgeContent={reviewCount} color='secondary' invisible={reviewCount == 0}>
```
a3957273, line 150: Always use strict equality `===` when comparing values.

a3957273 (changes requested):

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -91,7 +91,7 @@ export default function SideNavigation({
   }, [])
 
   async function fetchReviewCount() {
-    setReviewCount((await getReviewCount()).headers.get('x-count') as unknown as number)
+    setReviewCount((await getReviewCount()).headers.get('x-count') as string)
```
a3957273, line 94: I think it would be cleaner to parse it as a number here.  E.g:

```
setReviewCount(parseInt((await getReviewCount()).headers.get('x-count')))
```

Then reviewCount can remain a number, and you can compare it as a number

a3957273 (changes requested):

frontend/src/wrapper/SideNavigation.tsx
GB27247: Is it be preferable to raise an error on a null response or set the value to 0?

ARADDCC002: IMO we should probably send some sort of feedback if it can't find it and then set it to 0 so that it hides the bubble. 

We use the `useNotification` hook as defined [here](https://github.com/gchq/Bailo/blob/main/frontend/src/common/Snackbar.tsx) to show snackbar type feedback to user. There's an example of implementation (here)[https://github.com/gchq/Bailo/blob/main/frontend/src/Form/RenderButtons.tsx#L42-L46].

a3957273 (approved):

PR #810, opened 2023-10-25
implemented a mutatation function that updates the ui when selecting …

TT38665:
…a schema

### DIFF
--- a/frontend/pages/beta/model/[modelId]/schema.tsx
+++ b/frontend/pages/beta/model/[modelId]/schema.tsx
@@ -74,7 +74,7 @@ interface SchemaButtonProps {
 }
 
 function SchemaButton({ modelId, schema }: SchemaButtonProps) {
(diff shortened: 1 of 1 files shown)


frontend/pages/beta/model/[modelId]/schema.tsx
```diff
@@ -90,6 +90,7 @@ function SchemaButton({ modelId, schema }: SchemaButtonProps) {
   async function createModelUsingSchema(newSchema: SchemaInterface) {
     if (currentUser && model) {
       await postFromSchema(model.id, newSchema.id)
+      await mutateModel()
```
Comment (ARADDCC002) [inline frontend/pages/beta/model/[modelId]/schema.tsx:93]: You don't need to await this mutate as we aren't expecting a response, it should happen in the background.

ARADDCC002 (changes requested):

ARADDCC002 (approved):

PR #811, opened 2023-10-25
Feature/bai 953 update model name desc and visibility

ARADDCC002:


frontend/src/common/PageWithTabs.tsx
```diff
@@ -36,9 +39,15 @@ export default function PageWithTabs({
 
   const handleChange = (_event: SyntheticEvent, newValue: string) => {
     setCurrentTab(newValue)
-    router.replace({
-      query: { ...router.query, tab: newValue },
-    })
+    if (requiredUrlParams) {
```
a3957273, line 42: If you set a default value for `requiredUrlParams` of `{}` then you can remove the 'if' statement and simplify the code.

frontend/src/model/beta/Settings.tsx
```diff
@@ -1,21 +1,34 @@
...
 export default function Settings({ model }: SettingsProps) {
-  const [selectedCategory, setSelectedCategory] = useState<SettingsCategory>('general')
+  const router = useRouter()
+
+  const { section } = router.query
+
+  const [selectedCategory, setSelectedCategory] = useState<SettingsCategory>('details')
+
+  useEffect(() => {
+    section ? setSelectedCategory(section as SettingsCategory) : setSelectedCategory('details')
```
a3957273, line 24: This type assertion is unsafe, you should use a type guard to make sure this is okay.  This line can also be replaced with:

```
setSelectedCategory(section ?? 'details')
```

Using the null coalescing operator.

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -52,9 +57,23 @@ export default function ModelAccess({ model }: ModelAccessProps) {
     }
   }
 
-  // TODO - add a request to update the model's collaborators field
   async function updateAccessList() {
-    await patchModel(model.id, { collaborators: accessList })
+    const res = await patchModel(model.id, { collaborators: accessList })
```
a3957273, line 61: Always handle errors first and keep the indentation size of the 'success' branch linear.  For instance, in this case you could achieve that with:

```javascript
const res = await patchModel(model.id, { collaborators: accessList })

if (!res.ok) {
  return sendNotification({
    variant: 'success',
    msg: error,
    anchorOrigin: { horizontal: 'center', vertical: 'bottom' },
  })
}

sendNotification({
  variant: 'success',
  msg: 'Model access list updated',
  anchorOrigin: { horizontal: 'center', vertical: 'bottom' },
})
mutateModel()
```

This matters less for short functions like this.

frontend/src/model/beta/settings/ModelDetails.tsx
```diff
@@ -0,0 +1,154 @@
...
+  FormControlLabel,
+  Radio,
+  RadioGroup,
+  Stack,
+  TextField,
+  Tooltip,
+  Typography,
+} from '@mui/material'
+import { useTheme } from '@mui/material/styles'
+import _ from 'lodash-es'
```
a3957273, line 16: I think this breaks tree-shaking.  Explicitly import the things you want.  Also I think this is actually unused??

a3957273 (changes requested):

a3957273 (approved):

PR #812, opened 2023-10-25
Fixed ReviewsList keys and links

ARADDCC012:


frontend/src/reviews/ReviewItem.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { timeDifference } from 'utils/dateUtils'
+
+type ReviewItemProps = {
+  review: ReviewRequestInterface
+}
+
+export default function ReviewItem({ review }: ReviewItemProps) {
+  const router = useRouter()
+
+  function handleListItemClick() {
```
ARADDCC012, line 13: This function has been changed, the rest of the file is the same as it was when the component was in `ReviewsList.tsx`

ARADDCC002 (approved):

PR #815, opened 2023-10-26
added a context for handling unsaved changes

ARADDCC002:


frontend/src/contexts/unsavedChangesContext.ts
```diff
@@ -0,0 +1,11 @@
+import { createContext } from 'react'
+
+import { UnsavedChangesHook } from '../../utils/hooks/useUnsavedChanges'
+
+const UnsavedChangesContext = createContext<UnsavedChangesHook>({
+  unsavedChanges: false,
+  setUnsavedChanges: () => null,
```
ARADDCC012, line 7: Super nit-picky but a standard JS function that doesn't return anything would return `undefined` instead of `null`. `null` could suggest we are returning that value for a reason, but in this case we just want a function that does nothing.

frontend/utils/hooks/useUnsavedChanges.ts
```diff
@@ -0,0 +1,44 @@
...
+export type UnsavedChangesHook = {
+  unsavedChanges: boolean
+  setUnsavedChanges: (newValue: boolean) => void
+  sendWarning: () => void
+}
+
+export default function useUnsavedChanges(): UnsavedChangesHook {
+  const [unsavedChanges, setUnsavedChanges] = useState(false)
+
+  const warningText = 'Any unsaved changes will be lost - are you sure you wish to leave this page?'
```
ARADDCC012, line 13: Minor, but this could be defined outside of the hook so it's not recreated on each render cycle.

frontend/src/common/PageWithTabs.tsx
```diff
@@ -35,10 +37,14 @@ export default function PageWithTabs({
...
 
   const handleChange = (_event: SyntheticEvent, newValue: string) => {
-    setCurrentTab(newValue)
-    router.replace({
-      query: { ...router.query, tab: newValue },
-    })
+    if (unsavedChanges) {
+      sendWarning()
+    } else {
+      setCurrentTab(newValue)
```
ARADDCC012, line 43: If the user confirms that they do want to change tabs does this logic get hit? It looks like it wouldn't to me.

Maybe we need `sendWarning` to return the user's response to the popup window? We can then conditionally continue with the next piece of logic as normal, if the user confirmed, or skip the logic, if the user canceled.

ARADDCC012 (approved):

PR #816, opened 2023-10-27
Bugfix/bai 966 add loading button

TT38665:


frontend/pages/beta/model/new.tsx
```diff
@@ -39,6 +40,7 @@ export default function NewModel() {
 
   async function onSubmit(event) {
     event.preventDefault()
+    setLoading(true)
```
a3957273, line 43: You never run `setLoading(false)`!  If the 'createModel' request fails (e.g. with a missing field) you'll never be able to resubmit it.

frontend/pages/beta/model/new.tsx
```diff
@@ -151,9 +153,16 @@ export default function NewModel() {
               <Box sx={{ textAlign: 'right' }}>
                 <Tooltip title={!formValid ? 'Please make sure all required fields are filled out' : ''}>
                   <span>
-                    <Button variant='contained' disabled={!formValid} type='submit' data-test='createModelButton'>
+                    <LoadingButton
+                      variant='contained'
+                      disabled={!formValid}
+                      type='submit'
+                      data-test='createModelButton'
+                      onClick={onSubmit}
```
a3957273, line 161: You shouldn't need this `onSubmit` here, because the form has an `onSubmit` call.

frontend/src/common/ReviewWithComment.tsx
```diff
@@ -83,6 +85,7 @@ export default function ReviewWithComment({
 
   function submitForm(decision: ResponseTypeKeys) {
     setShowError(false)
+    setLoading(true)
```
a3957273, line 88: Similar comment, no `setLoading(false)` call.

frontend/src/model/beta/Releases.tsx
```diff
@@ -30,15 +31,20 @@ export default function Releases({ model }: { model: ModelInterface }) {
...
+  function handleButtonClick() {
+    setOpenDraftNewRelease(true)
+    setLoading(true)
+  }
   return (
     <Box sx={{ maxWidth: '900px', mx: 'auto', my: 4 }}>
       <Stack spacing={4}>
         <Box sx={{ textAlign: 'right' }}>
-          <Button variant='outlined' onClick={() => setOpenDraftNewRelease(true)} disabled={!model.card}>
+          <LoadingButton variant='outlined' onClick={handleButtonClick} loading={loading} disabled={!model.card}>
```
a3957273, line 45: This button doesn't make any network requests or anything.  It just opens a modal.  I don't think there's a need to make this a loading button.

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -55,6 +56,7 @@ export default function ModelAccess({ model }: ModelAccessProps) {
   // TODO - add a request to update the model's collaborators field
   async function updateAccessList() {
     await patchModel(model.id, { collaborators: accessList })
+    setLoading(true)
```
a3957273, line 59: Loading never disabled!

a3957273 (changes requested):

frontend/src/model/beta/Releases.tsx
```diff
@@ -30,15 +31,20 @@ export default function Releases({ model }: { model: ModelInterface }) {
 
   function handleDraftNewReleaseClose() {
     setOpenDraftNewRelease(false)
+    setLoading(false)
   }
 
+  function handleButtonClick() {
```
ARADDCC012, line 37: Rename to `handleDraftNewRelease` for clarity?

frontend/src/model/beta/Releases.tsx
ARADDCC012, line 45: This LoadingButton isn't needed as it's only being used to open a dialog. There isn't anything we're waiting for here.

frontend/src/model/beta/Settings.tsx
```diff
@@ -48,9 +50,9 @@ export default function Settings({ model }: SettingsProps) {
             <Typography variant='h6' component='h2'>
               Danger Zone!
             </Typography>
-            <Button variant='contained' disabled>
+            <LoadingButton variant='contained' disabled onClick={() => setLoading(true)} loading={loading}>
```
ARADDCC012, line 53: Can we pull the handleClick logic into a separate function with a TODO? Makes it clearer what our intentions are in future.
```
const handleDelete = () => {
  setLoading(true)
  
  // TODO - Delete model API request and setLoading(false) on error
}

onClick={handleDeleteModel}
```

frontend/src/model/beta/overview/TemplatePage.tsx
```diff
@@ -68,9 +72,9 @@ export default function TemplatePage({ model }: TemplatePageProps) {
                 Create from scratch
               </Typography>
               <Typography variant='body1'>Create a model from scratch using a predefined schema.</Typography>
-              <Button variant='contained' onClick={handleCreateFromScratchClick}>
+              <LoadingButton variant='contained' onClick={handleCreateFromScratchClick} loading={loading}>
```
ARADDCC012, line 75: This button also isn't needed, there's no API interaction here so we're not waiting for anything. All we're doing is changing the route (which just changes which page the user is on).

frontend/src/model/beta/overview/TemplatePage.tsx
```diff
@@ -12,9 +14,11 @@ type TemplatePageProps = {
 export default function TemplatePage({ model }: TemplatePageProps) {
   const theme = useTheme()
   const router = useRouter()
+  const [loading, setLoading] = useState(false)
 
   function handleCreateFromScratchClick() {
     router.push(`/beta/model/${model.id}/schema`)
+    setLoading(true)
```
ARADDCC012, line 21: Quick note that this line is pointless. `router.push` will change the page the user is on, so will render a different set of components. As this component is no longer being rendered (and therefore the button is no longer being rendered) the loading state won't be shown to the user.

frontend/src/model/beta/settings/AccessRequestSettings.tsx
```diff
@@ -14,7 +16,9 @@ export default function AccessRequestSettings() {
         <Typography>Allow users to make ungoverned access requests</Typography>
       </Stack>
       <div>
-        <Button>Save</Button>
+        <LoadingButton onClick={() => setLoading(true)} loading={loading}>
```
ARADDCC012, line 19: Again, can we pull this out into a separate function with a TODO?
```
const handleSave = () => {
  setLoading(true)
  
  TODO - Save settings API request and setLoading(false) on error
}

<LoadingButton onClick={handleSave} loading={loading}>
```

frontend/src/model/beta/settings/ModelAccess.tsx
ARADDCC012, line 59: We should always set the loading state before making the API request (that's the thing we're waiting for!). This should be moved to the line above `await patchModel`...

It's already been pointed out, but you'll need to `setLoading(false)` if `patchModel` errors. Unfortunately this request doesn't have any error handling at the moment (although it definitely should!), so you'll need to add that in also:
```
// You'll need a new state variable
const [patchModelErrorMessage, setPatchModelErrorMessage] = useState('')

async function updateAccessList() {
    setLoading(true)

    const response = await patchModel(model.id, { collaborators: accessList })

    if (!response.ok) {
      const error = await getErrorMessage(response)
      setLoading(false)  // Notice we have asynchronous logic on the line above, so we also want to wait for that to finish before setting the loading state back to false
     setPatchModelErrorMessage(error)
    }

    // TODO - Provide user feedback on success
}

// You'll also need to add the following (I've included the isUsersError logic so you know where to put it):
if (patchModelErrorMessage) {
  return <MessageAlert message={patchModelErrorMessage} severity='error' />
}

if (isUsersError) {
  return <MessageAlert message={isUsersError.info.message} severity='error' />
}
```

ARADDCC012: It's already been pointed out, but you'll need to `setLoading(false)` if `patchModel` errors. Unfortunately this request doesn't have any error handling at the moment (although it definitely should!), so you'll need to add that in also (the TODO is not for you to do in this ticket):
```
// You'll need a new state variable
const [patchModelErrorMessage, setPatchModelErrorMessage] = useState('')

async function updateAccessList() {
    setLoading(true)

    const response = await patchModel(model.id, { collaborators: accessList })

    if (!response.ok) {
      const error = await getErrorMessage(response)
      setLoading(false)  // Notice we have asynchronous logic on the line above, so we also want to wait for that to finish before setting the loading state back to false
      setPatchModelErrorMessage(error)
    }

    // TODO - Provide user feedback on success
}

// You'll also need to add the following (I've included the isUsersError logic so you know where to put it):
if (patchModelErrorMessage) {
  return <MessageAlert message={patchModelErrorMessage} severity='error' />
}

if (isUsersError) {
  return <MessageAlert message={isUsersError.info.message} severity='error' />
}
```

frontend/src/common/ReviewWithComment.tsx
```diff
@@ -83,10 +85,13 @@ export default function ReviewWithComment({
 
   function submitForm(decision: ResponseTypeKeys) {
     setShowError(false)
+    setLoading(false)
     if (invalidComment() && decision === ResponseTypes.RequestChanges) {
       setShowError(true)
+      setLoading(false)
     } else {
       onSubmit(decision, reviewComment, reviewRequest.role)
```
ARADDCC012, line 93: You don't need all of these `setLoading` calls. We should manage the loading state in the same place we're interacting with the API. In this case that's not in this component. All this component does is tell the parent component that it wants to submit some data. It's then up to that parent component to decide what to do with that data.

So in this case `ReviewBanner` is taking this data and sending it off to the backend. That means you'll want to manage the loading state in `ReviewBanner` also.

I suggest you change `ReviewWithComment` to have a new prop called `loading`. This will just be a boolean that you can pass to your `LoadingButton`s. Then you can have `const [loading, setLoading] = useState(false)` and your `setLoading(true)` and `setLoading(false)` in `ReviewBanner`. You can then pass this `loading` state value from `ReviewBanner` into `ReviewWithComment`.

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -118,9 +120,11 @@ export default function FormEditPage({ model }: FormEditPageProps) {
               <Button variant='outlined' onClick={onCancel}>
                 Cancel
               </Button>
-              <Button variant='contained' onClick={onSubmit}>
-                Save
-              </Button>
+              {!loading && (
```
ARADDCC012, line 123: Why are you conditionally rendering your `LoadingButton`? This can be removed as we always want to render the `LoadingButton`, regardless of whether or not we're in a loading state.

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -54,11 +57,24 @@ export default function ModelAccess({ model }: ModelAccessProps) {
...
+      const error = await getErrorMessage(response)
+      setLoading(false)
+      setPatchModelErrorMessage(error)
+    }
+
+    // TODO - Provide user feedback on success
 
-  if (isUsersError) {
-    return <MessageAlert message={isUsersError.info.message} severity='error' />
+    if (patchModelErrorMessage) {
```
ARADDCC012, line 71: The following should not be inside this function:
```
    if (patchModelErrorMessage) {
      return <MessageAlert message={patchModelErrorMessage} severity='error' />
    }

    if (isUsersError) {
      return <MessageAlert message={isUsersError.info.message} severity='error' />
    }
```

They are used to return a `MessageAlert` instead of the main `return` block at the bottom in the event of an error occurring. If they are inside this function they will return the `MessageAlert` to whatever called `updateAccessList`

ARADDCC012 (changes requested):
My other comments will need to be addressed before this can be merged.

frontend/src/model/beta/reviews/ReviewBanner.tsx
```diff
@@ -85,6 +87,7 @@ export default function ReviewBanner({ release, accessRequest, square = false }:
       mutateReleases()
       mutateAccessRequests()
       setIsReviewOpen(false)
+      setLoading(false)
```
ARADDCC012, line 90: This should be moved outside of this `if` statement. We want to `setLoading(false)` regardless of whether or not we get a success/error response. We're just using `loading` to track the time spent waiting for the API to respond.

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -53,6 +54,7 @@ export default function FormEditPage({ model }: FormEditPageProps) {
       }
       setSplitSchema({ reference: schema.id, steps })
       setIsEdit(false)
+      setLoading(false)
```
ARADDCC012, line 57: This isn't needed

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -21,22 +22,24 @@ type FormEditPageProps = {
...
 
   async function onSubmit() {
+    setLoading(true)
     if (schema) {
       const data = getStepsData(splitSchema, true)
       const res = await putModelCard(model.id, data)
       if (res.status && res.status < 400) {
         setIsEdit(false)
-
+        setLoading(false)
```
ARADDCC012, line 42: This should be moved outside of this `if` statement. WE want to `setLoading(false)` when we get a response, regardless of what that response is.

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -58,11 +60,13 @@ export default function ModelAccess({ model }: ModelAccessProps) {
   }
 
   async function updateAccessList() {
+    setLoading(true)
     const res = await patchModel(model.id, { collaborators: accessList })
     if (!res.ok) {
+      setLoading(false)
```
ARADDCC012, line 66: This can be removed. We're already doing this on line 80 (which is the correct place to do it). This one is in the wrong place and adding no value.

frontend/pages/beta/model/new.tsx
```diff
@@ -50,6 +52,7 @@ export default function NewModel() {
 
     if (!response.ok) {
       const error = await getErrorMessage(response)
+      setLoading(false)
```
ARADDCC012, line 55: This should be moved outside of this `if` statement. We always want to `setLoading(false) when we get a response.

ARADDCC012 (approved):

PR #818, opened 2023-10-27
Add an endpoint to download a file

a3957273:


backend/src/clients/s3.ts
```diff
@@ -31,3 +31,21 @@ export async function putObjectStream(bucket: string, key: string, body: Readabl
     fileSize,
   }
 }
+
+export async function getObjectStream(bucket: string, key: string, range?: { start: number; end: number }) {
```
JR40159, line 35: Tests for this method?

backend/src/clients/s3.ts
```diff
@@ -31,3 +31,21 @@ export async function putObjectStream(bucket: string, key: string, body: Readabl
...
+  const client = await getS3Client()
```
JR40159, line 36: Do we need to create a new S3 Client for every S3 request?

backend/src/routes/v2/model/file/getDownloadFile.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { FileInterface } from '../../../../models/v2/File.js'
+import { downloadFile, getFileById } from '../../../../services/v2/file.js'
+import { BadReq, InternalError } from '../../../../utils/v2/error.js'
+import { parse } from '../../../../utils/validate.js'
+
+export const getDownloadFileSchema = z.object({
+  params: z.object({
+    modelId: z.string({
+      required_error: 'Must specify model id as param',
```
JR40159, line 14: The standard Zod error here would be sufficient?

backend/src/routes/v2/model/file/getDownloadFile.ts
```diff
@@ -0,0 +1,56 @@
...
+    } = parse(req, getDownloadFileSchema)
+
+    const file = await getFileById(req.user, fileId)
+
+    // required to support utf-8 file names
+    res.set('Content-Disposition', contentDisposition(file.name, { type: 'inline' }))
+    res.set('Content-Type', file.mime)
+    res.set('Cache-Control', 'public, max-age=604800, immutable')
+
+    if (req.headers.range) {
```
JR40159, line 38: Do this first before logic for response?

backend/src/routes/v2/model/file/getDownloadFile.ts
```diff
@@ -0,0 +1,56 @@
...
+    // res.set('Accept-Ranges', 'bytes')
+    res.writeHead(200)
+
+    const stream = await downloadFile(req.user, fileId)
+    if (!stream.Body) {
+      throw InternalError('We were not able to retrieve the body of this file', { fileId })
+    }
+
+    // The AWS library doesn't seem to properly type 'Body' as being pipeable?
+    ;(stream.Body as any).pipe(res)
```
JR40159, line 54: Can you raise this issue on their repo?

backend/src/routes/v2/model/file/getDownloadFile.ts
```diff
@@ -0,0 +1,56 @@
...
+  },
```
JR40159, line 55: Am I missing something, where is the method's return statement?

backend/src/services/v2/file.ts
```diff
@@ -37,6 +37,18 @@ export async function uploadFile(user: UserDoc, modelId: string, name: string, m
   return file
 }
 
+export async function downloadFile(user: UserDoc, fileId: string, range?: { start: number; end: number }) {
```
JR40159, line 40: Tests?

JR40159 (changes requested):

backend/src/clients/s3.ts
a3957273: Our S3 credentials expire on some platforms relatively frequently.  Tracking whether the current S3 client has expired and replacing it seems too complex for a function that's likely to only get called a few hundred - thousand times a day.

backend/src/routes/v2/model/file/getDownloadFile.ts
a3957273: Range requests will include the same headers (Content-Type, Cache-Control and Content-Disposition) seen above.  The idea behind placing it here is that when we _do_ support it you can just write the implementation in here and not need to change anything else.

I could move the error up (rule: do error checking first), but it feels like a waste for now.

PR #821, opened 2023-10-30
Improve dictionary typing

GB27247:
This is a proposed improvement.

_Note_ : The precommit config supports Python 3.8-3.11, hence correcting the typing. See [PEP-585](https://peps.python.org/pep-0585/). Collections from `typing` are no longer standard from Python 3.9 onwards.

### DIFF
--- a/lib/python-beta/src/bailo/core/utils.py
+++ b/lib/python-beta/src/bailo/core/utils.py
@@ -1,0 +1,12 @@
+from __future__ import annotations
+
+
+def filter_none(json: dict[str, str | None]) -> dict[str, str]:
+    """
+    Removes None attributes from a dictionary.
+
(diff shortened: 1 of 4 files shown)


lib/python-beta/src/bailo/core/utils.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def filter_none(json: dict[str, str | None]) -> dict[str, str]:
+    """
+    Removes None attributes from a dictionary.
+
+    :param json: Dictionary to filter
+    :return: Dictionary with removed Nonetypes
+    """
+    res = {k: v for k, v in json.items() if v is not None}
```
a3957273, line 11: Nice job checking for `None` specifically.

a3957273 (approved):

PR #823, opened 2023-10-30
Feature/bai 950 allow users to select docker images for releases

ARADDCC002:
When creating a release you can now select (optionally) to include any model images:

image

Images (if applicable) will appear under the files of a release display:

image

You can now also view all the model's images with a new tab (I think this view needs some work, it's very boring):

image

Clicking the "Upload Image" button will open a dialog with information on how to push images to the registry:

image


backend/src/models/v2/Release.ts
```diff
@@ -15,7 +17,7 @@ export interface ReleaseInterface {
   draft: boolean
 
   fileIds: Array<string>
-  images: Array<string>
+  images: Array<FlattenedModelImage>
```
JR40159, line 20: For similar things we've defined interfaces in the same file. See `CollaboratorEntry`, `AccessRequestMetadata`, `ReviewResponse ` for examples

frontend/src/model/beta/images/UploadNewImageDialog.tsx
```diff
@@ -0,0 +1,127 @@
...
+              <Stack spacing={1}>
+                <Typography fontWeight='bold'>Logging in</Typography>
+                <Stack spacing={2}>
+                  <CodeLine line={`docker login ${uiConfig.registry.host} -u ${currentUser.id}`} />
+                </Stack>
+              </Stack>
+              <Stack spacing={1}>
+                <Typography fontWeight='bold'>Pushing an image to the registry</Typography>
+                <Stack spacing={2}>
+                  <CodeLine line={`docker tag <image> ${window.location.host}/<namespace>/<model>:<version>`} />
```
JR40159, line 110: Since the `<namespace>` is the model ID, can we populate that?

frontend/src/model/beta/images/UploadNewImageDialog.tsx
```diff
@@ -0,0 +1,127 @@
...
+                  <CodeLine line={`docker push ${window.location.host}/<namespace>/<model>:<version>`} />
+                  <CodeLine line={`${window.location.host}/api/v2/model/myModelId/images`} />
```
JR40159, line 112: This isn't a bash command?

JR40159 (dismissed):

backend/src/models/v2/Release.ts
JR40159: This has also already been defined https://github.com/gchq/Bailo/blob/709b0a4d477c71508a2394e0b5922a2b6d1a1880/backend/src/services/v2/registry.ts#L18-L22 
I would suggest moving the existing definition above into this file.
I would also suggest we rename `model` to be `name` which is how that field is referred to in the Docker docs. It also seems a little less confusing

frontend/src/model/beta/images/UploadNewImageDialog.tsx
ARADDCC002: Ah no this is the URL for listing all the images, I forgot to remove the dollar sign. I'll see if I can also make it clearer.

JR40159: Ah okay. Yes, making it clear that it is separate from the commands used to push to the registry sounds like a good idea as well. Could you also add copy buttons to the bash commands?

frontend/src/model/beta/images/UploadNewImageDialog.tsx
JR40159: Can we also rename `<model>` to `<name>`? It's how that part of the Docker URI is referred to in the Docker docs and is less confusing as the namespace is the Model ID

frontend/src/common/ModelImageList.tsx
```diff
@@ -0,0 +1,59 @@
...
+  value: FlattenedModelImage[]
+  setImages: Dispatch<SetStateAction<FlattenedModelImage[]>>
+}
+
+export default function ModelImageList({ model, value, setImages }: ModelImageListProps) {
+  const { modelImages, isModelImagesLoading, isModelImagesError } = useGetModelImages(model.id)
+  const [flattenedImageList, setFlattenedImageList] = useState<FlattenedModelImage[]>([])
+
+  useEffect(() => {
+    if (modelImages) {
```
ARADDCC012, line 22: I think this will always be an array, even if an empty one. So this check will always pass

frontend/src/model/beta/images/CodeLine.tsx
```diff
@@ -0,0 +1,35 @@
...
+      style={{
+        cursor: 'pointer',
+      }}
+      role='button'
+      tabIndex={0}
+      onClick={() => {
+        handleButtonClick()
+      }}
+      onKeyDown={(e) => {
+        if (e.key === 'Enter') {
```
ARADDCC012, line 25: This should also work for spacebar. Both spacebar and enter are used to activate buttons.

frontend/src/model/beta/images/CodeLine.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const theme = useTheme()
+  const sendNotification = useNotification()
+
+  const handleButtonClick = () => {
+    navigator.clipboard.writeText(line)
+    sendNotification({ variant: 'success', msg: 'Copied to clipboard' })
+  }
+
+  return (
+    <div
```
ARADDCC012, line 15: Could this be simplified to use a single Box?:
```
<Tooltip title='Copy to clipboard' arrow>
  <Box
    role='button'
    tabIndex={0}
    onClick={() => handleButtonClick()}
     onKeyDown={(e) => {
      if (e.key === 'Enter' || e.key === ' ') {
              handleButtonClick()
        }
    }
    sx={{
      cursor: 'pointer',
      backgroundColor: theme.palette.container.main,
      p: 1,
      borderRadius: 2,
    }}
  >
    $ {line}
  </Box>
</Tooltip>
```

frontend/src/model/beta/images/ModelImageDisplay.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+type ModelImageDisplayProps = {
+  modelImage: ModelImage
+}
+
+export default function ModelImageDisplay({ modelImage }: ModelImageDisplayProps) {
+  const theme = useTheme()
+
+  return (
+    <>
```
ARADDCC012, line 15: This fragment can be removed

frontend/pages/beta/model/[modelId].tsx
```diff
@@ -29,6 +30,7 @@ export default function Model() {
             { title: 'Overview', path: 'overview', view: <Overview model={model} /> },
             { title: 'Releases', path: 'releases', view: <Releases model={model} /> },
             { title: 'Access Requests', path: 'access', view: <AccessRequests model={model} /> },
+            { title: 'Images', path: 'images', view: <ModelImages model={model} /> },
```
Comment (a3957273) [inline frontend/pages/beta/model/[modelId].tsx:33]: Can we call this 'Registry', instead of 'images'? :)

frontend/src/model/beta/images/ModelImageDisplay.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <Box
+        sx={{
+          borderWidth: '1px',
+          borderStyle: 'solid',
+          borderColor: theme.palette.primary.main,
+          width: '100%',
+          borderRadius: 4,
+        }}
+      >
+        <Box p={2}>
```
ARADDCC012, line 25: Can this Box be removed and the padding applied to the Box above?

frontend/src/model/beta/images/ModelImageDisplay.tsx
```diff
@@ -0,0 +1,41 @@
...
+          <Stack direction='column' spacing={1}>
+            <Typography component='h2' variant='h6' color='primary'>
+              {modelImage.model}
+            </Typography>
+            <Typography fontWeight='bold'>Versions</Typography>
+            {modelImage.versions.map((imageVersion) => (
```
ARADDCC012, line 31: Consider using `useMemo` if there could be a lot of versions

frontend/src/model/beta/images/UploadNewImageDialog.tsx
ARADDCC002: You can actually click the lines to copy them, and a tooltip appears like so:

image

I've also added some icons to the left of each to make it clearer

frontend/src/model/beta/images/UploadNewImageDialog.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+  const theme = useTheme()
+
+  const [displayToken, setDisplayToken] = useState(false)
+  const [displayedToken, setDisplayedToken] = useState('')
+
+  const regenerateToken = async () => {
+    const { token } = await fetch('/api/v1/user/token', {
+      method: 'POST',
+    }).then((res) => res.json())
```
ARADDCC012, line 40: Shouldn't be using promise chains:
```
const res = await fetch('/api/v1/user/token', {
  method: 'POST',
})

if (!response.ok) {
  // Show error to user
  return // Either return nothing or a string perhaps?
}

let token

try {
  const body = await res.json()
  token = body.token
} catch (error) {
  // Show some sort of error stating we received an invalid response from the server (I'm not sure what our standard process is for this)
}

return token
```

frontend/src/model/beta/images/UploadNewImageDialog.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    return token
+  }
+
+  const showToken = async () => {
+    const token = await regenerateToken()
```
ARADDCC012, line 46: Given that `regenerateToken()` can error, we should also ensure that `token` is valid before setting it.
```
if (token) {
    setDisplayedToken(token)
    setDisplayToken(true)
}
```

frontend/src/model/beta/images/UploadNewImageDialog.tsx
```diff
@@ -0,0 +1,127 @@
...
+              <Typography fontWeight='bold'>User authentication token</Typography>
+              <Typography>Use the token below to authenticate when you try and run the docker login command</Typography>
+              <Stack direction='row'>
+                <Button sx={{ mr: 2 }} variant='outlined' onClick={showToken} data-test='showTokenButton'>
+                  Regenerate Token
+                </Button>
+                <Box
+                  sx={{
+                    backgroundColor: theme.palette.container.main,
+                    pr: 2,
```
ARADDCC012, line 76: Super minor but could combine these two padding values into one `px: 2`

frontend/src/model/beta/images/UploadNewImageDialog.tsx
```diff
@@ -0,0 +1,127 @@
...
+                >
+                  <Box component={Stack} direction='column' justifyContent='center'>
+                    <Typography variant='body1' data-test='dockerPassword'>
+                      {displayToken ? displayedToken : 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxxx'}
+                    </Typography>
+                  </Box>
+                </Box>
+                <Tooltip title='Regenerate & Copy to clipboard'>
+                  <IconButton
+                    onClick={async () => {
```
ARADDCC012, line 90: Consider pulling this function out of the JSX, there's a bit too much going on for it to be inline imo.

Also will need to check token is valid as `regenerateToken()` can return `undefined` if it errors:
```
if (token) {
  setDisplayedToken(token)
  navigator.clipboard.writeText(token)
}
```

frontend/src/model/beta/images/UploadNewImageDialog.tsx
```diff
@@ -0,0 +1,127 @@
...
+                      const token = await regenerateToken()
+                      setDisplayedToken(token)
+                      navigator.clipboard.writeText(token)
+                    }}
+                    aria-label='regenerate and copy to clipboard'
```
ARADDCC012, line 95: Clarify this is talking about a token?
`aria-label='regenerate token and copy to clipboard'`

frontend/src/model/beta/ModelImages.tsx
```diff
@@ -0,0 +1,53 @@
...
+export default function ModelImages({ model }: AccessRequestsProps) {
+  const { modelImages, isModelImagesLoading, isModelImagesError } = useGetModelImages(model.id)
+
+  const [openUploadImageDialog, setOpenUploadImageDialog] = useState(false)
+
+  const modelImageList = useMemo(
+    () =>
+      modelImages.length ? (
+        modelImages.map((modelImage) => (
+          <ModelImageDisplay modelImage={modelImage} key={`${modelImage.namespace}-${modelImage.model}`} />
```
ARADDCC012, line 25: Given that this comment (https://github.com/gchq/Bailo/pull/823/files#r1376614866) states that `modelImage.namespace` is the mode id and I'm guessing `modelImage.model` is also the model id, is `key={`${modelImage.namespace}-${modelImage.model}`}` a unique key across all images for the same model?

frontend/src/model/beta/images/UploadNewImageDialog.tsx
```diff
@@ -0,0 +1,127 @@
...
+                <Stack spacing={2}>
+                  <CodeLine line={`docker tag <image> ${window.location.host}/<namespace>/<model>:<version>`} />
+                  <CodeLine line={`docker push ${window.location.host}/<namespace>/<model>:<version>`} />
+                  <CodeLine line={`${window.location.host}/api/v2/model/myModelId/images`} />
+                </Stack>
+              </Stack>
+            </Stack>
+          </DialogContent>
+          <Box sx={{ mx: 3 }}>
+            <Divider sx={{ margin: 'auto' }} />
```
ARADDCC012, line 118: I don't think `sx={{ margin: 'auto' }}` is needed here

frontend/src/model/beta/releases/ModelReleaseDisplay.tsx
```diff
@@ -105,33 +105,41 @@ export default function ModelReleaseDisplay({
...
+                      key={file._id}
+                      direction={{ sm: 'row', xs: 'column' }}
+                      justifyContent='space-between'
+                      alignItems='center'
+                      spacing={2}
+                    >
+                      <Link href='/beta'>{file.name}</Link>
+                      <Typography variant='caption'>{prettyBytes(file.size)}</Typography>
+                    </Stack>
+                  ))}{' '}
```
ARADDCC012, line 124: Is this space `{' '}` intentional?

frontend/src/model/beta/releases/ModelReleaseDisplay.tsx
```diff
@@ -105,33 +105,41 @@ export default function ModelReleaseDisplay({
...
-                >
-                  <Link href='/beta'>{image}</Link>
-                  {/* TODO - Add file size here */}
-                  {/* <Typography variant='caption'>123GB</Typography> */}
-                </Stack>
-              ))}
+            <Box>{(release.files.length > 0 || release.images.length > 0) && <Divider />}</Box>
+            <Stack spacing={2}>
+              {release.files.length > 0 && (
+                <Box>
```
ARADDCC012, line 111: Can this be a fragment instead?

frontend/src/model/beta/releases/ModelReleaseDisplay.tsx
```diff
@@ -105,33 +105,41 @@ export default function ModelReleaseDisplay({
...
+                      spacing={2}
+                    >
+                      <Link href='/beta'>{file.name}</Link>
+                      <Typography variant='caption'>{prettyBytes(file.size)}</Typography>
+                    </Stack>
+                  ))}{' '}
+                </Box>
+              )}
+              {release.images.length > 0 && (
+                <Box>
```
ARADDCC012, line 128: Can this be a fragment instead?

frontend/src/model/beta/images/UploadNewImageDialog.tsx
a3957273: I don't think we need to list the API endpoint within our frontend?

frontend/src/model/beta/releases/ModelReleaseDisplay.tsx
```diff
@@ -105,33 +105,41 @@ export default function ModelReleaseDisplay({
...
+                  <Typography fontWeight='bold'>Docker images</Typography>
+                  {release.images.map((image) => (
+                    <Stack
+                      key={`${image.model}-${image.version}`}
+                      direction={{ sm: 'row', xs: 'column' }}
+                      justifyContent='space-between'
+                      alignItems='center'
+                      spacing={2}
+                    >
+                      <Link href='/beta'>{`${image.model}-${image.version}`}</Link>
```
a3957273, line 138: This should probably not link anywhere, but instead when clicked 'copy' the text.  It should also be of the format:

```
publicaddress/modelId/name:tag
```

So that people can just paste it in to `docker pull`.

a3957273 (changes requested):

frontend/src/common/ModelImageList.tsx
```diff
@@ -19,22 +19,24 @@ export default function ModelImageList({ model, value, setImages }: ModelImageLi
...
+          tag: modelImageVersion,
+        })
       }
-      setFlattenedImageList(updatedImageList)
     }
+    setFlattenedImageList(updatedImageList)
   }, [modelImages])
 
-  const handleChange = (_event: SyntheticEvent<Element, Event>, FlattenedImageList: FlattenedModelImage[]) => {
+  const sortedImageListByName = useMemo(() => {
```
ARADDCC012, line 35: Sorry it's on me for not being clear. I meant:
```
// This sort of thing can be pulled out into an arrayUtils file if we think we might want to sort by name elsewhere
const sortByNameAscending = <T extends { name: string }>(a: T, b: T) => {
  return a.name.localeCompare(b.name)
}

const sortedImageList = useMemo(() => flattenedImageList.sort(sortByNameAscending), [])
```

frontend/src/model/beta/ModelImages.tsx
```diff
@@ -21,8 +21,8 @@ export default function ModelImages({ model }: AccessRequestsProps) {
   const modelImageList = useMemo(
     () =>
       modelImages.length ? (
-        modelImages.map((modelImage) => (
-          <ModelImageDisplay modelImage={modelImage} key={`${modelImage.namespace}-${modelImage.model}`} />
+        modelImages.map((modelImage, index) => (
```
ARADDCC012, line 24: Index isn't a stable key. See: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

Specifically:
"The key is used by React to [identify which items have changed, are added, or are removed and should be stable](https://facebook.github.io/react/docs/lists-and-keys.html#keys).

It's a bad idea to use the array index since it doesn't uniquely identify your elements. In cases where the array is sorted or an element is added to the beginning of the array, the index will be changed even though the element representing that index may be the same. This results in unnecessary renders."

frontend/src/model/beta/images/CodeLine.tsx
```diff
@@ -1,35 +1,44 @@
...
-        }
-      }}
-    >
-      <Tooltip title='Copy to clipboard' arrow>
-        <Box sx={{ backgroundColor: theme.palette.container.main, p: 1, borderRadius: 2 }}>$ {line}</Box>
-      </Tooltip>
-    </div>
+        }}
+        onKeyDown={(e) => {
+          if (e.key === 'Enter' || e.key === 'Spacebar') {
```
ARADDCC012: For all modern browsers the key of spacebar will be `" "`. See: https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#whitespace_keys

If we want to support older versions of Firefox (pre version 37) then `'Spacebar'` will also be required.

frontend/src/model/beta/images/CodeLine.tsx
```diff
@@ -1,35 +1,44 @@
...
+            handleButtonClick()
+          }
+        }}
+      >
+        <Stack direction='row' spacing={2} alignItems='center'>
+          {icon}
+          <Typography>{`${line}`}</Typography>
```
ARADDCC012: In the old version the $ was used as a string character, not as a template string. So I think this should be:
```
<Typography>$ {line}</Typography>
```

frontend/src/model/beta/images/UploadNewImageDialog.tsx
```diff
@@ -19,25 +19,43 @@ import { useState } from 'react'
...
   const regenerateToken = async () => {
-    const { token } = await fetch('/api/v1/user/token', {
+    const res = await fetch('/api/v1/user/token', {
       method: 'POST',
-    }).then((res) => res.json())
+    })
+
+    if (!res.ok) {
+      setTokenErrorText(await getErrorMessage(res))
+      return // Either return nothing or a string perhaps?
```
ARADDCC012: This comment can be removed

frontend/src/model/beta/releases/ModelReleaseDisplay.tsx
```diff
@@ -108,37 +115,39 @@ export default function ModelReleaseDisplay({
...
+                </>
               )}
               {release.images.length > 0 && (
-                <Box>
+                <>
                   <Typography fontWeight='bold'>Docker images</Typography>
                   {release.images.map((image) => (
                     <Stack
-                      key={`${image.model}-${image.version}`}
+                      key={`${image.name}-${image.tag}`}
```
ARADDCC012, line 139: In theory I think we could have two images for different models, both with the same name and tag. Is that correct?

If so, consider adding the model id to the key also.

frontend/src/model/beta/ModelImages.tsx
ARADDCC012: Assuming these images have the same properties, could you use the same key as suggested in the following comment? https://github.com/gchq/Bailo/pull/823/files/709b0a4d477c71508a2394e0b5922a2b6d1a1880..12dc11db8adf0e8136f6f7d2906a5f82c265e584?diff=unified&w=1#r1377668028

frontend/src/model/beta/images/CodeLine.tsx
ARADDCC002: I've added the icon to remove the need for $ as I wanted it to have the choice if any icon

frontend/src/common/ModelImageList.tsx
```diff
@@ -32,9 +32,11 @@ export default function ModelImageList({ model, value, setImages }: ModelImageLi
     setFlattenedImageList(updatedImageList)
   }, [modelImages])
 
-  const sortedImageListByName = useMemo(() => {
-    return flattenedImageList.sort((a, b) => a.name.localeCompare(b.name))
-  }, [flattenedImageList])
+  const sortByNameAscending = <T extends { name: string }>(a: T, b: T) => {
```
ARADDCC012, line 35: If this is left inside this component it should also be in the `useMemo` dependency array. That would cause it to infinitely rerender (as functions are given a new reference on every render), so we should instead move this function outside of the component.

frontend/src/model/beta/registry/CodeLine.tsx
```diff
@@ -29,14 +29,14 @@ export default function CodeLine({ line, icon = <CodeIcon /> }: CodeLineProps) {
...
-          if (e.key === 'Enter' || e.key === 'Spacebar') {
+          if (e.key === 'Enter' || e.key === 'Spacebar' || e.key === ' ') {
             handleButtonClick()
           }
         }}
       >
         <Stack direction='row' spacing={2} alignItems='center'>
           {icon}
-          <Typography>{`${line}`}</Typography>
+          <Typography>{line}</Typography>
```
ARADDCC012, line 39: I think the $ was intended to make it stand out as a line of code (like a terminal), do we not want to keep it?

frontend/src/model/beta/registry/ModelImageDisplay.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const theme = useTheme()
+
+  const { uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
+
+  const modelImageTags = useMemo(
+    () =>
+      uiConfig &&
+      modelImage.tags.map((imageTag) => (
+        <CodeLine
+          key={`${modelImage.name}-${imageTag}`}
```
ARADDCC012, line 25: This will also need the repository to make it unique
`key={`${modelImage.repository}-${modelImage.name}-${imageTag}`}`

PR #845, opened 2023-11-02
added the ability to edit access requests

ARADDCC002:
Access requests can now be viewed on their own page. On this page they can be edited. Any review responses will also be displayed at the bottom of the page.

image


frontend/actions/review.ts
```diff
@@ -34,11 +34,11 @@ type SemverOrAccessRequestId =
     }
   | {
       semver?: never
-      accessRequestId: AccessRequestInterface['id']
+      accessRequestId: AccessRequestInterface['id'] | undefined
```
ARADDCC012, line 37: Use `?` to be consistent? e.g. `accessRequestId?: AccessRequestInterface['id'] `

frontend/actions/review.ts
```diff
@@ -34,11 +34,11 @@ type SemverOrAccessRequestId =
...
     }
 
 type GetReviewRequestsForModelQuery = {
-  modelId: ModelInterface['id']
+  modelId: ModelInterface['id'] | undefined
```
ARADDCC012, line 41: Change to `modelId?: ModelInterface['id']` for consistency?

frontend/actions/review.ts
ARADDCC012, line 37: I came across this same problem with releases. If we make these optional, we lose the guidance/safety that this type provides.

Instead we can do the following:
```
type SemverOrAccessRequestId =
  | {
      semver: ReleaseInterface['semver'] | ''
      accessRequestId?: never
    }
  | {
      semver?: never
      accessRequestId: AccessRequestInterface['id'] | ''
    }
```
That way we can use it like so (in your `[accessRequestId].tsx` component) and still keep the either/or behaviour for semver/requestId:
```
  const {
    reviews: activeReviews,
    isReviewsLoading: isActiveReviewsLoading,
    isReviewsError: isActiveReviewsError,
  } = useGetReviewRequestsForModel({
    modelId: modelId,
    accessRequestId: accessRequestId || '',
    isActive: true,
  })
```

If we tried to provide a `semver` along with the above, we'll get the error saying that it's invalid. Which is what we want :)

frontend/pages/beta/model/[modelId]/access-request/[accessRequestId].tsx
```diff
@@ -0,0 +1,74 @@
...
+    <Wrapper
+      title={accessRequest ? (accessRequest.metadata.overview.name as string) : 'Loading...'}
+      page='access-request'
+      fullWidth
+    >
+      <Container maxWidth='md'>
+        <Card sx={{ p: 4, m: 'auto', mb: 2 }}>
+          {isAccessRequestLoading && isActiveReviewsLoading && <Loading />}
+          {!accessRequest && !isAccessRequestLoading && (
+            <EmptyBlob text={`Oh no, it looks like access request ${accessRequest} doesn't exist!`} />
```
Comment (ARADDCC012) [inline frontend/pages/beta/model/[modelId]/access-request/[accessRequestId].tsx:47]: Is this useful? If we can't find the accessRequest the API should've returned a 404. So we'll be showing an error message instead of this anyway. I think this can probs be removed.

frontend/src/model/beta/reviews/ReviewBanner.tsx
```diff
@@ -98,7 +100,7 @@ export default function ReviewBanner({ release, accessRequest }: ReviewBannerPro
         borderWidth: '1px',
         borderStyle: 'solid',
         borderColor: theme.palette.primary.main,
-        borderRadius: '13px 13px 0px 0px',
+        borderRadius: removeBorderRadius ? 'none' : '13px 13px 0px 0px',
```
ARADDCC012, line 103: None isn't a valid borderRadius value. Instead you can do: 
```
...(!removeBorderRadius && { borderRadius: '13px 13px 0px 0px' }),
```

frontend/src/model/beta/reviews/ReviewBanner.tsx
```diff
@@ -19,13 +19,15 @@ type ReviewBannerProps =
   | {
       release: ReleaseInterface
       accessRequest?: never
+      removeBorderRadius?: boolean
     }
   | {
       release?: never
       accessRequest: AccessRequestInterface
+      removeBorderRadius?: boolean
```
ARADDCC012, line 27: Not 100% on the name here, how about `square`? Also you don't need to add this new prop twice. You can do the following:
```
type ReleaseOrAccessRequest =
  | {
      release: ReleaseInterface
      accessRequest?: never
    }
  | {
      release?: never
      accessRequest: AccessRequestInterface
    }

type ReviewBannerProps = {
  square?: boolean
} & ReleaseOrAccessRequest
```

frontend/src/common/ReviewComments.tsx
ARADDCC012, line 1: Should this file be moved to `frontend/src/common/reviews/ReviewComments.tsx`?

frontend/pages/beta/model/[modelId]/access-request/[accessRequestId].tsx
```diff
@@ -0,0 +1,74 @@
...
+                spacing={2}
+                divider={<Divider flexItem orientation='vertical' />}
+              >
+                <Link href={`/beta/model/${modelId}?tab=access`}>
+                  <Button sx={{ width: 'fit-content' }} startIcon={<ArrowBack />}>
+                    Back to model
+                  </Button>
+                </Link>
+                <Typography variant='h6' color='primary' component='h2'>
+                  {accessRequest ? (accessRequest.metadata.overview.name as string) : 'Loading...'}
```
Comment (ARADDCC012) [inline frontend/pages/beta/model/[modelId]/access-request/[accessRequestId].tsx:63]: This type assertion shouldn't be needed as we've typed `accessRequest.metadata.overview.name` as a string

frontend/src/common/ReviewComments.tsx
ARADDCC012, line 1: This file also looks like it should be split into 3 components. Happy to do this on my branch though as I'll be using it in the releases page.

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+import { useGetSchema } from '../../../../actions/schema'
+import { useGetUiConfig } from '../../../../actions/uiConfig'
+import { AccessRequestInterface, SplitSchemaNoRender } from '../../../../types/interfaces'
+import { getStepsData, getStepsFromSchema } from '../../../../utils/beta/formUtils'
+import Loading from '../../../common/Loading'
+import ModelCardForm from '../../../Form/beta/ModelCardForm'
+import MessageAlert from '../../../MessageAlert'
+
+type FormEditPageProps = {
```
ARADDCC012, line 16: Rename to `AccessRequestFormEditPageProps`?

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
```diff
@@ -0,0 +1,127 @@
...
+        setErrorText('')
+        mutateAccessRequest()
+      }
+    }
+  }
+
+  function onCancel() {
+    if (schema) {
+      mutateAccessRequest()
+      const steps = getStepsFromSchema(schema, {}, ['properties.contacts'], accessRequest.metadata)
```
ARADDCC012, line 49: We already do run this exact logic to update the SplitSchema in the `useEffect` below. I think it can be removed from here because `mutateAccessRequest()` will trigger that effect to run if needed.

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    setSplitSchema({ reference: schema.id, steps })
+  }, [schema, accessRequest])
+
+  useEffect(() => {
+    setUnsavedChanges(isEdit)
+  }, [isEdit, setUnsavedChanges])
+
+  if (isSchemaError) {
```
ARADDCC012, line 74: We've already spoken about this, but I'm happy to update these to use `MultipleErrorWrapper` in my ticket, so feel free to leave as is for now :)

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const [errorText, setErrorText] = useState('')
+
+  const { schema, isSchemaLoading, isSchemaError } = useGetSchema(accessRequest.schemaId)
+  const { mutateAccessRequest } = useGetAccessRequest(accessRequest.modelId, accessRequest.id)
+  const { uiConfig: _uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
+
+  const { setUnsavedChanges } = useContext(UnsavedChangesContext)
+  const theme = useTheme()
+
+  async function onSubmit() {
```
ARADDCC012, line 32: Consider renaming to `handleSubmit`

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
```diff
@@ -0,0 +1,127 @@
...
+        setErrorText(await getErrorMessage(res))
+      } else {
+        setIsEdit(false)
+        setErrorText('')
+        mutateAccessRequest()
+      }
+    }
+  }
+
+  function onCancel() {
```
ARADDCC012, line 46: Consider renaming to `handleCancel`

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
```diff
@@ -0,0 +1,127 @@
...
+                  Save
+                </Button>
+              </Stack>
+              <Typography variant='caption' color={theme.palette.error.main}>
+                {errorText}
+              </Typography>
+            </Stack>
+          )}
+        </Stack>
+        <ModelCardForm splitSchema={splitSchema} setSplitSchema={setSplitSchema} canEdit={isEdit} />
```
ARADDCC012, line 123: Is this the wrong form?

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
ARADDCC002: The `mutateAccessRequest` isn't being called onCancel if we leave this code out

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
ARADDCC002: I've updated it to be `JsonSchemaForm`

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
ARADDCC012: I meant keep the mutate here and remove the steps/split schema logic. Can it be trimmed down to this?
```
function handleCancel () {
 ()
  setIsEdit(false)
}
```

That way we call mutate and set the component back to readonly mode and the `useEffect` on line 58 should handle the steps/split schema logic due to `accessRequest` being in it's dependency array (the `accessRequest` will be "changed" due to the mutate call, triggering the effect)

ARADDCC012: I meant keep the mutate here and remove the steps/split schema logic. Can it be trimmed down to this?
```
function handleCancel () {
 mutateAccessRequest()
  setIsEdit(false)
}
```

That way we call mutate and set the component back to readonly mode and the `useEffect` on line 58 should handle the steps/split schema logic due to `accessRequest` being in it's dependency array (the `accessRequest` will be "changed" due to the mutate call, triggering the effect)

ARADDCC012: I meant keep the mutate here and remove the steps/split schema logic. Can it be trimmed down to this?
```
function handleCancel () {
  mutateAccessRequest()
  setIsEdit(false)
}
```

That way we call mutate and set the component back to readonly mode and the `useEffect` on line 58 should handle the steps/split schema logic due to `accessRequest` being in it's dependency array (the `accessRequest` will be "changed" due to the mutate call, triggering the effect)

ARADDCC012: Also, if this logic is taken from elsewhere, can we make the same change there too. This split schema logic is notoriously confusing and if we can simplify it a bit I think we should.

frontend/actions/review.ts
```diff
@@ -29,16 +29,16 @@ export function useGetReviewRequestsForUser(isActive = true) {
 
 type SemverOrAccessRequestId =
   | {
-      semver: ReleaseInterface['semver']
+      semver: ReleaseInterface['semver'] | ''
```
ARADDCC012, line 32: Really sorry, but I've realised the two additions of `|| ''` in this file aren't needed as semver and id are already strings. It's the change we made in another file that allowed us to get rid of the `|| undefined`.

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
ARADDCC012: Due to the accessRequest not being changed while working on the form, the mutate call wasn't having any effect. The following should work as we want:
```
  const resetForm = useCallback(() => {
    if (schema) {
      const steps = getStepsFromSchema(schema, {}, ['properties.contacts'], accessRequest.metadata)

      for (const step of steps) {
        step.steps = steps
      }

      setSplitSchema({ reference: schema.id, steps })
    }
  }, [accessRequest.metadata, schema])

  function handleCancel() {
    setIsEdit(false)
    resetForm()
  }

  useEffect(() => {
    resetForm()
  }, [resetForm])
```

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
```diff
@@ -0,0 +1,127 @@
...
+      const steps = getStepsFromSchema(schema, {}, ['properties.contacts'], accessRequest.metadata)
+      for (const step of steps) {
+        step.steps = steps
+      }
+      setSplitSchema({ reference: schema.id, steps })
+      setIsEdit(false)
+    }
+  }
+
+  useEffect(() => {
```
ARADDCC012, line 58: This effect hook can now be removed

frontend/src/model/beta/accessRequests/AccessRequestFormEditPage.tsx
```diff
@@ -0,0 +1,134 @@
...
+  function handleCancel() {
+    setIsEdit(false)
+    resetForm()
+  }
+
+  useEffect(() => {
+    resetForm()
+  }, [resetForm])
+
+  useEffect(() => {
```
ARADDCC012, line 65: This effect hook can now be removed

PR #850, opened 2023-11-03
Various bugfixes and clean up

ARADDCC012:


ARADDCC002 (approved):

backend/src/models/v2/Model.ts
```diff
@@ -38,6 +38,7 @@ export interface ModelInterface {
   id: string
 
   name: string
+  team: string
```
JR40159, line 41: Would a more accurate name be `teamId`?

backend/src/routes/v2/model/postModel.ts
```diff
@@ -12,6 +12,7 @@ export const postModelSchema = z.object({
     name: z.string({
       required_error: 'Must specify model name',
```
JR40159, line 13: For consistency and as you are editing this file, could you also remove the other `required_error` fields?

backend/src/services/v2/model.ts
```diff
@@ -19,7 +19,7 @@ import { BadReq, Forbidden, NotFound } from '../../utils/v2/error.js'
 import { convertStringToId } from '../../utils/v2/id.js'
 import { findSchemaById } from './schema.js'
 
-export type CreateModelParams = Pick<ModelInterface, 'name' | 'description' | 'visibility'>
+export type CreateModelParams = Pick<ModelInterface, 'name' | 'team' | 'description' | 'visibility'>
```
JR40159, line 22: Should the backend do validation to ensure that the team exists?

JR40159 (changes requested):

backend/src/models/v2/Model.ts
ARADDCC012: This isn't always a team ID. It's a select field with the ability to enter any freeform text. If the user selects something from the list it'll be an ID, otherwise it won't.

JR40159: What's the benefit of allowing users to enter any text they like? Will this not cause problems when managing the relationships between models and teams when the teams attached to models might or might not be teams that exist within BAILO?

ARADDCC012: After discussion we've decided to got with teamId. Changed!

backend/src/routes/v2/model/postModel.ts
ARADDCC012: Done

backend/src/services/v2/model.ts
ARADDCC012: Decided to leave this for when teams have been fully implemented in the backend

JR40159 (approved):

PR #851, opened 2023-11-04
Automatically set model card revision if none is provided

a3957273:


backend/src/services/v2/release.ts
```diff
@@ -43,6 +48,17 @@ export async function createRelease(user: UserDoc, releaseParams: CreateReleaseP
 
   const model = await getModelById(user, releaseParams.modelId)
 
+  if (releaseParams.modelCardVersion) {
+    // Ensure that the requested model card version exists.
+    await getModelCardRevision(user, releaseParams.modelId, releaseParams.modelCardVersion)
+  } else {
+    if (!model.card) {
+      throw BadReq('This model does not have a model card associated with it yet.', { modelId: model.id })
```
JR40159, line 56: Test?

JR40159 (changes requested):

PR #854, opened 2023-11-04
Add authorisation check for listing images from the registry

a3957273:


backend/src/services/v2/registry.ts
```diff
@@ -52,6 +56,19 @@ export async function listImageTags(user: UserDoc, imageRef: RepoRef) {
 }
 
 export async function listModelImages(user: UserDoc, modelId: string) {
```
JR40159, line 58: I was going to recommend adding a test for this but I can't seem to find any registry service tests...?

JR40159 (changes requested):

JR40159 (approved):

PR #862, opened 2023-11-07
Add Release, AccessRequest and Review classes

GB27247:


a3957273 (changes requested):

lib/python-beta/tests/test_model.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import annotations
+
+from bailo import Client, Release
+from semantic_version import Version
+
+
+def test_release_version():
+    client = Client("https://example.com")
+
+    release = Release(client, 'test_id', '1.0.0', 1, "notes", [], [], draft=True)
```
a3957273, line 10: When there are this many parameters they should be named.  This is unintelligable!  Which of the last two arrays are images vs files, etc.

lib/python-beta/src/bailo/core/agent.py
```diff
@@ -7,10 +7,12 @@ class Agent:
     def __init__(self):
         self.get = requests.get
         self.patch = requests.patch
-        self.post = requests.post
         self.put = requests.put
         self.delete = requests.delete
 
+    def post(self, *args, **kwargs):
```
a3957273, line 13: a) Not all of our post requests send and receive JSON (e.g. post file)
b) Instead of setting this, you should likely be using the `json` key on a `requests.post`?

lib/python-beta/src/bailo/helper/models.py
a3957273, line 1: This file is called `models` but seems to be to do with `releases`? :)

lib/python-beta/src/bailo/helper/models.py
```diff
@@ -0,0 +1,113 @@
+from __future__ import annotations
+
+import datetime
+from typing import Any
+
+import dateutil.parser
+from bailo.core import Client
+from semantic_version import Version
```
a3957273, line 8: This looks like a pip dependency, but no dependencies have been added to the requirements list.

lib/python-beta/src/bailo/helper/models.py
```diff
@@ -0,0 +1,113 @@
...
+        self.draft = draft
+        self.files = files
+
+        self.created_at = created_at
+        self.updated_at = updated_at
+
+
+    @classmethod
+    def get_release(
+        cls,
```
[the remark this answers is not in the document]
GB27247: While I agree the syntax looks cleaner, `staticmethod`s are not inherited. Given an example hypothetically I wanted to create some different types of review down the line I would have to create seperate functionality?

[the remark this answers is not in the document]
a3957273: Okay, that seems like a good reason.  But instead of calling it `cls` are we okay calling it `release_class` or equivalent?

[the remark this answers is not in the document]
GB27247: Unfortunately [Python's Style Guide](https://peps.python.org/pep-0008/) details
'Always use `cls` for the first argument to class methods.' by convention

PR #864, opened 2023-11-09
BAI-972 Model Class (and minor core fixes)

bw27492:
Core functionality implemented (including updates to core). Model class can now seamlessly conduct the following operations:

- Create a new model.
- Retrieve an existing model.
- Create and update model cards.
- Download model files (uploading not currently supported but will be for release).

Still to do for the model class:

- Properly utilise the Schema object on completion.
- Implement the upload method.
- Handle new releases.


lib/python-beta/src/bailo/core/client.py
```diff
@@ -37,13 +37,11 @@ def post_model(
...
+        filtered_json = filter_none({"name": name, "description": description, "visibility": visibility})
+
         return self.agent.post(
             f"{self.url}/v2/models",
-            json={
-                "name": name,
-                "description": description,
-                "visibility": visibility,
-            },
+            json=filtered_json,
```
a3957273, line 44: I don't know whether splitting it out into it's own variable is nicer than:

```python
return self.agent.post(
  f"{self.url}/v2/models",
  json=filter_none({
    "name": name,
    "description": description,
    "visibility": visibility
  }),
)
```

lib/python-beta/src/bailo/core/client.py
```diff
@@ -260,6 +258,23 @@ def get_files(
             f"{self.url}/v2/model/{model_id}/files",
         ).json()
 
+    def get_download_file(
```
a3957273, line 261: I'm tempted to say this should be special.  Maybe take in a `stream` parameter that you pass to the request and then return the request itself, instead of `request.content`.

That way downstream users can decide whether they want to stream or not.

a3957273: Example: https://stackoverflow.com/questions/16694907/download-large-file-in-python-with-requests

lib/python-beta/src/bailo/helper/model.py
```diff
@@ -0,0 +1,105 @@
...
+
+            if self.__mc_change == 1:
+                if self.__model_card_original is None:
+                    self.client.model_card_from_schema(model_id=self.model_id, schema_id=self.model_card_schema)
+
+                self.client.put_model_card(model_id=self.model_id, metadata=self.__model_card)
+
+            # Check files, new release? (TBC)
+
+        self.refresh()
```
a3957273, line 73: Don't refresh here, the response of `put_model_card()` should be the updated release.

lib/python-beta/src/bailo/helper/model.py
```diff
@@ -0,0 +1,105 @@
...
+    ):
+        self.client = client
+        self.name, self.description, self.visibility = name, description, visibility
+        self.model_id = None
+        self.model_version = None
+        self.model_card_version = None
+        self.model_card_schema = 'minimal-general-v10-beta'
+        self.__model_card_original = model_card
+        self.__model_card = model_card
+        self.__local_dir = f"./temp_{name.strip(' ')}"  # property tbc
```
a3957273, line 27: Using the `tempfile` module to get a temporary store will likely work better.  In a lot of cases (e.g. when deployed to a Docker container) the local directory may not be writable.

https://docs.python.org/3/library/tempfile.html

lib/python-beta/src/bailo/helper/model.py
```diff
@@ -0,0 +1,105 @@
...
+from bailo.helper.utils import retrieve_model
+
+
+class Model:
+    def __init__(
+        self,
+        client: Client,
+        name: str,
+        description: str,
+        visibility: ModelVisibility | None = ModelVisibility.Public,
```
a3957273, line 16: You're encoding backend logic in here.  Just use `visibility: ModelVisibility` here, as the backend will always return it.

lib/python-beta/src/bailo/helper/model.py
```diff
@@ -0,0 +1,105 @@
...
+        model.model_id = model_id
+        model.model_card_version = model_card_version
+        model.model_card_schema = model_card_schema
+
+        return model
+
+    def refresh(self):
+        self.name, self.description, self.visibility, self.__model_card, self.model_card_version, self.model_card_schema = retrieve_model(client=self.client, model_id=self.model_id)
+
+    def publish(self):
```
a3957273, line 50: Instead of a `.publish()` function, I think it would be better to have a static method called `create` or equivalent.  I believe this has been done on the `Releases`?

It'll make things like the 'from_id' function above cleaner as well.

lib/python-beta/src/bailo/helper/model.py
```diff
@@ -0,0 +1,105 @@
...
+        self.model_version = None
+        self.model_card_version = None
+        self.model_card_schema = 'minimal-general-v10-beta'
+        self.__model_card_original = model_card
+        self.__model_card = model_card
+        self.__local_dir = f"./temp_{name.strip(' ')}"  # property tbc
+        self.__mc_change = 0
+        self.__file_change = 0  # property tbc
+
+    @classmethod
```
a3957273, line 31: Use a static method instead of a class method here, and create the class by name 'Model'.  It's more obvious than using the generic `cls` throughout the code.

lib/python-beta/src/bailo/helper/model.py
```diff
@@ -0,0 +1,105 @@
...
+
+import os
+from typing import Any
+
+from bailo.core import Client, ModelVisibility
+from bailo.helper.utils import retrieve_model
+
+
+class Model:
+    def __init__(
```
a3957273, line 11: The init function should take an `id`.

lib/python-beta/src/bailo/helper/model.py
```diff
@@ -0,0 +1,105 @@
...
+        self.client = client
+        self.name, self.description, self.visibility = name, description, visibility
+        self.model_id = None
+        self.model_version = None
+        self.model_card_version = None
+        self.model_card_schema = 'minimal-general-v10-beta'
+        self.__model_card_original = model_card
+        self.__model_card = model_card
+        self.__local_dir = f"./temp_{name.strip(' ')}"  # property tbc
+        self.__mc_change = 0
```
a3957273, line 28: I'm not a fan of having these private methods to track whether properties have changed or not.  Can we just make everything idempotent?  

E.g. when a user runs `add_file()` it _immediately_ updates the release with the new file.  It means we don't have to track internal state.

lib/python-beta/src/bailo/helper/model.py
```diff
@@ -0,0 +1,105 @@
...
+    def from_id(cls, client: Client, model_id: str):
+        name, description, visibility, model_card, model_card_version, model_card_schema = retrieve_model(client=client, model_id=model_id)
+        model = cls(
+            client=client,
+            name=name,
+            description=description,
+            visibility=visibility,
+            model_card=model_card
+        )
+        model.model_id = model_id
```
a3957273, line 41: Below I talk about changing `.publish()` to a static `.create()` method, which would allow these to be provided into the constructor of the function.

lib/python-beta/src/bailo/helper/model.py
```diff
@@ -0,0 +1,105 @@
...
+        model.model_card_version = model_card_version
+        model.model_card_schema = model_card_schema
+
+        return model
+
+    def refresh(self):
```
a3957273, line 47: I think this is an anti-pattern.  I think instead data should either be updated from our responses to each function or by destroying and recreating the object.

lib/python-beta/src/bailo/helper/model.py
```diff
@@ -0,0 +1,105 @@
...
+        self.name, self.description, self.visibility, self.__model_card, self.model_card_version, self.model_card_schema = retrieve_model(client=self.client, model_id=self.model_id)
+
+    def publish(self):
+        if self.model_id is None:
+            res = self.client.post_model(name=self.name, description=self.description, visibility=self.visibility)
+            self.model_id = res['model']['id']
+
+            # Upload model card
+            if self.__model_card is not None:
+                self.client.model_card_from_schema(model_id=self.model_id, schema_id=self.model_card_schema)
```
a3957273, line 57: I think there should be individual methods on this to do things like update the model card.

lib/python-beta/src/bailo/helper/model.py
```diff
@@ -0,0 +1,105 @@
...
+        self.refresh()
+        self.__mc_change = 0
+        self.__file_change = 0
+
+    def download(self):
+        os.makedirs(self.__local_dir, exist_ok=True)
+        res = self.client.get_files(model_id=self.model_id)
+        files = res['files']
+        for file in files:
+            file_path = f"{self.__local_dir}/{file['name']}"
```
a3957273, line 82: This might not be safe, as the file name is _technically_ user provided.  I don't know whether we want to do things about this.

lib/python-beta/src/bailo/helper/utils.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
+from bailo.core import Client, ModelVisibility
+
+def retrieve_model(client: Client, model_id: str):
```
a3957273, line 4: This seems like it should be handled in the `Model.from_id()` function.

a3957273 (changes requested):

lib/python-beta/src/bailo/helper/model.py
a3957273: In the `release` PR it was mentioned this wouldn't be inherited by classes that extend this.  So maybe keeping it as a `classmethod` is better.  Could we at least give the variable name a better name than `cls` though? :)

PR #867, opened 2023-11-10
Feature/bai 951 model contributors and owners should be able to edit releases

ARADDCC012:


frontend/src/common/MultiFileInput.tsx
```diff
@@ -47,43 +49,39 @@ export default function MultiFileInput({
...
-          disabled={disabled}
+    <Box sx={{ ...(fullWidth && { width: '100%' }) }}>
+      {!readOnly && (
+        <Box mb={2}>
+          <label htmlFor={htmlId}>
+            <Button
+              component='span'
+              variant='outlined'
+              disabled={disabled}
+              sx={{ ...(fullWidth && { width: '100%' }) }}
```
ARADDCC002, line 68: I think `<Button />` has a `fullWidth` property so we can probably do `<Button fullWidth={fullWidth} />`

frontend/src/model/beta/releases/EditableRelease.tsx
```diff
@@ -0,0 +1,138 @@
...
+  useEffect(() => {
+    resetForm()
+  }, [resetForm])
+
+  useEffect(() => {
+    setUnsavedChanges(isEdit)
+  }, [isEdit, setUnsavedChanges])
+
+  if (isModelError) {
+    return <MessageAlert message={isModelError.info.message} severity='error' />
```
ARADDCC002, line 50: Did we decide on how we wanted to handle errors? Happy to leave this for another ticket if that's what we decided

frontend/src/model/beta/releases/EditableRelease.tsx
```diff
@@ -0,0 +1,138 @@
...
+    setIsLoading(false)
+    setIsEdit(false)
+    mutateReleases()
+  }
+
+  return (
+    <Box py={1}>
+      <EditableFormHeading
+        heading={
+          <div>
```
ARADDCC002, line 106: Does this need to be a div or can we use a fragment?

frontend/src/model/beta/releases/ReleaseForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const handleSemverChange = (event: ChangeEvent<HTMLInputElement>) => {
+    onSemverChange(event.target.value)
+  }
+
+  const handleMinorReleaseChange = (_event: ChangeEvent<HTMLInputElement>, checked: boolean) => {
+    onMinorReleaseChange(checked)
+  }
+
+  const releaseNotesLabel = (
+    <Typography component='label' sx={{ fontWeight: 'bold' }} htmlFor={'new-model-description'}>
```
ARADDCC002, line 63: Can we replace the `sx={}` prop with fontWieght prop instead `fontWeight: 'bold'`

ARADDCC002 (changes requested):

frontend/src/model/beta/releases/EditableRelease.tsx
ARADDCC012: Yeah this is what we decided on. Any components that don't include the `Wrapper` can safely use a `MessageAlert`. Any of the page components with their own `Wrapper` will need to use the `MultipleErrorWrapper`

frontend/src/model/beta/releases/EditableRelease.tsx
ARADDCC012: Internally the `EditableFormHeading` puts the `heading` inside of a `Stack direction='row'` so the `div` is needed so that these are rendered on top of each other instead of side-by-side. I've had to do the same for access requests too.

I thought about managing this within `EditableFormHeading`, but I decided to let us decide how the heading will look on a case-by-case basis instead.

frontend/src/model/beta/releases/ReleaseForm.tsx
ARADDCC012: I've changed this across the app for consistency

ARADDCC002 (approved):

PR #875, opened 2023-11-13
BAI-974 Schema Class

bw27492:


lib/python-beta/src/bailo/helper/schema.py
```diff
@@ -0,0 +1,62 @@
...
+        name = res['name']
+        kind = res['kind']
+        json_schema = res['jsonSchema']
+
+        if kind == 'model':
+            kind = SchemaKind.Model
+        if kind == 'accessRequest':
+            kind = SchemaKind.AccessRequest
+
+        return schema_id, name, kind, json_schema
```
GB27247, line 50: The object deconstruction is looking good. However I don't think think this a good use case for a `staticmethod` as the functionality is very object specific. I'd consider modifying your `classmethod` to use something `cls.__unpack()` after instantiating an object?

GB27247: Being a bit picky here but two prefixes and suffix underscores in the method name usually denotes magic methods. Private methods just have two prefixed underscores.

PR #876, opened 2023-11-14
updated currentUser actions to use v2

ARADDCC002:


frontend/actions/user.ts
```diff
@@ -1,30 +1,46 @@
+import qs from 'querystring'
 import useSWR from 'swr'
 
-import { User } from '../types/types'
 import { ErrorInfo, fetcher } from '../utils/fetcher'
 
-export function useListUsers() {
+interface EntityKind {
```
ARADDCC012, line 6: Does `EntityGroup` or `EntityResults` make more sense as a name? The entity kind is the string, so it feels strange to also refer to the whole object as that.

frontend/src/MuiForms/EntitySelectorBeta.tsx
```diff
@@ -27,11 +29,14 @@ export default function EntitySelectorBeta(props: EntitySelectorBetaProps) {
   const entities = useMemo(() => {
     if (!users) return []
 
-    return users.map((user) => user.id)
+    const userKind = users.find((userGroup) => userGroup.kind === 'user')
```
ARADDCC012: Again not sure about `userKind` as a name. `userGroup` makes more sense, but then you'd need to rename the param inside the `find` to maybe something like `userGrp` or `usrGroup`?

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -45,13 +44,22 @@ export default function ModelAccess({ model }: ModelAccessProps) {
...
 
-  function onUserChange(_event: React.SyntheticEvent<Element, Event>, newValue: User | null) {
-    if (
-      newValue &&
-      !accessList.find(({ entity }) => entity === `user:${newValue.id}` || entity === `group:${newValue.id}`)
-    ) {
+  const entities = useMemo(() => {
+    if (!users) return []
+
+    const userKind = users.find((userGroup) => userGroup.kind === 'user')
```
ARADDCC012, line 50: See https://github.com/gchq/Bailo/pull/876/files#r1392858477

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -45,13 +44,22 @@ export default function ModelAccess({ model }: ModelAccessProps) {
...
+    if (userKind) {
+      // return userKind.entities.map((entity) => entity.split(':')[1])
```
ARADDCC012: Commented out code?

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -96,10 +104,10 @@ export default function ModelAccess({ model }: ModelAccessProps) {
               setOpen(false)
             }}
             // we might get a string or an object back
```
ARADDCC012: Is this comment still true?

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -47,10 +47,9 @@ export default function ModelAccess({ model }: ModelAccessProps) {
   const entities = useMemo(() => {
     if (!users) return []
 
-    const userKind = users.find((userGroup) => userGroup.kind === 'user')
-    if (userKind) {
-      // return userKind.entities.map((entity) => entity.split(':')[1])
-      return userKind.entities
+    const usrGroup = users.find((usrGroup) => usrGroup.kind === 'user')
```
ARADDCC012, line 50: Although not a problem due to scoping being a thing, could we change to the following to avoid confusion between the two vars?
```
const userGroup = users.find((usrGroup) => usrGroup.kind === 'user')
```

ARADDCC012 (approved):

PR #878, opened 2023-11-15
updated model access list so that it uses a query to find users

ARADDCC002:


JR40159 (changes requested):
We shouldn't be including 'functions' within react components that are passed to props.  Otherwise, on every re-render the function definition changes which causes a complete re-render of the tree.

Instead, we should be using functions like `useCallback()` which are guaranteed to be stable across rerenders.

This applies in both files.

JR40159:
You've also removed `endAdornment`, does this mean that there's no feedback to the user anymore when that query is running?

JR40159 (approved):

PR #885, opened 2023-11-17
added a formatter for displaying user dns

ARADDCC002:


ARADDCC002:
Decided this would be better to handle on the backend so this can be closed.

PR #889, opened 2023-11-17
implemented model ui filtering#

ARADDCC002:


TT38665 (approved):

frontend/pages/beta/index.tsx
```diff
@@ -17,44 +17,32 @@ import {
...
-  const [selectedLibrary, setSelectedLibrary] = useState('')
+  const [selectedLibraries, setSelectedLibraries] = useState<string[]>([])
   const [selectedTask, setSelectedTask] = useState('')
-  const [selectedType, setSelectedType] = useState<MarketPlaceModelSelectType | ''>('')
+  const [selectedTypes, setSelectedTypes] = useState<string[]>([])
   const debouncedFilter = useDebounce(filter, 250)
 
-  const { models, isModelsError, mutateModels } = useListModels(group, debouncedFilter)
+  const { models, isModelsError } = useListModels(
+    selectedTypes.join(),
```
a3957273, line 38: Why do you have `.join()` here.  This will turn it into a string like `JAX,Transformers` which isn't how the search works.

frontend/pages/beta/index.tsx
```diff
@@ -125,17 +113,19 @@ export default function ExploreModels() {
...
           </Box>
           <Box>
-            <ChipSelector<MarketPlaceModelSelectType>
+            <ChipSelector
               label='Other'
-              tags={[MarketPlaceModelSelectType.MY_MODELS, MarketPlaceModelSelectType.FAVOURITES]}
-              onChange={setSelectedType}
-              selectedTags={selectedType}
+              multiple
+              tags={['mine']}
```
a3957273, line 126: Everything else starts with a capital, can we make this an object so that we can display it as 'Mine' but send the code as 'mine'?

a3957273 (changes requested):

frontend/pages/beta/index.tsx
ARADDCC002: You can either do:

`list=a,b,c,d` **or** `list=a&list=b&list=c&=d` I believe both are valid depending how the backend is set up, I wasn't sure which way we were doing it so made a guess. I'll change it to the latter approach now, though!

PR #891, opened 2023-11-18
BAI-940 Create Helper Classes

bw27492:


lib/python-beta/src/bailo/helper/release.py
```diff
@@ -0,0 +1,214 @@
...
+        minor: bool = False,
+        draft: bool = True,
+    ) -> Release:
+        """ Builds a release from Bailo and uploads it
+
+        :param client: A client object used to interact with Bailo
+        :param model_id: A Unique Model ID
+        :param version: A semantic version of a model release
+        """
+        if type(version) == Version:
```
a3957273, line 73: This should probably compare it to a string at a guess?

lib/python-beta/src/bailo/helper/release.py
```diff
@@ -0,0 +1,214 @@
...
+            version,
+            model_card_version,
+            notes,
+            files,
+            images,
+            minor,
+            draft
+        )
+
+    def download(self, local_dir: str) -> None:
```
a3957273, line 126: Maybe split into two functions, one to download one file and one to download multiple.

lib/python-beta/src/bailo/helper/reviews.py
a3957273, line 1: Maybe review this `reviews.py`.

lib/python-beta/src/bailo/helper/utils.py
a3957273, line 1: We don't need this anymore!

lib/python-beta/test.txt
a3957273, line 1: Interesting file.  I don't think we need this.

lib/python-beta/tests/test_access_request.py
```diff
@@ -0,0 +1,45 @@
...
+        end_date=end_date
+    )
+
+    # Get access request
+    get_ar = AccessRequest.from_id(integration_client, example_model.model_id, ar.access_request_id)
+
+    # Check they're the same access request
+    assert ar == get_ar
+
+    ar.delete()
```
a3957273, line 45: Do we want to check that this works?

lib/python-beta/tests/test_client.py
```diff
@@ -96,10 +97,10 @@ def test_post_release(requests_mock):
     client = Client("https://example.com")
     result = client.post_release(
         model_id="test_id",
-        model_card_version=1,
+        model_card_version=1.0,
```
a3957273, line 100: I don't think this should be a float!

a3957273 (changes requested):

lib/python-beta/src/bailo/core/client.py
```diff
@@ -260,27 +291,48 @@ def get_files(
...
+        Downloads a specific file.
+
+        :param model_id: Unique model ID
+        :param file_id: Unique file ID
+        :param local_file: (optional) The file on the local machine to be copied to
+        :return: A file object
+        """
+        with self.agent.get(f"{self.url}/v2/model/{model_id}/file/{file_id}/download", stream=True) as r:
+            if localfile_name is None:
+                fp = tempfile.NamedTemporaryFile()
```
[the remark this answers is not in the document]
GB27247: For now I'm just going to use ByteStreams that can be used within a context manager. This will also make them file compatible as well eg. if a user want's to download files
```python
with open("example.txt", "wb") as f:
    release.download(<file-id>, f)
```

a3957273 (approved):

PR #893, opened 2023-11-21
edited userlist query so that it only fires when a search param of at…

ARADDCC002:
… least 3 chars is given#

### DIFF
--- a/frontend/src/model/beta/settings/ModelAccess.tsx
+++ b/frontend/src/model/beta/settings/ModelAccess.tsx
@@ -26,6 +26,11 @@ import useNotification from '../../../common/Snackbar'
 import MessageAlert from '../../../MessageAlert'
 import EntityItem from './EntityItem'
 
+type EntityForGrouping = {
+  kind: string
+  label: string
+}
+
 type ModelAccessProps = {
   model: ModelInterface
 }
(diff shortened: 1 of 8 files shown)


JR40159 (approved):

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -51,19 +56,23 @@ export default function ModelAccess({ model }: ModelAccessProps) {
   const entities = useMemo(() => {
     if (!users) return []
 
-    const userGroup = users.find((usrGroup) => usrGroup.kind === 'user')
-    if (userGroup) {
-      return userGroup.entities
-    } else {
-      return []
+    const userList = users.find((usrGroup) => usrGroup.kind === 'user')
```
JR40159, line 59: We need to review what the backend is returning to avoid:
1. Having to do unnecessary computation on parsing/ sorting the response.
2. Hard-coding Entity Kinds here. Ideally the frontend should just present the information it receives from the backend without needing to know or work-out what Entity Kinds are included in the response.

JR40159 (changes requested):

frontend/src/model/beta/settings/ModelAccess.tsx
```diff
@@ -51,19 +56,23 @@ export default function ModelAccess({ model }: ModelAccessProps) {
...
+    const groupList = users.find((usrGroup) => usrGroup.kind === 'group')
```
a3957273, line 60: There's not going to be a 'group' type, instead there will be dynamically defined group labels as defined by each deployment.  Forcing people to change these values here seems like a drag.

a3957273 (changes requested):

JR40159 (approved):

ARADDCC012 (approved):

PR #894, opened 2023-11-21
Refactor Release Dialog

ARADDCC012:


frontend/pages/beta/model/[modelId]/access-request/schema.tsx
```diff
@@ -32,7 +32,7 @@ export default function NewSchemaSelection() {
       {schemas && !isSchemasLoading && (
         <Container maxWidth='md'>
           <Card sx={{ mx: 'auto', my: 4, p: 4 }}>
-            <Link href={`/beta/model/${modelId}`}>
+            <Link href={`/beta/model/${modelId}?tab=access`}>
```
Comment (ARADDCC002) [inline frontend/pages/beta/model/[modelId]/access-request/schema.tsx:35]: I'm not sure about this - what if someone links you this page? or if you were on a different tab and then hit the "Request Access" button above the tabbed view?

frontend/pages/beta/model/[modelId]/release/[semver].tsx
```diff
@@ -34,34 +34,35 @@ export default function Release() {
 
   if (error) return error
 
+  if (!release || (isReleaseLoading && isActiveReviewsLoading)) {
```
Comment (ARADDCC002) [inline frontend/pages/beta/model/[modelId]/release/[semver].tsx:37]: Should we be handling `!release` differently?

frontend/pages/beta/model/[modelId]/release/new.tsx
```diff
@@ -0,0 +1,151 @@
...
+        <Card sx={{ my: 4, p: 4 }}>
+          <Box component='form' onSubmit={handleSubmit}>
+            <Stack spacing={4}>
+              <Link href={`/beta/model/${modelId}?tab=releases`}>
+                <Button sx={{ width: 'fit-content' }} startIcon={<ArrowBack />}>
+                  Back to model
+                </Button>
+              </Link>
+              <Stack spacing={2} alignItems='center' justifyContent='center'>
+                <Typography variant='h6' color='primary'>
```
Comment (ARADDCC002) [inline frontend/pages/beta/model/[modelId]/release/new.tsx:107]: I appreciate this may not have been you, but if we put `variant='h6'` do we also need to use the `component` prop so that it renders the appropriate heading level under the hood?

frontend/src/common/MultiFileInput.tsx
```diff
@@ -75,22 +75,24 @@ export default function MultiFileInput({
...
             <Button fullWidth component='span' variant='outlined' disabled={disabled}>
               {label}
             </Button>
           </label>
           <Input multiple id={htmlId} type='file' onInput={handleFileChange} accept={accepts} disabled={disabled} />
-        </Box>
+        </>
+      )}
+      {files.length > 0 && (
+        <Stack spacing={1} mt={2}>
```
ARADDCC002, line 88: TIL you can use system props outside of `sx`!

frontend/src/model/beta/Releases.tsx
```diff
@@ -28,13 +28,10 @@ export default function Releases({ model }: { model: ModelInterface }) {
     }
   }, [model, releases])
 
-  function handleDraftNewReleaseClose() {
-    setOpenDraftNewRelease(false)
-  }
-
   function handleDraftNewRelease() {
```
ARADDCC002, line 31: Instead of using the `router.push()` function could we wrap the button in  `<Link />` to keep consistency with button navigations

ARADDCC002 (changes requested):

frontend/pages/beta/model/[modelId]/release/[semver].tsx
Comment (ARADDCC012) [reply frontend/pages/beta/model/[modelId]/release/[semver].tsx:37]: If we don't have a release and it hasn't errored then it just means it's loading, so makes sense to put in this block imo. Also means we don't have conditional logic down below in our main JSX block.

frontend/pages/beta/model/[modelId]/release/new.tsx
Comment (ARADDCC012) [reply frontend/pages/beta/model/[modelId]/release/new.tsx:107]: Good spot, I'll fix this for access requests too

frontend/src/common/MultiFileInput.tsx
ARADDCC012: Most, if not all, MUI components support system props now :)

ARADDCC002 (approved):

PR #902, opened 2023-11-27
navigated to newly created access request page

ARADDCC002:


frontend/src/common/Loading.tsx
```diff
@@ -8,7 +8,7 @@ type LoadingProps = {
 export default function Loading({ size = 40 }: LoadingProps): ReactElement {
   return (
     <Box sx={{ display: 'flex', justifyContent: 'center' }}>
-      <CircularProgress size={size} />
+      <CircularProgress color='primary' size={size} />
```
ARADDCC012, line 11: primary is the default so can be removed

frontend/src/model/beta/AccessRequests.tsx
```diff
@@ -19,9 +19,11 @@ export default function AccessRequests({ model }: AccessRequestsProps) {
   const accessRequestsList = useMemo(
     () =>
       accessRequests.length ? (
-        accessRequests.map((accessRequest) => (
-          <AccessRequestDisplay accessRequest={accessRequest} key={accessRequest.metadata.overview.name} />
-        ))
+        accessRequests
+          .sort((a, b) => b.createdAt.localeCompare(a.createdAt))
```
ARADDCC012, line 23: This sort can be replaced by the reusable `sortByCreatedAtDescending` from `dateUtils.ts`

ARADDCC012 (approved):

PR #904, opened 2023-11-28
Bugfix/bai 1004 show existing version number

TT38665:


frontend/src/model/beta/releases/ReleaseForm.tsx
```diff
@@ -10,6 +12,8 @@ import { FileWithMetadata, FlattenedModelImage } from 'types/interfaces'
 import { ModelInterface } from 'types/v2/types'
 import { isValidSemver } from 'utils/stringUtils'
 
+//import { ReleaseInterface } from '../../../../types/types'
```
ARADDCC012, line 15: Commented out code

frontend/src/model/beta/releases/ReleaseForm.tsx
```diff
@@ -54,6 +58,8 @@ export default function ReleaseForm({
   isEdit = false,
 }: ReleaseFormProps) {
   const isReadOnly = useMemo(() => editable && !isEdit, [editable, isEdit])
+  const [latestRelease, setLatestRelease] = useState<string>('')
```
ARADDCC012, line 61: This can be simplified to `useState('')`. It'll automatically type is as a string as you've initialised it as so

frontend/src/model/beta/releases/ReleaseForm.tsx
```diff
@@ -54,6 +58,8 @@ export default function ReleaseForm({
...
+  const { releases, isReleasesLoading } = useGetReleasesForModelId(model.id)
```
ARADDCC012, line 62: You're not handling the error from this

frontend/src/model/beta/releases/ReleaseForm.tsx
```diff
@@ -63,93 +69,112 @@ export default function ReleaseForm({
     onMinorReleaseChange(checked)
   }
 
+  useEffect(() => {
```
ARADDCC012, line 72: You could replace the `useState` and `useEffect` with a single `useMemo` like so:
```
const latestRelease = useMemo(() => releases && releases.length > 0 ? releases[0].semver : '', [])
```

frontend/src/model/beta/releases/ReleaseForm.tsx
```diff
@@ -63,93 +69,112 @@ export default function ReleaseForm({
...
+    if (model && releases.length > 0) {
```
ARADDCC012, line 73: This won't matter if you use the code from https://github.com/gchq/Bailo/pull/904/files#r1407761074, but there's still a learning point here.

The model isn't needed to get the latest release semver, so that isn't needed in this check.

frontend/src/model/beta/releases/ReleaseForm.tsx
```diff
@@ -63,93 +67,112 @@ export default function ReleaseForm({
...
-          </>
-        ) : (
-          <FormControl>
-            <FormControlLabel
-              control={<Checkbox size='small' checked={formData.isMinorRelease} onChange={handleMinorReleaseChange} />}
-              label='Minor release - No significant changes, does not require release re-approval'
+        <Stack>
+          {!editable && (
+            <>
+              <Typography fontWeight='bold'>Latest semantic version</Typography>
```
ARADDCC012, line 102: Up for discussion but I'd vote for this simply saying "Latest version". We don't need to drill home that it's a semver at every opportunity imo :)

ARADDCC012 (approved):

PR #911, opened 2023-11-29
added date picker widget for forms

ARADDCC002:


frontend/src/MuiForms/DateSelector.tsx
```diff
@@ -0,0 +1,60 @@
...
+      onChange(dateInput.toDate().toISOString().split('T')[0])
+    }
+    if (dateInput === null) {
+      onChange('')
+    }
+  }
+
+  const displayValue = (dateValue: string) => {
+    const [year, month, date] = dateValue.split('-')
+    return `${date}/${month}/${year}`
```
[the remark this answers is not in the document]
a3957273: I also don't think we should be changing the display value of this.  If we're storing it as `DD-MM-YYYY` then we should display it as `DD-MM-YYYY`.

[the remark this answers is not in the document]
a3957273: Also, just in general, you don't want to format dates yourself but use the internalisation libraries within JavaScript.  Different countries will format dates very differently.

frontend/src/MuiForms/DateSelector.tsx
```diff
@@ -0,0 +1,60 @@
...
+  InputProps?: any
+}
+
+export default function DateInput(props: CustomTextInputProps) {
+  const { onChange, value, label, formContext } = props
+
+  const theme = useTheme()
+  const handleChange = (dateInput: Dayjs | null) => {
+    if (dateInput && dateInput.isValid()) {
+      onChange(dateInput.toDate().toISOString().split('T')[0])
```
a3957273, line 28: Using `.format()` is likely clearer than this.  E.g. `dateInput.format('YYYY-MM-DD')`.

frontend/src/MuiForms/DateSelector.tsx
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+
+  const displayValue = (dateValue: string) => {
+    const [year, month, date] = dateValue.split('-')
+    return `${date}/${month}/${year}`
+  }
+
+  return (
+    <Fragment key={label}>
```
a3957273, line 41: You don't use required anywhere here, so there's no way to tell if it's mandatory or not.

frontend/src/MuiForms/DateSelector.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <Typography fontWeight='bold'>{label}</Typography>
+      {formContext.editMode && (
+        <LocalizationProvider dateAdapter={AdapterDayjs} adapterLocale='en-gb'>
```
a3957273, line 44: This can probably be far higher in the React tree, e.g. up at the top in `_app.tsx`.

a3957273 (changes requested):

frontend/src/MuiForms/DateSelector.tsx
ARADDCC002: I thought the documentation meant you had to had to individually wrap the date pickers with these providers but if we can do it at the top level that'd be even better.

a3957273: On a different project we managed to just put it at the top level, so can probably do that here too unless they've changed something.

a3957273 (approved):

PR #912, opened 2023-11-30
Revamp authorisation to support returning information and batching

a3957273:


a3957273:
We haven't updated or created any tests for this, so it won't pass inspection, but consideration over the approach would be useful.

JR40159 (dismissed):

backend/src/connectors/v2/authorisation/base.ts
```diff
@@ -0,0 +1,336 @@
...
+          }
+        }
+
+        return { success: true, id: access.name }
+      }),
+    )
+  }
+}
+
+// type Response = { success: true; info: undefined } | { success: false; info: string }
```
ARADDCC002: Is there a reason this is commented out? If so can we add a "todo" to this part of the code, ideally with a Jira ticket if possible.

backend/src/scripts/migrateV2.ts
```diff
@@ -0,0 +1,177 @@
...
+      roles: ['msro', 'owner'],
+    })
+  }
+
+  for (const reviewer of model.latestVersion.metadata.contacts.reviewer) {
+    const entity = `${reviewer.kind}:${identityConversion(reviewer.id)}`
+
+    const search = collaborators.find((entry) => entry.entity === entity)
+    if (search) {
+      search.roles.push('mtr')
```
ARADDCC002, line 73: Are these role values `mtr` etc typed anywhere?

backend/src/services/v2/model.ts
```diff
@@ -116,13 +119,14 @@ export async function searchModels(
     }
   }
 
-  const results = ModelModel
+  const results = await ModelModel
     // Find only matching documents
```
ARADDCC002, line 123: I don't know if these comments really add much, it seems quite self explanatory to me

ARADDCC002 (approved):

PR #913, opened 2023-11-30
Fix broken access request onSubmit function

a3957273:


frontend/pages/beta/model/[modelId]/access-request/new.tsx
```diff
@@ -51,32 +52,41 @@ export default function NewAccessRequest() {
   async function onSubmit() {
     setSubmissionErrorText('')
     setSubmitButtonLoading(true)
+
+    if (!modelId || !schemaId) {
+      setSubmissionErrorText(`Please wait until the page has hydrated before attempting to submit.`)
```
Comment (ARADDCC002) [inline frontend/pages/beta/model/[modelId]/access-request/new.tsx:57]: Is the term "hydration" a little too technical here?

ARADDCC002 (changes requested):

frontend/pages/beta/model/[modelId]/release/new.tsx
```diff
@@ -19,71 +19,69 @@ export default function NewRelease() {
...
+      minor: isMinorRelease,
+      fileIds: fileIds,
+      images: imageList,
+    }
+
+    const response = await postRelease(release)
+
+    if (!response.ok) {
+      setLoading(false)
+      return setErrorMessage(await getErrorMessage(response))
```
Comment (ARADDCC012) [inline frontend/pages/beta/model/[modelId]/release/new.tsx:78]: The way we were getting the error message and setting the loading state is preferred here as we `setLoading(false)` once all of the asynchronous logic has finished

Comment (ARADDCC012) [inline frontend/pages/beta/model/[modelId]/release/new.tsx:78]: The way we were getting the error message and setting the loading state previously is preferred here as we were calling `setLoading(false)` once all of the asynchronous logic had finished

frontend/pages/beta/model/[modelId]/release/new.tsx
```diff
@@ -19,71 +19,69 @@ export default function NewRelease() {
...
-            fileIds.push(res.file._id)
-          } else {
-            setLoading(false)
-            return setErrorMessage(await getErrorMessage(postArtefactResponse))
-          }
-        }
 
-        setLoading(false)
+    if (!model) {
+      return
```
Comment (ARADDCC012) [inline frontend/pages/beta/model/[modelId]/release/new.tsx:36]: This isn't something that we should ever reach, but if we do we should show an error message that encourages users to report what is most definitely a bug (I know we'd missed this already with the old logic). Something like "Failed to submit, model not found. Please contact Bailo support"

frontend/src/common/MultiFileInputFileDisplay.tsx
```diff
@@ -21,7 +22,7 @@ export default function MultiFileInputFileDisplay({ file, handleDelete, onChange
 
   return (
     <Grid container spacing={1} alignItems='center'>
-      <Grid item xs={4}>
+      <Grid item>
```
ARADDCC012, line 25: Grid items should always specify how much space they want to use. E.g. `xs={4}` means it wants to use 4 out of the 12 available columns for any screen size `xs` and above (you can also specify higher breakpoints if you want the layout to change for different screen sizes), `xs` means fill up the remaining columns or share evenly with any other grid items looking to do that same.

Aside from the file name being a Chip here and a Link in `ReleaseDispay`, we should probably make this look the same as the grid in `ReleaseDisplay` if we want to show the file size here too.

frontend/src/model/beta/releases/EditableRelease.tsx
```diff
@@ -35,8 +35,8 @@ export default function EditableRelease({ release }: EditableReleaseProps) {
     setSemver(release.semver)
     setReleaseNotes(release.notes)
     setIsMinorRelease(!!release.minor)
-    setArtefacts([]) // TODO - Reset the release artefact files (BAI-1026)
-    setArtefactsMetadata([])
+    setFiles([]) // TODO - Reset the release file files (BAI-1026)
```
ARADDCC012, line 38: This comment no longer reads correctly. Change to `// TODO - Reset the release files (BAI-1026)`?

ARADDCC012 (changes requested):

ARADDCC012 (approved):

PR #934, opened 2023-12-01
Various bugfixes and code cleanup

ARADDCC012:


frontend/src/model/beta/reviews/ReviewBanner.tsx
```diff
@@ -106,7 +106,7 @@ export default function ReviewBanner({ release, accessRequest }: ReviewBannerPro
         borderWidth: '1px',
         borderStyle: 'solid',
         borderColor: theme.palette.primary.main,
-        borderRadius: '4px 4px 0px 0px',
+        borderRadius: 0,
```
ARADDCC002, line 109: If we don't want a radius can we delete this? Or is it inheriting a default value from somewhere else?

PR #937, opened 2023-12-04
Feature/bai 1033 add enum v2 form widget

ARADDCC002:


frontend/src/MuiForms/Dropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+            label: {
+              WebkitTextFillColor:
+                theme.palette.mode === 'light' ? theme.palette.common.black : theme.palette.common.white,
+            },
+            '& .MuiInputBase-input.Mui-disabled': {
+              WebkitTextFillColor: disabledWebkitTextFillColor,
+            },
+          }}
+          onChange={handleChange}
+          value={value || (!formContext.editMode ? 'Unanswered' : '')}
```
ARADDCC012, line 60: Setting the `value` to `'Unanswered'` will cause a react error as it's not one of the options. You should instead set that as a `placeholder` on the `renderInput` `TextField` below

ARADDCC012 (approved):

PR #947, opened 2023-12-08
Bai 1008 send email after review response

TT38665:


backend/src/services/v2/review.ts
```diff
@@ -132,9 +139,38 @@ export async function respondToReview(
   if (!update) {
     throw GenericError(500, `Adding response to Review was not successful`, { modelId, reviewIdQuery, role })
   }
+  sendReviewNotification(update, user)
   return update
 }
 
+export async function sendReviewNotification(review: ReviewDoc, user: UserDoc) {
```
JR40159, line 146: Could we change this to `sendReviewResponseNotification` please? A 'Review' is normally what we create when someone creates a release or an access request, here we wanting to send a notification when someone responds to that review.

backend/src/services/v2/review.ts
```diff
@@ -132,9 +139,38 @@ export async function respondToReview(
...
+  let reviewIdQuery
+  switch (review.kind) {
+    case ReviewKind.Access: {
+      if (!review.accessRequestId) {
+        log.info('Cannot find modelId')
```
JR40159, line 151: I'd be tempted to upgrade this log to an error. We've only added this if statement to be type safe and it shouldn't ever happen. Also maybe change the log message to 'Unable to send notification for review response. Cannot find access request ID.

backend/src/services/v2/smtp/smtp.ts
```diff
@@ -86,6 +88,88 @@ export async function requestReviewForAccessRequest(
...
 
+export async function requestResponsePostReleaseReviewResponse(review: ReviewDoc, release: ReleaseDoc) {
+  if (!config.smtp.enabled) {
+    log.info('Not sending email due to SMTP disabled')
+    return
+  }
+  const reviewResponse = review.responses[0]
+
+  if (!reviewResponse) {
+    log.info('response not found')
```
JR40159, line 99: I'd probably say that same here as with the log statements in `review.ts`

backend/src/services/v2/smtp/smtp.ts
```diff
@@ -86,6 +88,88 @@ export async function requestReviewForAccessRequest(
...
+    return
+  }
+  const emailContent = buildEmail(
+    `Release ${release.semver} has been reviewed by ${reviewResponse?.user}`,
+    [
+      { title: 'Model ID', data: release.modelId },
+      { title: 'Reviewed on', data: review.createdAt.toDateString() },
```
JR40159, line 106: I don't think that date is when the person responded to the review but when the review was created. I wouldn't worry too much about included that info here anyway as the email should get sent pretty soon after someone responds to the review so the email timestamp could be used.

backend/src/services/v2/smtp/smtp.ts
```diff
@@ -86,6 +88,88 @@ export async function requestReviewForAccessRequest(
   await Promise.all(sendEmailResponses)
 }
 
+export async function requestResponsePostReleaseReviewResponse(review: ReviewDoc, release: ReleaseDoc) {
```
JR40159, line 91: Unlike the other previous emails we've added here, these emails aren't _requesting_ the recipient to do something. Maybe `notifyReviewResponseForRelease` instead, although that is along a bit long.

backend/src/services/v2/smtp/smtp.ts
```diff
@@ -86,6 +88,88 @@ export async function requestReviewForAccessRequest(
...
+  const reviewResponse = review.responses[0]
+
+  if (!reviewResponse) {
+    log.info('response not found')
+    return
+  }
+  const emailContent = buildEmail(
+    `Access request for model ${
+      accessRequest.modelId
+    } has been reviewed by ${accessRequest.metadata.overview.entities.toString()}`,
```
JR40159, line 146: `accessRequest.metadata.overview.entities` won't be the user who has just reviewed the access request. I think you're looking for `reviewResponse.user`?

JR40159 (changes requested):

backend/src/services/v2/smtp/smtp.ts
```diff
@@ -86,6 +88,80 @@ export async function requestReviewForAccessRequest(
...
+
+  if (!reviewResponse) {
+    log.info('response not found')
+    return
+  }
+  const emailContent = buildEmail(
+    `Release ${release.semver} has been reviewed by ${reviewResponse?.user}`,
+    [
+      { title: 'Model ID', data: release.modelId },
+      { title: 'Decision', data: reviewResponse.decision.replace(/_/g, ' ') },
```
JR40159, line 106: Maybe for both access requests and releases, you could have 'Reviewer Role' instead of calling what you had before 'Your Role'?

JR40159 (changes requested):

JR40159 (approved):

PR #950, opened 2023-12-11
Bugfix/bai 918 model card errors

ARADDCC012:


frontend/pages/beta/model/[modelId]/schema.tsx
```diff
@@ -34,9 +37,17 @@ export default function NewSchemaSelection() {
...
     if (currentUser && model) {
-      await postFromSchema(model.id, newSchema.id)
-      await mutateModel()
-      router.push(`/beta/model/${modelId}`)
+      setLoading(true)
+      setErrorMessage('')
+
+      const response = await postFromSchema(model.id, newSchema.id)
+
+      if (response.status && response.status < 400) {
```
Comment (ARADDCC002) [inline frontend/pages/beta/model/[modelId]/schema.tsx:45]: Should this not be if(!response.ok) ?

frontend/src/model/beta/overview/FormEditPage.tsx
```diff
@@ -34,12 +35,15 @@ export default function FormEditPage({ model }: FormEditPageProps) {
 
   async function onSubmit() {
     if (schema) {
+      setErrorMessage('')
       setLoading(true)
       const data = getStepsData(splitSchema, true)
       const res = await putModelCard(model.id, data)
       if (res.status && res.status < 400) {
```
ARADDCC002, line 42: Can we use `if (!res.ok)`

frontend/utils/fetcher.ts
```diff
@@ -66,6 +66,7 @@ export const getErrorMessage = async (res: Response) => {
     messageError = `${res.statusText}: ${(await res.json()).error.message}`
   } catch (e) {
     // unable to identify error message, possibly a network failure
+    return 'Unknown error - Please contact Bailo support'
```
ARADDCC002, line 69: Is it possible to put a link directly to our help page here?

frontend/pages/beta/model/[modelId]/schema.tsx
Comment (ARADDCC012) [reply frontend/pages/beta/model/[modelId]/schema.tsx:45]: It's an axios request so the response doesn't have an `ok` property. The `response.statusText` will be `'OK'` if the `response.status` is `200`, but we parse the response through `handleAxiosError` which ignores the `statusText`. If we want to change this I'd argue we can easily do it later on.

frontend/src/model/beta/overview/FormEditPage.tsx
ARADDCC012: See https://github.com/gchq/Bailo/pull/950#discussion_r1425246559

frontend/utils/fetcher.ts
ARADDCC012: It would be nice to, but this is just a util function that returns a string. We'd need additional logic everywhere it's being used to route the user to that page in the event of this happening.

I'm not against us doing that, but this isn't strictly a part of this ticket. I just noticed we weren't providing any feedback in this very niche scenario, so figured I'd put at least something in.

We can raise a ticket to address this if you want?

ARADDCC002 (approved):

PR #952, opened 2023-12-12
Bugfix/bai 1041 use full entity name for access requests

ARADDCC002:


frontend/src/MuiForms/EntitySelectorBeta.tsx
```diff
@@ -68,18 +66,18 @@ export default function EntitySelectorBeta(props: EntitySelectorBetaProps) {
             setOpen(false)
           }}
           disableClearable
-          isOptionEqualToValue={(option: string, value: string) => option === value}
-          getOptionLabel={(option) => option}
-          value={currentValue || []}
+          isOptionEqualToValue={(option: any, value: any) => option.id === value.id}
```
ARADDCC012, line 69: Can these `any`s be changed for `EntityObject`?

ARADDCC012, line 69: You should be able to provide a type to the Autocomplete like so
```
<Autocomplete<EntityObject, true, true>
  multiple
  open={open}
...
```
You can then remove these `any`s and let the Autocomplete figure out the type for us (which will be `EntityObject`)

a3957273 (approved):

PR #959, opened 2023-12-15
added file downloads to release page

ARADDCC002:


frontend/src/model/beta/releases/ReleaseForm.tsx
```diff
@@ -69,6 +71,12 @@ export default function ReleaseForm({
     onMinorReleaseChange(checked)
   }
 
+  const getFileId = (file: File | FileInterface) => {
+    if (isFileInterface(file)) {
```
a3957273, line 75: What if it's not a file interface?

a3957273 (changes requested):

a3957273 (approved):

PR #960, opened 2023-12-17
Initial publishing job config

bw27492:


.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,74 @@
...
+        with:
+          fetch-depth: 0
+
+      # Check for changed init file (for new version)
+      - name: Get changed init files
+        id: files-changed
+        uses: tj-actions/changed-files@v40
+        with:
+          files: |
+            lib/python-beta/src/bailo/__init__.py
```
GB27247, line 25: I just want to point out, this will try and publish to PyPi if it detects any change in this file (Not just the version number).  However, I'm satisfied that it wouldn't be an issue with correct versioning.

GB27247 (approved):

PR #989, opened 2024-01-04
Enforce unique file names on releases

a3957273:


a3957273:
No tests yet.

backend/src/routes/v2/model/file/getDownloadFile.ts
JR40159, line 1: Does this need an Open API spec?

backend/src/routes/v2/model/file/getDownloadFile.ts
```diff
@@ -25,9 +36,15 @@ interface GetDownloadFileResponse {
...
+
+    let fileId: string
+    if ('semver' in params) {
+      const file = await getFileByReleaseFileName(req.user, params.modelId, params.semver, params.fileName)
+      fileId = String(file._id)
+    } else {
+      fileId = params.fileId
+    }
 
     const file = await getFileById(req.user, fileId)
```
JR40159, line 49: If a semver is provided, will this be the second time you retrieve the same file?

backend/src/services/v2/release.ts
```diff
@@ -55,6 +59,19 @@ async function validateRelease(user: UserDoc, model: ModelDoc, release: ReleaseD
         )
       }
     }
+
+    const uniqueDuplicates = [...new Set(findDuplicates(fileNames))]
+    if (uniqueDuplicates.length) {
+      throw BadReq(
+        `Releases cannot have multiple files with the same name.  The duplicates in this file are: '${JSON.stringify(
```
JR40159, line 66: Should models have multiple files with the same name? When a user creates a release the UI uploads their files first and then creates the release with those files. Throwing this error from the second request in this process might get a little messy. I assume upon receiving this error a user would normally rename the file and then attempt to again to create the release but under the hood that would duplicate all the files from the first attempt to create the release?

JR40159 (changes requested):

backend/src/services/v2/release.ts
JR40159, line 66: Missing test case?

JR40159 (changes requested):

JR40159 (approved):

PR #995, opened 2024-01-04
changed how reviews work and removed active property

ARADDCC002:
Users can now make as many reviews on a release as they want as the "active" property of a review response has been removed. This means that users can change their mind if they want to, and also means once a release has been removed by one person it does not block another user making their own review. 

Review responses also have timestamps now, and comments have been merged with their respective decision box, as you can see below:

image

TODO:

The release display needs to be updated and only show the latest review for each role. I've temporarily removed it from this section:

image


frontend/src/reviews/ReviewDecision.tsx
```diff
@@ -1,20 +1,20 @@
...
+  response: ReviewResponse
 }
 
-export default function ReviewDecision({ user, decision }: ReviewDecisionProps) {
-  const isApproved = useMemo(() => decision === 'approve', [decision])
+export default function ReviewDecision({ response }: ReviewDecisionProps) {
+  const isApproved = useMemo(() => response.decision === 'approve', [response.decision])
 
-  const username = user.split(':')[0]
+  const username = response.user.split(':')[0]
```
ARADDCC012, line 17: Based off of a comment on your other PR, it think this might need to be `response.user.split(':')[1]`, but I'd double check before trusting me on that

frontend/src/reviews/ReviewsList.tsx
```diff
@@ -1,36 +1,55 @@
...
+export default function ReviewsList({ kind = 'all' }: ReviewsListProps) {
+  const { reviews, isReviewsLoading, isReviewsError } = useGetReviewRequestsForUser()
+  const { currentUser, isCurrentUserLoading, isCurrentUserError } = useGetCurrentUser()
   const [filteredReviews, setFilteredReviews] = useState<ReviewRequestInterface[]>([])
 
+  const doesNotContainUserResponse = useCallback(
+    (review: ReviewRequestInterface) => {
+      return (
+        currentUser &&
+        review.responses.filter(
```
ARADDCC012, line 24: Slight performance improvement if you use a `find` so it exits early if it can:
```
      return (
        currentUser &&
        !review.responses.find(
          (response) => response.user === `user:${currentUser.dn}` && response.decision === Decision.Approve,
        )
      )
```

frontend/src/reviews/ReviewsList.tsx
```diff
@@ -1,36 +1,55 @@
...
 }
 
-export default function ReviewsList({ isActive = true, kind = 'all' }: ReviewsListProps) {
-  const { reviews, isReviewsLoading, isReviewsError } = useGetReviewRequestsForUser(isActive)
+export default function ReviewsList({ kind = 'all' }: ReviewsListProps) {
+  const { reviews, isReviewsLoading, isReviewsError } = useGetReviewRequestsForUser()
+  const { currentUser, isCurrentUserLoading, isCurrentUserError } = useGetCurrentUser()
   const [filteredReviews, setFilteredReviews] = useState<ReviewRequestInterface[]>([])
 
+  const doesNotContainUserResponse = useCallback(
```
ARADDCC012, line 20: This var name and the logic below doesn't match. The logic is checking that the user hasn't approved the review, but the var name suggests it should be looking to see if the user has responded at all.

I'm guessing the logic is correct and the name should probs be changed to something like `doesNotContainUserApproval`. Otherwise the logic will need updating to remove the `response.decision === Decision.Approve` check.

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -79,39 +81,43 @@ export default function SideNavigation({
...
 }: SideNavigationProps) {
   const [reviewCount, setReviewCount] = useState(0)
   const theme = useTheme()
+  const { reviews, isReviewsLoading, isReviewsError } = useGetReviewRequestsForUser()
+
+  const doesNotContainUserResponse = useCallback(
+    (review: ReviewRequestInterface) => {
+      return (
+        currentUser &&
+        review.responses.filter(
```
ARADDCC012, line 96: Slight performance improvement if you use a `find` so it exits early if it can:
```
      return (
        currentUser &&
        !review.responses.find(
          (response) => response.user === `user:${currentUser.dn}` && response.decision === Decision.Approve,
        )
      )
```

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -79,39 +81,43 @@ export default function SideNavigation({
...
-  onError,
   onResetErrorMessage,
   drawerOpen = false,
   pageTopStyling = {},
 }: SideNavigationProps) {
   const [reviewCount, setReviewCount] = useState(0)
   const theme = useTheme()
+  const { reviews, isReviewsLoading, isReviewsError } = useGetReviewRequestsForUser()
+
+  const doesNotContainUserResponse = useCallback(
```
ARADDCC012, line 92: This var name and the logic below doesn't match. The logic is checking that the user hasn't approved the review, but the var name suggests it should be looking to see if the user has responded at all.

I'm guessing the logic is correct and the name should probs be changed to something like `doesNotContainUserApproval`. Otherwise the logic will need updating to remove the `response.decision === Decision.Approve` check.

ARADDCC012 (changes requested):

frontend/src/model/beta/reviews/ReviewBanner.tsx
```diff
@@ -112,26 +104,18 @@ export default function ReviewBanner({ release, accessRequest }: ReviewBannerPro
...
-      </Grid>
+      <Stack
+        direction='row'
+        justifyContent='space-between'
+        alignItems='center'
+        spacing={2}
+        sx={{ px: 2, width: '100%' }}
+      >
+        <ReviewRequiredIcon />
+        <Button variant='outlined' color='inherit' size='small' onClick={handleReviewOpen} data-test='reviewButton'>
```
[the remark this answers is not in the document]
ARADDCC002: I don't think so - it only opens a Dialog, it doesn't perform any action which could be accidentally duplicated.

frontend/src/reviews/ReviewDecision.tsx
ARADDCC002: This is currently correct for this case as it uses the full user:user name, I'm personally ok leaving this until the entity naming stuff is done on the backed as I don't to change much more of the review code.

frontend/src/reviews/ReviewsList.tsx
ARADDCC002: Updated name

frontend/src/model/beta/reviews/ReviewBanner.tsx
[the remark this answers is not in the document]
ARADDCC012: I forgot we were using a dialog for reviews, ignore me :)

ARADDCC012 (approved):

backend/src/models/v2/Review.ts
```diff
@@ -70,9 +73,16 @@ const ReviewSchema = new Schema<ReviewInterface>(
 
     responses: [
       {
-        user: { type: String, required: true },
-        decision: { type: String, enum: Object.values(Decision), required: true },
-        comment: { type: String, required: false },
+        type: new Schema<ReviewResponse>(
```
a3957273, line 76: Well I've never seen this before, it's exciting.

a3957273 (approved):

PR #996, opened 2024-01-05
Feature/create webhook

JR40159:


backend/src/routes/v2/model/accessRequest/postAccessRequest.ts
```diff
@@ -69,6 +71,12 @@ export const postAccessRequest = [
     const accessRequest = await createAccessRequest(req.user, modelId, body)
 
     await audit.onCreateAccessRequest(req, accessRequest)
```
JR40159, line 73: To simplify route methods and group similar things together, a dedicator service method could be used for all post request stuff? Things like auditing, uploading to document store, sending emails, webhooks.

backend/src/services/v2/webhook.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function sendWebhooks(
+  modelId: string,
+  eventKind: WebhookEventKeys,
+  eventTitle: string,
+  content: { release: ReleaseDoc } | { review: ReviewInterface } | { accessRequest: AccessRequestDoc },
+) {
+  const webhooks = await WebhookModel.find({ modelId, events: eventKind })
+
+  await Promise.all(
+    webhooks.map((webhook) =>
```
JR40159, line 38: Add error handling

backend/src/models/v2/Webhook.ts
```diff
@@ -0,0 +1,52 @@
+import { Document, model, Schema } from 'mongoose'
+import MongooseDelete from 'mongoose-delete'
+
+// Share this enum with auditing?
```
JR40159, line 4: No- this will be the list of supported webhook events

backend/src/services/v2/accessRequest.ts
```diff
@@ -61,6 +63,13 @@ export async function createAccessRequest(
     log.warn('Error when creating Release Review Requests. Approval cannot be given to this Access Request', error)
   }
 
+  sendWebhooks(
```
a3957273, line 66: If we're not awaiting, we should at least add a `.catch` to each of these with an error message.  Otherwise I think it'll get hidden?

a3957273 (changes requested):

backend/src/services/v2/accessRequest.ts
JR40159: Rather than add error handling to every place this function is called, I added the error handling inside the function to avoid duplication. Happy to add more error handling inside the function if you think it's not sufficient.

a3957273 (approved):

PR #997, opened 2024-01-05
Added get_download_file_token method to client

bw27492:


lib/python-beta/src/bailo/core/client.py
```diff
@@ -315,6 +315,31 @@ def get_download_file(
...
+        Downloads a specific file, using an API token.
+
+        :param model_id: Unique model ID
+        :param file_id: Unique file ID
+        :param buffer: BytesIO object for bailo to write to
+        :return: The unique file ID
+
+        ..note:: This method requires the TokenAgent agent.
+        """
+        if isinstance(self.agent, TokenAgent):
```
GB27247, line 334: `isinstance` is a built in function that considers inheritance eg `isinstance(Agent(), TokenAgent)` will be `True`. Instead you want something like `type(self.agent) == TokenAgent`

lib/python-beta/src/bailo/core/client.py
```diff
@@ -315,6 +315,31 @@ def get_download_file(
...
+            req = self.agent.get(
+                f"{self.url}/v2/token/model/{model_id}/file/{file_id}/download", stream=True, timeout=10_000
+            )
+            shutil.copyfileobj(req.raw, buffer)
+            return file_id
+        else:
+            raise Exception("get_download_file_token requires the use of TokenAgent")
```
GB27247, line 341: Does this need to be a separate method? I'd consider adding this to `get_download_file` and switching the endpoint given the Agent instance

GB27247 (changes requested):

a3957273 (approved):

PR #999, opened 2024-01-08
added the ability to make comments on releases

ARADDCC002:


frontend/src/reviews/ReviewCommentDisplay.tsx
```diff
@@ -0,0 +1,40 @@
...
+import UserAvatar from 'src/common/UserAvatar'
+import { EntityKind, ReviewComment } from 'types/types'
+import { formatDateString } from 'utils/dateUtils'
+
+type ReviewCommentsProps = {
+  response: ReviewComment
+}
+
+export default function ReviewComment({ response }: ReviewCommentsProps) {
+  const username = response.user.split(':')[0]
```
a3957273, line 11: This is the ID type, not the username!  It'll call all users 'user'.  You probably want `[1]` instead.

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +40,76 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
+
+  async function submitReviewComment() {
+    if (release && currentUser) {
+      const updatedRelease: ReleaseInterface = release
+      if (!updatedRelease.comments) {
+        updatedRelease.comments = []
+      }
+      const newComment: ReviewComment = {
+        comment: newReviewComment,
+        user: currentUser.dn,
```
a3957273, line 69: This is insecure and would let users update other peoples comments or impersonate other people.  Instead, a 'createReviewComment' endpoint should be made which gets the user from the `req.user` field.  See the 'review' endpoint for how this is done.

In addition, the 'editRelease' functionality is also only available to model owners, so won't be available to other people.

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +40,76 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
 
+  if (isCurrentUserError) {
+    return <MessageAlert message={isCurrentUserError.info.message} severity='error' />
+  }
+
   return (
     <>
       {reviews.length > 0 && <Divider />}
-      {isReviewsLoading && <Loading />}
+      {isReviewsLoading && isCurrentUserLoading && <Loading />}
```
a3957273, line 96: You still display the below stack even when the user is loading.  If a person submits a comment there the code will break.

The button should prevent duplicate clicks with a loading icon or similar.

These text fields should likely use markdown rendering / a markdown editor.

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +40,76 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
+            size='small'
+            sx={{ width: '100%' }}
+            placeholder='Add a comment'
+            value={newReviewComment}
+            onChange={(e) => setNewReviewComment(e.target.value)}
+          ></TextField>
+          <Button sx={{ mt: 1, float: 'right' }} variant='contained' onClick={submitReviewComment}>
+            Submit comment
+          </Button>
+          <Typography variant='caption'>{commentSubmissionError}</Typography>
```
a3957273, line 110: Why are we using a caption here, instead of a proper error type?

a3957273 (changes requested):

frontend/src/reviews/ReviewComments.tsx
ARADDCC002: Oh duh I completely forgot about the edit model permission! Will update now.

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +45,107 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
+    return decisionsAndComments.map((response) => {
+      if (isReviewResponse(response)) {
+        return <ReviewDecisionDisplay key={response.createdAt} response={response as ReviewResponse} />
+      } else {
+        return <ReviewCommentDisplay key={response.createdAt} response={response as ReviewComment} />
+      }
+    })
+  }, [reviews, release, accessRequest])
+
+  async function submitReviewComment() {
```
ARADDCC012, line 69: We should have a loading state while this is being submitted

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +45,107 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
+          mutateAccessRequest()
+          setNewReviewComment('')
+        } else {
+          setCommentSubmissionError(await res.json())
+        }
+      } else {
+        setCommentSubmissionError('There was a problem submitting this comment, please try again later.')
+      }
+    } else {
+      setCommentSubmissionError('There was a fetching user information, please try again later.')
```
ARADDCC012, line 115: Should this be `'There was a problem fetching user information, please try again later.'`?

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +45,107 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
+    }
+    if (currentUser) {
+      const newComment: ReviewComment = {
+        comment: newReviewComment,
+        user: currentUser.dn,
+        createdAt: new Date().toISOString(),
+      }
+      if (release) {
+        const updatedRelease: ReleaseInterface = release
+        if (!updatedRelease.comments) {
```
ARADDCC012, line 83: This logic is strange, shouldn't there always be a `release.comments` array? We can create a migration script to add the array to any existing data and should be able to guarantee that any new releases are created with an empty comments array.

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +45,107 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
+          updatedRelease.comments = []
+        }
+        updatedRelease.comments.push(newComment)
+        const res = await putRelease(updatedRelease)
+        if (res.ok) {
+          mutateRelease()
+          setNewReviewComment('')
+        } else {
+          setCommentSubmissionError(await res.json())
```
ARADDCC012, line 92: I think you'll need to do the following to get the error message:
```
setCommentSubmissionError(await getErrorMessage(res))
```

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +45,107 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
+        }
+      } else if (accessRequest) {
+        const updatedAccessRequest: AccessRequestInterface = accessRequest
+        if (!updatedAccessRequest.comments) {
```
ARADDCC012, line 96: Same here, shouldn't there always be an `accessRequest.comments` array?

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +45,107 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
+        const res = await submitAccessRequestComment(
+          accessRequest.modelId,
+          accessRequest.id,
+          updatedAccessRequest.comments,
+        )
+        if (res.ok) {
+          mutateAccessRequest()
+          setNewReviewComment('')
+        } else {
+          setCommentSubmissionError(await res.json())
```
ARADDCC012, line 109: I think you'll need to do the following to get the error message:
```
setCommentSubmissionError(await getErrorMessage(res))
```

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +45,107 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
       {reviewDetails}
+      <>
+        <Stack spacing={1} justifyContent='center' alignItems='flex-end'>
+          <TextField
+            size='small'
+            sx={{ width: '100%' }}
+            placeholder='Add a comment'
+            value={newReviewComment}
+            onChange={(e) => setNewReviewComment(e.target.value)}
+          ></TextField>
```
ARADDCC012, line 140: Can use `<TextField />` instead of `<TextField></TextField>` as you're not providing any children

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +45,107 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
+          <Button sx={{ mt: 1, float: 'right' }} variant='contained' onClick={submitReviewComment}>
```
ARADDCC012, line 141: We should avoid using `float`, instead use `ml: 'auto'` (you'll need to make sure the wrapping container has the `display: 'flex'` property)

ARADDCC012, line 141: Use a `LoadingButton` here

frontend/types/interfaces.ts
```diff
@@ -140,6 +141,7 @@ export interface AccessRequestInterface {
   schemaId: string
   deleted: boolean
   metadata: AccessRequestMetadata
+  comments?: Array<ReviewComment>
```
ARADDCC012, line 144: I don't think this should be optional

frontend/types/types.ts
```diff
@@ -382,6 +394,7 @@ export type ReleaseInterface = {
   minor?: boolean
   draft?: boolean
   fileIds: Array<string>
+  comments?: Array<ReviewComment>
```
ARADDCC012, line 397: I don't think this should be optional. We can always provide an array, even if it's empty

ARADDCC012 (changes requested):

backend/src/routes/v2/model/accessRequest/createAccessRequestComment.ts
```diff
@@ -0,0 +1,63 @@
...
+    comment: z.string(),
+  }),
+  params: z.object({
+    modelId: z.string(),
+    accessRequestId: z.string(),
+  }),
+})
+
+registerPath({
+  method: 'patch',
```
a3957273, line 23: This is a create method, so I would expect it to be `post`.

backend/src/services/v2/accessRequest.ts
```diff
@@ -118,6 +118,28 @@ export async function updateAccessRequest(
     accessRequest.markModified('metadata')
   }
 
+  if (diff.comments) {
+    accessRequest.comments = diff.comments
+    accessRequest.markModified('comments')
```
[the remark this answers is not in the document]
a3957273: I don't think we want to be able to edit comments like this at all.  This still lets the owner impersonate any messages and comments on their model access requests.

backend/src/services/v2/release.ts
```diff
@@ -144,6 +144,35 @@ export async function updateRelease(user: UserDoc, modelId: string, semver: stri
...
+
+  release.comments.push({
+    comment,
+    user: user.dn,
+    createdAt: new Date().toISOString(),
+  })
+
+  await validateRelease(user, model, release)
+
+  const auth = await authorisation.release(user, model, release, ReleaseAction.View)
```
a3957273, line 159: You don't need to check if we can view the release, we've already checked this when we fetch it!

backend/src/services/v2/release.ts
```diff
@@ -144,6 +144,35 @@ export async function updateRelease(user: UserDoc, modelId: string, semver: stri
   return updatedRelease
 }
 
+export async function newReleaseComment(user: UserDoc, modelId: string, semver: string, comment: string) {
```
a3957273, line 147: A better implementation of this function might be:

```typescript
export async function newReleaseComment(user: UserDoc, modelId: string, semver: string, message: string) {
  
  const release = await getReleaseBySemver(user, modelId, semver)

  const comment = {
    message,
    user: user.dn,
    createdAt: new Date().toISOString(),
  }

  const updatedRelease = await Release.findOneAndUpdate({ _id: release._id }, {
    $push: {
      comments: comment
  })
  return updatedRelease
```

a3957273: (Completely untested)

backend/src/models/v2/Release.ts
```diff
@@ -17,13 +17,21 @@ export interface ReleaseInterface {
...
+
   deleted: boolean
 
   createdBy: string
   createdAt: Date
   updatedAt: Date
 }
 
+export interface ReviewComment {
+  comment: string
```
a3957273, line 30: Can we rename this to something like `message`.  It's generally not a good thing to have both the object named the same as an attribute on the object.

a3957273 (dismissed):

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +46,92 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
-        review.responses.map((response) => <>{response.decision && <ReviewDecision response={response} />}</>),
-      ),
-    [reviews],
-  )
+  const reviewDetails = useMemo(() => {
+    let decisionsAndComments: Array<ReviewResponseKind> = []
+    reviews.forEach((review) => {
+      review.responses.forEach((response) => decisionsAndComments.push(response))
+    })
+    if (release && release.comments) {
```
ARADDCC012, line 54: Is the check for `&& release.comments` needed anymore? Looks like it's been updated so that the array is always there now

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -31,23 +46,92 @@ export default function ReviewComments({ release, accessRequest }: ReviewComment
...
+      decisionsAndComments = [...decisionsAndComments, ...release.comments]
+    }
+    if (accessRequest && accessRequest.comments) {
```
ARADDCC012, line 57: Is the check for `&& accessRequest.comments` needed anymore?

ARADDCC012 (changes requested):

frontend/actions/accessRequest.ts
```diff
@@ -49,3 +49,11 @@ export function patchAccessRequest(modelId: string, accessRequestId: string, for
     body: JSON.stringify({ metadata: { ...form } }),
   })
 }
+
+export function postAccessRequestComments(modelId: string, accessRequestId: string, comment: string) {
```
ARADDCC012, line 53: This only allows one comment to be sent, consider renaming to `postAccessRequestComment`

ARADDCC012 (changes requested):

frontend/src/model/beta/releases/ReleaseDisplay.tsx
```diff
@@ -56,15 +57,14 @@ export default function ReleaseDisplay({
...
-          //console.log(groupedResponses[user].sort(sortByCreatedAtAscending)[groupedResponses[user].length - 1])
           latestResponses.push(groupedResponses[user].sort(sortByCreatedAtAscending)[groupedResponses[user].length - 1])
         })
         reviewResult.responses = latestResponses
         latestReviews.push(reviewResult)
       })
       setReviewsWithLatestResponses(latestReviews)
     }
-  }, [reviews, isReviewsLoading])
+  }, [reviews, isReviewsLoading, release.semver])
```
ARADDCC012, line 67: `release.semver` isn't needed as a dependency

ARADDCC012 (changes requested):

ARADDCC012 (approved):

backend/src/services/v2/release.ts
```diff
@@ -170,6 +170,28 @@ export async function updateRelease(user: UserDoc, modelId: string, semver: stri
...
+  }
+  const updatedRelease = await Release.findOneAndUpdate(
+    { _id: release._id },
+    {
+      $push: {
+        comments: comment,
+      },
+    },
+  )
+  if (!updatedRelease) {
```
JR40159, line 189: You might as well do this just after line 174 to save you waiting until you do another mongo query to find out of the release doesn't exist.

JR40159 (changes requested):

JR40159 (approved):

PR #1004, opened 2024-01-12
Feature/migrate release and review

JR40159:


backend/src/scripts/migrateV2.ts
```diff
@@ -153,6 +174,169 @@ async function migrateModel(modelId: string) {
...
+        role = 'mtr'
+      }
+
+      const responses: ReviewResponse[] = []
+      for (let i = 0; i < approval.approvers.length; i++) {
+        const approver = approval.approvers[i]
+        if (approval.status === ApprovalStates.Accepted) {
+          responses.push({
+            user: toEntity('user', approver.id),
+            decision: Decision.Approve,
```
JR40159, line 310: Add a comment to both of these responses to note that the response has been migrated from V1

a3957273 (approved):

PR #1006, opened 2024-01-12
Improvements to user token management

ARADDCC012:


ARADDCC002:
When you add your token to lots of models the box they appear in gets quite big and on smaller screens can mean lots of scrolling. Can we cap the height and add a scroll bar?

image

ARADDCC002:
Is it worth adding an "All permissions" button here? 

image

ARADDCC012:
> Is it worth adding an "All permissions" button here?
> 
> image

I considered this, but as there's only two options it feels unnecessary to me. I'd be keen to hold off on an ALL option unless we add more options.

ARADDCC002 (approved):

PR #1010, opened 2024-01-15
Add the ability to update and delete a webhook and get all model webhooks

JR40159:


backend/src/routes.ts
```diff
@@ -245,6 +248,9 @@ if (config.experimental.v2) {
   server.delete('/api/v2/model/:modelId/file/:fileId', ...deleteFile)
 
   server.post('/api/v2/model/:modelId/webhooks', ...postWebhook)
+  server.get('/api/v2/model/:modelId/webhooks', ...getWebhooks)
+  server.put('/api/v2/model/:modelId/webhooks/:webhookId', ...putWebhook)
```
a3957273, line 252: These two paths should be singular, `/webhook/:webhookId`.

backend/src/routes/v2/model/webhook/putWebhook.ts
```diff
@@ -0,0 +1,63 @@
...
+export const putWebhookSchema = z.object({
+  params: z.object({
+    modelId: z.string(),
+    webhookId: z.string(),
+  }),
+  body: z.object({
+    name: z.string(),
+
+    uri: z.string(),
+    token: z.string().optional(),
```
a3957273, line 19: Traditionally this is generated by the server.  If it's easy to change, I might suggest we change it to that?  Otherwise, happy to keep it provided by the client.

backend/src/services/v2/webhook.ts
```diff
@@ -36,6 +36,43 @@ export async function createWebhook(user: UserDoc, webhookParams: CreateWebhookP
...
+  if (!webhook) {
+    throw NotFound(`The requested webhook was not found.`, { webhookId })
+  }
+
+  return webhook
+}
+
+export async function getWebhooksByModel(user: UserDoc, modelId: string) {
+  const model = await getModelById(user, modelId)
+  const auth = await authorisation.model(user, model, ModelAction.Update)
```
a3957273, line 57: Is `Update` the correct permission here?

a3957273 (changes requested):

a3957273 (approved):

PR #1013, opened 2024-01-17
added a maxwidth to the jsonscheama form page on larger screensizes

ARADDCC002:


ARADDCC012 (approved):

frontend/src/Form/beta/JsonSchemaForm.tsx
```diff
@@ -120,35 +107,38 @@ export default function JsonSchemaForm({
...
-              }
-            : { ArrayFieldTemplate }
-        }
-      >
-        {/* eslint-disable-next-line react/jsx-no-useless-fragment */}
-        <></>
-      </Form>
-    </Stack>
+      </Grid>
+      <Divider orientation='vertical' flexItem sx={{ mr: '-1px' }} />
```
ARADDCC012, line 111: This should be inside of one of the Grid items, all children of a Grid container should be Grid items

ARADDCC012, line 111: An easier way of achieving this is to apply a `borderRight` to the left Grid item.

All children of a `<Grid container />` should be <Grid item>`s. That's why you've had to put a negative margin to make it work like this, because it's not designed to work that way.

ARADDCC012, line 111: All children of a Grid container should be Grid items, so this goes against how the component is designed to be used. That's why you've had to apply a negative margin to get it to work.

Instead you should apply a right border to the left Grid item and use the theme to set it to the colour of a divider.

ARADDCC012 (changes requested):
See my other comment

ARADDCC012 (approved):

PR #1020, opened 2024-01-19
Update errors for bad connectors

JR40159:


backend/src/utils/v2/error.ts
```diff
@@ -43,3 +43,7 @@ export function NotFound(message: string, context?: BailoError['context'], logge
 export function InternalError(message: string, context?: BailoError['context'], logger?: Logger) {
   return GenericError(500, message, context, logger)
 }
+
+export function ConfigurationError(message: string, context?: BailoError['context'], logger?: Logger) {
+  return GenericError(500, `BAILO configuration error: ${message}`, context, logger)
```
a3957273, line 48: Let's change this to a 503 error, signalling 'Service Unavailability'.  500 errors are not only super generic, but also usually suggest the error is request specific.  A 503 error will make it clear that it's likely the entire service isn't functional, which is likely true if the backend is facing a configuration error.

a3957273 (approved):

PR #1021, opened 2024-01-19
conditional rendering has been done to make sure that the latest versi…

TT38665:
…on does not show when viewing a release


a3957273 (changes requested):
We also probably don't want to display this when _editing_ a model, because the release semver can't be created there either!

TT38665:
> We also probably don't want to display this when _editing_ a model, because the release semver can't be created there either!

It's covered in the conditional rendering. The latest version does not show when you click on a release or when editing a model.

ARADDCC012:
> We also probably don't want to display this when _editing_ a model, because the release semver can't be created there either!

Can confirm that when the form is in `editable` mode, the form is being used for View and Edit. If it's not in `editable` mode the form is being used for Create only. So by conditionally showing the latest version when not in `editable` mode, it'll only show on the create release page.

a3957273 (approved):

PR #1030, opened 2024-01-24
Feature/allow models to be exported as pdfs

ARADDCC012:
Example PDF using a model card with the minimal schema:
image
Screenshot 2024-01-24 at 11 55 25
Screenshot 2024-01-24 at 11 55 47


backend/src/routes/v2/model/modelcard/getModelCardExport.ts
```diff
@@ -0,0 +1,57 @@
...
+            format: 'binary',
+          },
+        },
+      },
+    },
+  },
+})
+
+interface GetModelCardExportResponse {
+  file: FileInterface
```
ARADDCC012, line 41: I don't think this is right, but I've followed the logic already set out in `backend/src/routes/v2/model/file/getDownloadFile.ts` as they're both returning a file as a stream.

ARADDCC002:
Can we make the model title and description text bigger on the first page?

ARADDCC002 (approved):

a3957273 (changes requested):

PR #1033, opened 2024-01-24
disabled the releases and access requests tab on model page if no sch…

ARADDCC002:
…ema selected

### DIFF
--- a/frontend/src/common/PageWithTabs.tsx
+++ b/frontend/src/common/PageWithTabs.tsx
@@ -94,7 +94,6 @@ export default function PageWithTabs({
               disabled={tab.disabled}
               value={tab.path}
               data-test={`${tab.path}Tab`}
-              sx={{ pointerEvents: 'auto' }}
               label={
                 <Tooltip title={tab.disabledText}>
                   <span>{tab.title}</span>
--- a/frontend/src/common/PageWithTabs.tsx
+++ b/frontend/src/common/PageWithTabs.tsx
@@ -1,4 +1,4 @@
-import { Box, Button, Divider, Stack, Tab, Tabs, Tooltip, Typography } from '@mui/material'
+import { Box, Button, Divider, Stack, Tab, Tabs, Typography } from '@mui/material'
 import { grey } from '@mui/material/colors/'
 import { useTheme } from '@mui/material/styles'
 import { useRouter } from 'next/router'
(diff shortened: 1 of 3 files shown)


frontend/src/common/PageWithTabs.tsx
```diff
@@ -90,10 +91,15 @@ export default function PageWithTabs({
           return (
             <Tab
               key={tab.title}
-              label={tab.title}
               disabled={tab.disabled}
               value={tab.path}
               data-test={`${tab.path}Tab`}
+              style={{ pointerEvents: 'auto' }}
```
ARADDCC012, line 97: Can we use `sx` here?

ARADDCC012 (changes requested):

frontend/src/common/PageWithTabs.tsx
```diff
@@ -90,10 +91,15 @@ export default function PageWithTabs({
           return (
             <Tab
               key={tab.title}
-              label={tab.title}
               disabled={tab.disabled}
               value={tab.path}
               data-test={`${tab.path}Tab`}
+              sx={{ pointerEvents: 'auto' }}
```
ARADDCC012, line 97: Is this needed? I believe the default behaviour of `Tab` should handle this for us.

ARADDCC012 (changes requested):

frontend/src/model/beta/ModelImages.tsx
```diff
@@ -51,12 +51,7 @@ export default function ModelImages({ model }: AccessRequestsProps) {
       <Container sx={{ my: 2 }}>
         <Stack spacing={4}>
           <Box sx={{ textAlign: 'right' }}>
-            <Button
-              variant='outlined'
-              onClick={() => setOpenUploadImageDialog(true)}
-              disabled={!model.card}
```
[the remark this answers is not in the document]
ARADDCC002: Nah we discussed this and we reckon that Images can be added before a schema is set

frontend/src/common/PageWithTabs.tsx
```diff
@@ -12,6 +12,7 @@ export interface PageTab {
   view: ReactElement
   disabled?: boolean
   datatest?: string
+  disabledText?: string
```
ARADDCC012, line 15: I don't think this is being used is it?

ARADDCC012 (approved):

PR #1036, opened 2024-01-25
Bugfix/bai 1117 fix archived section review page

ARADDCC002:


TT38665 (approved):

frontend/src/reviews/ReviewsList.tsx
```diff
@@ -6,38 +6,33 @@ import EmptyBlob from 'src/common/EmptyBlob'
 import Loading from 'src/common/Loading'
 import MessageAlert from 'src/MessageAlert'
 import ReviewItem from 'src/reviews/ReviewItem'
-import { Decision, ReviewRequestInterface } from 'types/interfaces'
+import { ReviewRequestInterface } from 'types/interfaces'
 
 type ReviewsListProps = {
-  kind?: 'release' | 'access' | 'all'
+  kind?: 'release' | 'access' | 'all' | 'archived'
```
ARADDCC012, line 12: There's no need for 'all' anymore, should we remove it and make this a required prop?

frontend/src/reviews/ReviewsList.tsx
```diff
@@ -6,38 +6,33 @@ import EmptyBlob from 'src/common/EmptyBlob'
...
 }
 
 export default function ReviewsList({ kind = 'all' }: ReviewsListProps) {
```
ARADDCC012, line 15: We're always providing this prop now it can be made required and this default can be removed.

frontend/src/reviews/ReviewsList.tsx
```diff
@@ -6,38 +6,33 @@ import EmptyBlob from 'src/common/EmptyBlob'
...
-  const doesNotContainUserApproval = useCallback(
+  const containsUserApproval = useCallback(
     (review: ReviewRequestInterface) => {
-      return (
-        currentUser &&
-        !review.responses.find(
-          (response) => response.user === `user:${currentUser.dn}` && response.decision === Decision.Approve,
-        )
-      )
+      return currentUser && review.responses.find((response) => response.user === `user:${currentUser.dn}`)
```
ARADDCC012, line 22: We're no longer looking for approvals, but instead just looking for a response from the user. Is this a mistake? We could have a "requested changes" response, which this logic now treats as an approval.

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -81,12 +81,7 @@ export default function SideNavigation({
 
   const doesNotContainUserApproval = useCallback(
     (review: ReviewRequestInterface) => {
-      return (
-        currentUser &&
-        !review.responses.find(
-          (response) => response.user === `user:${currentUser.dn}` && response.decision === Decision.Approve,
-        )
-      )
+      return currentUser && !review.responses.find((response) => response.user === `user:${currentUser.dn}`)
```
ARADDCC012, line 84: We're no longer looking for approvals, is this a mistake?

ARADDCC012 (changes requested):

frontend/src/wrapper/SideNavigation.tsx
ARADDCC002: See above

ARADDCC012 (approved):

PR #1040, opened 2024-01-26
Feature/bai 881 implement oauth authentication

JR40159:


backend/src/clients/cognito.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Make this into a client function queryEmails that returns an array of emails
+ * Handle errors like missing userpool Id
+ */
+export async function listUsers(query: string) {
+  const dnName = config.oauth.cognito.userIdAttribute
+  const client = new CognitoIdentityProviderClient(config.oauth.cognito.identityProviderClient)
+
+  const command = new ListUsersCommand({
+    UserPoolId: config.oauth.cognito.userPoolId,
+    Filter: `"${dnName}"^="${query}"`,
```
a3957273, line 15: Add a parameter to change this to absolute match, otherwise emails that share prefixes will break.

a3957273 (changes requested):

a3957273 (approved):

PR #1043, opened 2024-01-31
removed link component, to stop the double tabbing through the sidebar

TT38665:


frontend/src/wrapper/NavMenuItem.tsx
```diff
@@ -1,7 +1,6 @@
 import { Badge, ListItem, ListItemButton, ListItemIcon, ListItemText, Tooltip } from '@mui/material'
 import { useTheme } from '@mui/material/styles'
 import { ReactElement } from 'react'
-import Link from 'src/Link'
```
a3957273, line 4: You probably don't want to disable tabbing down the sidebar.  It's important that users without a mouse are still able to navigate.

a3957273 (dismissed):

ARADDCC012 (approved):

PR #1045, opened 2024-01-31
Bugfix/catch email error

JR40159:


backend/src/services/v2/review.ts
```diff
@@ -166,39 +166,47 @@ export async function sendReviewResponseNotification(review: ReviewDoc, user: Us
       }
 
       const access = await getAccessRequestById(user, review.accessRequestId)
-      notifyReviewResponseForAccess(review, access)
+      try {
+        notifyReviewResponseForAccess(review, access)
```
a3957273, line 170: This is an async function that's not being 'awaited', therefore the `try {} catch{}` won't do anything here.  Either await this, or turn the try / catch into a `.catch()` statement.

backend/src/services/v2/review.ts
```diff
@@ -166,39 +166,47 @@ export async function sendReviewResponseNotification(review: ReviewDoc, user: Us
...
+      } catch (error) {
+        log.warn('Error when notifying collaborators about review response.', { error })
+      }
       break
     }
     case ReviewKind.Release: {
       if (!review.semver) {
-        log.error('Unable to send notification for review response.Cannot find semver')
+        log.error('Unable to send notification for review response. Cannot find semver.')
```
a3957273, line 178: All of these logs should include a model.id and an accessRequest.id.

backend/src/services/v2/review.ts
```diff
@@ -166,39 +166,47 @@ export async function sendReviewResponseNotification(review: ReviewDoc, user: Us
...
 
-function getRoleEntites(roles: string[], collaborators: CollaboratorEntry[]) {
+function getRoleEntities(roles: string[], collaborators: CollaboratorEntry[]) {
   return roles.map((role) => {
-    const entites = collaborators
+    const entities = collaborators
       .filter((collaborator) => collaborator.roles.includes(role))
       .map((collaborator) => collaborator.entity)
-    if (entites.length === 0) {
+    if (entities.length === 0) {
```
a3957273, line 200: Thank you for renaming this everywhere!

backend/src/services/v2/review.ts
```diff
@@ -166,39 +166,47 @@ export async function sendReviewResponseNotification(review: ReviewDoc, user: Us
...
       throw BadReq('Unable to create Review Request. Could not find any entities for the role.', { role })
     }
-    return { role, entites }
+    return { role, entities: entities }
```
a3957273, line 203: `entities` should be short hand here.

a3957273 (changes requested):

a3957273 (approved):

PR #1068, opened 2024-02-05
added start of testing frontend v2

ARADDCC002:


frontend/test/model/accessRequests/AccessRequestDisplay.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+vi.mock('../../../actions/review', () => ({
+  useGetReviewRequestsForModel: vi.fn(),
+}))
+vi.mock('src/model/beta/reviews/ReviewBanner.tsx', () => ({ default: (_props: ReviewBannerProps) => <></> }))
+vi.mock('src/common/UserDisplay.tsx', () => ({ default: (_props: UserDisplayProps) => <></> }))
+vi.mock('src/model/beta/reviews/ReviewDisplay.tsx', () => ({ default: (_props: ReviewDisplayProps) => <></> }))
+
+describe('AccessRequestDisplay', () => {
+  it('renders a AccessRequestDisplay component', async () => {
```
ARADDCC012, line 21: This isn't testing anything useful. We should be testing that something happens/doesn't happen in a given scenario. See `HelpPopover.spec.tsx` for a good example of this. It breaks down the core functionality of the component into simple unit tests.

frontend/test/model/accessRequests/EditableAccessRequestForm.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return {
+    getStepsFromSchema: vi.fn(),
+    widgets: {},
+    setStepState: vi.fn(),
+  }
+})
+vi.mock('utils/beta/formUtils.ts', () => mockFormUtils)
+
+describe('EditableAccessRequestForm', () => {
+  it('renders a EditableAccessRequestForm component', async () => {
```
ARADDCC012, line 28: This isn't testing anything useful. See my other comment for more info.

frontend/test/model/common/SchemaButton.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { MarkdownDisplayProps } from 'src/common/MarkdownDisplay'
+import SchemaButton from 'src/model/beta/common/SchemaButton'
+import { lightTheme } from 'src/theme'
+import { testAccessRequestSchema } from 'utils/test/testModels'
+import { describe, expect, it, vi } from 'vitest'
+
+vi.mock('src/common/MarkdownDisplay.tsx', () => ({ default: (_props: MarkdownDisplayProps) => <></> }))
+
+describe('SchemaButton', () => {
+  it('renders a SchemaButton component', async () => {
```
ARADDCC012, line 12: This isn't testing anything useful. See my other comment for more info.

frontend/test/model/common/SchemaButton.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <SchemaButton schema={testAccessRequestSchema} onClick={() => undefined} />
+      </ThemeProvider>,
+    )
+
+    await waitFor(async () => {
+      expect(await screen.findByText('Select schema')).not.toBeUndefined()
+    })
+  })
+
+  it('renders a loading SchemaButton component', async () => {
```
ARADDCC012, line 24: This is a decent unit test, but it would be good to follow a naming convention such as "Something is true/false when something is the case". So here it would be `it('enters a loading state when the loading prop is true').

frontend/test/model/common/SchemaButton.spec.tsx
ARADDCC012, line 1: Another good unit test for this component would be:
- onClick is called when the Select Schema button is clicked

frontend/test/model/common/ValidationErrorIcon.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { render, screen, waitFor } from '@testing-library/react'
+import { MarkdownDisplayProps } from 'src/common/MarkdownDisplay'
+import ValidationErrorIcon from 'src/model/beta/common/ValidationErrorIcon'
+import { lightTheme } from 'src/theme'
+import { testAccessRequestSchemaStepNoRender } from 'utils/test/testModels'
+import { describe, expect, it, vi } from 'vitest'
+
+vi.mock('src/common/MarkdownDisplay.tsx', () => ({ default: (_props: MarkdownDisplayProps) => <></> }))
+
+describe('Va;idationError', () => {
```
ARADDCC012, line 11: Typo

frontend/test/model/accessRequests/AccessRequestDisplay.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+      mutateReviews: vi.fn(),
+    })
+    render(
+      <ThemeProvider theme={lightTheme}>
+        <AccessRequestDisplay accessRequest={testAccessRequest} />
+      </ThemeProvider>,
+    )
+
+    await waitFor(async () => {
+      expect(await screen.findByText(testAccessRequest.metadata.overview.name)).not.toBeUndefined()
```
ARADDCC012, line 35: Instead of using a double negative `.not.toBeUndefined()`, can we pelase use `.toBeDefined()`? This double negative is used throughout this PR.

frontend/test/model/common/ValidationErrorIcon.spec.tsx
ARADDCC012, line 11: The describe should probably be the exact name of the component: `ValidationErrorIcon` so that it's easier to debug if needed.

frontend/test/model/common/ValidationErrorIcon.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+  it('renders a ValidationError component', async () => {
```
ARADDCC012, line 12: This isn't testing anything useful. See my other comment for more info.

frontend/test/model/common/ValidationErrorIcon.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <ValidationErrorIcon step={testAccessRequestSchemaStepNoRender} />
+      </ThemeProvider>,
+    )
+
+    await waitFor(async () => {
+      expect(await screen.findByLabelText('This step is unfinished')).not.toBeUndefined()
+    })
+  })
+
+  it('does not render a ValidationError component when step is marked as complete', async () => {
```
ARADDCC012, line 24: This isn't the name of the component. I also find it's useful to get into the habit of not using the component name to describe the individual test for the following reasons:
- The describe already contains the component name.
- It forces you to be more descriptive with your test naming.

In this case it already is descriptive, but you're repeating the name unnecessarily. Consider renaming to something like `'does not render anything when step is marked as complete'`

ARADDCC012 (changes requested):

frontend/test/model/accessRequests/AccessRequestDisplay.spec.tsx
ARADDCC012, line 1: This component should have more tests.

frontend/test/model/accessRequests/EditableAccessRequestForm.spec.tsx
ARADDCC012, line 1: This component should have more tests.

frontend/test/model/accessRequests/AccessRequestDisplay.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+vi.mock('../../../actions/review', () => ({
+  useGetReviewRequestsForModel: vi.fn(),
+}))
+vi.mock('src/model/beta/reviews/ReviewBanner.tsx', () => ({ default: (_props: ReviewBannerProps) => <></> }))
+vi.mock('src/common/UserDisplay.tsx', () => ({ default: (_props: UserDisplayProps) => <></> }))
+vi.mock('src/model/beta/reviews/ReviewDisplay.tsx', () => ({ default: (_props: ReviewDisplayProps) => <></> }))
+
+describe('AccessRequestDisplay', () => {
+  it('displays the metadata for an access request metadata object when display component is rendered', async () => {
```
ARADDCC012, line 21: Repeated yourself here. Maybe `'displays access request metadata when not loading and no errors'`?

frontend/test/model/accessRequests/AccessRequestDisplay.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      mutateReviews: vi.fn(),
+    })
+    render(
+      <ThemeProvider theme={lightTheme}>
+        <AccessRequestDisplay accessRequest={testAccessRequest} />
+      </ThemeProvider>,
+    )
+
+    await waitFor(async () => {
+      expect(await screen.findByText(testAccessRequest.metadata.overview.name)).toBeDefined()
```
ARADDCC012, line 35: If the goal of this test is to check that the metadata is being displayed, then we should check all of the properties are being displayed. This test is currently only testing that the name is displayed.

frontend/test/model/accessRequests/AccessRequestDisplay.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+    })
+  })
+
+  it('displays a comments icon when the provided access request has comment responses', async () => {
```
ARADDCC012, line 39: Can we try to keep test names concise and to the point? We want to know what the test is doing at a glance, so we can cut out the extra words: `displays comment icon when access request has comments`

frontend/test/model/accessRequests/AccessRequestDisplay.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      mutateReviews: vi.fn(),
+    })
+    render(
+      <ThemeProvider theme={lightTheme}>
+        <AccessRequestDisplay accessRequest={testAccessRequestWithComments} />
+      </ThemeProvider>,
+    )
+
+    await waitFor(async () => {
+      expect(await screen.findByLabelText('Comments')).toBeDefined()
```
ARADDCC012, line 53: Can we use `ByTestId` queries? This will use the same `data-test` attributes that we use in cypress. This should be our primary way of finding elements in tests as it's much more specific. You can read more about the different queries here: https://testing-library.com/docs/react-testing-library/cheatsheet/ and here: https://testing-library.com/docs/queries/about/#priority

As an example, this test would still pass if we happened to put the text "Comments" anywhere else in this component, even if we removed the icon. So the test isn't providing us the security it was written for. If we use `findByTestId` then it will only pass so long as that specific icon is present.

In this case we could use `findByTestId`, which returns the found element if found and throws and error when not found.

I won't flag all of these up, but if you agree, we should apply this to all of the unit tests in this PR.

frontend/test/model/accessRequests/AccessRequestDisplay.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      expect(await screen.findByText('1 comment')).toBeDefined()
+    })
+  })
+
+  it('does not display a comments icon when the provided access request has no comment responses', async () => {
```
ARADDCC012, line 58: Reword to: `does not display comment icon when access request does not have comments`?

frontend/test/model/accessRequests/AccessRequestDisplay.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      isReviewsError: undefined,
+      mutateReviews: vi.fn(),
+    })
+    render(
+      <ThemeProvider theme={lightTheme}>
+        <AccessRequestDisplay accessRequest={testAccessRequest} />
+      </ThemeProvider>,
+    )
+    await waitFor(async () => {
+      const commentsIcon = screen.queryByLabelText('Comments')
```
ARADDCC012, line 71: Instead we could use `queryByTestId` as that returns `null` when the element is not found, instead of throwing an error like `findByTestId` would.

frontend/test/model/common/SchemaButton.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+import SchemaButton from 'src/model/beta/common/SchemaButton'
+import { lightTheme } from 'src/theme'
+import { testAccessRequestSchema } from 'utils/test/testModels'
+import { describe, expect, it, vi } from 'vitest'
+
+vi.mock('src/common/MarkdownDisplay.tsx', () => ({ default: (_props: MarkdownDisplayProps) => <></> }))
+const consoleMock = vi.spyOn(console, 'log').mockImplementation(() => undefined)
+
+describe('SchemaButton', () => {
+  it('enters loading state when the loading prop is true', async () => {
```
ARADDCC012, line 13: rename? `displays a loading spinner when loading prop is true`?

frontend/test/model/common/SchemaButton.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <SchemaButton schema={testAccessRequestSchema} onClick={() => undefined} loading />
+      </ThemeProvider>,
+    )
+
+    await waitFor(async () => {
+      expect(await screen.findByRole('progressbar')).toBeDefined()
+    })
+  })
+
+  it('does not enter a loading state when the loading prop is false', async () => {
```
ARADDCC012, line 25: rename? `does not display a loading spinner when loading prop is false`

frontend/test/model/common/SchemaButton.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+      expect(loadingSpinner).toBeNull()
+    })
+  })
+
+  it('fires the onClick event when the schema button is clicked', async () => {
+    const testMessage = 'The schema button has been clicked!'
+
+    function onClick() {
+      // eslint-disable-next-line no-console
+      console.log(testMessage)
```
ARADDCC012, line 43: This can be done in a much more simple way.
```
// Create a spy
const handleClickSpy = vi.fn()

// Pass your spy as a prop
    render(
      <ThemeProvider theme={lightTheme}>
        <SchemaButton schema={testAccessRequestSchema} onClick={handleClickSpy} />
      </ThemeProvider>,
    )

// Assert that the spy has been called on click
    await waitFor(async () => {
      const schemaButton = screen.getByTestId('selectSchemaButton-access-request-schema')
      fireEvent.click(schemaButton)
      expect(handleClickSpy).toHaveBeenCalledOnce()
    })
```

frontend/test/model/common/ValidationErrorIcon.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { MarkdownDisplayProps } from 'src/common/MarkdownDisplay'
+import ValidationErrorIcon from 'src/model/beta/common/ValidationErrorIcon'
+import { lightTheme } from 'src/theme'
+import { testAccessRequestSchemaStepNoRender } from 'utils/test/testModels'
+import { describe, expect, it, vi } from 'vitest'
+
+vi.mock('src/common/MarkdownDisplay.tsx', () => ({ default: (_props: MarkdownDisplayProps) => <></> }))
+
+describe('ValidationErrorIcon', () => {
+  it('shows a validation warning message and icon when the form step is marked as incomplete', async () => {
```
ARADDCC012, line 12: You've said it should show a message and icon, but are only asserting that the message is shown. You should add another assertion to check for the icon.

ARADDCC012, line 12: Quite a minor point, but both of us have swapped between "display" and "shows". Can we pick one and be consistent?

frontend/test/model/common/ValidationErrorIcon.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <ValidationErrorIcon step={testAccessRequestSchemaStepNoRender} />
+      </ThemeProvider>,
+    )
+
+    await waitFor(async () => {
+      expect(await screen.findByLabelText('This step is unfinished')).toBeDefined()
+    })
+  })
+
+  it('does not display a validation warning message and icon when the form step is marked as complete', async () => {
```
ARADDCC012, line 24: Small rewording suggestion: `message and icon` -> `message or icon`

frontend/test/model/common/ValidationErrorIcon.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const updatedStep = testAccessRequestSchemaStepNoRender
```
ARADDCC012, line 25: This can be:
```
const step = {
  ... testAccessRequestSchemaStepNoRender,
  isComplete: () => true,
}
```

ARADDCC012 (changes requested):

frontend/test/model/accessRequests/AccessRequestDisplay.spec.tsx
ARADDCC002: I've changed this where I can, there's a few places where I've left by "getByRole" like for the  loading spinner because the element will technically always be there, but when the LoadingButton is in a loading state it only has its role changed.

I'm also going to try my best to update the form fields, we _might_ still need to do text checking in the document. In the past I've managed to pass data-test IDs into the custom form field components as props but they don't always get attached to the correct element because of how MUI works.

frontend/test/model/common/ValidationErrorIcon.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+      ...testAccessRequestSchemaStepNoRender,
+      isComplete: () => true,
+    }
+    const { container } = render(
+      <ThemeProvider theme={lightTheme}>
+        <ValidationErrorIcon step={step} />
+      </ThemeProvider>,
+    )
+
+    screen.debug()
```
ARADDCC012, line 35: This can be removed

ARADDCC012 (approved):

PR #1074, opened 2024-02-08
Fix navbar logo imports

a3957273:


ARADDCC002 (approved):
Approved but this will only affect v1 if that's intentional?

PR #1085, opened 2024-02-13
Bai/1148 unit test for review role display

TT38665:


frontend/src/reviews/ReviewRoleDisplay.spec.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }
+})
+vi.mock('utils/beta/roles.ts', () => mockRoleUtils)
+
+vi.mock('actions/model', () => ({
+  useGetModelRoles: vi.fn(),
+}))
+
+describe('ReviewRoleDisplay', () => {
+  const testMessageAccess = 'This access needs to be reviewed by the Manager.'
```
ARADDCC012, line 25: This highlights that we should update the component so that the message says "access request" instead of just "access". Not for you to do here though :)

frontend/src/reviews/ReviewRoleDisplay.tsx
```diff
@@ -25,7 +25,7 @@ export default function ReviewRoleDisplay({ review }: ReviewRoleDisplayProps) {
   return (
     <>
       {isModelRolesLoading && <Loading />}
-      <Stack direction='row' alignItems='center' justifyContent='center' spacing={1}>
+      <Stack direction='row' alignItems='center' justifyContent='center' spacing={1} data-test='reviewRoleMessage'>
```
ARADDCC012, line 28: Looks like this `data-test` isn't being used in your tests so it can be removed.

ARADDCC012 (changes requested):

ARADDCC012 (approved):

ARADDCC002:
I think before we merge this in we wait until my v2 testing PR is merged in, and then we can update this one so that the test file is in the new testing directory. Happy to sort that for you though.

ARADDCC012 (changes requested):
Hold off on merging until https://github.com/gchq/Bailo/pull/1089 has been merged into main

PR #1087, opened 2024-02-13
Add schema config

JR40159:


backend/config/default.cjs
```diff
@@ -139,6 +139,26 @@ module.exports = {
         schema: require('../src/scripts/example_schemas/minimal_deployment_schema.json'),
       },
     ],
+    v2: {
+      model: [
```
a3957273, line 143: Nitpick: Could we call these `modelCards` and `accessRequests`?

a3957273 (changes requested):

a3957273 (approved):

PR #1097, opened 2024-02-15
Add inferencing routes 

GB27247:


backend/src/services/v2/inference.ts
```diff
@@ -0,0 +1,113 @@
...
+        modelId: modelId,
+        image: inferenceParams.image,
+        tag: inferenceParams.tag,
+      })
+    }
+
+    throw error
+  }
+
+  const auth = await authorisation.model(user, model, ModelAction.Update)
```
JR40159, line 77: It's a good idea to move this check as high up as possible to avoid any unnecessary computation should the user not be authorised. At the very least, it should effeminately be above the saving of the mongo document.

JR40159 (changes requested):

backend/src/services/v2/inference.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  const inference = await getInferenceByImage(user, modelId, inferenceParams.image, inferenceParams.tag)
+  const auth = await authorisation.model(user, model, ModelAction.Update)
+
+  if (!auth.success) {
+    throw Forbidden(auth.info, { userDn: user.dn, modelId: modelId })
+  }
+
+  const updatedInference = await Inference.findOneAndUpdate(
+    { modelId: inference.modelId, image: inference.image, tag: inference.tag },
```
JR40159, line 106: What would happen here if you tried to update the inference to be the same as an existing one?

JR40159: Also, should you allow the user to update the model ID on the inference setting? You've only checked their permissions on the existing model and the route is specific for the existing model ID. I would assume that they would have to delete the existing one and make another request to create an inference service on another model.

JR40159 (changes requested):

backend/test/services/inference.spec.ts
JR40159, line 1: Some missing test cases here, use the vitest coverage report to help add the missing tests

JR40159 (changes requested):

JR40159 (approved):

PR #1101, opened 2024-02-17
Block bad semver values from releases

a3957273:


backend/test/services/release.spec.ts
JR40159, line 1: Bad semver test?

JR40159 (changes requested):

JR40159 (approved):

PR #1103, opened 2024-02-19
added a reusable text input dialog component. added a button to rende…

ARADDCC002:
image
image


a3957273 (changes requested):
The JSON we're asking users to fill in is quite large, could we increase the default width / height of the input box?  Also, some extra padding around the input box could improve the design.

Finally, adding a title seems like a good addition, the modal looks rather anemic at the moment!

frontend/src/model/overview/FormEditPage.tsx
```diff
@@ -74,6 +79,26 @@ export default function FormEditPage({ model }: FormEditPageProps) {
     setUnsavedChanges(isEdit)
   }, [isEdit, setUnsavedChanges])
 
+  function handleJsonFormOnSubmit(formData: string) {
+    setJsonUploadDialogOpen(false)
+    try {
+      const jsonInput = JSON.parse(formData)
+      if (schema) {
+        const steps = getStepsFromSchema(schema, {}, ['properties.contacts'], JSON.parse(jsonInput))
```
a3957273, line 87: You're using 'JSON.parse' twice here.

a3957273, line 87: `properties.contacts` does not exist in V2 I don't think?  This can just be an empty array.

a3957273 (changes requested):

a3957273 (approved):

PR #1108, opened 2024-02-20
improved styling for block quotes in markdown

ARADDCC002:


a3957273:
This doesn't seem to be for block quotes, isn't this for code?

ARADDCC002:
> This doesn't seem to be for block quotes, isn't this for code?

Whoops! This is updated:

image

a3957273 (approved):

PR #1109, opened 2024-02-20
various small fixes

ARADDCC002:


JR40159 (changes requested):
Marketplace errors don't look quite right. Padding still might need work.

JR40159 (approved):

PR #1117, opened 2024-02-22
Removed old Python client and renamed new client

bw27492:


lib/python/test.ipynb
GB27247, line 1: I'm not sure this is supposed to be in this PR?

GB27247 (changes requested):
Looks good. Apart from the test notebook:
1) Can you reflect the folder change in `.prettierignore` 
2) Can you update the `pyproject.toml` file to point to the `python` folder in `.vscode/settings.json`

GB27247 (approved):

PR #1119, opened 2024-02-23
Initial removal of V1 code

a3957273:


ARADDCC002 (approved):
Had a manual test through and everything seems to be in order so approved. But feel free to hold for additional approvals if you want as it's such a big change!

backend/config/default.cjs
JR40159, line 1: I don't think the enabled config value for oauth is required now in V2.

PR #1125, opened 2024-02-28
added save buttons to top and bottom of model card and editable acces…

ARADDCC002:
…s request form


a3957273 (approved):
In the future, might be nice to separate out the component into another file.

PR #1149, opened 2024-03-04
Dev/add theme picker

ARADDCC002:
image

image


frontend/src/settings/display/ExampleDisplay.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { patchCurrentUserSettings, useGetCurrentUserSettings } from 'actions/user'
+import { ThemeKey } from 'src/theme'
+
+interface ExampleDisplayProps {
+  theme: ThemeKey
+}
+export default function ExampleDisplay({ theme }: ExampleDisplayProps) {
+  const { mutateUserSettings } = useGetCurrentUserSettings()
+
+  async function setTheme() {
```
ARADDCC012, line 12: I'd prefer to keep the `setSomething` naming convention for updating react state vars. Consider renaming this function to something like `handleUpdateTheme`

frontend/src/settings/display/ExampleDisplay.tsx
```diff
@@ -0,0 +1,35 @@
...
+    const updatedSettings = await patchCurrentUserSettings({ theme: theme.key })
+    if (updatedSettings.status === 200) {
```
ARADDCC012, line 14: No error handling

frontend/src/settings/display/ExampleDisplay.tsx
```diff
@@ -0,0 +1,35 @@
...
+      mutateUserSettings()
+    }
+  }
+
+  return (
+    <ThemeProvider theme={theme}>
```
ARADDCC012, line 20: This should be able to be removed as it's already included in `_app.tsx`

frontend/src/theme.ts
```diff
@@ -38,6 +38,11 @@ const defaultComponentOverrides: ThemeOptions['components'] = {
   },
 }
 
+export interface ThemeKey {
```
ARADDCC012, line 41: This is more of a mapping than a key, consider renaming to ThemeMapping

frontend/src/settings/display/ExampleDisplay.tsx
```diff
@@ -0,0 +1,35 @@
+import { Box, Button } from '@mui/material'
+import { ThemeProvider } from '@mui/material/styles'
+import { patchCurrentUserSettings, useGetCurrentUserSettings } from 'actions/user'
+import { ThemeKey } from 'src/theme'
+
+interface ExampleDisplayProps {
+  theme: ThemeKey
```
ARADDCC012, line 7: This prop name is confusing as it's not a `Theme`. Consider renaming to `themeMapping` in line with one of my other comments.

frontend/types/types.ts
```diff
@@ -373,3 +373,8 @@ export type ReviewRequestInterface = {
   createdAt: string
   updatedAt: string
 } & PartialReviewRequestInterface
+
+export interface UserSettings {
+  dn: string
+  theme: string
```
ARADDCC012, line 379: It's hard to keep track of what `theme` means throughout this PR. Can we clarify that this is a `themeKey`? Sorry if this results in wide sweeping changes, but being able to tell that this is this is linked to the `key` will help with readability.

frontend/src/theme.ts
```diff
@@ -234,3 +239,21 @@ export const darkTheme = createTheme({
     },
   },
 })
+
+export const ThemeNames = {
```
ARADDCC012, line 243: Generally the naming convention for enums is to use singular names, not plural. The reason being is that when you use the enum in code, you're selecting one thing from the choices available.

I think the same rules should apply here as it's our own way of using enums without all of the TS headache.

frontend/src/theme.ts
```diff
@@ -234,3 +239,21 @@ export const darkTheme = createTheme({
...
+  Light: 'light',
+  Dark: 'dark',
+} as const
+
+const lightThemeKey = {
```
ARADDCC012, line 248: If you agree with my other comment, rename this to `lightThemeMapping`

frontend/src/theme.ts
```diff
@@ -234,3 +239,21 @@ export const darkTheme = createTheme({
...
+  key: ThemeNames.Light,
+  theme: lightTheme,
+  title: 'Light',
+}
+const darkThemeKey = {
```
ARADDCC012, line 253: If you agree with my other comment, rename this to `darkThemeMapping`

ARADDCC012 (changes requested):

frontend/actions/user.ts
```diff
@@ -91,3 +99,26 @@ export function deleteUserToken(accessKey: TokenInterface['accessKey']) {
...
+interface UserSettingsResponse {
+  settings: UserSettings
+}
+
+export function useGetCurrentUserSettings() {
+  const { data, error, mutate } = useSWR<UserSettingsResponse, ErrorInfo>('/api/v2/user/settings', fetcher)
+
+  return {
+    mutateUserSettings: mutate,
+    userSettings: data?.settings || undefined,
```
ARADDCC012, line 112: Can be simplified to: `userSettings: data?.settings`

frontend/src/settings/ProfileTab.tsx
```diff
@@ -8,9 +7,11 @@ type ProfileTabProps = {
 
 export default function ProfileTab({ user }: ProfileTabProps) {
   return (
-    <Box sx={{ px: 2, py: 4 }}>
-      <Typography fontWeight='bold'>Name</Typography>
-      <Typography>{user.dn}</Typography>
-    </Box>
+    <Container maxWidth='md'>
+      <Box>
```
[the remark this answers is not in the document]
ARADDCC002: This keeps the content centered to be consistent with the other tabs

frontend/src/settings/display/ExampleDisplay.tsx
ARADDCC002: Each example display has its own theme (so you can see what the colours are!).

image

Ideally I'll improve it so they're not just buttons

PR #1155, opened 2024-03-06
Feature/infrastructure updates

JR40159:


backend/src/routes/middleware/expressLogger.ts
```diff
@@ -9,15 +9,16 @@ import log from '../../services/log.js'
...
-        {
-          url: tokens.url(req, res),
-          method: tokens.method(req, res),
-          'response-time': tokens['response-time'](req, res),
-          status: tokens.status(req, res),
-          code: 'approval',
-        },
-        tokens.dev(morgan, req, res),
+        info,
+        `${info.method} ${info.url} ${info.status} ${info['response-time']} ms - ${info['content-length']}`,
```
a3957273, line 21: I don't think you want a space between response time and 'ms' here.

a3957273 (approved):

backend/src/routes/middleware/expressLogger.ts
JR40159: I just blindly copied the morgan format but we're not restricted by that so I'll update to remove the space.

PR #1157, opened 2024-03-06
BAI-1093 Experiment Tracking

bw27492:


lib/python/src/bailo/helper/model.py
```diff
@@ -263,3 +275,126 @@ def __unpack_mc(self, res):
...
+        for run in self.raw:
+            if run["run"] == run_id:
+                sel_run = run
+            else:
+                continue
+
+        values = []
+
+        for metric in sel_run['metrics']:
+            for k, v in metric.items():
```
TT38665, line 395: What is this for loop doing?

lib/python/src/bailo/core/utils.py
```diff
@@ -11,3 +11,29 @@ def filter_none(json: dict[str, Any]) -> dict[str, Any]:
     """
     res = {k: v for k, v in json.items() if v is not None}
     return res
+
+
+class NestedDict(dict):
+    def __getitem__(self,keytuple):
```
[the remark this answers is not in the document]
bw27492: Allows the user to programmatically access a dict e.g. with a tuple of keys instead of `model_card["performance"]["performance metrics"]`

lib/python/src/bailo/helper/model.py
bw27492: This loops through all the model metrics for a selected experiment run, and formats them so they can be published to a Bailo model card

lib/python/src/bailo/helper/release.py
```diff
@@ -132,23 +134,52 @@ def from_version(cls, client: Client, model_id: str, version: Version | str) ->
...
+        """Upload a file to the release.
+
+        :param path: The path, or name of file or directory to be uploaded
+        :param data: A BytesIO object if not loading from disk
 
         :return: The unique file ID of the file uploaded
+        ..note:: If path provided is a directory, it will be uploaded as a zip
         """
-        res = self.client.simple_upload(self.model_id, name, file).json()
+        name = path.split("/")[-1]
```
GB27247, line 165: Be careful about this. This wouldn't necessarily work for Windows OS.

lib/python/tests/test_model.py
```diff
@@ -1,23 +1,25 @@
...
+def test_experiment(local_model):
+    experiment = Experiment(model=local_model)
+
+    assert isinstance(experiment, Experiment)
+
+
+def test_create_experiment_from_model(local_model):
+    experiment = local_model.create_experiment()
+
+    assert isinstance(experiment, Experiment)
```
GB27247, line 22: I don't think it's necessary to test the `local_model` fixture.

lib/python/tests/test_model.py
```diff
@@ -121,3 +123,53 @@ def test_create_release_without_model_card(integration_client):
...
+
+    assert expected_accuracy == actual_accuracy
+
+
+@pytest.mark.integration
+def test_publish_experiment_without_valid_id(standard_experiment):
+    run_id = "random_incorrect_id"
+
+    with pytest.raises(NameError):
+        standard_experiment.publish(mc_loc="performance.performanceMetrics", run_id=run_id)
```
GB27247, line 166: Can there also be a test for publishing with no runs.

lib/python/src/bailo/helper/model.py
```diff
@@ -263,3 +282,175 @@ def __unpack_mc(self, res):
...
+    @classmethod
+    def create(
+        cls,
+        model: Model,
+    ) -> Experiment:
+        """Create an experiment locally.
+
+        :param model: A Bailo model object which the experiment is being run on
+        :return: Experiment object
+        """
```
GB27247, line 311: I think it would be good to have a code block and some extra detail in this doc string for some more clarity on what experiment tracking includes.

lib/python/src/bailo/helper/release.py
```diff
@@ -132,23 +134,52 @@ def from_version(cls, client: Client, model_id: str, version: Version | str) ->
...
+            with open(path, "wb") as f:
+                f.write(res.content)
 
-    def upload(self, name: str, file: BytesIO) -> str:
-        """Upload files in a given directory to the release.
+        return res
 
-        :param name: The name of the file to upload to bailo
-        :param f: A BytesIO object
+    def upload(self, path: str, data: BytesIO | None = None) -> str:
```
a3957273, line 156: instead of downloading mlflow to the local directory, download to temporary directories instead.  (e.g. using tempfiles modules)

lib/python/src/bailo/helper/release.py
```diff
@@ -132,23 +134,52 @@ def from_version(cls, client: Client, model_id: str, version: Version | str) ->
...
+            if os.path.isdir(path):
+                shutil.make_archive(name, "zip", path)
+                path = f"{name}.zip"
+                name = path
+                is_zip = True
+
+            with open(path, "rb") as f:
+                res = self.client.simple_upload(self.model_id, name, f).json()
+
+            if is_zip:
```
GB27247, line 178: To avoid using is zip like this you can use the assignment expression operator(`:=`) eg.
```python
if is_zip := os.path.isdir(path):
    if os.path.isdir(path):
        shutil.make_archive(name, "zip", path)
        path = f"{name}.zip"
        name = path

    with open(path, "rb") as f:
                res = self.client.simple_upload(self.model_id, name, f).json()

    if is_zip:
   ```

File: lib/python/src/bailo/helper/model.py [hunk 9]
```diff
@@ -263,3 +282,175 @@ def __unpack_mc(self, res):
...
+
+        values = []
+
+        for metric in sel_run["metrics"]:
+            for k, v in metric.items():
+                values.append({"name": k, "value": v})
+
+        # Updating the model card
+        parsed_values = [{"dataset": sel_run["dataset"], "datasetMetrics": values}]
+        mc[tuple(mc_loc.split("."))] = parsed_values
```
GB27247, line 435: I'd be tempted to add this funtionality to a protected method on the `NestedDict` class. Instead of here, where it's not very readable or reusable.

lib/python/src/bailo/helper/model.py
```diff
@@ -263,3 +282,175 @@ def __unpack_mc(self, res):
...
+
+    def from_mlflow(self, tracking_uri: str, experiment_id: str):
+        """Imports information from an MLFlow Tracking experiment.
+
+        :param tracking_uri: MLFlow Tracking server URI
+        :param experiment_id: MLFlow Tracking experiment ID
+        :raises ImportError: Import error if MLFlow not installed
+        """
+        if ml_flow == 1:
+            client = mlflow.tracking.MlflowClient(tracking_uri=tracking_uri)
```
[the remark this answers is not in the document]
bw27492: For readability I'd argue it makes sense to keep it as `import mlfow`. Since there are other components used (e.g. `mlflow.artifacts.download_artifacts`. Likely to use more components in future too

lib/python/src/bailo/helper/model.py
```diff
@@ -263,3 +282,175 @@ def __unpack_mc(self, res):
...
+            # Create new release
+            try:
+                release_latest_version = self.model.get_latest_release().version
+                release_new_version = release_latest_version.next_minor()
+            except:
+                release_new_version = semver
+
+            run_id = sel_run["run"]
+            notes = f"{notes} (Run ID: {run_id})"
+            release_new = self.model.create_release(version=release_new_version, minor=True, notes=notes)
```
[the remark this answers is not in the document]
bw27492: Which variable here is reused?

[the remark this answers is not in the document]
GB27247: Apologies. This was from a previous version. It looks like it's been resolved.

lib/python/src/bailo/helper/model.py
```diff
@@ -263,3 +282,175 @@ def __unpack_mc(self, res):
...
+            self.run = self.run + 1
+        except:
+            self.run = 0
+
+        self.run_data = {"run": self.run, "params": [], "metrics": [], "artifacts": [], "dataset": ""}
+
+        self.raw.append(self.run_data)
+
+        if not is_mlflow:
+            print(f"Bailo tracking run {self.run}.")
```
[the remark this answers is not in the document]
bw27492: Agreed but I think logging should be applied across the board so I'll make a separate ticket

lib/python/src/bailo/helper/release.py
bw27492: TIL this is also called the walrus operator

lib/python/src/bailo/helper/release.py
```diff
@@ -132,23 +134,50 @@ def from_version(cls, client: Client, model_id: str, version: Version | str) ->
...
+            with open(path, "wb") as f:
+                f.write(res.content)
 
-    def upload(self, name: str, file: BytesIO) -> str:
-        """Upload files in a given directory to the release.
+        return res
 
-        :param name: The name of the file to upload to bailo
-        :param f: A BytesIO object
+    def upload(self, path: str, data: BytesIO = None) -> str:
```
GB27247, line 156: Same typing problem here.

GB27247 (approved):

PR #1158, opened 2024-03-08
Feature/bai 849 use open telemetry to send instrumentation info

JR40159:


backend/config/default.cjs
```diff
@@ -180,4 +180,9 @@ module.exports = {
       registry: 'registry',
     },
   },
+
+  instrumentation: {
+    enabled: false,
+    serviceName: 'backend',
```
a3957273, line 186: This should probably include `authenticationToken`.  The config should include every possible field, even if it defaults to empty.

a3957273 (changes requested):

JR40159:
Add the ability to turn on OT debug model via config

ARADDCC013 (approved):

PR #1163, opened 2024-03-13
Improve personal access token dialog

ARADDCC012:


ARADDCC002 (changes requested):

JR40159:
The warning 'You will never be able to access this token again. Make sure to copy it to a safe place.' should probably persist across all the tabs

a3957273 (changes requested):
internal notes

a3957273 (approved):

PR #1167, opened 2024-03-19
Use roles in authorisation

JR40159:


backend/src/connectors/authorisation/base.ts
```diff
@@ -143,16 +144,22 @@ export class BasicAuthorisationConnector {
...
-        if (!isNamed && !hasModelRole && [AccessRequestAction.Delete, AccessRequestAction.Update].includes(action)) {
+        /**
+         * Reject if:
+         *  - user is not named on the access request
+         *  - user is not the owner of the model
+         *  - the user is trying to delete or update an existing AR
+         */
+        if (
+          !isNamed &&
+          (await missingRequiredRole(user, model, ['owner'])) &&
```
a3957273, line 160: I think this should be either of 'owner' or 'contributor'.  I would expect someone with the permissions to upload a model to also download it.

a3957273 (changes requested):

backend/src/connectors/authorisation/base.ts
JR40159: Download a model? This is for the permissions when trying to update or delete an access request?

a3957273 (approved):

PR #1172, opened 2024-03-26
added auto focus to first question of json schema forms

ARADDCC002:


ARADDCC002:
This is still in draft as some of the variable names need refining and some values should be properly typed. Currently we're using `any` in a few places and I'd rather we avoid that at all costs.

ARADDCC012 (changes requested):

ARADDCC012 (approved):

PR #1208, opened 2024-04-10
Fix download error handling

GB27247:


frontend/src/model/releases/FileDownload.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+type FileDownloadProps = {
+  modelId: string
+  file: FileInterface
+}
+
+export default function FileDownload({ modelId, file }: FileDownloadProps) {
+  return (
+    <>
+      {isFileInterface(file) && (
```
a3957273, line 13: What if it's not a file interface?

frontend/src/model/releases/FileDownload.tsx
```diff
@@ -0,0 +1,33 @@
...
+        <Grid container alignItems='center' key={file.name}>
+          <Grid item xs={11}>
+            {isFileInterface(file) && (
```
a3957273, line 16: This seems like a double check.

frontend/src/model/releases/FileDownload.tsx
GB27247: Nothing will be rendered otherwise.

a3957273 (approved):

PR #1209, opened 2024-04-11
Disable create and edit buttons on failure

GB27247:


frontend/src/model/settings/ModelAccess.tsx
```diff
@@ -31,6 +31,7 @@ export default function ModelAccess({ model }: ModelAccessProps) {
   const [userListQuery, setUserListQuery] = useState('')
   const [loading, setLoading] = useState(false)
   const [errorMessage, setErrorMessage] = useState('')
+  const [roleError, onRoleError] = useState(false)
```
ARADDCC002, line 34: As this is a boolean I would rename this to `isRoleError` and `setIsRoleError`.

frontend/src/model/settings/ModelAccess.tsx
```diff
@@ -143,7 +145,13 @@ export default function ModelAccess({ model }: ModelAccessProps) {
         </TableHead>
         <TableBody>{accessListEntities}</TableBody>
       </Table>
-      <LoadingButton variant='contained' aria-label='Save access list' onClick={updateAccessList} loading={loading}>
+      <LoadingButton
```
ARADDCC002, line 148: If this is disabled due to the `roleError` value being set to `true` I would add some error text below the button explaining the error. 

It might be more work, but you could replace the boolean and have it as a string, and when you set `roleError` inside `UserDisplay` set it to error message created by `isUserInformationError`.

frontend/src/entry/model/releases/ReleaseForm.tsx
```diff
@@ -41,6 +41,7 @@ type ReleaseFormProps = {
   filesMetadata: FileWithMetadata[]
   onFilesMetadataChange: (value: FileWithMetadata[]) => void
   onImageListChange: (value: FlattenedModelImage[]) => void
+  setRegistryError: (value: boolean) => void
```
ARADDCC012, line 44: This is an event, consider renaming to `onRegistryError`

frontend/src/entry/model/releases/EditableRelease.tsx
```diff
@@ -43,6 +43,7 @@ export default function EditableRelease({ release, isEdit, onIsEditChange }: Edi
   const [imageList, setImageList] = useState<FlattenedModelImage[]>(release.images)
   const [errorMessage, setErrorMessage] = useState('')
   const [isLoading, setIsLoading] = useState(false)
+  const [registryError, setRegistryError] = useState(false)
```
ARADDCC012, line 46: At a glance I would assume this is a string that holds the text for an error. Consider renaming to `isRegistryError` and `setIsRegistryError` to clarify that it is a boolean.

frontend/src/entry/model/releases/EditableRelease.tsx
```diff
@@ -214,6 +216,7 @@ export default function EditableRelease({ release, isEdit, onIsEditChange }: Edi
         onFilesChange={(value) => setFiles(value)}
         onFilesMetadataChange={(value) => setFilesMetadata(value)}
         onImageListChange={(value) => setImageList(value)}
+        setRegistryError={(value) => setRegistryError(value)}
```
ARADDCC012, line 219: As this function is being used in a `useEffect` it must be stable. Currently the function is recreated on every render, which will cause an infinite loop. To prevent this I would add the following logic to this file (I have suggested renaming in other comments, but have left the names untouched in the below snippet):
```
// Change this line to
setRegistryError={handleRegistryError}

// Add the following to this component to make this function stable
const handleRegistryError = useCallback((value: boolean) => setRegistryError(value), [])
```

frontend/pages/model/[modelId]/release/new.tsx
```diff
@@ -156,6 +157,7 @@ export default function NewRelease() {
                   filesMetadata={filesMetadata}
                   onFilesMetadataChange={(value) => setFilesMetadata(value)}
                   onImageListChange={(value) => setImageList(value)}
+                  setRegistryError={(value) => setRegistryError(value)}
```
Comment (ARADDCC012) [inline frontend/pages/model/[modelId]/release/new.tsx:160]: As this function is being used in a `useEffect` it must be stable. Currently the function is recreated on every render, which will cause an infinite loop. To prevent this I would add the following logic to this file (I have suggested renaming in other comments, but have left the names untouched in the below snippet):
```
// Change this line to
setRegistryError={handleRegistryError}

// Add the following to this component to make this function stable
const handleRegistryError = useCallback((value: boolean) => setRegistryError(value), [])
```

frontend/pages/model/[modelId]/release/new.tsx
```diff
@@ -26,6 +26,7 @@ export default function NewRelease() {
   const [imageList, setImageList] = useState<FlattenedModelImage[]>([])
   const [errorMessage, setErrorMessage] = useState('')
   const [loading, setLoading] = useState(false)
+  const [registryError, setRegistryError] = useState(false)
```
Comment (ARADDCC012) [inline frontend/pages/model/[modelId]/release/new.tsx:29]: Consider renaming to `isRegistryError` and `setIsRegistryError` to clarify that it's a boolean

frontend/src/entry/model/ModelImageList.tsx
```diff
@@ -22,12 +22,26 @@ type PartialModelImageListProps =
...
+  readOnly = false,
+  multiple,
+}: ModelImageListProps) {
   const { modelImages, isModelImagesLoading, isModelImagesError } = useGetModelImages(model.id)
 
+  useEffect(() => {
+    if (setRegistryError) {
+      setRegistryError(!!isModelImagesError)
+    }
+  }, [isModelImagesError, setRegistryError])
```
ARADDCC012, line 43: As a learning point, be aware that it can be dangerous to put function props in any dependency arrays (`setRegistryError` in this case). JS functions are recreated on every render and will therefore always trigger hooks that depend on it unless you wrap it with a `useCallback` that itself doesn't rely on an unstable function.

I've already added comments for where those `useCallbacks` are needed in this PR, but thought it worth putting this here so that you can get the full picture.

frontend/src/entry/model/settings/EntityItem.tsx
```diff
@@ -15,13 +15,13 @@ type EntityItemProps = {
   accessList: CollaboratorEntry[]
   onAccessListChange: (value: CollaboratorEntry[]) => void
   model: EntryInterface
+  setRoleError: (value: string) => void
```
ARADDCC012, line 18: Two things here. One is that this is an event, so should be prefixed with `on...`. The other is that this should be generic as it is used for two different errors. Consider renaming to `onError` as beyond this we don't care where the error originated, we simply care that there is an error to display to the user.

frontend/src/entry/model/settings/EntityItem.tsx
```diff
@@ -99,9 +100,10 @@ function EntityIcon({ entity }: EntityIconProps) {
 
 type EntityNameDisplayProps = {
   entity: CollaboratorEntry
+  setRoleError: (value: string) => void
```
ARADDCC012, line 103: Consider renaming to `onUserError`

frontend/src/entry/model/settings/ModelAccess.tsx
```diff
@@ -30,6 +31,7 @@ export default function ModelAccess({ model }: ModelAccessProps) {
   const [userListQuery, setUserListQuery] = useState('')
   const [loading, setLoading] = useState(false)
   const [errorMessage, setErrorMessage] = useState('')
+  const [roleError, setRoleError] = useState('')
```
ARADDCC012, line 34: This isn't needed because we already have `errorMessage` that we can use to display your new errors being returned. We always use a `MessageAlert` to display errors to the user, so opting for a `Tooltip` here adds inconsistency to our app.

I suggest using `errorMessage` to handle the displaying of the error and have a boolean `isEntityError` to handle disabling the button.

frontend/src/entry/model/settings/ModelAccess.tsx
```diff
@@ -142,9 +145,19 @@ export default function ModelAccess({ model }: ModelAccessProps) {
         </TableHead>
         <TableBody>{accessListEntities}</TableBody>
       </Table>
-      <LoadingButton variant='contained' aria-label='Save access list' onClick={updateAccessList} loading={loading}>
-        Save
-      </LoadingButton>
+      <Tooltip title={roleError}>
```
ARADDCC012, line 148: This `Tooltip` and `span` won't be required if you use `errorMessage` to display the error.

ARADDCC012 (changes requested):

frontend/src/Form/EditableFormHeading.tsx
```diff
@@ -11,6 +11,7 @@ type EditableFormHeadingProps = {
   onEdit: () => void
   onCancel: () => void
   onSubmit: () => void
+  registryError?: boolean
```
ARADDCC012, line 14: Consider renaming to `isRegistryError`

frontend/src/entry/model/ModelImageList.tsx
```diff
@@ -22,12 +22,26 @@ type PartialModelImageListProps =
...
+  model,
+  value,
+  onChange,
+  onRegistryError,
+  readOnly = false,
+  multiple,
+}: ModelImageListProps) {
   const { modelImages, isModelImagesLoading, isModelImagesError } = useGetModelImages(model.id)
 
+  const handleRegistryError = useCallback((value: boolean) => onRegistryError(value), [onRegistryError])
```
ARADDCC012, line 39: This callback doesn't prevent infinite rendering because it's dependent on `onRegistryError`. You can change the below `useEffect` to be like it was before:
```
  useEffect(() => {
    onRegistryError(!!isModelImagesError)
  }, [isModelImagesError, handleRegistryError])
```

For context, the reason why the `useCallback`s in the release and inferencing form components work is because they are only dependent on a `useState` setter function, which React guarantees is a stable function (also why it makes no difference as to whether or not you include setter functions in the dependency array)

frontend/pages/model/[modelId]/access-request/[accessRequestId]/review.tsx
```diff
@@ -100,7 +100,7 @@ export default function AccessRequestReview() {
             <Divider />
             <Stack spacing={1} direction='row' justifyContent='space-between' sx={{ mb: 2 }}>
               <Typography variant='caption'>
-                Created by {<UserDisplay dn={accessRequest.createdBy} />} on
+                Created by {<UserDisplay onUserInformationError={setErrorMessage} dn={accessRequest.createdBy} />} on
```
Comment (ARADDCC012) [inline frontend/pages/model/[modelId]/access-request/[accessRequestId]/review.tsx:103]: `onUserInformationError` not needed

frontend/src/common/UserDisplay.tsx
ARADDCC012, line 1: The changes to this file can be reverted

frontend/src/entry/model/settings/EntityItem.tsx
ARADDCC012, line 1: The changes to this file can be reverted

frontend/src/entry/model/settings/ModelAccess.tsx
```diff
@@ -44,6 +44,7 @@ export default function ModelAccess({ model }: ModelAccessProps) {
           accessList={accessList}
           onAccessListChange={setAccessList}
           model={model}
+          onError={setErrorMessage}
```
ARADDCC012, line 47: This change can be reverted

ARADDCC012 (changes requested):

ARADDCC012 (approved):

PR #1213, opened 2024-04-16
Export model cards and releases to S3

JR40159:


backend/src/clients/kms.ts
```diff
@@ -0,0 +1,42 @@
+import { DescribeKeyCommand, KMSClient, SignCommand, SignCommandOutput } from '@aws-sdk/client-kms'
+import { ASN1HEX } from 'jsrsasign'
+
+import config from '../utils/config.js'
+import { InternalError } from '../utils/error.js'
+
+const uint8ArrayFromHexString = (hexstring) =>
```
a3957273, line 7: Is there a reason you're doing regular expression matching over a simpler `Buffer.from(hexstring, 'hex')` here?

backend/src/routes/v2/model/postRequestExport.ts
```diff
@@ -0,0 +1,65 @@
...
+      .array(z.string())
+      .optional()
+      .openapi({ example: ['0.0.1', '0.0.2'] }),
+    disclaimerAgreement: z.boolean(),
+  }),
+})
+
+registerPath({
+  method: 'post',
+  path: '/api/v2/model/:modelId/export',
```
a3957273, line 26: I think we should have separate concepts of 'exporting', where we give the user a zip / pdf / etc and 'mirroring', where we transfer the model ourselves to another location.

backend/src/services/mirroredModel.ts
```diff
@@ -0,0 +1,224 @@
...
+  // unknown release semver
+  if (releaseSemvers && releaseSemvers.length > 0) {
+    zipFiles.push({
+      filename: `${modelId}/releases.zip`,
+      file: await generateReleaseZip(user, model, releaseSemvers),
+      metadata: { modelId, exportKind: ExportKind.Releases },
+    })
+  }
+
+  return Promise.all(zipFiles.map((zipFile) => uploadZipFileToS3(zipFile)))
```
a3957273, line 69: Have you checked that all of this definitely uses streaming, and will handle sufficiently large machine learning models?

backend/src/services/mirroredModel.ts
```diff
@@ -0,0 +1,224 @@
...
+  const s3Stream = new PassThrough()
+  const hashStream = new PassThrough()
+
+  zip.file.pipe(hashStream)
+  zip.file.pipe(s3Stream)
+
+  let signatures = {}
+  if (config.modelMirror.export.kmsSignature.enabled) {
+    try {
+      const messageDigest = await generateDigest(hashStream)
```
a3957273, line 82: I think this will break.

You create a 'zip file' of the zip files and model cards.  Then you wait for it to _entirely_ create before generating a hash.  We won't have enough memory to reliably do this.  Also, during this time the 's3Stream' hasn't done anything, so if your file is larger than the HighWatermark the upload will stop?

backend/src/services/mirroredModel.ts
```diff
@@ -0,0 +1,224 @@
...
+  const releases = await getReleasesForExport(user, model.id, semvers)
+  if (semvers.length > 1) {
+    await checkTotalFileSize(model.id, semvers)
+  }
+
+  const zip = archiver('zip')
+  const errors: any[] = []
+  // Using a .catch here to ensure all errors are returned, rather than just the first error.
+  await Promise.all(releases.map((release) => addReleaseToZip(user, model, release, zip).catch((e) => errors.push(e))))
+  await zip.finalize()
```
a3957273, line 145: Archiver isn't an 'async / await' library.  The 'await' doesn't do anything here.  Also, I'm fairly confident you have to listen to events from the 'zip' created prior to finalizing or they could return early.

a3957273 (changes requested):

a3957273 (approved):

PR #1217, opened 2024-04-17
Remove Release and Access Request banner

GB27247:


frontend/pages/model/[modelId].tsx
```diff
@@ -17,24 +18,30 @@ export default function Model() {
   const router = useRouter()
   const { modelId }: { modelId?: string } = router.query
   const { model, isModelLoading, isModelError } = useGetModel(modelId)
+  const { currentUser, isCurrentUserLoading, isCurrentUserError } = useGetCurrentUser()
   const { uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
 
+  const roles = useMemo(
```
Comment (ARADDCC012) [inline frontend/pages/model/[modelId].tsx:24]: It's a bit verbose, but it might help to clarify what roles these are. We also have the concept of modelroles, which is an array of objects that may not be related to the current user. Consider changing all of the `roles` in this PR to be `currentUserRoles` or similar.

Also, when working with arrays it tends to make logic easier down the line if you always return an array. So you could do:
```
  const currentUserRoles = useMemo(
    () =>
      model?.collaborators.find((collaborator) => collaborator.entity.split(':')[1] === currentUser?.dn)?.roles || [],
    [model, currentUser],
  )
```
You could then remove the `roles` check in the `tabs` memo hook below.

frontend/pages/model/[modelId].tsx
```diff
@@ -62,13 +69,14 @@ export default function Model() {
...
     isModelError,
+    isCurrentUserError,
     isUiConfigError,
   })
   if (error) return error
 
   return (
     <Wrapper title={model ? model.name : 'Loading...'} page='marketplace' fullWidth>
-      {(isModelLoading || isUiConfigLoading) && <Loading />}
+      {isModelLoading && isCurrentUserLoading && isUiConfigLoading && <Loading />}
```
Comment (ARADDCC012) [inline frontend/pages/model/[modelId].tsx:79]: This should be `{(isModelLoading || isCurrentUserLoading || isUiConfigLoading) && <Loading />}` because we want to show the loading component if any of them are loading.

frontend/src/model/AccessRequests.tsx
```diff
@@ -23,12 +24,16 @@ export default function AccessRequests({ model }: AccessRequestsProps) {
         accessRequests
           .sort(sortByCreatedAtDescending)
           .map((accessRequest) => (
-            <AccessRequestDisplay accessRequest={accessRequest} key={accessRequest.metadata.overview.name} />
+            <AccessRequestDisplay
+              accessRequest={accessRequest}
+              key={accessRequest.metadata.overview.name}
+              hideReviewBanner={!roles?.some((role) => ['msro', 'mtr'].includes(role))}
```
ARADDCC012, line 30: Can we move this logic into `frontend/utils/roles.ts` as a reusable function?
```
export const hasRole = (userRoles: string[], validRoles: string[]) => {
 return userRoles.some((role) => validRoles.includes(role))
}
```
You can then use it here like so:
```
hideReviewBanner={!hasRole(roles, ['msro', 'mtr'])}
```

frontend/src/model/Releases.tsx
```diff
@@ -18,9 +23,15 @@ export default function Releases({ model }: { model: ModelInterface }) {
   const releaseDisplays = useMemo(
     () =>
       releases.map((release) => (
-        <ReleaseDisplay key={release.semver} model={model} release={release} latestRelease={latestRelease} />
+        <ReleaseDisplay
+          key={release.semver}
+          model={model}
+          release={release}
+          latestRelease={latestRelease}
+          hideReviewBanner={!roles?.some((role) => ['msro', 'mtr'].includes(role))}
```
ARADDCC012, line 31: See other comment about using reusable logic for this.

frontend/src/model/AccessRequests.tsx
```diff
@@ -12,9 +12,10 @@ import Loading from '../common/Loading'
 
 type AccessRequestsProps = {
   model: ModelInterface
+  roles: string[]
```
ARADDCC012, line 15: Consider renaming to `currentUserRoles`

frontend/src/model/Releases.tsx
```diff
@@ -9,7 +9,12 @@ import EmptyBlob from '../common/EmptyBlob'
 import Loading from '../common/Loading'
 import ReleaseDisplay from './releases/ReleaseDisplay'
 
-export default function Releases({ model }: { model: ModelInterface }) {
+type ReleasesProps = {
+  model: ModelInterface
+  roles: string[]
```
ARADDCC012, line 14: Consider renaming to `currentUserRoles`

ARADDCC012 (changes requested):

ARADDCC012 (approved):

PR #1229, opened 2024-04-23
Chore/update backend types entry

ARADDCC002:


backend/src/routes/v2/model/postModel.ts
```diff
@@ -4,14 +4,15 @@ import { z } from 'zod'
...
-import { EntryVisibility, ModelInterface } from '../../../models/Model.js'
+import { EntryKind, EntryVisibility, ModelInterface } from '../../../models/Model.js'
 import { createModel } from '../../../services/model.js'
 import { modelInterfaceSchema, registerPath } from '../../../services/specification.js'
 import { parse } from '../../../utils/validate.js'
 
 export const postModelSchema = z.object({
   body: z.object({
     name: z.string(),
+    kind: z.nativeEnum(EntryKind).optional().default(EntryKind.Model),
```
ARADDCC012, line 15: Should we force this to be specified so that API users can't accidentally create a model when they intended to make a data card?

frontend/src/marketplace/EntryList.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { ModelSearchResult } from 'actions/model'
+import Link from 'next/link'
+import { Fragment } from 'react'
+import ChipSelector from 'src/common/ChipSelector'
+import EmptyBlob from 'src/common/EmptyBlob'
+
+interface EntryListProps {
+  entries: ModelSearchResult[]
+  selectedChips: string[]
+  handleChipsOnChange: (chips: string[]) => void
```
ARADDCC012, line 12: This is effectively an event prop as we are just passing it through, so consider renaming to `onSelectedChipsChange`.

I believe this is the agreed upon convention. `on...` denotes an event, `handle...` denotes a handler for an event.

ARADDCC012 (changes requested):

backend/src/migrations/002_update_model_kinds.ts
```diff
@@ -0,0 +1,15 @@
+import ModelModel from '../models/Model.js'
+
+export async function up() {
+  const models = await ModelModel.find({})
+  for (const model of models) {
+    if (model.kind === undefined) {
+      model.kind = 'model'
+      model.save()
```
a3957273, line 8: You need an `await`!

ARADDCC002: whoops duh! Added

ARADDCC012 (approved):

PR #1235, opened 2024-04-30
added an outdated label to outdated review responses and added the ro…

ARADDCC002:
image

### DIFF
--- a/frontend/src/reviews/ReviewDecisionDisplay.tsx
+++ b/frontend/src/reviews/ReviewDecisionDisplay.tsx
@@ -1,52 +1,79 @@
 import Done from '@mui/icons-material/Done'
 import HourglassEmpty from '@mui/icons-material/HourglassEmpty'
 import { Card, Divider, Stack, Typography } from '@mui/material'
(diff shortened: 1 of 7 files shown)


frontend/src/reviews/ReviewDecisionDisplay.tsx
```diff
@@ -1,52 +1,81 @@
...
+                  <UserDisplay dn={username} />
+                  {` ${isApproved ? 'has approved' : 'has requested changes'}`}
+                </Typography>
+                {isApproved ? (
+                  <Done color='success' fontSize='small' />
+                ) : (
+                  <HourglassEmpty color='warning' fontSize='small' />
+                )}
+                {response.outdated && (
+                  <Typography sx={{ backgroundColor: theme.palette.warning.light, borderRadius: 1, px: 0.5 }}>
```
ARADDCC012, line 62: Subjective, but I think the standard `theme.palette.warning.light` looks better. The colours we use elsewhere are vibrant and this looks much more dull in comparison. What do you think of using a border like this?

image

ARADDCC012 (approved):

PR #1236, opened 2024-04-30
Enable local replica set + enable basic authentication

GCHQDeveloper36:
Whilst working on transactions, I noticed the local instance wasn't quite in replica set mode, possibly after my switch to the Bitnami image.

Authentication seems to be required, I wasn't able to force it to be disabled during dev but happy to try again or take suggestions on that. If it remains enabled, I've followed the pattern for other services of `serviceadmin` for user/pass here but happy to use something else.

It may require dropping the mongo Docker volume if you struggle to bring it up when testing these changes, though the Bitnami image 'should' adapt automatically.

### DIFF
--- a/.github/workflows/build.js.yml
+++ b/.github/workflows/build.js.yml
@@ -313,6 +313,13 @@ jobs:
               secretKey: 'minioadmin',
             },
 
+            mongo: {
+              uri: 'mongodb://mongo:27017/bailo?replicaSet=rs0&authSource=admin',
+
+              user: 'mongoadmin',
+              pass: 'mongoadmin',
+            },
+
             smtp: {
               auth: {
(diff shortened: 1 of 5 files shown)


GCHQDeveloper36:
I've put this back into draft as I noticed upon restarting the Mongo container it seems to struggle to become primary. Should be fixable but I won't be able to look at it till Friday. Happy to take feedback on the PR as a whole though and I can incorporate anything into it on the next go-around.

GCHQDeveloper36:
Must have been something weird locally as it seems fine today. Tests now pass, missed it was using the prod docker compose in CI/CD. If desired, I can tweak the prod compose to be a three-member replica set.

a3957273 (approved):

PR #1237, opened 2024-04-30
Enable Mongo transactions for updating model cards

GCHQDeveloper36:
(Depends on #1236)

Suggested implementation for transactions around updating model cards - possibly consider this WIP until there's been some further testing done by others.


backend/src/services/model.ts
```diff
@@ -271,9 +274,21 @@ export async function _setModelCard(
...
 
   const revision = new ModelCardRevisionModel({ ...newDocument, modelId, createdBy: user.dn })
-  await revision.save()
 
-  await ModelModel.updateOne({ id: modelId }, { $set: { card: newDocument } })
+  if (!isReplicaSet()) {
+    throw InternalError('Database is not in replica set mode, cannot use transactions')
+  }
+
+  await mongoose.connection
```
JR40159, line 282: To reduce added duplication in the future, it would be nice if we could wrap this transaction logic into a reusable function that takes in the database actions as callbacks and executes them in a transaction if available or just executes them sequentially if transactions isn't available

GCHQDeveloper36:
I've refactored based on the latest comments/discussions, as well as added some basic tests in. I don't think there's much I can do about requiring 'knowledge' of the session within the backend service code without rewriting and introducing a heap of 'new' code to override the existing mongoose methods. But, if there's no session, it should be silently dropped and unused.

backend/src/utils/transactions.ts
```diff
@@ -0,0 +1,48 @@
...
+ * if transactions are allowed - see {@link useTransactions} for the conditions.
+ * Otherwise, the actions will be executed as normal.
+ *
+ * @param actions to execute in the provided order
+ * @returns the results of each action as an array
+ */
+export async function useTransaction<T extends any[]>(actions: {
+  [K in keyof T]: TransactionCallback<T[K]>
+}): Promise<T> {
+  if (useTransactions()) {
```
ARADDCC002, line 42: We normally "fail first" when it comes to error/issue checking. It's a bit of a preference but I feel we should reverse this if statement that we're checking to see if `useTransactions()` is false/undefined, then return the error.

backend/src/utils/transactions.ts
```diff
@@ -0,0 +1,48 @@
...
+    return await mongoose.connection.transaction((session) => executeInsideTransaction(actions, session))
+  } else {
+    log.trace(transactionsNotEnabled)
+    return executeInsideTransaction(actions)
```
ARADDCC002, line 46: If `useTransactions()` is false/undefined by this point, the comment at the top of the file suggests that we do this work outside of a transaction, but the function name `executeInsideTransaction" implies it will be a transaction still? Apologies if I'm missing something here.

ARADDCC002 (changes requested):

backend/src/utils/transactions.ts
GCHQDeveloper36: Yeah this was poorly named - I've renamed the function to just 'execute'

GCHQDeveloper36:
Rebased against main (v2.8.1)

GCHQDeveloper36:
Rebased against main (v2.9.0)

PR #1263, opened 2024-05-08
Switch hashes to SHA-256 and implement token limitations

a3957273:


backend/src/models/Token.ts
```diff
@@ -10,12 +11,37 @@ export const TokenScope = {
 export type TokenScopeKeys = (typeof TokenScope)[keyof typeof TokenScope]
 
 export const TokenActions = {
-  ImageRead: 'image:read',
+  ModelRead: 'model:read',
```
JR40159, line 14: These all seem very similar to the permissions defined in the authorisation actions. My concern here is duplication and confusion in how they are similar but not the same- eg in the authorisation actions there's `access_request:create` and `access_request:view` and in this file you've defined `access_request:write` and `access_request:read`.

JR40159 (changes requested):

backend/src/models/Token.ts
JR40159: To avoid duplication these token actions should be exposed via the API so clients know what token actions are available to them

JR40159 (changes requested):

PR #1265, opened 2024-05-08
changed how reviews are displayed on review page

ARADDCC002:
image

### DIFF
--- a/frontend/src/reviews/ReviewsListContainer.tsx
+++ b/frontend/src/reviews/ReviewsListContainer.tsx
@@ -17,7 +17,7 @@ export default function ReviewsListContainer({ status }: ReviewsListContainerPro
 
   useEffect(() => {
     if (isReviewKind(category)) {
-      setSelectedCategory(category ?? ReviewKind.ACCESS)
+      setSelectedCategory(category)
     }
   }, [category])
 
--- a/frontend/src/reviews/ReviewsListContainer.tsx
+++ b/frontend/src/reviews/ReviewsListContainer.tsx
@@ -1,0 +1,69 @@
+import { Container, Divider, List, Stack } from '@mui/material'
+import router from 'next/router'
+import { useEffect, useState } from 'react'
+import SimpleListItemButton from 'src/common/SimpleListItemButton'
+import ReviewsList from 'src/reviews/ReviewsList'
+import { ReviewListStatusKeys } from 'types/types'
+
+export const ReviewListContainerCategory = {
+  RELEASE: 'release',
+  ACCESS: 'access',
+} as const
+
+export type ReviewListContainerCategoryKeys =
+  (typeof ReviewListContainerCategory)[keyof typeof ReviewListContainerCategory]
+
+function isReviewListContainerCategory(value: string | string[] | undefined): value is ReviewListContainerCategoryKeys {
+  return value === ReviewListContainerCategory.RELEASE || value === ReviewListContainerCategory.ACCESS
+}
+
+export interface ReviewsListContainerProps {
+  status: ReviewListStatusKeys
+}
+
+export default function ReviewsListContainer({ status }: ReviewsListContainerProps) {
+  const { category } = router.query
+
+  const [selectedCategory, setSelectedCategory] = useState<ReviewListContainerCategoryKeys>(
+    ReviewListContainerCategory.RELEASE,
+  )
+
+  useEffect(() => {
+    if (isReviewListContainerCategory(category)) {
+      setSelectedCategory(category ?? ReviewListContainerCategory.RELEASE)
+    }
+  }, [category])
+
+  const handleListItemClick = (category: ReviewListContainerCategoryKeys) => {
+    setSelectedCategory(category)
+    router.replace({
+      query: { ...router.query, category },
+    })
+  }
+  return (
+    <Stack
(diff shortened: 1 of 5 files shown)


frontend/src/reviews/ReviewsListContainer.tsx
```diff
@@ -0,0 +1,59 @@
+import { Container, Divider, List, Stack } from '@mui/material'
+import router from 'next/router'
+import { useEffect, useState } from 'react'
+import SimpleListItemButton from 'src/common/SimpleListItemButton'
+import ReviewsList from 'src/reviews/ReviewsList'
+import { ReviewKind, ReviewKindKeys, ReviewListStatusKeys } from 'types/types'
+
+function isReviewListContainerCategory(value: string | string[] | undefined): value is ReviewKindKeys {
```
ARADDCC012, line 8: As this is a type guard for checking if it's a review kind, consider renaming to `isReviewKind`, moving into `types.ts` (next to the relevant type) and changing `value` to be `value: unknown` to make it more generic.

frontend/src/reviews/ReviewsListContainer.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return value === ReviewKind.RELEASE || value === ReviewKind.ACCESS
+}
+
+export interface ReviewsListContainerProps {
+  status: ReviewListStatusKeys
+}
+
+export default function ReviewsListContainer({ status }: ReviewsListContainerProps) {
+  const { category } = router.query
```
ARADDCC012, line 17: Shouldn't this be using `useRouter`?
```
const router = useRouter()
const { category } = router.query
```

frontend/src/reviews/ReviewsListContainer.tsx
```diff
@@ -0,0 +1,59 @@
...
+          selected={selectedCategory === ReviewKind.RELEASE}
+          onClick={() => handleListItemClick(ReviewKind.RELEASE)}
+        >
+          Releases
+        </SimpleListItemButton>
+        <SimpleListItemButton
+          selected={selectedCategory === ReviewKind.ACCESS}
+          onClick={() => handleListItemClick(ReviewKind.ACCESS)}
+        >
+          Access requests
```
ARADDCC012, line 50: Should Access Requests come first? They're more prominent than releases. Also this would be ordering them alphabetically.

Also consider using title case here, so `Access Requests`.

frontend/src/reviews/ReviewsListContainer.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export interface ReviewsListContainerProps {
+  status: ReviewListStatusKeys
+}
+
+export default function ReviewsListContainer({ status }: ReviewsListContainerProps) {
+  const { category } = router.query
+
+  const [selectedCategory, setSelectedCategory] = useState<ReviewKindKeys>(ReviewKind.RELEASE)
```
ARADDCC012, line 19: As access requests are more prominent, should we default to showing those?

frontend/src/reviews/ReviewsListContainer.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  useEffect(() => {
+    if (isReviewListContainerCategory(category)) {
+      setSelectedCategory(category ?? ReviewKind.RELEASE)
```
ARADDCC012, line 23: Same here, should we default to showing access requests instead?

frontend/src/reviews/ReviewsListContainer.tsx
```diff
@@ -0,0 +1,59 @@
...
+        </SimpleListItemButton>
+        <SimpleListItemButton
+          selected={selectedCategory === ReviewKind.ACCESS}
+          onClick={() => handleListItemClick(ReviewKind.ACCESS)}
+        >
+          Access requests
+        </SimpleListItemButton>
+      </List>
+      <Container sx={{ my: 2 }}>
+        {selectedCategory === ReviewKind.RELEASE && <ReviewsList kind={ReviewKind.RELEASE} status={status} />}
```
ARADDCC012, line 54: Could simplify these two lines to the following:
```
<ReviewsList kind={selectedCategory} status={status} />
```

ARADDCC012 (changes requested):

frontend/src/reviews/ReviewsListContainer.tsx
```diff
@@ -1,26 +1,23 @@
...
   const { category } = router.query
 
-  const [selectedCategory, setSelectedCategory] = useState<ReviewKindKeys>(ReviewKind.RELEASE)
+  const [selectedCategory, setSelectedCategory] = useState<ReviewKindKeys>(ReviewKind.ACCESS)
 
   useEffect(() => {
-    if (isReviewListContainerCategory(category)) {
-      setSelectedCategory(category ?? ReviewKind.RELEASE)
+    if (isReviewKind(category)) {
+      setSelectedCategory(category ?? ReviewKind.ACCESS)
```
ARADDCC012, line 20: This can be changed to `setSelectedCategory(category)` as it already has passed the `isReviewKind` check

ARADDCC012 (changes requested):

ARADDCC012 (approved):

PR #1270, opened 2024-05-13
added isadmin property to getCurrentUser call so we can hide certain …

ARADDCC002:
…pages from nonadmins

Note this is used for client-side validation ONLY. This should not be used for checking for anything important - it's main purpose is to hide pages/sections that normal users are currently not able to use anyway.

### DIFF
--- a/backend/src/routes/v2/entities/getCurrentUser.ts
+++ b/backend/src/routes/v2/entities/getCurrentUser.ts
@@ -2,6 +2,8 @@ import bodyParser from 'body-parser'
 import { Request, Response } from 'express'
 import { z } from 'zod'
 
+import { Roles } from '../../../connectors/authentication/Base.js'
+import authentication from '../../../connectors/authentication/index.js'
 import { UserInterface } from '../../../models/User.js'
 import { registerPath, userInterfaceSchema } from '../../../services/specification.js'
 import { parse } from '../../../utils/validate.js'
@@ -34,7 +36,6 @@ export const getCurrentUser = [
   bodyParser.json(),
   async (req: Request, res: Response<GetCurrentUserResponses>) => {
     const _ = parse(req, getCurrentUserSchema)
-
-    return res.json({ user: req.user })
+    return res.json({ user: { ...req.user, isAdmin: await authentication.hasRole(req.user, Roles.Admin) } })
   },
 ]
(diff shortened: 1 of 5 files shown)


frontend/pages/schemas/list.tsx
```diff
@@ -1,10 +1,29 @@
...
 import SchemaTab from 'src/schemas/SchemaTab'
 
 export default function SchemasPage() {
+  const { currentUser, isCurrentUserLoading, isCurrentUserError } = useGetCurrentUser()
+
+  if (isCurrentUserLoading) return <Loading />
+
+  if (!currentUser || !currentUser.isAdmin) {
+    return (
+      <Forbidden errorMessage='If you think this is in error please contact the model owners.' noMargin hideNavButton />
```
a3957273, line 18: Should say "Bailo admins" or equivalent, not model owners.

frontend/pages/schemas/list.tsx
```diff
@@ -1,10 +1,29 @@
...
+    )
+  }
+
+  const error = MultipleErrorWrapper(`Unable to load model page`, {
```
a3957273, line 22: This is not a model page.

frontend/types/types.ts
```diff
@@ -161,6 +161,7 @@ export interface PostSimpleUpload {
 
 export interface User {
   dn: string
+  isAdmin?: boolean
```
a3957273, line 164: Why is this optional?

backend/src/routes/v2/entities/getCurrentUser.ts
```diff
@@ -34,7 +36,6 @@ export const getCurrentUser = [
   bodyParser.json(),
   async (req: Request, res: Response<GetCurrentUserResponses>) => {
     const _ = parse(req, getCurrentUserSchema)
-
-    return res.json({ user: req.user })
+    return res.json({ user: { ...req.user, isAdmin: await authentication.hasRole(req.user, Roles.Admin) } })
```
a3957273, line 39:
```typescript
const _ = parse(req, getCurrentUserSchema)
const isAdmin = await authentication.hasRole(req.user, Roles.Admin)
return res.json({ user: { ...req.user, isAdmin }})
```

Splitting it out onto a separate line improves the readability, especially if we add more values.

a3957273 (changes requested):

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -152,7 +152,7 @@ export default function SideNavigation({
             />
             <Divider />
             {/* TODO Once currentUser api has been updated to use roles we should check if they're admin */}
```
ARADDCC012, line 154: This comment can be removed now

ARADDCC012 (changes requested):

a3957273 (approved):

PR #1280, opened 2024-05-21
completely resstructed responses by having them in their own collection

ARADDCC002:
Comments / Review responses are now stored in a single collection, and are populated as and when they need to be by the API.

This will improve accessing them for the purpose of editing. 

This in draft state until the migration script is finished.


backend/src/connectors/authorisation/base.ts
```diff
@@ -108,6 +116,16 @@ export class BasicAuthorisationConnector {
     )
   }
 
+  async responses(user: UserInterface, response: ResponseDoc, action: ResponseActionKeys): Promise<Response> {
+    if (action === ResponseAction.Update && toEntity('user', user.dn) !== response.user) {
+      return { id: user.dn, success: false, info: 'Only the author can update a comment' }
+    }
+    return {
```
a3957273, line 123: We should probably make sure that if we're creating a response, the user has access to the 'access request' they're creating the response against.

backend/src/migrations/004_comment_response_refactor.ts
```diff
@@ -0,0 +1,73 @@
...
+import ResponseModel, { ResponseKind } from '../models/Response.js'
+import ReviewModel from '../models/Review.js'
+import { toEntity } from '../utils/entity.js'
+
+export async function up() {
+  // Resolve release comments
+  const releases: any = await ReleaseModel.find({})
+  for (const release of releases) {
+    release.commentIds = []
+    if (release.get('comments') !== undefined) {
```
a3957273, line 12: Question: What does `.get` do on a model?  Why not `release.comments` here?

backend/src/migrations/004_comment_response_refactor.ts
```diff
@@ -0,0 +1,73 @@
...
+      for (const releaseComment of release.get('comments')) {
+        const newComment = new ResponseModel({
+          user: toEntity('user', releaseComment.user),
```
a3957273, line 15: The age old argument of "Should this be an entity or not".  In this case, it's _always_ going to be a user, so maybe it should just use `releaseComment.user`, instead of an entity.

backend/src/migrations/004_comment_response_refactor.ts
```diff
@@ -0,0 +1,73 @@
...
+          kind: ResponseKind.Comment,
+          comment: releaseComment.message,
+          createdAt: releaseComment.createdAt,
```
a3957273, line 18: You should also likely add `updatedAt`, which should be set to the `createdAt` value.

backend/src/migrations/004_comment_response_refactor.ts
a3957273: Oh, perhaps because it's not actually on the Mongoose schema anymore.

backend/src/migrations/004_comment_response_refactor.ts
```diff
@@ -0,0 +1,73 @@
...
+        })
+        await newComment.save()
+        release.commentIds.push(newComment._id)
+        release.set('comments', undefined, { strict: false })
```
a3957273, line 22: This is a fun way of doing this.  I always set it and then have to run `.markModified`.  `.set()` appears to do both!

backend/src/migrations/004_comment_response_refactor.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  // Resolve access request comments
+  const accessRequests: any = await AccessRequestModel.find({})
+  for (const accessRequest of accessRequests) {
+    accessRequest.commentIds = []
+    if (accessRequest.get('comments') !== undefined) {
+      for (const accessCommment of accessRequest.get('comments')) {
+        const newComment = new ResponseModel({
+          user: toEntity('user', accessCommment.user),
```
a3957273, line 35: Identical comments to above.

backend/src/migrations/004_comment_response_refactor.ts
```diff
@@ -0,0 +1,73 @@
...
+          kind: ResponseKind.Comment,
```
a3957273, line 36: It might be useful to have whether it's a comment from an 'access request' or a 'release' on this model.  At the moment it's tricky to go from a comment to the object it's placed on, which might be a useful operation.  I'm less sure on whether this is a good or bad idea, so will raise it with the rest of the team.

backend/src/migrations/004_comment_response_refactor.ts
```diff
@@ -0,0 +1,73 @@
...
+          comment: accessCommment.message,
+          createdAt: accessCommment.createdAt,
+        })
+        await newComment.save()
+        accessRequest.commentIds.push(newComment._id)
+        accessRequest.set('comments', undefined, { strict: false })
```
a3957273, line 42: This is a dangerous operation.  Could we comment out this line initially, then add it in as a later migration once we're confident that the prior migration worked as expected?

a3957273 (changes requested):

backend/src/migrations/004_comment_response_refactor.ts
ARADDCC002: I've done the same for releases and also for reviews where I previously did `delete review.comments`

backend/src/migrations/004_comment_response_refactor.ts
ARADDCC002: I'm conflicted - I was trying to keep it simple by not having this two way relationship between the two models, and just having releases/access requests store a list of commentIds. 

Can you think of a useful use case of where we'd want to go to the release or access request from the comment itself? I think from a UI perspective we don't have comments on their own page anywhere, they are only displayed on the release or access request page. 

We also might want to add comments to model cards or other places in the future which would increase complexity if we were to go with your approach, whereas it would work "out of the box" in its current format.

I'm happy to be persuaded to change if the rest of the team think it's useful, though. :)

backend/src/services/smtp/smtp.ts
```diff
@@ -95,13 +96,19 @@ export async function notifyReviewResponseForRelease(review: ReviewDoc, release:
     log.info('Not sending email due to SMTP disabled')
     return
   }
-  const reviewResponse = review.responses[0]
+
+  const reviewResponse = await findResponseById(review.responseIds[0])
```
a3957273, line 100: I don't think this is accurate anymore.  With `comments` this might not represent the latest decision.  I think this _might_ need to be the last item in the array.

backend/src/services/smtp/smtp.ts
```diff
@@ -124,12 +131,17 @@ export async function notifyReviewResponseForAccess(review: ReviewDoc, accessReq
     log.info('Not sending email due to SMTP disabled')
     return
   }
-  const reviewResponse = review.responses[0]
+  const reviewResponse = await findResponseById(review.responseIds[0])
```
a3957273, line 134: Again, I think the `0` here is inaccurate.

backend/src/services/specification.ts
```diff
@@ -131,13 +131,18 @@ export const reviewInterfaceSchema = z.object({
...
-  ),
+  responses: z.array(z.string().optional()),
+
+  createdAt: z.string().openapi({ example: new Date().toISOString() }),
+  updatedAt: z.string().openapi({ example: new Date().toISOString() }),
+})
+
+export const responseInterfaceSchema = z.object({
+  user: z.string().optional().openapi({ example: 'Joe Bloggs' }),
+  kind: z.nativeEnum(ResponseKind).openapi({ example: 'comment' }),
```
a3957273, line 142: Use the enum values here, don't hard code them!

a3957273 (changes requested):

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -53,34 +47,51 @@ export default function ReviewComments({ release, accessRequest, isEdit }: Revie
...
   })
+  const { responses, isResponsesLoading, isResponsesError, mutateResponses } = useGetResponses([
+    release ? release._id : accessRequest._id,
+    ...reviews.map((review) => review._id),
+  ])
 
   const hasResponseOrComment = useMemo(() => {
-    const hasReviewResponse = !!reviews.find((review) => review.responses.length > 0)
-    const hasComment = release ? release.comments.length > 0 : accessRequest.comments.length > 0
+    const hasReviewResponse = responses.filter((response) => response.kind === ResponseKind.Review).length > 0
```
ARADDCC012, line 56: To avoid having to iterate over the entire array you could do:
```
const hasReviewResponse = !!responses.find((response) => response.kind === ResponseKind.Review)
```

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -53,34 +47,51 @@ export default function ReviewComments({ release, accessRequest, isEdit }: Revie
...
+    const hasComment = responses.filter((response) => response.kind === ResponseKind.Comment).length > 0
```
ARADDCC012, line 57: To avoid having to iterate over the entire array you could do:
```
const hasComment = !!responses.find((response) => response.kind === ResponseKind.Comment)
```

frontend/src/reviews/ReviewDecisionDisplay.tsx
```diff
@@ -31,8 +31,10 @@ export default function ReviewDecisionDisplay({ response, modelId }: ReviewDecis
   return (
     <>
       {isModelRolesLoading && <Loading />}
-      <Stack direction='row' spacing={2} alignItems='center'>
-        <UserAvatar entity={{ kind: EntityKind.USER, id: username }} size='chip' />{' '}
+      <Stack direction='row' spacing={2} alignItems='flex-start'>
+        <div style={{ marginTop: 5 }}>
```
ARADDCC012, line 35: We should use theme spacing here `<Box mt={1}>`

frontend/src/reviews/ReviewDecisionDisplay.tsx
```diff
@@ -63,13 +65,15 @@ export default function ReviewDecisionDisplay({ response, modelId }: ReviewDecis
                   </Typography>
                 )}
               </Stack>
-              <Typography variant='caption'>as {getRoleDisplay(response.role, modelRoles)}</Typography>
+              <Typography variant='caption'>
+                as {getRoleDisplay(response.role || 'Error fetching model role', modelRoles)}
```
ARADDCC012, line 69: This is an unusual way for us to handle errors. Why does the "get response" endpoint not error in this scenario? If it did then we could handle the error as we normally would.

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -75,33 +76,38 @@ export default function SideNavigation({
...
-  )
-
   useEffect(() => {
     async function fetchReviewCount() {
       onResetErrorMessage()
       if (reviews) {
-        setReviewCount(reviews.filter((filteredReview) => doesNotContainUserResponse(filteredReview)).length)
+        setReviewCount(
+          reviews.filter(
+            (filteredReview) =>
```
ARADDCC012, line 89: Super picky, but the review we're iterating over hasn't actually been filtered at this point. So `filteredReview` is a bit of a confusing name. Consider renaming to `review`

frontend/types/types.ts
```diff
@@ -65,27 +65,34 @@ export interface FileInterface {
...
+export type ResponseKindKeys = (typeof ResponseKind)[keyof typeof ResponseKind]
 
-export type ReviewResponseKind = ReviewComment | ReviewResponse
+export interface ResponseInterface {
+  entity: string
+  kind: ResponseKindKeys
+  parentId: string
+  outdated?: boolean
+  decision?: DecisionKeys
+  comment?: string
```
ARADDCC012, line 80: `comment` is a little confusing as a property name now considering the response can itself be a comment or a review. Consider renaming to something like `message` or `text`?

frontend/utils/test/testModels.ts
```diff
@@ -21,11 +21,11 @@ const accessRequestSchemaId = 'my-request-schema'
 const modelcardSchemaId = 'my-model-schema'
 
 export const testAccessRequest: AccessRequestInterface = {
+  _id: '123125123',
```
ARADDCC012, line 24: Not super important, but could we make this more readable by giving it a value like `testAccessRequestId`. This would make it easier to see how the test data is linked to one another.

ARADDCC012 (changes requested):

frontend/src/reviews/ReviewDecisionDisplay.tsx
ARADDCC002: It's not a problem getting the response object but getting the role. In our typing `role` is optional because regular comments do not have roles, only review responses.

ARADDCC002: In theory the `role` property will always be there so we _could_ do `getRoleDisplay(response.role as string, modelRoles)` but I'm not keen on this approach?

ARADDCC012: I see. I think the best approach is to do the following then:
```
{response.role && (
    <Typography variant='caption'>as {getRoleDisplay(response.role, modelRoles)}</Typography>
)}
```

ARADDCC012 (approved):

PR #1282, opened 2024-05-22
Bai 1247 create a basic read only model

TT38665:


backend/src/services/model.ts
```diff
@@ -29,6 +29,10 @@ export async function createModel(user: UserInterface, modelParams: CreateModelP
         roles: ['owner'],
       },
     ],
+    settings: {
+      ungovernedAccess: false,
+      mirroredModelId: 'demo1-7zwbdq',
```
JR40159, line 34: Something you've added for testing?

backend/src/services/model.ts
```diff
@@ -144,12 +148,18 @@ export async function getModelCard(
   version: number | GetModelCardVersionOptionsKeys,
 ) {
   if (version === GetModelCardVersionOptions.Latest) {
-    const card = (await getModelById(user, modelId)).card
+    const model = await getModelById(user, modelId)
```
JR40159, line 151: The changes in this function are no longer needed as we're not using authorisation anymore

backend/src/services/model.ts
```diff
@@ -222,6 +230,11 @@ export async function _setModelCard(
   }
 
   const revision = new ModelCardRevisionModel({ ...newDocument, modelId, createdBy: user.dn })
+  const auth = await authorisation.model(user, model, ModelAction.Write)
```
JR40159, line 233: No need to update authorisation here as we're no longer using it to check mirrored models.

backend/src/services/model.ts
```diff
@@ -235,10 +248,8 @@ export async function updateModelCard(
   metadata: unknown,
 ): Promise<ModelCardRevisionDoc> {
   const model = await getModelById(user, modelId)
-
-  const auth = await authorisation.model(user, model, ModelAction.Update)
-  if (!auth.success) {
-    throw Forbidden(auth.info, { userDn: user.dn, modelId })
+  if (model.settings.mirror.sourceModelId !== '' || model.settings.mirror.destinationModelId !== '') {
```
JR40159, line 251: This check can be the same are the other ones- if the source ID has been set, don't allow it. No need to take destination ID into account

backend/src/services/model.ts
```diff
@@ -268,6 +279,9 @@ export type UpdateModelParams = Pick<
 >
 export async function updateModel(user: UserInterface, modelId: string, diff: Partial<UpdateModelParams>) {
   const model = await getModelById(user, modelId)
+  if (model.settings.mirror.sourceModelId !== '' || model.settings.mirror.destinationModelId !== '') {
```
JR40159, line 282: You should be able to update a model regardless of whether it is a mirrored model or it is going to be copied to a mirrored model. Instead, you stop any update that is trying to set the source model ID (trying to change an existing model to be a mirrored model) and stop a destination model ID being set if the source model ID is already set (stop a mirrored model being mirrored somewhere else)

JR40159 (changes requested):

backend/src/services/model.ts
```diff
@@ -14,7 +14,7 @@ import { BadReq, Forbidden, NotFound } from '../utils/error.js'
 import { convertStringToId } from '../utils/id.js'
 import { findSchemaById } from './schema.js'
 
-export type CreateModelParams = Pick<ModelInterface, 'name' | 'kind' | 'teamId' | 'description' | 'visibility'>
+export type CreateModelParams = Pick<ModelInterface, 'name' | 'teamId' | 'description' | 'visibility' | 'settings'>
```
JR40159, line 17: You will want to check that when a user creates a model, they can't set both source and destination model ID simultaneously

backend/src/services/model.ts
JR40159, line 1: You will probably want to add the mirrored model check to `createModelCardFromSchema`

backend/src/services/release.ts
JR40159, line 1: Missing mirrored model check on: `newReleaseComment` , `updateReleaseComment`

JR40159 (changes requested):

backend/src/services/model.ts
```diff
@@ -268,6 +273,25 @@ export type UpdateModelParams = Pick<
 >
 export async function updateModel(user: UserInterface, modelId: string, diff: Partial<UpdateModelParams>) {
   const model = await getModelById(user, modelId)
+  if (diff.settings && diff.settings.mirror && diff.settings.mirror.sourceModelId) {
+    throw BadReq('Cannot change mirror model to a standard model.')
```
JR40159, line 277: Other way around- Cannot change standard model to be a mirrored model.

backend/src/services/model.ts
```diff
@@ -268,6 +273,25 @@ export type UpdateModelParams = Pick<
...
+  }
+  if (
+    model.settings.mirror.sourceModelId &&
+    diff.settings &&
+    diff.settings.mirror &&
+    diff.settings.mirror.destinationModelId
+  ) {
+    throw BadReq('Cannot change settings for this model.')
```
JR40159, line 285: 'Cannot set a destination model ID for a mirrored model.'

JR40159 (changes requested):

backend/src/routes/v2/model/patchModel.ts
```diff
@@ -18,7 +18,10 @@ export const patchModelSchema = z.object({
     settings: z
       .object({
         ungovernedAccess: z.boolean().optional().default(false).openapi({ example: true }),
-        mirroredModelId: z.string().optional().openapi({ example: 'yolo-v4-abcdef' }),
+        mirror: z.object({
+          sourceModelId: z.string().openapi({ example: 'yolo-v4-abcdef' }),
```
JR40159, line 22: Make optional

backend/src/routes/v2/model/patchModel.ts
```diff
@@ -18,7 +18,10 @@ export const patchModelSchema = z.object({
...
+          destinationModelId: z.string().openapi({ example: 'yolo-v4-abcdef' }),
```
JR40159, line 23: Make optional

JR40159 (changes requested):

backend/src/services/model.ts
```diff
@@ -32,6 +32,9 @@ export async function createModel(user: UserInterface, modelParams: CreateModelP
   })
 
   const auth = await authorisation.model(user, model, ModelAction.Create)
+  if (model.settings.mirror.sourceModelId && model.settings.mirror.destinationModelId) {
+    throw BadReq('You cannot select both settings simultaneously.')
```
JR40159, line 36: "You cannot select both mirror settings simultaneously."

backend/src/services/model.ts
```diff
@@ -268,6 +273,25 @@ export type UpdateModelParams = Pick<
...
+  ) {
+    throw BadReq('Cannot change settings for this model.')
+  }
+  if (
+    diff.settings &&
+    diff.settings.mirror &&
+    diff.settings.mirror.destinationModelId &&
+    diff.settings.mirror.sourceModelId
+  ) {
+    throw BadReq('Cannot select both settings for this model simultaneously.')
```
JR40159, line 293: "You cannot select both mirror settings simultaneously."

JR40159 (changes requested):

backend/src/routes/v2/model/postModel.ts
```diff
@@ -19,8 +19,13 @@ export const postModelSchema = z.object({
     settings: z
       .object({
         ungovernedAccess: z.boolean().optional().default(false).openapi({ example: true }),
+        mirror: z.object({
+          sourceModelId: z.string().openapi({ example: 'yolo-v4-abcdef' }),
```
JR40159, line 23: Make optional

backend/src/routes/v2/model/postModel.ts
```diff
@@ -19,8 +19,13 @@ export const postModelSchema = z.object({
...
+          destinationModelId: z.string().openapi({ example: 'yolo-v4-abcdef' }),
```
JR40159, line 24: Make optional

backend/src/services/model.ts
```diff
@@ -262,19 +275,33 @@ export async function updateModelCard(
...
-export type UpdateModelParams = Pick<
-  ModelInterface,
-  'name' | 'description' | 'visibility' | 'collaborators' | 'settings'
->
-export async function updateModel(user: UserInterface, modelId: string, diff: Partial<UpdateModelParams>) {
+export type UpdateModelParams = Pick<ModelInterface, 'name' | 'description' | 'visibility' | 'collaborators'>
+export async function updateModel(
+  user: UserInterface,
+  modelId: string,
+  modelDiff: Partial<UpdateModelParams>,
```
a3957273, line 282: Why not include settings in the model?  What makes settings special?

a3957273: Ah.  Presumably because we only use `Object.assign()` below, which isn't recursive.  I would prefer a recursive merging here, as opposed to treating 'settings' as a special case.  I think something like:

```
.mergeWith({}, model, modelDiff, (a, b) => 
  _.isArray(b) ? b : undefined
);
```

Might work.

backend/src/services/model.ts
```diff
@@ -32,10 +32,21 @@ export async function createModel(user: UserInterface, modelParams: CreateModelP
...
 
   const auth = await authorisation.model(user, model, ModelAction.Create)
+
   if (!auth.success) {
     throw Forbidden(auth.info, { userDn: user.dn })
   }
 
+  if (
+    (model.settings.mirror.sourceModelId && model.settings.mirror.destinationModelId) ||
+    (modelParams.settings &&
```
a3957273, line 42: This is equivalent to `modelParams?.settings?.mirror?.destinationModelId && modelParams?.settings?.mirror?.sourceModelId`.

Also, I think checking _both_ sets here are unnecessarily because the `modelParams` are directly used to fill in the `model` value.

Also, we should split this out into a separate function and use it on both `create` and `update` events to reduce duplication.

a3957273 (changes requested):

JR40159 (approved):

a3957273 (approved):

PR #1284, opened 2024-05-30
BAI-1246 Python Logging

bw27492:


lib/python/src/bailo/helper/model.py
```diff
@@ -182,6 +194,10 @@ def get_latest_release(self):
         releases = self.get_releases()
         if releases == []:
             raise BailoException("This model has no releases.")
+
+        latest_release = max(releases)
+        logger.info(f"Latest release ({str(latest_release.version)}) for {self.model_id} retrieved successfully.")
+
         return max(releases)
```
GB27247, line 201: Could we change this to `latest_release`?

lib/python/src/bailo/helper/model.py
```diff
@@ -319,6 +339,14 @@ def from_mlflow(self, tracking_uri: str, experiment_id: str):
         if ml_flow:
             client = mlflow.tracking.MlflowClient(tracking_uri=tracking_uri)
             runs = client.search_runs(experiment_id)
+            if len(runs) > 0:
```
GB27247, line 342: We can get rid of `>0` It doesn't have any effect on this condition.

lib/python/src/bailo/helper/release.py
```diff
@@ -204,6 +217,9 @@ def download_all(self, path: str = os.getcwd(), include: list | str = None, excl
                 file for file in file_names if not any([fnmatch.fnmatch(file, pattern) for pattern in exclude])
             ]
 
+        logger.info(
+            f"Downloading {len(file_names)} of {len(orig_file_names)} files for version {str(self.version)} of {self.model_id}..."
+        )
```
GB27247, line 222: We can use f-strings with this although logging has the intended behaviour of formatting additional arguments into the log. We can use something like this. This applies to all the other places this is used.

```python
logger.info(
    f"Downloading %d of %d files for version %s of %s", len(file_names), len(orig_file_names), str(self.version), self.model_id
)
```

lib/python/src/bailo/core/agent.py
```diff
@@ -106,18 +115,27 @@ def __init__(
...
         super().__init__()
+        logger.info("TokenAgent initiated successfully.")
 
         if access_key is None:
+            logger.error("Access key not provided. Trying other sources...")
             try:
                 access_key = os.environ["BAILO_ACCESS_KEY"]
+                logger.info("Access key acquired from BAILO_ACCESS_KEY environment variable.")
             except KeyError:
+                logger.error("Access key not found in BAILO_ACCESS_KEY environment variable. Requires user input.")
```
GB27247, line 126: Normally logging errors are given when something is unrecoverable within the application. I think info should be used here instead

lib/python/src/bailo/core/agent.py
```diff
@@ -25,9 +29,12 @@ def __init__(
         """
         self.verify = verify
 
+        logger.debug("Base Agent initiated successfully.")
+
     def __request(self, method, *args, **kwargs):
         kwargs["verify"] = self.verify
 
+        logger.debug(f"{method} to {args[0]}")
```
GB27247, line 37: I don't think this is needed. Requests provide there own logging that does the same thing but with a bit more detail in terms of sessions and ports.

GB27247 (changes requested):

lib/python/src/bailo/core/agent.py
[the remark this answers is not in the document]
bw27492: Done

lib/python/src/bailo/core/agent.py
```diff
@@ -25,9 +29,12 @@ def __init__(
         """
         self.verify = verify
 
+        logger.debug("Base Agent initiated successfully.")
```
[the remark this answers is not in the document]
bw27492: Done - object creation logs removed. Other debug messages changed to info

lib/python/src/bailo/helper/model.py
bw27492: Done

lib/python/src/bailo/helper/model.py
bw27492: Done

lib/python/src/bailo/core/agent.py
bw27492: Done

lib/python/src/bailo/core/agent.py
bw27492: Done

lib/python/src/bailo/helper/release.py
bw27492: Done

GB27247 (changes requested):

lib/python/src/bailo/helper/release.py
```diff
@@ -234,12 +239,12 @@ def upload(self, path: str, data: BytesIO | None = None) -> str:
...
         ..note:: If path provided is a directory, it will be uploaded as a zip
         """
-        logger.info(f"Uploading file(s) to version {str(self.version)} of {self.model_id}...")
+        logger.info(f"Uploading file(s) to version %s of %s...", str(self.version), self.model_id)
         name = os.path.split(path)[-1]
 
         if data is None:
             if is_zip := os.path.isdir(path):
-                logger.warning(f"Given path ({path}) is a directory. This will be converted to a zip file for upload.")
+                warnings.warn(f"Given path ({path}) is a directory. This will be converted to a zip file for upload.")
```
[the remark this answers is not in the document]
bw27492: Done

lib/python/src/bailo/helper/model.py
```diff
@@ -319,6 +336,16 @@ def from_mlflow(self, tracking_uri: str, experiment_id: str):
...
+            if len(runs):
+                logger.info(
+                    f"Successfully retrieved MLFlow experiment %s from tracking server. %d were found.",
+                    experiment_id,
+                    len(runs),
+                )
+            else:
+                warnings.warn(
+                    f"MLFlow experiment {experiment_id} does not have any runs and publishing requires at least one valid run. Are you sure the ID is correct?"
+                )
```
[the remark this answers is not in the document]
bw27492: This is not expected behaviour - MLFlow does not return an error if you request an invalid experiment, instead it will send an empty list. Hence this warning

lib/python/src/bailo/__init__.py
```diff
@@ -19,3 +19,9 @@
 from bailo.helper.model import Experiment, Model
 from bailo.helper.release import Release
 from bailo.helper.schema import Schema
+
+
+logging.basicConfig()
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+logger.propagate = False
```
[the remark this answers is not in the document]
bw27492: Done

GB27247 (approved):

PR #1304, opened 2024-06-03
added a new common component for dialog components on button press. a…

ARADDCC002:
image

image


frontend/src/entry/model/settings/ModelRolesInfo.tsx
```diff
@@ -0,0 +1,76 @@
...
+              </Stack>
+            </Grid>
+            <Grid item sm={6}>
+              <Stack spacing={1}>
+                <Box>
+                  <Typography component='h3' variant='h6' fontWeight='bold'>
+                    Dynamic Roles
+                  </Typography>
+                  <Typography variant='caption'>
+                    The following roles are specified by this model&apos;s schema
```
TT38665, line 65: Is this 'model&apos;s' meant to be in the text?

TT38665, line 65: Why use 'model&apos;s' and not 'model's'?

ARADDCC012: Agreed, I think using the following is easier to read
```
`The following roles are specified by this model's schema`
```

frontend/src/common/HelpDialog.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+function HelpDialog({ infoLabel, dialogTitle, content, icon }: Props) {
+  const [open, setOpen] = React.useState(false)
+
+  return (
+    <>
+      <Button size='small' onClick={() => setOpen(true)} endIcon={icon ? icon : <HelpOutlineIcon />}>
+        <Typography variant='caption'>{infoLabel}</Typography>
+      </Button>
+      <Dialog open={open} onClose={() => setOpen(false)} maxWidth='md' TransitionComponent={Transition}>
```
ARADDCC012, line 30: It feels a bit out of place to be using a transition when we don't for any other dialog. I believe we should be consistent and either use a transition for all of them or none of them.

ARADDCC012, line 30: All of our other dialogs prevent closing via the escape key or from clicking the background. Instead we include a close button at the bottom. Consider making this dialog work the same for consistency.

frontend/src/common/HelpDialog.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <DialogTitle>{dialogTitle}</DialogTitle>
+        <DialogContent>{content}</DialogContent>
+      </Dialog>
+    </>
+  )
+}
+export default HelpDialog
```
ARADDCC012, line 37: Minor, but can we move this to the line where we define the component in the same way that we usually do?
```
export default function HelpDialog(...
```

frontend/src/common/HelpDialog.tsx
```diff
@@ -0,0 +1,37 @@
...
+  props: TransitionProps & {
+    children: React.ReactElement<any, any>
+  },
+  ref: React.Ref<unknown>,
+) {
+  return <Slide direction='up' ref={ref} {...props} />
+})
+
+type Props = {
+  infoLabel: string
```
ARADDCC012, line 16: Consider renaming to `label`

frontend/src/entry/model/settings/ModelRolesInfo.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { RoleKind } from 'types/types'
+
+interface ModelRolesInfoProps {
+  modelId: string
+}
+
+export default function ModelRolesInfo({ modelId }: ModelRolesInfoProps) {
+  const { modelRoles, isModelRolesLoading, isModelRolesError } = useGetModelRoles(modelId)
+
+  const roles = useCallback(
```
ARADDCC012, line 16: `useCallback` isn't providing any performance gains here. As it stands we'll still be running the internal logic on every render cycle because `roles()` is being called in your JSX block below.

I would `useMemo` for each `kind` and use those stable values in your JSX below. You can also use a reduce so that you don't have to loop over `modelRoles` multiple times. I've made the following generic for entries:
```
  const getFilteredRoles = useCallback(
    (roleKind: string) =>
      entryRoles.reduce<ReactNode[]>((filteredRoles, entryRole) => {
        if (entryRole.kind === roleKind) {
          filteredRoles.push(
            <Box key={entryRole.id}>
              <Typography fontWeight='bold'>{entryRole.name}</Typography>
              <Typography>{entryRole.description}</Typography>
            </Box>,
          )
        }
        return filteredRoles
      }, []),
    [entryRoles],
  )

  const modelRolesList = useMemo(() => getFilteredRoles(RoleKind.MODEL), [getFilteredRoles])
  const schemaRolesList = useMemo(() => getFilteredRoles(RoleKind.SCHEMA), [getFilteredRoles])
```

frontend/src/entry/model/settings/ModelRolesInfo.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <Stack spacing={2}>
+          <Typography>
+            Roles in Bailo are split into two categories; standard and dynamic. Standard roles are generic across
+            different schema and are used for determining model permissions for general purpose model upkeep, whereas
+            dynamic roles are created on a per schema basis and used as part of the review process. The dynamic roles
+            presented below are specified on the schema selected for this model and may not apply to other models using
+            a different schema.
+          </Typography>
+          <Grid container spacing={1}>
+            <Grid item sm={6}>
```
ARADDCC012, line 47: This only handles screen sizes of `sm` and above. Change `sm` for `xs` and it'll handle all of them.

frontend/src/entry/model/settings/ModelRolesInfo.tsx
```diff
@@ -0,0 +1,76 @@
...
+                <Box>
+                  <Typography component='h3' variant='h6' fontWeight='bold'>
+                    Standard Roles
+                  </Typography>
+                  <Typography variant='caption'>The following roles are generic across all models</Typography>
+                </Box>
+                {roles(RoleKind.MODEL)}
+              </Stack>
+            </Grid>
+            <Grid item sm={6}>
```
ARADDCC012, line 58: Change `sm` for `xs` to cover every screen size.

frontend/src/entry/model/settings/ModelRolesInfo.tsx
ARADDCC012, line 1: `ModelAccess.tsx` has been updated to be generic for all entry kinds (it's also been renamed to `EntryAccess.tsx` so you'll need to update your import name too). You'll need to update this component to also be generic. For all other generic entry components I've done the following:
- Replace "model" in variable names to be "entry".
- Add a `entryKind` prop if needed (will be needed in this case).
- Replace and user facing text containing "model" to use the entry kind along with `toTitleCase()` or `toSentenceCase().

backend/src/types/types.ts
```diff
@@ -2,10 +2,19 @@ export interface User {
   dn: string
 }
 
+export const RoleKind = {
+  MODEL: 'model',
```
ARADDCC012, line 6: What about data cards? Should this be `entry` so it can be used for both or do we need a third kind for data cards?

frontend/src/entry/model/settings/ModelRolesInfo.tsx
```diff
@@ -0,0 +1,76 @@
...
+    return <MessageAlert message={isModelRolesError.info.message} severity='error' />
+  }
+
+  return (
+    <>
+      {isModelRolesLoading && <Loading />}
+      {!isModelRolesLoading && (
+        <Stack spacing={2}>
+          <Typography>
+            Roles in Bailo are split into two categories; standard and dynamic. Standard roles are generic across
```
ARADDCC012, line 40: This block of text may need to be genericised to use the entry kind, instead of statically referring to "model". Either that or we need to conditionally show a model/data card specific block of text.

ARADDCC012 (changes requested):

frontend/src/common/HelpDialog.tsx
ARADDCC002: I feel the same but also wanted to do all the others outside of this ticket. I'll remove it from this PR and make a new ticket to add transitions to all other dialogs because I do quite like it!

ARADDCC002: I can see in some places we disable closing via the escape key but do we really disable by clicking outside? I don't think that's the case. There doesn't _seem_ to be much consistency in this. I can disable closing by the escape key for now, though.

frontend/src/common/HelpDialog.tsx
ARADDCC002: Not sure why I did it this way!

frontend/src/common/HelpDialog.tsx
ARADDCC002: Removed as we no longer need it due to using the IconButton component instead of a regular Button

frontend/src/entry/model/settings/ModelRolesInfo.tsx
ARADDCC002: I did this because I wanted `xs` to be set to `12` so it works on _really_ small screens, I just forgot to add `xs` whoops!!

frontend/src/entry/model/settings/ModelRolesInfo.tsx
ARADDCC002: See above

frontend/src/common/HelpDialog.tsx
ARADDCC012: Never mind, we're not consistent at all with the onClose behaviour. Happy with you leaving as is if you prefer.

frontend/src/entry/settings/EntryAccess.tsx
```diff
@@ -146,11 +146,7 @@ export default function EntryAccess({ entry }: EntryAccessProps) {
             <TableCell>
               <Stack direction='row' spacing={1} alignItems='center'>
                 <span>Roles</span>
-                <HelpDialog
-                  infoLabel='Roles explained'
-                  dialogTitle='Roles Explained'
-                  content={<ModelRolesInfo modelId={model.id} />}
-                />
+                <HelpDialog dialogTitle='Roles Explained' content={<EntryRolesInfo modelId={entry.id} />} />
```
ARADDCC012, line 149: Consider renaming the title to something like `What are roles?`

frontend/src/common/HelpDialog.tsx
```diff
@@ -1,37 +1,29 @@
...
+export default function HelpDialog({ dialogTitle, content }: HelpDialogProps) {
+  const [open, setOpen] = useState(false)
 
   return (
     <>
-      <Button size='small' onClick={() => setOpen(true)} endIcon={icon ? icon : <HelpOutlineIcon />}>
-        <Typography variant='caption'>{infoLabel}</Typography>
-      </Button>
-      <Dialog open={open} onClose={() => setOpen(false)} maxWidth='md' TransitionComponent={Transition}>
+      <IconButton size='small' onClick={() => setOpen(true)}>
```
ARADDCC012, line 15: I think this got lost in one of my other comments, but I think this would benefit from a tooltip. We could simply repurpose the `dialogTitle`, rename it to `title` and use that title for the tooltip and dialog.
```
  return (
    <>
      <Tooltip title={title}>
        <IconButton size='small' onClick={() => setOpen(true)}>
          <HelpOutlineIcon />
        </IconButton>
      </Tooltip>
      <Dialog open={open} onClose={() => setOpen(false)} maxWidth='md' disableEscapeKeyDown>
        <DialogTitle>{title}</DialogTitle>
        <DialogContent>{content}</DialogContent>
        <DialogActions>
          <Button color='secondary' variant='outlined' onClick={() => setOpen(false)}>
            Close
          </Button>
        </DialogActions>
      </Dialog>
    </>
  )
```

ARADDCC012 (changes requested):

PR #1310, opened 2024-06-06
Bugfix bai 1311 collaborator role cannot change settings

TT38665:


backend/src/connectors/authorisation/base.ts
```diff
@@ -98,9 +98,9 @@ export class BasicAuthorisationConnector {
         // Check a user has a role before allowing write actions
         if (
           [ModelAction.Write, ModelAction.Update].some((a) => a === action) &&
-          (await missingRequiredRole(user, model, ['owner', 'collaborator']))
+          (await missingRequiredRole(user, model, ['owner']))
         ) {
-          return { id: model.id, success: false, info: 'Only owners and collaborators can update a model.' }
+          return { id: model.id, success: false, info: 'Only the owner can update a model.' }
```
ARADDCC002, line 103: I would keep owner as plural because there can be multiple owners

ARADDCC002 (changes requested):

ARADDCC002 (approved):

PR #1314, opened 2024-06-10
Bai 1312 link to python client docs

TT38665:


frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -151,6 +152,15 @@ export default function SideNavigation({
               icon={<ContactSupportIcon />}
             />
             <Divider />
+            <NavMenuItem
```
a3957273, line 155: Could this be put just below the 'API' menu item?  I like the similarities between the 'API' and 'Python Client' docs.

a3957273 (changes requested):

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -151,6 +152,15 @@ export default function SideNavigation({
...
+              href='/docs/python/index.html'
+              selectedPage={page}
+              primaryText='Python Client Docs'
+              drawerOpen={drawerOpen}
+              menuPage=''
```
ARADDCC012, line 160: This is required so that we can compare it to `selectedPage`. If they match then we highlight this tab as the selected tab. Some unique text like `menuPage='pythonDocs'` should be good.

ARADDCC012 (changes requested):

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -141,6 +142,16 @@ export default function SideNavigation({
...
+            <NavMenuItem
+              href='/docs/python/index.html'
+              selectedPage={page}
+              primaryText='Python Client Docs'
+              drawerOpen={drawerOpen}
+              menuPage='pythonDocs'
+              title='Python Client Docs'
+              icon={<DescriptionIcon />}
+            />
+            <Divider />
```
ARADDCC012, line 154: This divider should be moved back to where it was before, below the help and support `NavMenuItem`.

ARADDCC012 (approved):

a3957273 (approved):

PR #1323, opened 2024-06-12
added optional advanced section on entry creation for adding users to…

ARADDCC002:
image

image


backend/src/services/model.ts
```diff
@@ -21,21 +21,32 @@ export function checkModelRestriction(model: ModelInterface) {
...
+  // Add the creator of the entry if they are not already in the collaborator list
+  const collaborators = modelParams.collaborators.find(
+    (collaborator) => (collaborator.entity === toEntity('user', user.dn)) !== undefined,
+  )
+    ? [...modelParams.collaborators]
+    : [
+        ...modelParams.collaborators,
+        {
+          entity: toEntity('user', user.dn),
+          roles: ['owner'],
```
JR40159, line 42: Won't we always want the user who is making this request to be owner?

backend/src/services/model.ts
```diff
@@ -21,21 +21,32 @@ export function checkModelRestriction(model: ModelInterface) {
   }
 }
 
-export type CreateModelParams = Pick<ModelInterface, 'name' | 'teamId' | 'description' | 'visibility' | 'settings'>
+export type CreateModelParams = Pick<
+  ModelInterface,
+  'name' | 'teamId' | 'description' | 'visibility' | 'settings' | 'collaborators'
```
JR40159, line 26: Could you also add 'kind' to this list while you're here? I think I accidentally removed it at some point...

JR40159 (changes requested):

JR40159 (approved):

PR #1326, opened 2024-06-13
added new property in necessary files to enable templating and implem…

TT38665:
…ented migration script.

### DIFF
--- a/backend/src/models/Model.ts
+++ b/backend/src/models/Model.ts
@@ -40,6 +40,7 @@ export interface ModelCardInterface {
 
 export interface Settings {
   ungovernedAccess: boolean
+  allowTemplating: boolean
   mirror: { sourceModelId?: string; destinationModelId?: string }
 }
 
@@ -94,6 +95,7 @@ const ModelSchema = new Schema<ModelInterface>(
     ],
     settings: {
       ungovernedAccess: { type: Boolean, required: true, default: false },
+      allowTemplating: { type: Boolean, required: true, default: false },
       mirror: { sourceModelId: { type: String }, destinationModelId: { type: String } },
     },
 
--- a/backend/src/routes/v2/model/patchModel.ts
+++ b/backend/src/routes/v2/model/patchModel.ts
@@ -18,6 +18,7 @@ export const patchModelSchema = z.object({
     settings: z
       .object({
         ungovernedAccess: z.boolean().optional().openapi({ example: true }),
+        allowTemplating: z.boolean().optional().openapi({ example: true }),
         mirror: z
           .object({
             sourceModelId: z.string().optional().openapi({ example: 'yolo-v4-abcdef' }),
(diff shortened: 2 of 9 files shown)


frontend/src/entry/model/settings/AccessRequestSettings.tsx
```diff
@@ -23,6 +23,7 @@ export default function AccessRequestSettings({ model }: AccessRequestSettingsPr
     const updatedModelSettings = {
       settings: {
         ungovernedAccess: allowUngoverned,
+        allowTemplating: false,
```
ARADDCC012, line 26: As we're not going to be changing this value in `AccessRequestSettings` it makes more sense to reuse the value already on the model. So I'd update this to `allowTemplating: model.settings.allowTemplating`

ARADDCC012 (changes requested):

backend/src/models/Model.ts
```diff
@@ -40,6 +40,7 @@ export interface ModelCardInterface {
 
 export interface Settings {
   ungovernedAccess: boolean
+  allowTemplating: boolean
   mirror: { sourceModelId?: string; destinationModelId?: string }
```
ARADDCC012, line 44: Not your doing, but this property isn't defined in the frontend `EntryInterface`. Should it be?

backend/src/routes/v2/model/patchModel.ts
```diff
@@ -18,6 +18,7 @@ export const patchModelSchema = z.object({
     settings: z
       .object({
         ungovernedAccess: z.boolean().optional().openapi({ example: true }),
+        allowTemplating: z.boolean().optional().openapi({ example: true }),
```
ARADDCC012, line 21: This will also need adding to the zod schema in `backend/src/routes/v2/model/postModel.ts`

backend/src/models/Model.ts
TT38665: It's declared in the next ticket. My bad

ARADDCC012: To be clear, I'm talking about the `mirror` property here. It looks like it was added to the backend interface at some point in the past, but wasn't replicated in the frontend.

You've correctly added your new `allowTemplating` property to the frontend interface in both of your PRs, so no need to worry about that.

ARADDCC012:
You've got some tests failing. All you should need to do is add your new property to the test data in the failing files.

backend/src/models/Model.ts
JR40159: We haven't done any work on the mirrored model stuff in the frontend yet, and I'm assuming updating the frontend interface can wait until then?

frontend/src/entry/model/settings/AccessRequestSettings.tsx
ARADDCC012: Unresolved this as it's still outstanding

ARADDCC012 (approved):

PR #1327, opened 2024-06-13
Implemented template tab in model settings. May need to  looka at han…

TT38665:
…dlesave function again.

### DIFF
--- a/frontend/src/entry/model/settings/AccessRequestSettings.tsx
+++ b/frontend/src/entry/model/settings/AccessRequestSettings.tsx
@@ -23,7 +23,7 @@ export default function AccessRequestSettings({ model }: AccessRequestSettingsPr
     const updatedModelSettings = {
       settings: {
         ungovernedAccess: allowUngoverned,
-        allowTemplating: false,
+        allowTemplating: model.settings.allowTemplating,
       },
     }
 
--- a/frontend/src/entry/model/settings/TemplateSettings.tsx
+++ b/frontend/src/entry/model/settings/TemplateSettings.tsx
@@ -12,7 +12,7 @@ type TemplateSettingsProps = {
 }
 
 export default function TemplateSettings({ model }: TemplateSettingsProps) {
-  const [allowTemplate, setAllowTemplate] = useState(model.settings.allowTemplating)
+  const [allowTemplating, setAllowTemplating] = useState(model.settings.allowTemplating)
   const [loading, setLoading] = useState(false)
   const [errorMessage, setErrorMessage] = useState('')
   const sendNotification = useNotification()
@@ -21,8 +21,8 @@ export default function TemplateSettings({ model }: TemplateSettingsProps) {
     setLoading(true)
     const updatedModelSettings = {
       settings: {
-        ungovernedAccess: false,
-        allowTemplating: allowTemplate,
+        ungovernedAccess: model.settings.ungovernedAccess,
+        allowTemplating: allowTemplating,
       },
     }
 
@@ -33,7 +33,7 @@ export default function TemplateSettings({ model }: TemplateSettingsProps) {
     } else {
       sendNotification({
         variant: 'success',
-        msg: 'Access request settings updated',
+        msg: 'Template settings updated',
         anchorOrigin: { horizontal: 'center', vertical: 'bottom' },
       })
     }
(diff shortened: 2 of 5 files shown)


frontend/src/entry/model/settings/AccessRequestSettings.tsx
```diff
@@ -23,6 +23,7 @@ export default function AccessRequestSettings({ model }: AccessRequestSettingsPr
     const updatedModelSettings = {
       settings: {
         ungovernedAccess: allowUngoverned,
+        allowTemplating: false,
```
ARADDCC012, line 26: Same comment I put in your other PR, but repeating it here for full coverage:

As we're not going to be changing this value in `AccessRequestSettings` it makes more sense to reuse the value already on the model. So I'd update this to `allowTemplating: model.settings.allowTemplating`

frontend/src/entry/model/settings/TemplateSettings.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const [allowTemplate, setAllowTemplate] = useState(model.settings.allowTemplating)
+  const [loading, setLoading] = useState(false)
+  const [errorMessage, setErrorMessage] = useState('')
+  const sendNotification = useNotification()
+
+  async function handleSave() {
+    setLoading(true)
+    const updatedModelSettings = {
+      settings: {
+        ungovernedAccess: false,
```
ARADDCC012, line 24: This should be reusing the value already stored on the model, otherwise we'll be overwriting it to `false` whenever they change the `allowTemplating` setting. Update this to `ungovernedAccess: model.settings.ungovernedAccess`

frontend/src/entry/model/settings/TemplateSettings.tsx
```diff
@@ -0,0 +1,63 @@
...
+import MessageAlert from 'src/MessageAlert'
+import { EntryInterface } from 'types/types'
+import { getErrorMessage } from 'utils/fetcher'
+
+type TemplateSettingsProps = {
+  model: EntryInterface
+}
+
+export default function TemplateSettings({ model }: TemplateSettingsProps) {
+  const [allowTemplate, setAllowTemplate] = useState(model.settings.allowTemplating)
```
ARADDCC012, line 15: Minor, but could this be renamed to `allowTemplating` and `setAllowTemplating` so that we don't have two slightly different names for the same property?

frontend/src/entry/model/settings/TemplateSettings.tsx
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    const response = await patchModel(model.id, updatedModelSettings)
+
+    if (!response.ok) {
+      setErrorMessage(await getErrorMessage(response))
+    } else {
+      sendNotification({
+        variant: 'success',
+        msg: 'Access request settings updated',
```
ARADDCC012, line 36: This message isn't correct. Should be something like `'Template settings updated'`

frontend/src/entry/model/settings/TemplateSettings.tsx
```diff
@@ -0,0 +1,63 @@
...
+      </Typography>
+      <FormControlLabel
+        label='Allow users to make a template'
+        control={
+          <Checkbox onChange={(event) => setAllowTemplate(event.target.checked)} checked={allowTemplate} size='small' />
+        }
+      />
+      <Divider />
+      <div>
+        <LoadingButton variant='contained' aria-label='Save templating' onClick={handleSave} loading={loading}>
```
ARADDCC012, line 56: I think this aria label could be clearer. Something like `aria-label='Save model template settings'`

frontend/src/entry/model/settings/TemplateSettings.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const [loading, setLoading] = useState(false)
+  const [errorMessage, setErrorMessage] = useState('')
+  const sendNotification = useNotification()
+
+  async function handleSave() {
+    setLoading(true)
+    const updatedModelSettings = {
+      settings: {
+        ungovernedAccess: false,
+        allowTemplating: allowTemplate,
```
ARADDCC012, line 25: If you rename `allowTemplate` to `allowTemplating` as I've suggested in another comment, this can be changed from `allowTemplating: allowTemplate,` to `allowTemplating`.

ARADDCC012 (changes requested):

frontend/src/entry/model/settings/TemplateSettings.tsx
ARADDCC012: Unresolved as this is still outstanding

ARADDCC012 (approved):

PR #1332, opened 2024-06-18
Map tokens to ui config

GB27247:


frontend/pages/settings/personal-access-tokens/new.tsx
```diff
@@ -18,79 +18,68 @@ import {
...
 
-export default function NewToken() {
   const theme = useTheme()
   const [description, setDescription] = useState('')
   const [isAllModels, setIsAllModels] = useState(true)
   const [selectedModels, setSelectedModels] = useState<EntrySearchResult[]>([])
-  const [isAllActions, setIsAllActions] = useState(true)
-  const [selectedActions, setSelectedActions] = useState<TokenActionKeys[]>(actionOptions)
+  const [isAllActions, setIsAllActions] = useState(false)
+  const [selectedActions, setSelectedActions] = useState<string[]>([])
```
ARADDCC012, line 46: Do we no longer want to default to all actions being selected?

frontend/pages/settings/personal-access-tokens/new.tsx
```diff
@@ -110,9 +99,26 @@ export default function NewToken() {
           />
         )
       }),
-    [selectedActions],
+    [selectedActions, TokenWriteAction, TokenReadAction],
   )
 
+  if (isUiConfigLoading) {
```
ARADDCC012, line 105: I'd recommend moving the loading and error handlers to just above your main return block at the bottom. Error handling should also come before loading.
```
  if (isUiConfigError) {
    return <MessageAlert message={isUiConfigError.info.message}></MessageAlert>
  }
  
  if (isUiConfigLoading) {
    return <Loading />
  }
  
  return (
     ...
   )
```

frontend/pages/settings/personal-access-tokens/new.tsx
```diff
@@ -18,79 +18,68 @@ import {
...
   const modelsAutocompletePlaceholder = useMemo(() => {
     if (isAllModels) return 'All models selected'
     if (selectedModels.length) return ''
     return 'Select models'
   }, [isAllModels, selectedModels.length])
 
-  const isGenerateButtonDisabled = useMemo(
-    () => !description || !(isAllModels || selectedModels.length) || !selectedActions.length,
-    [description, isAllModels, selectedModels.length, selectedActions.length],
+  const [TokenReadAction, TokenWriteAction] = actionOptions.reduce<Record<string, string>[]>(
```
ARADDCC012, line 57: This should probably be in a `useMemo`

frontend/pages/settings/personal-access-tokens/new.tsx
```diff
@@ -18,79 +18,68 @@ import {
...
 
-  const { models, isModelsLoading, isModelsError } = useListModels('model')
-
   const modelsAutocompletePlaceholder = useMemo(() => {
     if (isAllModels) return 'All models selected'
     if (selectedModels.length) return ''
     return 'Select models'
   }, [isAllModels, selectedModels.length])
 
-  const isGenerateButtonDisabled = useMemo(
```
ARADDCC012, line 82: I think this logic is still needed. We don't want a user to be able to create a token that doesn't work with any models/actions.

ARADDCC012 (dismissed):

frontend/pages/settings/personal-access-tokens/new.tsx
GB27247: This is pretty standard in other apps. One reason is there's a further barrier to make less restrictive tokens, but I'm happy to hear suggestions.

frontend/pages/settings/personal-access-tokens/new.tsx
```diff
@@ -18,79 +18,72 @@ import {
...
+  const { uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
 
-const getActionName = (action: TokenActionKeys) => {
-  const [name, _kind] = action.split(':')
-  return name
-}
+  const tokenActions = uiConfig?.tokenActions || {}
 
-const actionOptions = Object.values(TokenAction)
+  const actionOptions = Object.keys(tokenActions)
```
ARADDCC012, line 39: Minor, but as `uiConfig.tokenActions` is a decent size object I think it's worth `useMemo`ing this. You'll also have to useMemo `tokenActions` too to make that a stable dependency.

frontend/pages/settings/personal-access-tokens/new.tsx
ARADDCC012: I think with the improvements being made to the token action selection UI in a future PR, I think this makes sense to do. Happy for you to leave as is.

backend/src/utils/config.ts
```diff
@@ -187,4 +190,5 @@ export interface Config {
 }
 
 const config: Config = _config.util.toObject()
+config.ui.tokenActions = TokenDescriptions
```
JR40159, line 193: I think this is a clear move away from how the UI config is used and having things like this where the backend appends to the configuration that is provided by the configuration file could easily add confusion in the future. I think having to do this shows that this token information isn't suitable for the UI config. The UI config contains settings that can differ per deployment and need to be shared between the frontend and the backend. The token available actions and their descriptions fulfil the latter but not the former. Instead I'd suggest a simple token endpoint that returns an object of Bailo's available token actions with their descriptions so clients like the UI can display that information to users

backend/src/models/Token.ts
```diff
@@ -29,15 +29,32 @@ export const TokenActions = {
   ImageRead: 'image:read',
   ImageWrite: 'image:write',
 
-  SchemaRead: 'schema:read',
   SchemaWrite: 'schema:write',
+} as const
+
+export const TokenDescriptions = {
```
JR40159, line 35: I would prefer these two objects to be combined into one. 
```
{
  SchemaWrite: { 
    id: 'schema:write',
    description: 'Grants permissions to upload and modify schemas for administrators.'
  },
  ....
}
```

JR40159 (changes requested):

ARADDCC012:
https://github.com/gchq/Bailo/pull/1332#discussion_r1662388249 is outstanding

ARADDCC002 (approved):

PR #1334, opened 2024-06-19
Add initial export capability

a3957273:


backend/src/services/export.ts
JR40159, line 1: Could we update the name of this new service since we have different types of exports now?

JR40159, line 1: No tests :(

backend/src/routes/v2/model/modelcard/getModelCardHtml.ts
JR40159, line 1: No tests :(

JR40159 (changes requested):

JR40159 (approved):

ARADDCC002 (approved):

PR #1335, opened 2024-06-19
Prevent a token creating another token

a3957273:


backend/src/services/token.ts
```diff
@@ -17,6 +17,11 @@ interface CreateTokenProps {
 
 const nanoid = customAlphabet('0123456789ABCDEFGHIJKLMNOPQSRTUVWXYZ')
 export async function createToken(user: UserInterface, { description, scope, modelIds, actions }: CreateTokenProps) {
+  if (user.token) {
```
JR40159, line 20: Please could you create a quick unit test for this? :)

JR40159 (changes requested):

JR40159 (approved):

PR #1337, opened 2024-06-19
Add token forwarding to inference service login

GB27247:


infrastructure/helm/bailo/templates/bailo/bailo.configmap.yaml
```diff
@@ -191,6 +192,10 @@ data:
           registry: '{{ .Values.minio.registryBucket }}',
         },
       },
+
+      inference: {
+        tokenHeader: '{{ .Values.inference.tokenHeader }}'
```
JR40159, line 197: I'm assuming the is sensitive information and as such should be included in the `custom-environment-variables` and be populated via a secret.

backend/config/default.cjs
```diff
@@ -241,4 +242,8 @@ module.exports = {
       },
     },
   },
+
+  inference: {
+    tokenHeader: undefined,
```
JR40159, line 247: If you don't want to give an example value, I personally would prefer to put an empty string here as it is clearer what data type should be used.

JR40159 (changes requested):

backend/config/default.cjs
```diff
@@ -241,4 +242,8 @@ module.exports = {
       },
     },
   },
+
+  inference: {
+    tokenHeader: '',
```
a3957273, line 247: Could we call this something like `authorisationToken`?

backend/src/clients/inferencing.ts
```diff
@@ -10,12 +10,12 @@ interface InferenceService {
   port: number
 }
 
-export async function createInferenceService(inferenceServiceParams: InferenceService) {
+export async function createInferenceService(inferenceServiceParams: InferenceService, token: string) {
```
a3957273, line 13: Instead of passing the `token` in, we could just get it from `config` in this file instead.

backend/src/services/inference.ts
```diff
@@ -74,7 +75,13 @@ export async function createInference(user: UserInterface, modelId: string, infe
     port: inferenceParams.settings.port,
   }
 
-  await createInferenceService(inferenceService)
+  const tokenHeader = config.inference.tokenHeader
+
+  if (!tokenHeader) {
+    throw NotFound('No authentication key exists')
```
a3957273, line 81: Perhaps throw `Unauthorized` instead, as a `NotFound` responds with a 404.

frontend/src/entry/model/InferenceServices.tsx
```diff
@@ -15,6 +19,21 @@ type InferenceProps = {
...
+  const [errorMessage, setErrorMessage] = useState('')
+  const [healthCheck, setHealthCheck] = useState(false)
+
+  const { tokens, isTokensLoading, isTokensError, mutateTokens } = useGetUserTokens()
+  const { uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
+
+  useEffect(() => {
+    async function checkAuthentication() {
+      fetch(`${uiConfig?.inference.connection.host}/api/health`, { credentials: 'include' }).then((response) => {
+        setHealthCheck(response.status === 200)
```
a3957273, line 31: Instead use `response.ok`

frontend/src/entry/model/InferenceServices.tsx
```diff
@@ -27,6 +46,32 @@ export default function InferenceServices({ model }: InferenceProps) {
     [model, inferences],
   )
 
+  const handleCreateToken = async () => {
+    const error = MultipleErrorWrapper('Unable to create an authenication token', { isTokensError, isUiConfigError })
+    if (error) return error
```
a3957273, line 51: Instead of returning jsx here, perhaps use `setErrorMessage` with a string.

frontend/src/entry/model/InferenceServices.tsx
```diff
@@ -27,6 +46,32 @@ export default function InferenceServices({ model }: InferenceProps) {
...
+    const authorizationTokenName = uiConfig?.inference.authorizationTokenName
+    const authorizationAccessKeys = tokens.filter((value) => value.description === authorizationTokenName)
+    if (authorizationTokenName && authorizationAccessKeys) {
+      for (const token of authorizationAccessKeys) {
+        deleteUserToken(token.accessKey)
```
a3957273, line 56: You should await this, and ensure the response is `.ok`.

a3957273 (changes requested):

frontend/src/entry/model/InferenceServices.tsx
GB27247: This wraps mulitple errors within a jsx element in the case that both of them fail.

JR40159 (changes requested):

frontend/src/entry/model/InferenceServices.tsx
```diff
@@ -15,6 +19,30 @@ type InferenceProps = {
...
+
+  useEffect(() => {
+    async function checkAuthentication() {
+      try {
+        setErrorMessage('')
+        const response = await fetch(`${uiConfig?.inference.connection.host}/api/health`, { credentials: 'include' })
+        setHealthCheck(response.ok)
+        if (!response.ok) {
+          setErrorMessage(await getErrorMessage(response))
+          return setErrorMessage('Login failed when when accessing the inferencing service')
```
ARADDCC012, line 36: You're setting the error message twice here. The second will override the first. I'd suggest just doing
```
return setErrorMessage(await getErrorMessage(response))
```

frontend/src/entry/model/InferenceServices.tsx
```diff
@@ -27,6 +55,41 @@ export default function InferenceServices({ model }: InferenceProps) {
     [model, inferences],
   )
 
+  const handleCreateToken = async () => {
+    setErrorMessage('')
+    const error = MultipleErrorWrapper('Unable to create an authenication token', { isTokensError, isUiConfigError })
+    if (error) return error
+    const authorizationTokenName = uiConfig?.inference.authorizationTokenName
+    const authorizationAccessKeys = tokens.filter((value) => value.description === authorizationTokenName)
+    if (authorizationTokenName && authorizationAccessKeys) {
```
ARADDCC012, line 64: `authorizationAccessKeys` will always be an array so this part of the check will always be true. Either check the length `authorizationAccessKeys.length > 0` or remove it from the check, depending on whether or not it is required.

frontend/src/entry/model/InferenceServices.tsx
```diff
@@ -27,6 +55,41 @@ export default function InferenceServices({ model }: InferenceProps) {
...
+      for (const token of authorizationAccessKeys) {
+        const response = await deleteUserToken(token.accessKey)
+
+        if (!response.ok) {
+          setErrorMessage(await getErrorMessage(response))
```
ARADDCC012, line 69: I would think we'd want to stop processing any more tokens if this has failed. Consider doing:
```
if (!response.ok) {
    setErrorMessage(await getErrorMessage(response))
    mutateTokens()
    return
}
```

frontend/src/entry/model/InferenceServices.tsx
```diff
@@ -35,17 +98,35 @@ export default function InferenceServices({ model }: InferenceProps) {
     return <MessageAlert message={isInferencesError.info.message} severity='error' />
```
ARADDCC012, line 98: You will also need this for `isTokensError` and `isUiConfigError `:
```
  if (isInferencesError) {
    return <MessageAlert message={isInferencesError.info.message} severity='error' />
  }
  
    if (isTokensError) {
    return <MessageAlert message={isTokensError.info.message} severity='error' />
  }
  
    if (isUiConfigError) {
    return <MessageAlert message={isUiConfigError.info.message} severity='error' />
  }
```

frontend/src/entry/model/InferenceServices.tsx
```diff
@@ -35,17 +98,35 @@ export default function InferenceServices({ model }: InferenceProps) {
...
+        <Stack spacing={2}>
+          <Typography>
+            Access to inferencing services requires token a with model access. Are you sure you want to proceed?
+          </Typography>
+          <Stack spacing={2} direction='row'>
+            <Button variant='contained' onClick={handleCreateToken}>
+              Yes
+            </Button>
+          </Stack>
+          {errorMessage && <MessageAlert message={errorMessage} severity='error' />}
```
ARADDCC012, line 127: `MessageAlert` internally checks `if (!message) return null`, so you don't need to conditionally render it.

I'd also recommend pulling this outside of your conditional logic (move it to just above the </Container> a couple lines below) so that it's always shown if there is an error to be displayed. (Not an issue for what you've done, but someone could easily miss that it's in a conditional block when adding future error handling)

ARADDCC012 (changes requested):

ARADDCC012 (changes requested):

JR40159 (approved):

ARADDCC012 (approved):

PR #1348, opened 2024-06-27
Bai 1348/mirrored model creation

GB27247:


frontend/src/entry/model/releases/EditableRelease.tsx
```diff
@@ -219,25 +219,27 @@ export default function EditableRelease({ release, isEdit, onIsEditChange }: Edi
...
-        isEdit={isEdit}
-        isLoading={isLoading}
-        onEdit={handleEdit}
-        onCancel={handleCancel}
-        onSubmit={handleSubmit}
-        onDelete={() => setOpen(true)}
-        errorMessage={errorMessage}
-        isRegistryError={isRegistryError}
-      />
+      {(!model.settings.mirror || model.settings.mirror?.sourceModelId) && (
```
ARADDCC012, line 222: We'll still want the heading so the release name is shown. Are you doing this to hide the edit/delete buttons? If so, we should instead conditionally render those buttons. You'll just need to pass through a `readOnly` prop and handle it within `EditableFormHeading`.

frontend/src/entry/overview/FormEditPage.tsx
```diff
@@ -142,6 +143,7 @@ export default function FormEditPage({ entry }: FormEditPageProps) {
                 onClick={() => setIsEdit(!isEdit)}
                 sx={{ mb: { xs: 2 } }}
                 data-test='editEntryCardButton'
+                hidden={readOnly}
```
ARADDCC012, line 146: Again, we should be conditionally rendering this button.

However, due to your conditional logic in `frontend/src/entry/overview/Overview.tsx` I don't think we need to do hide anything in this component.

frontend/src/entry/overview/Overview.tsx
```diff
@@ -13,18 +14,24 @@ type OverviewPageKeys = (typeof OverviewPage)[keyof typeof OverviewPage]
...
+  return readOnly ? (
+    <MessageAlert
+      severity='warning'
+      message='This mirrored model has no model card. Please export the model card from the source model.'
+    />
+  ) : (
     <Container sx={{ my: 2 }}>
       {page === OverviewPage.TEMPLATE && <TemplatePage entry={entry} />}
-      {page === OverviewPage.FORM && <FormEditPage entry={entry} />}
+      {page === OverviewPage.FORM && <FormEditPage entry={entry} readOnly />}
```
ARADDCC012, line 34: Why are you passing `readOnly` into `FormEditPage`? Based on your logic `FormEditPage` will not be rendered in read-only mode, so the `readOnly` prop will always be `false`.

frontend/src/entry/overview/FormEditPage.tsx
```diff
@@ -17,9 +17,10 @@ import { getStepsData, getStepsFromSchema } from 'utils/formUtils'
 
 type FormEditPageProps = {
   entry: EntryInterface
+  readOnly?: boolean
```
ARADDCC012, line 20: This prop looks unnecessary to me due to your conditional logic in `frontend/src/entry/overview/Overview.tsx`

frontend/src/marketplace/EntryList.tsx
```diff
@@ -25,11 +25,14 @@ export default function EntryList({
...
-            {entry.name}
-          </MuiLink>
-        </Link>
+        <Stack direction='row'>
+          <Link style={{ textDecoration: 'none' }} href={`${entry.kind}/${entry.id}`} passHref>
+            <MuiLink variant='h5' sx={{ fontWeight: '500', textDecoration: 'none', color: theme.palette.primary.main }}>
+              {entry.name}
+            </MuiLink>
+          </Link>
+          {entry.readonly && <Typography>readonly</Typography>}
```
ARADDCC012, line 34: This is user-facing so consider doing:
```
{entry.readonly && <Typography> (read-only)</Typography>}
```

ARADDCC012 (changes requested):

frontend/src/common/PageWithTabs.tsx
```diff
@@ -110,30 +112,49 @@ export default function PageWithTabs({
...
+              left: 70,
+              right: 0,
+              top: 95,
+              backgroundColor: 'black',
+              borderRadius: 0,
+              color: 'white',
+            }}
+          >
+            <Typography justifyContent='center' align='center'>
+              [READONLY] Mirrored from {sourceModelId}
```
JR40159, line 129: I like making it clear where the model comes from but I'm not a big fan of using a banner for this, it's in danger of being over-used. Could we include this information next to the model name at the top or something?

frontend/pages/entry/new.tsx
JR40159, line 1: Now we have three options, I'm noticing that the page doesn't copy well with being resized

frontend/src/entry/overview/Overview.tsx
```diff
@@ -13,18 +14,24 @@ type OverviewPageKeys = (typeof OverviewPage)[keyof typeof OverviewPage]
...
   const page: OverviewPageKeys = useMemo(
     () => (entry.card && entry.card.schemaId ? OverviewPage.FORM : OverviewPage.TEMPLATE),
     [entry.card],
   )
 
-  return (
+  return readOnly ? (
+    <MessageAlert
+      severity='warning'
+      message='This mirrored model has no model card. Please export the model card from the source model.'
```
JR40159, line 29: If I manually a model card data into the DB then I still see this message rather than the model card?

frontend/pages/model/[modelId].tsx
```diff
@@ -32,11 +32,21 @@ export default function Model() {
...
-              view: <Releases model={model} currentUserRoles={currentUserRoles} />,
+              view: (
+                <Releases
+                  model={model}
+                  currentUserRoles={currentUserRoles}
+                  readOnly={!!model.settings.mirror?.sourceModelId}
+                />
+              ),
               disabled: !model.card,
               disabledText: 'Select a schema to view this tab',
```
Comment (JR40159) [inline frontend/pages/model/[modelId].tsx:51]: This still appears for mirrored models but isn't applicable since a user cannot select a schema

JR40159 (changes requested):

frontend/pages/entry/new.tsx
GB27247: I've modified the styling so the cards wrap around when the page is too small.

ARADDCC012 (approved):

JR40159 (changes requested):
On the create entry page for a mirrored model, can you make the source model ID mandatory?

JR40159 (approved):

frontend/src/entry/CreateEntry.tsx
```diff
@@ -228,7 +228,7 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
                 <span>
                   <LoadingButton
                     variant='contained'
-                    disabled={!isFormValid}
+                    disabled={!isFormValid || (!sourceModelId && createEntryKind === CreateEntryKind.MIRRORED_MODEL)}
```
ARADDCC012, line 231: This additional logic should be added to `isFormValid` instead of here.

ARADDCC012 (changes requested):

frontend/src/entry/overview/Overview.tsx
```diff
@@ -13,19 +14,25 @@ type OverviewPageKeys = (typeof OverviewPage)[keyof typeof OverviewPage]
...
+  return readOnly && !entry.card ? (
+    <MessageAlert
+      severity='warning'
+      message='This mirrored model has no model card. Please export the model card from the source model.'
+    />
+  ) : (
     <Container sx={{ my: 2 }}>
       {page === OverviewPage.TEMPLATE && <TemplatePage entry={entry} />}
-      {page === OverviewPage.FORM && <FormEditPage entry={entry} currentUserRoles={currentUserRoles} />}
+      {page === OverviewPage.FORM && <FormEditPage entry={entry} currentUserRoles={currentUserRoles} readOnly />}
```
ARADDCC012, line 35: Whoops, just noticed this one. Passing `readOnly` without a specified value in react is the same as doing `readOnly={true}` so you'll need to update this to the following:
```
{page === OverviewPage.FORM && <FormEditPage entry={entry} currentUserRoles={currentUserRoles} readOnly={readOnly} />}
```

ARADDCC012 (approved):

PR #1349, opened 2024-06-28
Add helper function to list schema ids

gc153671:


lib/python/src/bailo/helper/schema.py
```diff
@@ -107,3 +107,14 @@ def __unpack(self, res) -> None:
             self.kind = SchemaKind.ACCESS_REQUEST
 
         logger.info(f"Attributes for Schema ID %s successfully unpacked.", self.schema_id)
+
+    @classmethod
```
bw27492, line 111: I think this could actually be a `staticmethod` instead since we aren't using `cls` unless we want to create a list of `Schema()` objects. If we just want the IDs then `staticmethod` is best

bw27492 (changes requested):

bw27492 (approved):

PR #1351, opened 2024-06-28
Disable next telemetry

GCHQDeveloper36:
If desired, this disables telemetry about commands etc being sent back to the NextJS folks.

More info: https://nextjs.org/telemetry


a3957273 (approved):
Looks good to me!

PR #1377, opened 2024-07-01
Feature/conditionally hide UI based on roles

ARADDCC012:


frontend/pages/model/[modelId]/release/[semver].tsx
```diff
@@ -29,24 +33,41 @@ export default function Release() {
...
-      userReviews.some(
-        (userReview) =>
-          userReview.model.id === review.model.id && userReview.accessRequestId === review.accessRequestId,
-      ),
-    ).length > 0
+  const currentUserRoles = useMemo(() => getCurrentUserRoles(model, currentUser), [model, currentUser])
+
+  const userCanReview = useMemo(
+    () =>
+      hasRole(currentUserRoles, ['msro', 'mtr']) &&
```
Comment (JR40159) [inline frontend/pages/model/[modelId]/release/[semver].tsx:43]: I'm a little apprehensive about having authorisation now in both the frontend as well as the backend as it's duplication that will not be good for maintainability. Specifically here it's important to remember that in the near future that the roles for reviews will be dynamic and defined in the schema.

Comment (ARADDCC012) [reply frontend/pages/model/[modelId]/release/[semver].tsx:43]: Agreed. I was conscious of that when writing this and have made a mental note to prepare for the amount of added effort this work will add to dynamic roles. I've tried to limit that by having reusable logic, but unfortunately this is the best we can do for now.

frontend/pages/model/[modelId].tsx
```diff
@@ -51,15 +52,19 @@ export default function Model() {
...
               path: 'inferencing',
               view: <InferenceServices model={model} />,
               hidden: !uiConfig.inference.enabled,
             },
-            { title: 'Settings', path: 'settings', view: <Settings entry={model} /> },
+            {
+              title: 'Settings',
+              path: 'settings',
+              view: <Settings entry={model} currentUserRoles={currentUserRoles} />,
+            },
```
Comment (GB27247) [inline frontend/pages/model/[modelId].tsx:67]: Not sure how I feel about this. In my opinion the settings tab itself should be restricted?

GB27247 (changes requested):
There's also a couple of buttons in the inferencing tab that should be disabled for non owner or contributors.

frontend/src/Form/EditableFormHeading.tsx
```diff
@@ -16,7 +16,16 @@ type EditableFormHeadingProps = {
...
+} & (
+  | {
+      canUserEditOrDelete: boolean
+      actionButtonsTooltip: string
+    }
+  | {
+      canUserEditOrDelete?: never
+      actionButtonsTooltip?: never
+    }
+)
```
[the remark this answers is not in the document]
ARADDCC012: This is to make it easier to work with `EditableFormHeading` in future. It just forces either both props to be provided or neither.

Your code snippet doesn't provide this direction and is the same as doing:
```
{
    canUserEditOrDelete?: boolean
    actionButtonsTooltip?: string
}
```

frontend/pages/model/[modelId].tsx
Comment (ARADDCC012) [reply frontend/pages/model/[modelId].tsx:67]: I asked the same thing, but it's important for any user to be able to see the collaborators in the entry access tab. We could simplify a lot of the settings roles logic if we find another place to display that information. I'll make a ticket to explore this idea.

GB27247 (approved):

PR #1380, opened 2024-07-03
Python Improvements

bw27492:


lib/python/src/bailo/helper/model.py
GB27247, line 1: Could we have a `__repr__` method on this class. It's very hard to see what I've searched for without being able to view the model ID.

lib/python/src/bailo/helper/entry.py
```diff
@@ -46,11 +46,17 @@ def update(self) -> None:
...
         """Create a card using a schema on Bailo.
 
-        :param schema_id: A unique schema ID
+        :param schema_id: A unique schema ID, defaults to None. If None, either minimal-general-v10 or minimal-data-card-v10 is used
         """
+        if schema_id is None:
+            if self.kind == EntryKind.MODEL:
+                schema_id = "minimal-general-v10"
+            if self.kind == EntryKind.DATACARD:
+                schema_id = "minimal-data-card-v10"
```
GB27247, line 58: I'd be cautious about these schema's they're not guaranteed and are configurable.

lib/python/src/bailo/helper/model.py
```diff
@@ -456,6 +482,7 @@ def from_mlflow(self, tracking_uri: str, experiment_id: str):
 
             if len(mlflow.artifacts.list_artifacts(artifact_uri=artifact_uri)):
                 mlflow_dir = os.path.join(self.temp_dir, f"mlflow_{run_id}")
+                print(artifact_uri)
```
GB27247, line 485: Is this a print statement function from testing?

lib/python/tests/test_model.py
```diff
@@ -53,6 +53,14 @@ def test_create_get_from_id_and_update(
     assert model.model_id == get_model.model_id
 
 
+@pytest.mark.integration
+def test_search_models(integration_client):
+    models = Model.search(client=integration_client)
+
+    for model in models:
```
GB27247, line 60: A better way of doing this is `assert all(isinstance(model, Model) for model in models)` to avoid multiple assertions for the same test

lib/python/tests/test_model.py
```diff
@@ -53,6 +53,14 @@ def test_create_get_from_id_and_update(
...
+        assert isinstance(model, Model)
```
GB27247, line 61: As well as this could we test for failure and some edge cases?

Eg.
- Test a variety of each of arguments.
- Assert you can retrieve a specific model

lib/python/src/bailo/helper/model.py
```diff
@@ -111,14 +111,44 @@ def from_id(cls, client: Client, model_id: str) -> Model:
...
+        :param filters: Custom filters, defaults to None
+        :param search: String to be located in model cards, defaults to ""
+        :return: List of model objects
+        """
+        res = client.get_models(kind=EntryKind.MODEL, task=task, libraries=libraries, filters=filters, search=search)
+        models = []
+
+        for model in res["models"]:
+            res_model = client.get_model(model_id=model["id"])["model"]
+            model_obj = cls(client=client, model_id=model["id"], name=model["name"], description=model["description"])
```
GB27247, line 137: This does give warnings for each of the models on creation if their is no schema.

GB27247 (changes requested):

lib/python/src/bailo/helper/entry.py
bw27492: Good point, we should provide a warning in the logs. If a user is using custom schemas they should be providing their own ID anyway

lib/python/src/bailo/helper/access_request.py
```diff
@@ -69,15 +69,17 @@ def from_id(cls, client: Client, model_id: str, access_request_id: str) -> Acces
         )
 
     @classmethod
-    def create(cls, client: Client, model_id: str, schema_id: str, metadata: Any) -> AccessRequest:
+    def create(
+        cls, client: Client, model_id: str, metadata: Any, schema_id: str = "minimal-access-request-general-v10"
```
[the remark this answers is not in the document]
bw27492: Done

lib/python/src/bailo/helper/model.py
bw27492: Done

lib/python/src/bailo/helper/model.py
bw27492: Removed

lib/python/tests/test_model.py
bw27492: Done

lib/python/tests/test_model.py
bw27492: Added an additional test to look for specific models

GB27247: Could we test with different arrangements of task, libraries and filters?

lib/python/src/bailo/helper/entry.py
```diff
@@ -46,11 +46,18 @@ def update(self) -> None:
...
 
-        :param schema_id: A unique schema ID
+        :param schema_id: A unique schema ID, defaults to None. If None, either minimal-general-v10 or minimal-data-card-v10 is used
         """
+        if schema_id is None:
+            if self.kind == EntryKind.MODEL:
+                schema_id = MinimalSchema.MODEL
+            if self.kind == EntryKind.DATACARD:
+                schema_id = MinimalSchema.DATACARD
+            warnings.warn(f"Schema ID not provided, will attempt to default to {schema_id}.")
```
GB27247, line 59: This shouldn't be a warning it's expected behaviour.

GB27247 (changes requested):

lib/python/src/bailo/helper/entry.py
bw27492: Done

GB27247 (approved):

PR #1382, opened 2024-07-04
Add query parameters to roles

GB27247:


frontend/src/common/ReviewWithComment.tsx
```diff
@@ -60,6 +62,7 @@ export default function ReviewWithComment({
 
   const { responses, isResponsesLoading, isResponsesError } = useGetResponses([...reviews.map((review) => review._id)])
   const { modelRoles, isModelRolesLoading, isModelRolesError } = useGetModelRoles(modelId)
+  const { role } = router.query
   const [reviewRequest, setReviewRequest] = useState(reviews[0])
```
ARADDCC012, line 66: You can do all of this a little easier by just changing this line to
```
  const [reviewRequest, setReviewRequest] = useState<ReviewRequestInterface>(
    reviews.find((review) => review.role === role) || reviews[0],
  )
```
I've added in an explicit type for the `useState` as TS seems to be losing track of it.

frontend/src/common/ReviewWithComment.tsx
```diff
@@ -135,6 +144,7 @@ export default function ReviewWithComment({
               value={reviewRequest}
               getOptionLabel={(option) => getRoleDisplay(option.role, modelRoles)}
               options={reviews}
+              defaultValue={typeof role === 'string' ? reviews.find((review) => review.role === role) : null}
```
ARADDCC012, line 147: We should be setting the default value of the `reviewRequest` state variable instead of doing this. I've shown how to do this in another comment.

ARADDCC012 (changes requested):

frontend/src/common/ReviewWithComment.tsx
```diff
@@ -135,6 +145,7 @@ export default function ReviewWithComment({
               value={reviewRequest}
               getOptionLabel={(option) => getRoleDisplay(option.role, modelRoles)}
               options={reviews}
+              defaultValue={reviewRequest}
```
ARADDCC012, line 148: Sorry for not making this clear, but `defaultValue` isn't required at all. We're already controlling the value with the `value` prop

ARADDCC012 (approved):

PR #1384, opened 2024-07-05
Migrate access request schema select to main SchemaSelect component

GCHQDeveloper36:
Merges the access request schema select into the common SchemaSelect component as they're fundamentally the same.


frontend/src/schemas/SchemaSelect.tsx
```diff
@@ -104,9 +122,9 @@ export default function SchemaSelect({ entry }: SchemaSelectProps) {
       {!isLoadingData && (
         <Container maxWidth='md'>
           <Card sx={{ mx: 'auto', my: 4, p: 4 }}>
-            <Link href={`/${entry.kind}/${entry.id}`}>
+            <Link href={link}>
               <Button sx={{ width: 'fit-content' }} startIcon={<ArrowBack />}>
-                {`Back to ${EntryKindLabel[entry.kind]}`}
+                {`Back to ${SchemaKindLabel[schemaKind]}`}
```
ARADDCC012, line 127: This should refer to the entry the user will be taken to, so should be
```
{`Back to ${EntryKindLabel[entry.kind]}`}
```

frontend/src/schemas/SchemaSelect.tsx
```diff
@@ -56,6 +66,12 @@ export default function SchemaSelect({ entry }: SchemaSelectProps) {
     [currentUser, entry, mutateEntry, router],
   )
 
+  let selectionCallback = entryCallback
```
ARADDCC012, line 69: The references should still be stable with how this has been implemented, but to ensure they are we could do the following:
```
  const selectionCallback = useMemo(
    () => (schemaKind === SchemaKind.ACCESS_REQUEST ? accessRequestCallback : entryCallback),
    [schemaKind, accessRequestCallback, entryCallback],
  )
```

frontend/src/schemas/SchemaSelect.tsx
```diff
@@ -12,21 +12,23 @@ import Loading from 'src/common/Loading'
 import Link from 'src/Link'
 import MessageAlert from 'src/MessageAlert'
 import SchemaButton from 'src/schemas/SchemaButton'
-import { EntryInterface, EntryKind, EntryKindLabel, SchemaInterface, SchemaKind } from 'types/types'
+import { EntryInterface, EntryKind, SchemaInterface, SchemaKind, SchemaKindKeys, SchemaKindLabel } from 'types/types'
 
 type SchemaSelectProps = {
-  entry: EntryInterface
+  schemaKind: SchemaKindKeys
+  entry?: EntryInterface
```
ARADDCC012, line 19: I would prefer to keep `entry` a required prop as we need this to determine where to route the user to if the click the back button. It's also needed to know what text to put on that button. This would also mean that we don't need `id` as a prop.

I suggest updating `frontend/pages/model/[modelId]/access-request/schema.tsx` to fetch the model so that it can be passed into this component.

frontend/pages/model/[modelId]/access-request/schema.tsx
```diff
@@ -1,116 +1,17 @@
...
-import MultipleErrorWrapper from 'src/errors/MultipleErrorWrapper'
-import Link from 'src/Link'
-import SchemaButton from 'src/schemas/SchemaButton'
-import { SchemaInterface, SchemaKind } from 'types/types'
+import SchemaSelect from 'src/schemas/SchemaSelect'
+import { SchemaKind } from 'types/types'
 
 export default function AccessRequestSchema() {
   const router = useRouter()
   const { modelId }: { modelId?: string } = router.query
```
Comment (ARADDCC012) [inline frontend/pages/model/[modelId]/access-request/schema.tsx:9]: You could use this ID to fetch the model like so:
```
const { model, isModelLoading, isModelError } = useGetModel(modelId, EntryKind.MODEL)
```
You'll then need to add back in the error handling:
```
  const error = MultipleErrorWrapper(`Unable to load schema page`, {
    isModelError,
  })
  if (error) return error
```
And finally you'll need to manage the loading state just below the `Title`:
```
      <Title text='Select a schema' />
      {isModelLoading && <Loading />}
```
You can then pass the `model` into `SchemaSelect` via the `entry` prop.
```
<SchemaSelect schemaKind={SchemaKind.ACCESS_REQUEST} entry={model} />
```

frontend/src/schemas/SchemaSelect.tsx
```diff
@@ -36,7 +38,15 @@ export default function SchemaSelect({ entry }: SchemaSelectProps) {
   const activeSchemas = useMemo(() => schemas.filter((schema) => schema.active), [schemas])
   const inactiveSchemas = useMemo(() => schemas.filter((schema) => !schema.active), [schemas])
 
-  const createEntryCardUsingSchema = useCallback(
+  const accessRequestCallback = useCallback(
+    async (newSchema: SchemaInterface) => {
+      setLoading(true)
+      router.push(`/model/${id}/access-request/new?schemaId=${newSchema.id}`)
```
ARADDCC012, line 44: Just to future proof ourselves in case any new entry kinds allow access requests I would use the `entry.kind` in the URL:
```
router.push(`/${entry.kind}/${entry.id}/access-request/new?schemaId=${newSchema.id}`)
```

frontend/src/schemas/SchemaSelect.tsx
```diff
@@ -104,9 +122,9 @@ export default function SchemaSelect({ entry }: SchemaSelectProps) {
       {!isLoadingData && (
         <Container maxWidth='md'>
           <Card sx={{ mx: 'auto', my: 4, p: 4 }}>
-            <Link href={`/${entry.kind}/${entry.id}`}>
+            <Link href={link}>
```
ARADDCC012, line 125: This change can be reverted if you change `entry` to be a required prop

ARADDCC012 (changes requested):

ARADDCC002 (approved):

PR #1389, opened 2024-07-08
3. Minor updates to frontend dependencies.

a3957273:
Some minor frontend dependency bumps. Should be merged in after https://github.com/gchq/Bailo/pull/1388


frontend/utils/fetcher.ts
```diff
@@ -23,8 +23,8 @@ export const textFetcher = async (input: RequestInfo, init: RequestInit) => {
   return res.text()
 }
 
-export const fetcher = async (input: RequestInfo, init: RequestInit) => {
-  const res = await fetch(input, init)
+export const fetcher = async (path: any) => {
```
ARADDCC012, line 26: With how we've implemented our fetcher we should restrict `path` to being a `string`. Also consider using `url` instead of `path` as this is the common terminology used in the swr docs

ARADDCC012 (changes requested):

ARADDCC012 (approved):

PR #1390, opened 2024-07-08
4. Remove 'textFetcher' function

a3957273:
Remove the 'textFetcher' action, which isn't used after we removed the file listing from zip files.


ARADDCC012 (approved):

ARADDCC012:
Dependant on https://github.com/gchq/Bailo/pull/1389

PR #1394, opened 2024-07-08
Improve developer logging of requests

a3957273:
Before:

Screenshot 2024-07-08 at 23 54 06

After:

Screenshot 2024-07-08 at 23 52 48


backend/src/services/log.ts
```diff
@@ -15,6 +15,14 @@ interface BunyanLog {
   src: { file: string; line: number }
   msg: string
 }
+
+interface RequestCompletedLog extends BunyanLog {
+  method: string
+  url: string
+  status: number
+  'response-time': number
```
ARADDCC012, line 23: Is the a reason for kebab case here? Seems strange to not use camel case unless there's something I'm missing

backend/src/services/log.ts
```diff
@@ -76,6 +84,11 @@ export class Writer extends WritableStream {
     return typeof value === 'object' ? util.inspect(value) : String(value)
   }
 
+  static isRequestCompleted(data: any): data is RequestCompletedLog {
```
ARADDCC012, line 87: Should be using `unknown` instead of `any`. Looks like this is an issue in a number of places in this file, but I'll refrain from commenting on those.

If `data.msg` is guaranteed then we can at the very least use generics to improve the typing.

ARADDCC012 (changes requested):

PR #1395, opened 2024-07-09
Fix image upload authorisation

GB27247:


a3957273 (approved):

a3957273 (changes requested):
Please add a unit test to ensure this bug doesn't occur in the future.

backend/test/connectors/authorisation/base.spec.ts
```diff
@@ -185,4 +185,32 @@ describe('connectors > authorisation > base', () => {
       success: false,
     })
   })
+
+  test('image > push with no roles', async () => {
```
JR40159, line 189: Could you also add a test for the equivalent push condition please?

JR40159 (changes requested):

JR40159 (approved):

PR #1397, opened 2024-07-10
Bugfix/response refactor migrations

ARADDCC012:


backend/src/migrations/004_comment_response_refactor.ts
JR40159, line 1: We need to be careful of editing existing migrations as we don't want our environments to diverge. Are all these changes style changes?

JR40159 (changes requested):

JR40159 (approved):

PR #1402, opened 2024-07-15
added the ability to edt comments via the UI

ARADDCC002:
image

image

image

### DIFF
--- a/frontend/src/reviews/EditableReviewComment.tsx
+++ b/frontend/src/reviews/EditableReviewComment.tsx
@@ -38,7 +38,7 @@ export default function EditableReviewComment({
       )}
       {isEditMode && (
         <>
-          <RichTextEditor value={comment} onChange={(input) => onCommentChange(input)} />
+          <RichTextEditor value={comment} onChange={onCommentChange} />
           <Stack sx={{ textAlign: 'right', pt: 2 }} direction='row' spacing={1} justifyContent='right'>
             <Button onClick={onCancel}>Cancel</Button>
             <Button variant='contained' onClick={onSave}>
--- a/frontend/src/reviews/EditableReviewComment.tsx
+++ b/frontend/src/reviews/EditableReviewComment.tsx
@@ -7,26 +7,26 @@ import { formatDateTimeString } from 'utils/dateUtils'
 
 interface EditableReviewCommentProps {
   comment: string
-  setComment: (value: string) => void
+  onCommentChange: (value: string) => void
   response: ResponseInterface
   isEditMode: boolean
-  setIsEditMode: (value: boolean) => void
   editCommentErrorMessage: string
+  onCancel: () => void
   onSave: () => void
 }
 
 export default function EditableReviewComment({
   comment,
-  setComment,
+  onCommentChange,
   response,
   isEditMode,
-  setIsEditMode,
   editCommentErrorMessage,
+  onCancel,
   onSave,
 }: EditableReviewCommentProps) {
   return (
     <>
-      {comment && !isEditMode && (
+      {!isEditMode && (
         <Stack spacing={2}>
           <MarkdownDisplay>{comment}</MarkdownDisplay>
           {response.updatedAt !== response.createdAt && (
@@ -38,9 +38,9 @@ export default function EditableReviewComment({
       )}
       {isEditMode && (
         <>
-          <RichTextEditor value={comment} onChange={(input) => setComment(input)} />
+          <RichTextEditor value={comment} onChange={(input) => onCommentChange(input)} />
           <Stack sx={{ textAlign: 'right', pt: 2 }} direction='row' spacing={1} justifyContent='right'>
-            <Button onClick={() => setIsEditMode(false)}>Cancel</Button>
+            <Button onClick={onCancel}>Cancel</Button>
             <Button variant='contained' onClick={onSave}>
               Save
             </Button>
--- a/frontend/src/reviews/EditableReviewComment.tsx
+++ b/frontend/src/reviews/EditableReviewComment.tsx
@@ -1,0 +1,53 @@
+import { Button, Stack, Typography } from '@mui/material'
+import MarkdownDisplay from 'src/common/MarkdownDisplay'
+import RichTextEditor from 'src/common/RichTextEditor'
+import MessageAlert from 'src/MessageAlert'
+import { ResponseInterface } from 'types/types'
+import { formatDateTimeString } from 'utils/dateUtils'
+
+interface EditableReviewCommentProps {
+  comment: string
+  setComment: (value: string) => void
+  response: ResponseInterface
+  isEditMode: boolean
+  setIsEditMode: (value: boolean) => void
+  editCommentErrorMessage: string
+  onSave: () => void
+}
+
+export default function EditableReviewComment({
+  comment,
+  setComment,
+  response,
+  isEditMode,
+  setIsEditMode,
+  editCommentErrorMessage,
+  onSave,
+}: EditableReviewCommentProps) {
+  return (
+    <>
+      {comment && !isEditMode && (
+        <Stack spacing={2}>
+          <MarkdownDisplay>{comment}</MarkdownDisplay>
+          {response.updatedAt !== response.createdAt && (
+            <Typography variant='caption' sx={{ fontStyle: 'italic' }}>
+              Edited {formatDateTimeString(response.updatedAt)}
+            </Typography>
+          )}
+        </Stack>
+      )}
+      {isEditMode && (
+        <>
+          <RichTextEditor value={comment} onChange={(input) => setComment(input)} />
+          <Stack sx={{ textAlign: 'right', pt: 2 }} direction='row' spacing={1} justifyContent='right'>
+            <Button onClick={() => setIsEditMode(false)}>Cancel</Button>
+            <Button variant='contained' onClick={onSave}>
+              Save
+            </Button>
+          </Stack>
+          <MessageAlert message={editCommentErrorMessage} />
+        </>
+      )}
+    </>
+  )
+}
--- a/frontend/src/reviews/ReviewCommentDisplay.tsx
+++ b/frontend/src/reviews/ReviewCommentDisplay.tsx
@@ -45,6 +45,7 @@ export default function ReviewCommentDisplay({
   const handleEditOnCancel = () => {
     setIsEditMode(false)
     setEditCommentErrorMessage('')
+    setComment(response.comment || '')
   }
 
   const handleEditOnSave = async () => {
@@ -77,7 +78,7 @@ export default function ReviewCommentDisplay({
             </Typography>
             <Stack direction='row' alignItems='center' spacing={1}>
               <Typography fontWeight='bold'>{formatDateString(response.createdAt)}</Typography>
-              <IconButton onClick={(event) => setAnchorEl(event.currentTarget)}>
+              <IconButton onClick={(event) => setAnchorEl(event.currentTarget)} aria-label='Actions'>
                 <MoreHorizIcon />
               </IconButton>
             </Stack>
--- a/frontend/src/reviews/ReviewCommentDisplay.tsx
+++ b/frontend/src/reviews/ReviewCommentDisplay.tsx
@@ -42,6 +42,11 @@ export default function ReviewCommentDisplay({
     setIsEditMode(true)
   }
 
+  const handleEditOnCancel = () => {
+    setIsEditMode(false)
+    setEditCommentErrorMessage('')
+  }
+
   const handleEditOnSave = async () => {
     setEditCommentErrorMessage('')
     const res = await patchResponse(response._id, comment)
@@ -80,11 +85,11 @@ export default function ReviewCommentDisplay({
           <Divider sx={{ mt: 1, mb: 2 }} />
           <EditableReviewComment
             comment={comment}
-            setComment={setComment}
+            onCommentChange={setComment}
             response={response}
             isEditMode={isEditMode}
-            setIsEditMode={setIsEditMode}
             editCommentErrorMessage={editCommentErrorMessage}
+            onCancel={handleEditOnCancel}
             onSave={handleEditOnSave}
           />
         </Card>
--- a/frontend/src/reviews/ReviewCommentDisplay.tsx
+++ b/frontend/src/reviews/ReviewCommentDisplay.tsx
@@ -1,27 +1,60 @@
 import { Menu as MenuIcon } from '@mui/icons-material'
-import { Box, Card, Divider, IconButton, Menu, MenuItem, Stack, Typography } from '@mui/material'
+import { Box, Button, Card, Divider, IconButton, Menu, MenuItem, Stack, Typography } from '@mui/material'
+import { useTheme } from '@mui/material/styles'
+import { patchResponse } from 'actions/response'
 import { useState } from 'react'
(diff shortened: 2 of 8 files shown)


frontend/src/reviews/ReviewCommentDisplay.tsx
```diff
@@ -49,16 +83,33 @@ export default function ReviewCommentDisplay({ response, onReplyButtonClick }: R
...
+              )}
+            </Stack>
+          )}
+          {isEditMode && (
+            <>
+              <RichTextEditor value={updatedComment} onChange={(input) => setUpdatedComment(input)} />
+              <Box sx={{ float: 'right' }}>
+                <Button onClick={handleEditOnSave}>Save</Button>
+              </Box>
+              <Typography variant='caption' color={theme.palette.error.main}>
```
ARADDCC012, line 103: This should be a `MessageAlert`

frontend/src/reviews/ReviewCommentDisplay.tsx
```diff
@@ -25,6 +42,22 @@ export default function ReviewCommentDisplay({ response, onReplyButtonClick }: R
...
   }
 
+  const handleEditOnClick = () => {
+    setUpdatedComment(response.comment || '')
+    setIsEditMode(true)
+  }
+
+  const handleEditOnSave = async () => {
+    setEditCommentErrorMessage('')
+    const res = await patchResponse(response['_id'], updatedComment)
```
ARADDCC012, line 52: Minor but can we use the following syntax here for the ID? `response._id`

I'm guessing you've done it this way to circumvent our typing being off here, so I think we'll also need to add `_id` to the `ResponseInterface`

frontend/src/reviews/ReviewCommentDisplay.tsx
```diff
@@ -37,6 +70,7 @@ export default function ReviewCommentDisplay({ response, onReplyButtonClick }: R
             p: 1,
           }}
         >
+          {isEditMode}
```
ARADDCC012, line 73: This looks like debug code, can be removed

frontend/src/reviews/ReviewCommentDisplay.tsx
```diff
@@ -49,16 +83,33 @@ export default function ReviewCommentDisplay({ response, onReplyButtonClick }: R
...
+              <Typography variant='caption' color={theme.palette.error.main}>
+                {editCommentErrorMessage}
+              </Typography>
+            </>
           )}
         </Card>
       </Stack>
       <Menu anchorEl={anchorEl} open={open} onClose={() => setAnchorEl(null)}>
         <MenuItem onClick={() => handleReplyOnClick(response.comment)}>Reply</MenuItem>
+        {currentUser && currentUser.dn === username && <MenuItem onClick={handleEditOnClick}>Edit comment</MenuItem>}
```
ARADDCC012, line 112: Minor but I don't think specifying what's being edited is needed here. Consider changing the text to simply "Edit", but we'd need an `aria-label="Edit comment"` if you were to do that.

frontend/src/reviews/ReviewCommentDisplay.tsx
```diff
@@ -49,16 +83,33 @@ export default function ReviewCommentDisplay({ response, onReplyButtonClick }: R
...
+                  Updated {formatDateTimeString(response.updatedAt)}
+                </Typography>
+              )}
+            </Stack>
+          )}
+          {isEditMode && (
+            <>
+              <RichTextEditor value={updatedComment} onChange={(input) => setUpdatedComment(input)} />
+              <Box sx={{ float: 'right' }}>
+                <Button onClick={handleEditOnSave}>Save</Button>
```
ARADDCC012, line 101: We should also have a cancel button

frontend/src/reviews/ReviewCommentDisplay.tsx
```diff
@@ -49,16 +83,33 @@ export default function ReviewCommentDisplay({ response, onReplyButtonClick }: R
...
+                <Typography variant='caption' sx={{ fontStyle: 'italic' }}>
+                  Updated {formatDateTimeString(response.updatedAt)}
+                </Typography>
+              )}
+            </Stack>
+          )}
+          {isEditMode && (
+            <>
+              <RichTextEditor value={updatedComment} onChange={(input) => setUpdatedComment(input)} />
+              <Box sx={{ float: 'right' }}>
```
ARADDCC012, line 100: Floats should be avoided as they can result in layouts changing in unexpected ways. You can either do `ml='auto'` or probably `textAlign='right'`

frontend/src/reviews/ReviewCommentDisplay.tsx
```diff
@@ -1,21 +1,38 @@
...
+  currentUser,
+  mutateResponses,
+}: ReviewCommentDisplayProps) {
   const [entityKind, username] = response.entity.split(':')
 
+  const theme = useTheme()
+
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null)
+  const [isEditMode, setIsEditMode] = useState(false)
+  const [updatedComment, setUpdatedComment] = useState('')
```
ARADDCC012, line 33: With these kinds of editable inputs I find it's easier to always use one value, rather than alternating between using `release.comment` and `updatedComment` depending on the situation.

Consider renaming this to `comment` and using it throughout. You'll then just need to set it to either `release.comment` (on initial render and when cancelling edit mode) or let the `RichTextEditor` control it when in edit mode.

frontend/src/reviews/ReviewCommentDisplay.tsx
```diff
@@ -49,16 +83,33 @@ export default function ReviewCommentDisplay({ response, onReplyButtonClick }: R
...
-            <div>
-              <Divider sx={{ mt: 1, mb: 2 }} />
+          <Divider sx={{ mt: 1, mb: 2 }} />
+          {response.comment && !isEditMode && (
+            <Stack spacing={2}>
               <MarkdownDisplay>{response.comment}</MarkdownDisplay>
-            </div>
+              {response.updatedAt !== response.createdAt && (
+                <Typography variant='caption' sx={{ fontStyle: 'italic' }}>
+                  Updated {formatDateTimeString(response.updatedAt)}
```
ARADDCC012, line 92: For consistency consider using `Edited` instead of `Updated`

frontend/src/reviews/ReviewDecisionDisplay.tsx
```diff
@@ -69,16 +112,41 @@ export default function ReviewDecisionDisplay({ response, modelId }: ReviewDecis
...
+              <IconButton onClick={(event) => setAnchorEl(event.currentTarget)}>
+                <MenuIcon />
+              </IconButton>
+            </Stack>
           </Stack>
-          {response.comment && (
-            <div>
-              <Divider sx={{ mt: 1, mb: 2 }} />
+          <Divider sx={{ mt: 1, mb: 2 }} />
+          {response.comment && !isEditMode && (
```
ARADDCC012, line 123: The logic between here and line 142 looks exactly the same as in `ReviewCommentDisplay`. We should pull this out into a component rather than duplicating code.

frontend/src/reviews/ReviewDecisionDisplay.tsx
```diff
@@ -69,16 +112,41 @@ export default function ReviewDecisionDisplay({ response, modelId }: ReviewDecis
                 <Typography variant='caption'>as {getRoleDisplay(response.role, modelRoles)}</Typography>
               )}
             </Stack>
-            <Typography fontWeight='bold'>{formatDateString(response.createdAt)}</Typography>
+            <Stack direction='row' alignItems='center' spacing={1}>
+              <Typography fontWeight='bold'>{formatDateString(response.createdAt)}</Typography>
+              <IconButton onClick={(event) => setAnchorEl(event.currentTarget)}>
+                <MenuIcon />
```
ARADDCC012, line 118: I can see this is simply following `ReviewCommentDisplay`, but can we use the following icon instead for both components? https://mui.com/material-ui/material-icons/?query=more&selected=MoreHoriz

ARADDCC012 (changes requested):

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -80,18 +83,29 @@ export default function ReviewComments({ release, accessRequest, isEdit }: Revie
     decisionsAndComments.sort(sortByCreatedAtAscending)
     return decisionsAndComments.map((response) => {
       if (response.kind === ResponseKind.Review) {
-        return <ReviewDecisionDisplay key={response.createdAt} response={response} modelId={modelId} />
+        return (
+          <ReviewDecisionDisplay
+            key={response.createdAt}
```
ARADDCC012, line 88: If you do update `ResponseInterface` to have `_id`, we can use that `_id` as a guaranteed unique key.

frontend/src/reviews/ReviewComments.tsx
```diff
@@ -80,18 +83,29 @@ export default function ReviewComments({ release, accessRequest, isEdit }: Revie
...
+            modelId={modelId}
+            onReplyButtonClick={(quote) => setNewReviewComment(`${quote} \n\n ${newReviewComment}`)}
+            currentUser={currentUser}
+            mutateResponses={mutateResponses}
+          />
+        )
       } else {
         return (
           <ReviewCommentDisplay
             key={response.createdAt}
```
ARADDCC012, line 99: Same here, we could use `_id` as a guaranteed unique key.

frontend/src/reviews/ReviewCommentDisplay.tsx
ARADDCC002: I think this is fine for comment responses, but I think it's misleading for review responses. If I approve a release and I see the response at the bottom, and then click the menu icon and see "Edit", I'll assume that means I can also edit my response. Thoughts on keeping it as "Edit comment" for reviews?

frontend/src/reviews/ReviewCommentDisplay.tsx
ARADDCC002: Duh this would be so useful no idea why I didn't add it! :P

frontend/src/reviews/ReviewCommentDisplay.tsx
ARADDCC012: Yep totally agree! I hadn't thought about that.

frontend/src/reviews/EditableReviewComment.tsx
```diff
@@ -0,0 +1,53 @@
+import { Button, Stack, Typography } from '@mui/material'
+import MarkdownDisplay from 'src/common/MarkdownDisplay'
+import RichTextEditor from 'src/common/RichTextEditor'
+import MessageAlert from 'src/MessageAlert'
+import { ResponseInterface } from 'types/types'
+import { formatDateTimeString } from 'utils/dateUtils'
+
+interface EditableReviewCommentProps {
+  comment: string
+  setComment: (value: string) => void
```
ARADDCC012, line 10: `set...` should be reserved for react's set state functions. This is just a callback prop that happens to lead to a set state call, but this component should have no knowledge of how the value is to be used. Rename to `onCommentChange`?

frontend/src/reviews/EditableReviewComment.tsx
```diff
@@ -0,0 +1,53 @@
...
+  response: ResponseInterface
+  isEditMode: boolean
+  setIsEditMode: (value: boolean) => void
```
ARADDCC012, line 13: If an error is thrown when updating a comment, it might be useful to reset the error when "Cancel" is clicked. Consider changing this to an `onCancel` prop and in both `ReviewCommentDisplay` and `ReviewDecisionDisplay` we could both `setIsEditMode(false)` and `setEditCommentErrorMessage('')`. That way the error won't be present when they re-enter edit mode.

frontend/src/reviews/EditableReviewComment.tsx
```diff
@@ -0,0 +1,53 @@
...
+  setComment,
+  response,
+  isEditMode,
+  setIsEditMode,
+  editCommentErrorMessage,
+  onSave,
+}: EditableReviewCommentProps) {
+  return (
+    <>
+      {comment && !isEditMode && (
```
ARADDCC012, line 29: `comment` is defined as a required string, so the check `comment &&` is unnecessary.

ARADDCC012 (changes requested):

frontend/src/reviews/ReviewCommentDisplay.tsx
```diff
@@ -25,6 +38,26 @@ export default function ReviewCommentDisplay({ response, onReplyButtonClick }: R
     }
   }
 
+  const handleEditOnClick = () => {
+    setIsEditMode(true)
+  }
+
+  const handleEditOnCancel = () => {
```
ARADDCC012, line 45: You also need to `setComment(response.comment || '')` in here

frontend/src/reviews/ReviewDecisionDisplay.tsx
```diff
@@ -1,29 +1,75 @@
...
+    if (value) {
+      onReplyButtonClick(value.replace(/^/gm, '>'))
+    }
+  }
+
+  const handleEditOnClick = () => {
+    setIsEditMode(true)
+  }
+
+  const handleEditOnCancel = () => {
```
ARADDCC012, line 57: You also need to `setComment(response.comment || '')` in here

frontend/src/reviews/EditableReviewComment.tsx
```diff
@@ -0,0 +1,53 @@
...
+          {response.updatedAt !== response.createdAt && (
+            <Typography variant='caption' sx={{ fontStyle: 'italic' }}>
+              Edited {formatDateTimeString(response.updatedAt)}
+            </Typography>
+          )}
+        </Stack>
+      )}
+      {isEditMode && (
+        <>
+          <RichTextEditor value={comment} onChange={(input) => onCommentChange(input)} />
```
ARADDCC012, line 41: Super minor but this can just be `onChange={onCommentChange}`

frontend/src/reviews/ReviewCommentDisplay.tsx
```diff
@@ -45,20 +78,25 @@ export default function ReviewCommentDisplay({ response, onReplyButtonClick }: R
             <Stack direction='row' alignItems='center' spacing={1}>
               <Typography fontWeight='bold'>{formatDateString(response.createdAt)}</Typography>
               <IconButton onClick={(event) => setAnchorEl(event.currentTarget)}>
```
ARADDCC012, line 80: This should have an `aria-label='Actions'`

frontend/src/reviews/ReviewDecisionDisplay.tsx
```diff
@@ -69,16 +115,29 @@ export default function ReviewDecisionDisplay({ response, modelId }: ReviewDecis
                 <Typography variant='caption'>as {getRoleDisplay(response.role, modelRoles)}</Typography>
               )}
             </Stack>
-            <Typography fontWeight='bold'>{formatDateString(response.createdAt)}</Typography>
+            <Stack direction='row' alignItems='center' spacing={1}>
+              <Typography fontWeight='bold'>{formatDateString(response.createdAt)}</Typography>
+              <IconButton onClick={(event) => setAnchorEl(event.currentTarget)}>
```
ARADDCC012, line 120: This should have an `aria-label='Actions'`

ARADDCC012 (approved):

PR #1408, opened 2024-07-17
added a better description of general role permissions to UI

ARADDCC002:
image

### DIFF
--- a/frontend/src/entry/model/settings/EntryRolesInfo.tsx
+++ b/frontend/src/entry/model/settings/EntryRolesInfo.tsx
@@ -37,6 +37,21 @@ export default function EntryRolesInfo({ entry }: EntryRolesInfoProps) {
 
   const schemaRolesList = useMemo(() => getFilteredRoles(RoleKind.SCHEMA), [getFilteredRoles])
 
+  const permissionTableRows = useMemo(() => {
+    return rows.map((row) => (
+      <TableRow key={row.permission} sx={{ '&:last-child td, &:last-child th': { border: 0 } }}>
+        <TableCell component='th' scope='row'>
+          {row.permission}
+        </TableCell>
+        <TableCell align='center'>{row.roles.includes('owner') ? <Done /> : <></>}</TableCell>
+        <TableCell align='center'>{row.roles.includes('contributor') ? <Done /> : <></>}</TableCell>
+        <TableCell align='center'>{row.roles.includes('consumer') ? <Done /> : <></>}</TableCell>
+        <TableCell align='center'>{row.roles.includes('releaseReviewer') ? <Done /> : <></>}</TableCell>
+        <TableCell align='center'>{row.roles.includes('accessReviewer') ? <Done /> : <></>}</TableCell>
+      </TableRow>
+    ))
+  }, [])
+
   if (isEntryRolesError) {
     return <MessageAlert message={isEntryRolesError.info.message} severity='error' />
   }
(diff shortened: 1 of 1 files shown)


frontend/src/entry/model/settings/EntryRolesInfo.tsx
```diff
@@ -54,36 +54,105 @@ export default function EntryRolesInfo({ entry }: EntryRolesInfoProps) {
...
         </Stack>
       )}
     </>
   )
 }
+
+const rows = [
+  {
+    permission: 'View private model',
+    roles: ['owner', 'contributor', 'consumer', 'releaseReviewer', 'accessReviewer'],
```
ARADDCC012, line 114: Would be great if we could have a `RoleValue` "enum" for all of these instead of having to make sure we don't typo any of these string arrays. Feel free to rename if you can think of anything better!

frontend/src/entry/model/settings/EntryRolesInfo.tsx
```diff
@@ -54,36 +54,105 @@ export default function EntryRolesInfo({ entry }: EntryRolesInfoProps) {
...
+  {
+    permission: 'Edit or delete an inferencing form',
+    roles: ['owner', 'contributor'],
+  },
+  {
+    permission: 'Request access',
+    roles: ['owner', 'contributor', 'consumer'],
+  },
+  {
+    permission: 'Edit or delete any access request',
```
ARADDCC012, line 145: This is a bit of an awkward one because you can also edit or delete an access request if you are named in the additional contacts of said access request. Maybe we could use a `HelpPopover` in this permission cell to explain that?

Something like this perhaps?
"You can also edit/delete an access request if you are an Additional Contact"

frontend/src/entry/model/settings/EntryRolesInfo.tsx
```diff
@@ -34,9 +35,23 @@ export default function EntryRolesInfo({ entry }: EntryRolesInfoProps) {
...
-  const entryRolesList = useMemo(() => getFilteredRoles(RoleKind.ENTRY), [getFilteredRoles])
   const schemaRolesList = useMemo(() => getFilteredRoles(RoleKind.SCHEMA), [getFilteredRoles])
 
+  const permissionTableRows = useMemo(() => {
+    return rows.map((row) => (
+      <TableRow key={row.permission} sx={{ '&:last-child td, &:last-child th': { border: 0 } }}>
+        <TableCell component='th' scope='row'>
+          {row.permission}
+        </TableCell>
+        <TableCell align='center'>{row.roles.includes('owner') ? <Done /> : <></>}</TableCell>
```
ARADDCC012, line 46: Again, would be good to be using the same RoleValue "enum" for these strings 'owner', etc

ARADDCC012 (changes requested):

frontend/src/entry/model/settings/EntryRolesInfo.tsx
```diff
@@ -108,53 +138,60 @@ export default function EntryRolesInfo({ entry }: EntryRolesInfoProps) {
...
   {
     permission: 'Request access',
-    roles: ['owner', 'contributor', 'consumer'],
+    roles: [RoleValue.OWNER, RoleValue.CONTRIBUTOR, RoleValue.CONSUMER],
   },
   {
     permission: 'Edit or delete any access request',
-    roles: ['owner'],
+    roles: [RoleValue.OWNER],
+    helpText: 'You can also edit/delete an access request if you an Additional Contact',
```
ARADDCC012, line 183: Typo here. Should say "if you **are** an Additional Contact"

frontend/src/entry/model/settings/EntryRolesInfo.tsx
```diff
@@ -37,17 +52,32 @@ export default function EntryRolesInfo({ entry }: EntryRolesInfoProps) {
 
   const schemaRolesList = useMemo(() => getFilteredRoles(RoleKind.SCHEMA), [getFilteredRoles])
 
+  const showPopover = (row: Row) => {
```
ARADDCC012, line 55: `showPopover` and `canRoleDoAction` are just components written as normal functions. Consider doing the following (but feel free to keep them in the same file as they're very specific to `EntryRolesInfo`):
```
type RowHelpTextProps = {
  row: Row
}

function RowHelpText({ row }: RowHelpTextProps) {
  if (!row.helpText) {
    return null
  }

  return <HelpPopover>{row.helpText}</HelpPopover>
}

type RoleActionIconProps = {
  role: RoleValue
  roles: RoleValue[]
}

function RoleActionIcon({ role, roles }: RoleActionIconProps) {
  if (!roles.includes(role)) {
    return null
  }

  return <Done />
}
```
They would be used like so:
```
          <Stack direction='row' alignItems='center'>
            {row.permission}
            <RowHelpText row={row} />
          </Stack>
```
```
        <TableCell align='center'>
          <RoleActionIcon role={RoleValue.OWNER} roles={row.roles} />
        </TableCell>
```

ARADDCC012 (approved):

PR #1412, opened 2024-07-18
Added slide up transitions to all the current dialog modals.

IR96334:


frontend/src/common/ConfirmationDialogue.tsx
```diff
@@ -15,6 +18,15 @@ type ConfirmationDialogProps = {
   errorMessage?: string
 }
 
+const Transition = forwardRef(function Transition(
```
a3957273, line 21: Consider putting the `transition` in a single place and importing it, as this code is repeated verbatim in half a dozen places.

frontend/src/common/ConfirmationDialogue.tsx
```diff
@@ -24,7 +36,7 @@ export default function ConfirmationDialogue({
   dialogMessage = 'Are you sure you want to perform this action?',
 }: ConfirmationDialogProps) {
   return (
-    <Dialog fullWidth open={open} onClose={onCancel}>
+    <Dialog fullWidth open={open} onClose={onCancel} keepMounted disableEscapeKeyDown TransitionComponent={Transition}>
```
a3957273, line 39: We should ensure that we are WCAG 2.0 AA compliant, which requires the ability to use common key combos to escape dialogs.

a3957273 (changes requested):

frontend/src/common/ConfirmationDialogue.tsx
IR96334: I have removed this to be compliant

frontend/src/common/ConfirmationDialogue.tsx
IR96334: Thrown into utils folder

frontend/utils/transitions.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { TransitionProps } from '@mui/material/transitions'
+import { forwardRef } from 'react'
+
+export const Transition = forwardRef(function Transition(
+  props: TransitionProps & {
+    children: React.ReactElement<any, any>
+  },
+  ref: React.Ref<unknown>,
+) {
+  return <Slide direction={'up'} ref={ref} {...props} />
```
ARADDCC002, line 11: It would be nice if we could provide a direction as a prop, just in case we want to use different directions in certain situations.

We should leave `up` as the default though, that way we don't need to alter any of your other changes.

ARADDCC012: I don't know if I agree. The point of using the same transition component across the board was to keep everything feeling consistent. If we start allowing the direction to be customisable, we are moving away from the aim of being consistent.

ARADDCC012 (changes requested):

frontend/utils/transitions.tsx
ARADDCC002: I'm convinced by @ARADDCC012's comment, @IR96334 feel free to disregard my comment!

frontend/utils/transitions.tsx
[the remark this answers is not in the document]
IR96334: Done

ARADDCC002 (approved):

ARADDCC012 (approved):

PR #1419, opened 2024-07-29
added backend implementation for model templating

ARADDCC002:


GB27247 (changes requested):

backend/src/services/model.ts
```diff
@@ -347,9 +346,28 @@ export async function createModelCardFromSchema(
...
-  _schemaId: string,
+  user: UserInterface,
+  modelId: string,
+  templateId: string,
 ): Promise<ModelCardRevisionDoc> {
-  throw NotImplemented({}, 'This feature is not yet implemented')
+  const model = await getModelById(user, modelId)
+  checkModelRestriction(model)
+  const template = await getModelById(user, templateId)
+  checkModelRestriction(template)
```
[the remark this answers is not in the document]
ARADDCC002: I think it's necessary to check that new model is not a mirrored model - there could be an instance that someone makes a mirrored model but doesn't create a model card, then tries to base it off a template.

As we are not changing the metadata of the template, however, I think we can safely remove this check.

backend/src/services/model.ts
```diff
@@ -347,9 +346,28 @@ export async function createModelCardFromSchema(
...
+  const modelAuth = await authorisation.model(user, model, ModelAction.Write)
+  if (!modelAuth.success) {
+    throw Forbidden(modelAuth.info, { userDn: user.dn, modelId })
+  }
+  const templateAuth = await authorisation.model(user, template, ModelAction.View)
+  if (!templateAuth.success) {
+    throw Forbidden(templateAuth.info, { userDn: user.dn, modelId })
+  }
+
+  if (!template.card) {
```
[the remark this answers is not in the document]
ARADDCC002: When you first create a model `card` doesn't exist - I'm not sure where you're seeing it being set as an empty object, but if it is somehow being set let me know!

backend/src/services/model.ts
```diff
@@ -347,9 +346,27 @@ export async function createModelCardFromSchema(
...
+  user: UserInterface,
+  modelId: string,
+  templateId: string,
 ): Promise<ModelCardRevisionDoc> {
-  throw NotImplemented({}, 'This feature is not yet implemented')
+  if (modelId === templateId) {
+    throw BadReq('The model and template ID must be different', { modelId, templateId })
+  }
+  const model = await getModelById(user, modelId)
+  checkModelRestriction(model)
```
ARADDCC012, line 357: Do we need to `checkModelRestriction` for the model being used as a template also?

backend/src/services/model.ts
```diff
@@ -347,9 +346,27 @@ export async function createModelCardFromSchema(
...
+  const template = await getModelById(user, templateId)
+
+  // Check to make sure user can access the template. We already check for the model auth later on in _setModelCard
+  const templateAuth = await authorisation.model(user, template, ModelAction.View)
+  if (!templateAuth.success) {
+    throw Forbidden(templateAuth.info, { userDn: user.dn, modelId })
```
ARADDCC012, line 363: Shouldn't this be reporting the `templateId`?
```
throw Forbidden(templateAuth.info, { userDn: user.dn, modelId: templateId })
```

backend/src/services/model.ts
```diff
@@ -347,9 +346,27 @@ export async function createModelCardFromSchema(
...
-  _user: UserInterface,
-  _modelId: string,
-  _schemaId: string,
+  user: UserInterface,
+  modelId: string,
+  templateId: string,
 ): Promise<ModelCardRevisionDoc> {
-  throw NotImplemented({}, 'This feature is not yet implemented')
+  if (modelId === templateId) {
+    throw BadReq('The model and template ID must be different', { modelId, templateId })
```
ARADDCC012, line 354: We should have a unit test that covers this

backend/src/services/model.ts
ARADDCC012, line 363: We should have a unit test that covers this

backend/test/services/model.spec.ts
```diff
@@ -311,4 +312,17 @@ describe('services > model', () => {
     )
     expect(modelMocks.save).not.toBeCalled()
   })
+
+  test('crateModelCardFromTemplate > requesting to use a template without a model card will throw an error', async () => {
```
ARADDCC012, line 316: We should also have a "happy day" test

ARADDCC012 (changes requested):

backend/src/services/model.ts
ARADDCC002: I don't think so. All it does is check to see if it's "mirrored" which is basically read only. As we're not editing the template in any way, I think it's fine to use it if it's read only.

backend/src/services/model.ts
```diff
@@ -347,9 +346,25 @@ export async function createModelCardFromSchema(
...
+  // Check to make sure user can access the template. We already check for the model auth later on in _setModelCard
+  const templateAuth = await authorisation.model(user, template, ModelAction.View)
+  if (!templateAuth.success) {
+    throw Forbidden(templateAuth.info, { userDn: user.dn, templateId })
+  }
+
+  if (!template.card?.schemaId) {
+    throw BadReq('The template model is missing a model card', { modelId, templateId })
+  }
+  const revision = await _setModelCard(user, modelId, template.card.schemaId, 1, template.card.metadata)
```
JR40159, line 368: Is there anything to stop a user from calling this with a `modelId` that references a model that already has model card versions? If this were to happen then this line would set the first version to be the model card of the template and all later versions would be from the pre-existing model card.

JR40159 (changes requested):

JR40159 (approved):

PR #1420, opened 2024-07-29
updated email headers

ARADDCC002:


backend/src/services/smtp/emailBuilder.ts
```diff
@@ -17,9 +17,9 @@ export type actions = {
   url: string
 }
 
-export function buildEmail(title: string, metadata: Info[], actions: actions[]): EmailContent {
+export function buildEmail(title: string, metadata: Info[], actions: actions[], subject?: string): EmailContent {
   return {
-    subject: title,
+    subject: subject || title,
```
ARADDCC012, line 22: Instead of having to duplicate passing in the title with "Action Required" prepended, can we instead replace `subject` with an optional `actionRequired` boolean prop?

This not only saves us from typing out the same text twice, but also defines what an "action required" subject looks like in one place.

You could then do:
```
subject: actionRequired ? `ACTION REQUIRED: ${title}` : title
```

ARADDCC012 (changes requested):

ARADDCC012 (approved):

PR #1424, opened 2024-07-31
Feature/bai 1343 add announcement banner

ARADDCC002:


GB27247 (changes requested):

frontend/src/Annoucement.tsx
```diff
@@ -0,0 +1,79 @@
...
+            <Button variant='text' size='small' onClick={() => setShowFullText(!showFullText)}>
+              {showFullText ? 'Show less' : 'Show more'}
+            </Button>
+          </span>
+        </Typography>
+      </Stack>
+    ) : (
+      message
+    )
+  }
```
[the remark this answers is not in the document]
ARADDCC002: This is a mistake and should be memozied, updated!

frontend/src/Wrapper.tsx
```diff
@@ -105,6 +121,16 @@ export default function Wrapper({ children }: WrapperProps): ReactElement {
               </>
             )}
           </Box>
+          {isUiConfigLoading && <Loading />}
+          {uiConfig && (
+            <Box sx={{ mb: 2 }}>
+              {annoucementBannerOpen ? (
+                <Announcement message={uiConfig.announcement.text} onClose={handleAnnouncementOnClose} />
```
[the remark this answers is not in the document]
ARADDCC002: We discussed this and chose to go at the bottom to keep it consistent with other applications with similar messages. If enough users prefer it moved we can look to doing that.

GB27247 (dismissed):

frontend/src/Banner.tsx
```diff
@@ -24,7 +25,11 @@ function Banner() {
     return null
   }
 
-  return <Card sx={style}>{uiConfig.banner.text}</Card>
+  return (
+    <Card sx={style}>
+      <Stack>
```
ARADDCC012, line 30: This stack isn't needed as it only has one child

frontend/src/Wrapper.tsx
```diff
@@ -17,6 +19,8 @@ export type WrapperProps = {
   children?: ReactNode
 }
 
+const DISMISSED_COOKIE_NAME = 'dismissed-timestamp'
```
ARADDCC012, line 22: Can we move this into `frontend/utils/constants.ts`?

frontend/src/Wrapper.tsx
```diff
@@ -44,6 +51,16 @@ export default function Wrapper({ children }: WrapperProps): ReactElement {
     }
   }, [isUiConfigLoading, uiConfig, isDocsPage])
 
+  useEffect(() => {
+    if (uiConfig) {
+      setAnnouncementBannerOpen(
+        uiConfig &&
```
ARADDCC012, line 57: We've already checked `uiConfig` is truthy above so this check isn't needed

frontend/src/Wrapper.tsx
```diff
@@ -2,8 +2,10 @@ import Box from '@mui/material/Box'
 import { useTheme } from '@mui/material/styles'
 import Toolbar from '@mui/material/Toolbar'
 import { useGetUiConfig } from 'actions/uiConfig'
+import Cookies from 'js-cookie'
```
ARADDCC012, line 5: I know the library uses `Cookies` in their examples, but can we use `cookies` as it then follows our naming conventions?

frontend/src/Announcement.tsx
```diff
@@ -0,0 +1,80 @@
...
+          </Button>
+        </Typography>
+      </Stack>
+    ) : (
+      message
+    )
+  }, [message, showFullText])
+
+  return (
+    <>
```
ARADDCC012, line 32: This fragment isn't needed

frontend/src/Announcement.tsx
```diff
@@ -0,0 +1,80 @@
...
+          bottom: 0,
+          mb: 4,
+          borderStyle: 'solid',
+          borderWidth: 2,
+          borderColor: theme.palette.primary.main,
+          borderRadius: 1,
+          left: 0,
+          right: 0,
+          maxWidth: 'md',
+          ml: 'auto',
```
ARADDCC012, line 47: Could combine `ml` and `mr` by using `mx: 'auto'`

frontend/src/Announcement.tsx
```diff
@@ -0,0 +1,80 @@
...
+export default function Announcement({ message, onClose }: AnnoucementProps) {
+  const theme = useTheme()
+
+  const [showFullText, setShowFullText] = useState(false)
+
+  const announcementText = useMemo(() => {
+    return message.length > 100 ? (
+      <Stack>
+        <Typography>
+          {!showFullText ? `${message.slice(0, 100)}...` : message}
```
ARADDCC012, line 20: Can we reverse the logic here? It's a little confusing to have a ternary written this way round. Consider:
```
{showFullText ? message : `${message.slice(0, 100)}...`}
```

frontend/src/Wrapper.tsx
```diff
@@ -105,6 +127,14 @@ export default function Wrapper({ children }: WrapperProps): ReactElement {
               </>
             )}
           </Box>
+          {isUiConfigLoading && <Loading />}
+          {uiConfig && (
+            <Box sx={{ mb: 2 }}>
```
ARADDCC012, line 132: Is this box necessary? I don't think we want the banner to take up any space from the existing content, we want it to sit on top of it.

Also consider combining the two conditional blocks into:
```
{uiConfig && annoucementBannerOpen && ( ... )}
```

ARADDCC012 (changes requested):

ARADDCC012 (approved):

PR #1450, opened 2024-08-05
added a subheading and copy button to the pagewithtabs component

ARADDCC002:
image

### DIFF
--- a/frontend/pages/data-card/[dataCardId].tsx
+++ b/frontend/pages/data-card/[dataCardId].tsx
@@ -56,9 +56,11 @@ export default function DataCard() {
         <PageWithTabs
           showCopyButton
           title={dataCard.name}
+          subheading={`ID: ${dataCard.id}`}
           tabs={tabs}
           requiredUrlParams={{ dataCardId: dataCard.id }}
-          textToCopy={dataCard.id}
+          titleToCopy={dataCard.name}
+          subheadindToCopy={dataCard.id}
         />
       )}
     </>
--- a/frontend/pages/data-card/[dataCardId].tsx
+++ b/frontend/pages/data-card/[dataCardId].tsx
@@ -54,7 +54,6 @@ export default function DataCard() {
       {(isDataCardLoading || isCurrentUserLoading) && <Loading />}
       {dataCard && (
         <PageWithTabs
-          showCopyButton
           title={dataCard.name}
           subheading={`ID: ${dataCard.id}`}
           tabs={tabs}
--- a/frontend/pages/data-card/[dataCardId].tsx
+++ b/frontend/pages/data-card/[dataCardId].tsx
@@ -60,7 +60,7 @@ export default function DataCard() {
           tabs={tabs}
           requiredUrlParams={{ dataCardId: dataCard.id }}
           titleToCopy={dataCard.name}
-          subheadindToCopy={dataCard.id}
+          subheadingToCopy={dataCard.id}
         />
       )}
     </>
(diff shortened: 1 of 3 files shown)


frontend/pages/data-card/[dataCardId].tsx
```diff
@@ -56,9 +56,11 @@ export default function DataCard() {
         <PageWithTabs
           showCopyButton
           title={dataCard.name}
+          subheading={`ID: ${dataCard.id}`}
```
Comment (GB27247) [inline frontend/pages/data-card/[dataCardId].tsx:59]: Do we wan't the same with Access Requests and Schemas?

frontend/src/common/PageWithTabs.tsx
```diff
@@ -19,30 +19,35 @@ export interface PageTab {
...
   actionButtonOnClick,
   requiredUrlParams = {},
   showCopyButton = false,
-  textToCopy = '',
+  titleToCopy = '',
+  subheadingToCopy = '',
   sourceModelId = '',
 }: {
   title: string
+  subheading?: string
```
GB27247, line 34: I think this as an optional type looks like a typo?

frontend/src/common/PageWithTabs.tsx
```diff
@@ -19,30 +19,35 @@ export interface PageTab {
...
   tabs: PageTab[]
   actionButtonTitle?: string
   displayActionButton?: boolean
   actionButtonOnClick?: () => void
   requiredUrlParams?: ParsedUrlQuery
   showCopyButton?: boolean
-  textToCopy?: string
+  titleToCopy?: string
+  subheadingToCopy?: string
```
GB27247, line 42: Is it necessary to have extra `...ToCopy` props since they're likely just repeats of existing props? I realise it's done this way already with the title?

frontend/src/common/PageWithTabs.tsx
```diff
@@ -116,19 +121,33 @@ export default function PageWithTabs({
...
+          {subheading && (
+            <Stack direction='row' alignItems='center'>
+              <Typography variant='caption' sx={{ color: darken(theme.palette.primary.main, 0.4) }}>
+                {subheading}
+              </Typography>
+              <CopyToClipboardButton
+                textToCopy={subheadingToCopy ? subheadingToCopy : subheading}
+                notificationText='Copied to clipboard'
+                ariaLabel='copy to clipboard'
+              />
```
GB27247, line 149: Does this component need to be conditional on `showCopyButton`, like with the title page?

GB27247 (changes requested):

frontend/pages/data-card/[dataCardId].tsx
Comment (ARADDCC002) [reply frontend/pages/data-card/[dataCardId].tsx:59]: They don't use this component in the same way - the Schemas page for example just uses the static title "Schemas", the reason we need this change for entries is because they have dynamic names and IDs

frontend/src/common/PageWithTabs.tsx
ARADDCC002: Not a typo - we don't always need a subheading eg the schemas page.

frontend/src/common/PageWithTabs.tsx
ARADDCC002: This is something I thought about, and you _are_ right when it comes to the title for models and data cards, but it's slightly different for the subheading. In my current implementation, the subheading is "ID: my-model-123", but I don't want to copy to clipboard the bit that says "ID:". Equally, we don't want to hard code "ID:" into the `PageWithTabs` component, because subheadings might not always be IDs, they can be any string value.

frontend/src/common/PageWithTabs.tsx
ARADDCC002: Yes it does, good spot!

PR #1451, opened 2024-08-05
added release note validation to put and post release

ARADDCC002:
image


GB27247 (commented):
Just to note. This is not reflected on the Python Client I'm happy to resolve that. Also the front end edit form does not validate against this change.

ARADDCC002:
> Just to note. This is not reflected on the Python Client I'm happy to resolve that. Also the front end edit form does not validate against this change.

That's strange.. I updated `putRelease` to have the same validation but you're right that the UI allows for empty strings.

GB27247 (approved):

PR #1454, opened 2024-08-05
Feature/bai 1214 add metrics widget

ARADDCC002:
image

image


frontend/utils/stringUtils.ts
```diff
@@ -38,6 +38,11 @@ export const isValidPortNumber = (portNumber: string) => {
   )
 }
 
+export const isValidNumber = (numberOrString: string) => {
+  const numericString = Number(numberOrString)
+  return !isNaN(numericString) && Number.isInteger(numericString)
```
GB27247, line 43: I interpret metrics as floating point numbers as well?

frontend/src/MuiForms/Metrics.tsx
```diff
@@ -0,0 +1,170 @@
...
+      onChange(updatedMetricArray.map((metric) => ({ name: metric.name, value: metric.value })))
+    },
+    [metricsWithIds, onChange],
+  )
+
+  const handleAndValidateMetricValue = useCallback(
+    (metric: MetricValueWithId, value: string) => {
+      let valueAsStringOrNumber: string | number = value
+      setErrorMessage('')
+      if (!isValidNumber(value) || value.length === 0) {
```
GB27247, line 78: Since this only used in one place could we combine these conditions?

backend/src/scripts/example_schemas/minimal_model_schema.json
```diff
@@ -23,6 +23,25 @@
...
+              "name": {
+                "title": "Metric name",
+                "type": "string"
+              },
+              "value": {
+                "title": "Model performance metric value",
+                "type": "number"
+              }
+            }
+          }
```
GB27247, line 43: Should this have unique items?

GB27247 (changes requested):

backend/src/scripts/example_schemas/minimal_model_schema.json
ARADDCC002: I don't think that'll work because they're objects

frontend/src/MuiForms/MetricItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export default function MetricItem({ metric, onChange, onDelete }: MetricItemProps) {
+  const theme = useTheme()
+  const [errorMessage, setErrorMessage] = useState('')
+
+  const handleMetricValueOnChange = useCallback(
+    (value: string) => {
+      const valueAsStringOrNumber: string | number = value
+      parseInt(valueAsStringOrNumber)
```
GB27247, line 23: I don't think this is useful until the value is validated. Also `parseInt` will convert an empty string to `NaN` which is invalid and makes means you have to delete the entry.

frontend/src/MuiForms/MetricItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+      setErrorMessage('')
+      if (!isValidNumber(value)) {
+        setErrorMessage('Metric value must be a valid whole number')
+      }
+      onChange({ id: metric.id, name: metric.name, value: valueAsStringOrNumber })
```
GB27247, line 28: This should be in an else block. Also with the current implementation `valueAsStringOrNumber` will never be a string.

GB27247 (changes requested):

frontend/src/MuiForms/MetricItem.tsx
ARADDCC002: I've updated this, it effectively stops users from entering letters altogether, which might _not_ be a bad thing given the libraries existing constraints.

frontend/src/MuiForms/MetricItem.tsx
ARADDCC002: I've removed it as it wasn't needed, but I didn't see it cause `Nan` issues when entering strings before weirdly.

frontend/src/MuiForms/Metrics.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  const handleMetricItemOnChange = useCallback(
+    (updatedMetricItem: MetricValueWithId) => {
+      const updatedMetricArray = _.cloneDeep(metricsWithIds)
+      const index = metricsWithIds.findIndex((metric) => metric.id === updatedMetricItem.id)
+      updatedMetricArray[index] = updatedMetricItem
+      onChange(
+        updatedMetricArray.map((metric) => ({
+          name: metric.name,
+          value: isValidNumber(metric.value as string) ? parseInt(metric.value as string) : metric.value,
```
GB27247, line 67: `isValidNumber('1e+12')` is truthy as `1e+12` is casted to a `Number` however `parseInt('1e+12')` is `1`. So if a user enters a large number it'll change back to a small number.

GB27247 (changes requested):

GB27247 (approved):

PR #1455, opened 2024-08-06
Bai 1376 display entry collaborators outside of the settings tab

IR96334:


frontend/src/entry/overview/EntryCardRolesDialog.tsx
```diff
@@ -0,0 +1,22 @@
...
+type EntryCardRoleDialogProps = {
+  entry: EntryInterface
+  open: boolean
+  setOpen: (isOpen: boolean) => void
+}
+
+export default function EntryCardRolesDialog({ entry, open, setOpen }: EntryCardRoleDialogProps) {
+  return (
+    <>
+      <Dialog open={open} onClose={() => setOpen(false)} fullWidth maxWidth='sm' TransitionComponent={Transition}>
```
ARADDCC012, line 16: Entity names can get quite long in some cases so I reckon we should up this to `maxWidth='md'`

frontend/src/entry/overview/EntryCardRolesDialog.tsx
```diff
@@ -0,0 +1,22 @@
...
+        <DialogTitle>Roles</DialogTitle>
+        <EntryRoleList entry={entry} />
```
ARADDCC012, line 18: This should be wrapped in a `<DialogContent>` which handles the padding for you.
```
        <DialogContent>
          <EntryRoleList entry={entry} />
        </DialogContent>
```

frontend/src/entry/overview/EntryRoleList.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { EntryInterface } from 'types/types'
+
+type EntryCardRoleDialogProps = {
+  entry: EntryInterface
+}
+
+export default function EntryRoleList({ entry }: EntryCardRoleDialogProps) {
+  return (
+    <>
+      <Grid container spacing={2} paddingX={5} paddingBottom={2}>
```
ARADDCC012, line 12: `paddingX` and `paddingBottom` can be removed. This padding will be handled by using `<DialogContent>` as suggested in another comment.

frontend/src/entry/overview/EntryCardEntityRoleRow.tsx
```diff
@@ -0,0 +1,45 @@
...
+      </Grid>
+    </Fragment>
+  ))
+}
+
+type EntryCollaboratorProps = {
+  entity: CollaboratorEntry
+}
+
+function EntryCollaboratorIcon({ entity }: EntryCollaboratorProps) {
```
ARADDCC012, line 33: Instead of duplicating code by adding `EntryCollaboratorIcon` and `EntryCollaboratorName` in here you should instead move `EntityIcon` and `EntityNameDisplay` from `frontend/src/entry/settings/EntityItem.tsx` into new files in `frontend/src/entry/EntityIcon.tsx` and `frontend/src/entry/EntityIcon.tsx` respectively. I would then export them from there and use them here and in `EntityItem.tsx`.

While you're doing that I also suggest fixing the props. They're collaborator entries, not entities. So consider renaming them from `entity: CollaboratorEntry` to `entryCollaborator: CollaboratorEntry`.

frontend/src/entry/overview/EntryCardEntityRoleRow.tsx
```diff
@@ -0,0 +1,45 @@
...
+import UserDisplay from 'src/common/UserDisplay'
+import EntryCardRolesChipSet from 'src/entry/overview/EntryCardRolesChipSet'
+import { CollaboratorEntry } from 'types/types'
+
+type EntryCardEntityRoleRowProps = {
+  entryCollaborators: CollaboratorEntry[]
+}
+
+export default function EntryCardEntityRoleRow({ entryCollaborators }: EntryCardEntityRoleRowProps) {
+  return entryCollaborators.map((user) => (
```
ARADDCC012, line 14: Consider renaming `user` to `collaborator`. Collaborators can be groups, so calling them users isn't always true.

frontend/src/entry/overview/EntryRoleList.tsx
```diff
@@ -0,0 +1,23 @@
+import { Grid } from '@mui/material'
+import EntryCardEntityRoleRow from 'src/entry/overview/EntryCardEntityRoleRow'
+import { EntryInterface } from 'types/types'
+
+type EntryCardRoleDialogProps = {
+  entry: EntryInterface
+}
+
+export default function EntryRoleList({ entry }: EntryCardRoleDialogProps) {
```
ARADDCC012, line 9: I don't see the value in splitting up into quite so many components, though it's definitely better to overuse components than underuse them! Consider combining this and `EntryCardRoleReview`:
```
export default function EntryRoleList({ entry }: EntryCardRoleDialogProps) {
  const rows = useMemo(
    () =>
      entry.collaborators.map((collaborator) => (
        <Fragment key={collaborator.entity}>
          <Grid item xs={6}>
            <Stack direction='row' alignItems='center' spacing={1}>
              <EntryCollaboratorIcon entryCollaborator={collaborator} />
              <EntryCollaboratorName entryCollaborator={collaborator} />
            </Stack>
          </Grid>
          <Grid item xs={6}>
            <EntryCardRolesChipSet entryCollaborator={collaborator} />
          </Grid>
        </Fragment>
      )),
    [entry.collaborators],
  )

  return (
    <Grid container spacing={2}>
      <Grid item xs={6}>
        Entity
      </Grid>
      <Grid item xs={6}>
        Roles
      </Grid>
      {rows}
    </Grid>
  )
}
```

frontend/src/entry/overview/EntryCardEntityRoleRow.tsx
```diff
@@ -0,0 +1,45 @@
...
+  entity: CollaboratorEntry
+}
+
+function EntryCollaboratorIcon({ entity }: EntryCollaboratorProps) {
+  const isUser = useMemo(() => entity.entity.startsWith('user:'), [entity])
+  return isUser ? <UserIcon color='primary' /> : <GroupIcon color='secondary' />
+}
+
+function EntryCollaboratorName({ entity }: EntryCollaboratorProps) {
+  const entityName = useMemo(() => entity.entity.split('/').pop()!, [entity])
```
ARADDCC012, line 39: I've already suggested reusing `EntityNameDisplay` which will fix this issue. However I'm not quite sure what you're trying to do with:
```
const entityName = useMemo(() => entity.entity.split('/').pop()!, [entity])
```
It's confusing, but our entities are "magic" strings in the format "kind:dn", so the existing logic in `EntityNameDisplay` is exactly what you'll want to use (hence the suggestion to simply reuse it).

frontend/src/entry/overview/EntryCardRolesChipSet.tsx
```diff
@@ -0,0 +1,22 @@
+import Dangerous from '@mui/icons-material/DangerousOutlined'
+import { Chip } from '@mui/material'
+import { CollaboratorEntry } from 'types/types'
+
+type EntryCardChipSetProps = {
+  entryCollaborator: CollaboratorEntry
+}
+
+export default function EntryCardRolesChipSet({ entryCollaborator }: EntryCardChipSetProps) {
```
ARADDCC012, line 9: A few things here:
- Always use `sx` instead of `style` for MUI components. `sx` offers the same as `style`, but adds some MUI DX (developer experience) shortcuts.
- We should always be using colours defined in our theme, instead of defining colours inside components.
- The styling can be simplified a fair bit while still achieving the same effect.
- Having only one of the chips showing an icon looks a bit strange.
- We should `useMemo` the `.map` as mentioned in another comment (just a good habit to get into, though very minor in this specific case)

I've addressed all of the above here:
```
import { Chip } from '@mui/material'
import { useMemo } from 'react'
import { CollaboratorEntry } from 'types/types'

type EntryCardChipSetProps = {
  entryCollaborator: CollaboratorEntry
}

export default function EntryCardRolesChipSet({ entryCollaborator }: EntryCardChipSetProps) {
  const roleChips = useMemo(
    () => entryCollaborator.roles.map((role) => <Chip key={role} label={role} color='primary' sx={{ mr: 1 }} />),
    [entryCollaborator.roles],
  )

  return roleChips.length ? roleChips : <Chip label='No roles' color='warning' />
}

```

frontend/src/entry/overview/EntryCardEntityRoleRow.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <>
```
ARADDCC012, line 41: Again this will be solved by reusing `EntityNameDisplay`, but this fragment isn't needed.

The reason why we use fragments is because in react you must always return only one root element from any function. If we find ourselves in a situation where we have multiple root elements then that's where we need to use a fragment. You could also use something like a `div`, but fragments have the benefit of being removed by react before the DOM is updated, so they're always preferred in these scenarios.

Here's an example of the above:
```
// Invalid
return (
  <div>Hello</div>
  <div>World</div>
)

// Valid
return (
  <>
    <div>Hello</div>
    <div>World</div>
  </>
)

// Also valid
return (
  <Fragment>
    <div>Hello</div>
    <div>World</div>
  </Fragment>
)
```

ARADDCC012 (changes requested):

frontend/src/entry/overview/EntryRolesDialog.tsx
```diff
@@ -0,0 +1,22 @@
+import { DialogContent, DialogTitle } from '@mui/material'
+import Dialog from '@mui/material/Dialog'
+import { Transition } from 'src/common/Transition'
+import EntryRoleList from 'src/entry/overview/EntryRoleList'
+import { EntryInterface } from 'types/types'
+
+type EntryRoleDialogProps = {
```
ARADDCC012, line 7: Should be `EntryRolesDialogProps`, you're missing an "s" :)

frontend/src/entry/overview/FormEditPage.tsx
```diff
@@ -8,7 +8,7 @@ import Loading from 'src/common/Loading'
 import TextInputDialog from 'src/common/TextInputDialog'
 import UnsavedChangesContext from 'src/contexts/unsavedChangesContext'
 import EntryCardHistoryDialog from 'src/entry/overview/EntryCardHistoryDialog'
-import EntryCardRolesDialog from 'src/entry/overview/EntryCardRolesDialog'
+import EntryCardRolesDialog from 'src/entry/overview/EntryRolesDialog'
```
ARADDCC012, line 11: `EntryCardRolesDialog` should be renamed to match the new name of the component: `EntryRolesDialog`

frontend/src/entry/overview/EntryRoleList.tsx
```diff
@@ -1,23 +1,61 @@
...
-import { EntryInterface } from 'types/types'
+import GroupIcon from '@mui/icons-material/Groups'
+import UserIcon from '@mui/icons-material/Person'
+import { Grid, Stack } from '@mui/material'
+import { Fragment, useMemo } from 'react'
+import UserDisplay from 'src/common/UserDisplay'
+import EntryCardRolesChipSet from 'src/entry/overview/EntryCardRolesChipSet'
+import { CollaboratorEntry, EntryInterface } from 'types/types'
 
 type EntryCardRoleDialogProps = {
```
ARADDCC012, line 9: Should be `EntryRoleListProps`

frontend/src/entry/overview/ExportEntryCardDialog.tsx
```diff
@@ -60,7 +61,7 @@ export default function ExportEntryCardDialog({ entry, splitSchema, open, setOpe
   }, [splitSchema.steps])
 
   return (
-    <Dialog open={open} onClose={() => setOpen(false)} fullWidth maxWidth='md'>
+    <Dialog open={open} onClose={() => setOpen(false)} fullWidth maxWidth='md' TransitionComponent={Transition}>
```
ARADDCC012, line 64: Good catch!

frontend/src/entry/overview/EntryRoleList.tsx
```diff
@@ -1,23 +1,61 @@
...
+      </Grid>
+      <Grid item xs={6}>
+        Roles
       </Grid>
+      {rows}
+    </Grid>
+  )
+}
+
+type EntryCollaboratorProps = {
```
ARADDCC012, line 45: This can be deleted now

frontend/src/entry/overview/EntryRoleList.tsx
```diff
@@ -1,23 +1,61 @@
...
+  entryCollaborator: CollaboratorEntry
+}
+
+function EntryCollaboratorIcon({ entryCollaborator }: EntryCollaboratorProps) {
```
ARADDCC012, line 49: This can be deleted now

frontend/src/entry/overview/EntryRoleList.tsx
```diff
@@ -1,23 +1,61 @@
...
+  const isUser = useMemo(() => entryCollaborator.entity.startsWith('user:'), [entryCollaborator])
+  return isUser ? <UserIcon color='primary' /> : <GroupIcon color='secondary' />
+}
+
+function EntryCollaboratorName({ entryCollaborator }: EntryCollaboratorProps) {
```
ARADDCC012, line 54: This can be deleted now

frontend/src/entry/overview/EntryRoleList.tsx
```diff
@@ -1,23 +1,61 @@
...
 }
 
 export default function EntryRoleList({ entry }: EntryCardRoleDialogProps) {
+  const rows = useMemo(
+    () =>
+      entry.collaborators.map((collaborator) => (
+        <Fragment key={collaborator.entity}>
+          <Grid item xs={6}>
+            <Stack direction='row' alignItems='center' spacing={1}>
+              <EntryCollaboratorIcon entryCollaborator={collaborator} />
```
ARADDCC012, line 20: Sorry my earlier comments weren't clear. I didn't change the following in the code snippet I provided, even though I'd asked for it in another comment.

You should be using the new `EntityIcon` that you moved into `frontend/src/entry/EntityIcon.tsx`
```
// At the top of this file
import EntityIcon from 'src/entry/EntityIcon`

// Then replace this line with the following
<EntityIcon entryCollaborator={collaborator} />
```

frontend/src/entry/overview/EntryRoleList.tsx
```diff
@@ -1,23 +1,61 @@
...
+              <EntryCollaboratorName entryCollaborator={collaborator} />
```
ARADDCC012, line 21: Again, apologies for the confusion.

You should be using the new `EntityNameDisplay` that you moved into `frontend/src/entry/EntityNameDisplay.tsx`
```
// At the top of this file
import EntityNameDisplay from 'src/entry/EntityNameDisplay`

// Then replace this line with the following
<EntityNameDisplay entryCollaborator={collaborator} />
```

frontend/src/entry/overview/EntryCardRolesChipSet.tsx
```diff
@@ -1,22 +1,16 @@
-import Dangerous from '@mui/icons-material/DangerousOutlined'
 import { Chip } from '@mui/material'
+import { useMemo } from 'react'
 import { CollaboratorEntry } from 'types/types'
 
 type EntryCardChipSetProps = {
```
ARADDCC012, line 5: My bad for not noticing this the first time around. Should be `EntryCardRolesChipSetProps`

frontend/src/entry/overview/EntryCardEntityRoleRow.tsx
ARADDCC012, line 1: This component is now unused so this file can be deleted

ARADDCC012 (changes requested):

frontend/src/entry/overview/EntryRoleList.tsx
IR96334: Done?

frontend/src/entry/overview/EntryRoleList.tsx
IR96334: Done?

frontend/src/entry/overview/EntryRoleList.tsx
IR96334: Sorry, yeah just forgot to change it. You were clear, dw. Changed.

frontend/src/entry/overview/EntryRoleList.tsx
IR96334: See above. Done?

frontend/src/entry/overview/EntryCardRolesChipSet.tsx
IR96334: Changed

frontend/src/entry/overview/FormEditPage.tsx
ARADDCC002: Import name still needs updating

ARADDCC002 (approved):

PR #1459, opened 2024-08-09
readded all the theme picker code

ARADDCC002:
image


JR40159 (changes requested):
Did you consider using cookies/local storage to save the user's preferences rather than the backend/DB? It would save on the overhead of the backend changes for something that is UI specific and not used by other clients and we already use this for the announcement banner and that seems to work pretty well

ARADDCC002:
This needs a lot of changes before going in so it's best off closing and revisiting this in the future when I get more time.

PR #1461, opened 2024-08-14
added logs to entries and a table to the UI to display them

ARADDCC002:


JR40159 (changes requested):
It would be good to review the design of this to avoid duplication with our existing event auditing.

ARADDCC002:
Closing as we need to redesign how we want to manage the logs

PR #1462, opened 2024-08-15
Fix unstable swr array hooks

GB27247:


frontend/actions/model.ts
```diff
@@ -41,7 +43,7 @@ export function useListModels(
 
   return {
     mutateModels: mutate,
-    models: data ? data.models : [],
+    models: data ? data.models : emptyArray,
```
ARADDCC012, line 46: We shouldn't ever mutate the data returned by SWR, but mistakes can happen. I'd feel a lot more comfortable if each SWR call used its own empty array, so we never have to worry about data appearing in arrays being used in multiple places. It's a bit verbose, but I think the safety provided is worth it.

Consider doing the following for each SWR hook:
```
// Define the empty array for this SWR hook (try to make this reference the name of the hook)
const emptyModelList = []

// Use it in the returned object as you've already done
models: data ? data.models : emptyModelList,
```

ARADDCC012 (changes requested):

ARADDCC012 (approved):

PR #1466, opened 2024-08-20
updated config and audit for create model card£

ARADDCC002:


JR40159:
Looks good, just need to verify the reason for the change.

PE39806 (approved):

PR #1467, opened 2024-08-20
Updated review wording for new access requests and releases

ARADDCC002:


backend/src/services/smtp/smtp.ts
```diff
@@ -37,11 +37,11 @@ export async function requestReviewForRelease(entity: string, review: ReviewDoc,
   }
 
   const emailContent = buildEmail(
-    `Release ${release.semver} has been created for model ${release.modelId}`,
+    `Release ${release.semver} for model ${release.modelId} is ready for your review`,
```
JR40159, line 40: Under a previous ticket, BAI-1383, all emails that require a user to do something should now have ""ACTION REQUIRED:" prepended to the subject?

backend/src/services/smtp/smtp.ts
```diff
@@ -68,11 +72,10 @@ export async function requestReviewForAccessRequest(
   }
 
   const emailContent = buildEmail(
-    `Request for Entities '${accessRequest.metadata.overview.entities}' access to the model '${accessRequest.modelId}'`,
+    `${requestingEntitiesText(accessRequest.metadata.overview.entities.length)} requesting access to model ${accessRequest.modelId}`,
```
JR40159, line 75: I'm assuming you've done this because the raw list of entities isn't nice to look at. Normally I'd suggest using getUserInformation() so convert them all to friendly name but here we might have both users and groups. How do we present groups nicely in the frontend if we don't have a group equivalent for getUserInformation()?

JR40159 (changes requested):

JR40159 (approved):

PR #1468, opened 2024-08-21
Add model card actions

GB27247:


backend/src/connectors/authorisation/base.ts
```diff
@@ -105,8 +105,9 @@ export class BasicAuthorisationConnector {
 
         // Check a user has a role before allowing write actions
         if (
-          [ModelAction.Write, ModelAction.Update].some((a) => a === action) &&
-          (await missingRequiredRole(user, model, ['owner', 'mtr', 'msro']))
+          ([ModelAction.Write, ModelAction.Update].some((a) => a === action) &&
```
JR40159, line 108: For clarity I would probably split this into two if statements/conditions:
1. For `ModelAction.Write` that tells the user they don't have permission to update the model card if they don't have one of 'owner', 'mtr', 'msro', 'contributor'.
2. For `ModelAction.Update`  that tells the user they don't have permission to update the model if they don't have one of 'owner', 'mtr', 'msro'.

JR40159 (changes requested):

ARADDCC012 (approved):

JR40159 (approved):

PR #1472, opened 2024-08-22
Reversed the list, equivalent to sorting by date ASC

IR96334:


frontend/src/settings/authentication/AuthenticationTab.tsx
```diff
@@ -29,7 +29,7 @@ export default function AuthenticationTab() {
   const tokenList = useMemo(
     () =>
       tokens.length > 0 && !isTokensLoading ? (
-        tokens.map((token, index) => (
+        tokens.reverse().map((token, index) => (
```
ARADDCC002, line 32: Are these tokens guaranteed to be in order to begin with? We should have some util functions with organise things by a timestamp so I'd probably make sure it's ordered first before reversing it.

ARADDCC002 (changes requested):

frontend/src/settings/authentication/AuthenticationTab.tsx
ARADDCC002: My understanding is that it'll be in order depending on when the items were added. I think 99% of the time they'll be in order but I don't think the order is necessarily guaranteed in niche situations. `token` should have some sort of timestamp so I would order based on that as we have some util functions to order either ascending or descending :)

ARADDCC002 (approved):

frontend/src/settings/authentication/AuthenticationTab.tsx
IR96334: I submit!

PR #1476, opened 2024-08-28
Feature/bai 1249 as a user i want my model cards stored as a zip file in s3 to be viewable in bailo

JR40159:


GB27247 (changes requested):

backend/src/routes/v2/model/postRequestImport.ts
```diff
@@ -0,0 +1,59 @@
...
+            sourceModelId: z.string(),
+            modelCardVersions: z.array(z.string()),
+          }),
+        },
+      },
+    },
+  },
+})
+
+interface PostRequestExportResponse {
```
ARADDCC012, line 40: Rename to `PostRequestImportResponse`?

backend/src/services/mirroredModel.ts
```diff
@@ -77,6 +86,67 @@ export async function exportModel(
...
+  if (mirroredModelId === '') {
+    throw BadReq('Missing mirrored model ID.')
+  }
+  let sourceModelId
+
+  let res: Response
+  try {
+    res = await fetch(payloadUrl)
+  } catch (err) {
+    throw InternalError('Unable to get the file.', { err })
```
ARADDCC012, line 99: Is it worth logging the `payloadUrl` in these three internal errors?

backend/src/services/mirroredModel.ts
```diff
@@ -77,6 +86,67 @@ export async function exportModel(
...
+  if (!res.ok) {
+    throw InternalError('Unable to get zip file.', { response: { status: res.status, body: await res.text() } })
+  }
+
+  if (!res.body) {
+    throw InternalError('Unable to get the file.')
+  }
+
+  const modelCards: ModelCardRevisionInterface[] = []
+  const test = new Uint8Array(await res.arrayBuffer())
```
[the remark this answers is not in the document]
ARADDCC012: Perhaps also clarify that this is the zipped data?

backend/src/services/mirroredModel.ts
```diff
@@ -77,6 +86,67 @@ export async function exportModel(
...
+    }
+    modelCards.push(modelCard)
+  })
+
+  await Promise.all(modelCards.map((card) => saveImportedModelCard(card, sourceModelId)))
+  await setLatestImportedModelCard(mirroredModelId)
+  return { mirroredModelId, sourceModelId, modelCardVersions: modelCards.map((modelCard) => modelCard.version) }
+}
+
+function parseModelCard(modelCardString: string, mirroredModelId: string, sourceModelId?: string) {
```
ARADDCC012, line 133: For readability it might be worth clarifying that `modelCardString` is `modelCardJson`. At a glance "string" could refer to a number of things related to the model card.

backend/src/services/model.ts
```diff
@@ -376,3 +379,66 @@ export async function createModelCardFromTemplate(
...
+    modelCard,
+    {
+      upsert: true,
+    },
+  )
+}
+
+export async function setLatestImportedModelCard(modelId: string) {
+  // Add check for mirrored model
+  const latestModelCard = await ModelCardRevisionModel.findOne({ modelId }).sort({ version: -1 }).limit(1)
```
ARADDCC012, line 424: I don't think the `sort` and `limit` on this `findOne` is doing what you expect. `findOne` will return the first result that meets the condition, so you'll just be sorting a single result. I think if you replace `findOne` with `find` it'll work as expected.

Also consider adding in some error handling in the event we don't find anything.

backend/src/services/model.ts
```diff
@@ -376,3 +379,66 @@ export async function createModelCardFromTemplate(
...
+    { modelId: modelCard.modelId, version: modelCard.version },
+    modelCard,
+    {
+      upsert: true,
+    },
+  )
+}
+
+export async function setLatestImportedModelCard(modelId: string) {
+  // Add check for mirrored model
```
ARADDCC012, line 423: Does this comment still need to be addressed?

ARADDCC012 (changes requested):

backend/src/services/model.ts
```diff
@@ -376,3 +379,66 @@ export async function createModelCardFromTemplate(
   const revision = await _setModelCard(user, modelId, template.card.schemaId, 1, template.card.metadata)
   return revision
 }
+
+export async function saveImportedModelCard(modelCard: ModelCardRevisionInterface, sourceModelId: string) {
+  const model = await Model.findOne({
+    id: modelCard.modelId,
+  })
+  if (!model) {
+    throw NotFound(`The mirrored model ID found in the notification cannot be found.`, { modelId: modelCard.modelId })
```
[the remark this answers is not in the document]
JR40159: I also think this is confusing so I've simplified it.

backend/src/services/model.ts
JR40159: Updated but this method needs unit and manual testing

JR40159: Stuck with using `findOne` as this returns a single document rather than `find` as that returns a cursor. I have though removed the limit as that's redundant when using `findOne`

backend/src/services/mirroredModel.ts
```diff
@@ -254,36 +321,44 @@ async function addReleaseToZip(user: UserInterface, model: ModelDoc, release: Re
...
+    if (totalFileSize > config.modelMirror.export.maxSize) {
+      throw BadReq('Requested export is too large.', {
+        size: totalFileSize,
+        maxSize: config.modelMirror.export.maxSize,
+      })
+    }
   }
-}
 
-async function checkReleaseFiles(
```
GB27247, line 272: Could we catch if AV scanning is enabled?

GB27247 (changes requested):

backend/src/services/model.ts
```diff
@@ -420,9 +420,18 @@ export async function saveImportedModelCard(modelCard: ModelCardRevisionInterfac
 }
 
 export async function setLatestImportedModelCard(modelId: string) {
-  // Add check for mirrored model
-  const latestModelCard = await ModelCardRevisionModel.findOne({ modelId }).sort({ version: -1 }).limit(1)
-  await ModelModel.updateOne({ id: modelId }, { $set: { card: latestModelCard } })
+  const latestModelCard = await ModelCardRevisionModel.findOne({ modelId }).sort({ version: -1 })
+  if (!latestModelCard) {
+    throw NotFound('Cannot find latest model card.')
```
ARADDCC012, line 425: Worth logging the `modelId` with this?

backend/src/services/model.ts
```diff
@@ -420,9 +420,18 @@ export async function saveImportedModelCard(modelCard: ModelCardRevisionInterfac
...
+  }
+
+  const result = await ModelModel.findOneAndUpdate(
+    { id: modelId, 'settings.mirror.sourceModelId': { $exists: true, $ne: '' } },
+    { $set: { card: latestModelCard } },
+  )
+  if (!result) {
+    throw InternalError('Unable update model with latest model card ID.')
```
ARADDCC012, line 433: Could just be me, but this message is a little confusing. The message refers to "latest model card ID", but we're not using that ID in the query above.

ARADDCC012 (approved):

GB27247 (approved):

PR #1507, opened 2024-09-04
added a popover for av results on files

ARADDCC002:


frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -1,30 +1,98 @@
...
-              <Link href={`/api/v2/model/${modelId}/file/${file._id}/download`} data-test={`fileLink-${file.name}`}>
-                <Typography noWrap textOverflow='ellipsis'>
-                  {file.name}
-                </Typography>
-              </Link>
-            </Tooltip>
-          </Grid>
-          <Grid item xs={1} textAlign='right'>
-            <Typography variant='caption'>{prettyBytes(file.size)}</Typography>
+        <>
```
ARADDCC012, line 77: This fragment can be removed

ARADDCC012 (changes requested):

backend/src/utils/config.ts
```diff
@@ -135,6 +135,10 @@ export interface Config {
       text: string
       startTimestamp: string
     }
+
+    avScanning: {
+      enabled: boolean
```
JR40159, line 140: You need to remove the existing `avScanning.enabled` configuration, as they're now duplicated.

backend/config/default.cjs
JR40159, line 1: The configuration to enable av scanning should be in the Docker Compose configuration file because you need to use Docker Compose to have the AV scanner running

JR40159 (changes requested):

backend/config/default.cjs
```diff
@@ -191,6 +190,10 @@ module.exports = {
       text: '',
       startTimestamp: '',
     },
+
+    avScanning: {
```
JR40159, line 194: By default, we don't want avScanning enabled because you need to have an ClamAV instance running somewhere

backend/src/services/file.ts
JR40159, line 1: You'll need to update the mocked config

JR40159 (changes requested):

JR40159 (approved):

PR #1514, opened 2024-09-11
File scanning connector to handle multiple AV tools

ARADDCC002:


backend/src/services/file.ts
```diff
@@ -38,37 +35,11 @@ export async function uploadFile(user: UserInterface, modelId: string, name: str
...
-        avScan: { state: ScanState.Error },
-      })
-    })
-    avStream.on('timeout', async (error) => {
-      log.error({ error, modelId, fileId: file._id, name }, 'Scan timed out.')
-      await file.update({
-        avScan: { state: ScanState.Error },
-      })
+  if (config.ui.avScanning.enabled) {
+    const fileScanConnectors = getFileScanningConnectors()
```
JR40159, line 39: It would be good if this new connector type could have the same interface as other connectors and this could just be `scanners.scan(file)`. Ideally we want to encapsulate the initialisation and the calling of each individual scanner inside the connector logic, not here in the service layer every time we want to use the scanners

backend/src/connectors/fileScanning/clamAv.ts
```diff
@@ -0,0 +1,88 @@
...
+let av: NodeClam
+const toolName = 'Clam AV'
+
+export class ClamAvFileScanningConnector extends BaseFileScanningConnector {
+  constructor() {
+    super()
+  }
+
+  async init() {
+    if (!av) {
```
JR40159, line 20: I don't think this check is necessary because the `init()` function is only called to initialise the av scanner

backend/src/connectors/fileScanning/clamAv.ts
```diff
@@ -0,0 +1,88 @@
...
+      try {
+        av = await new NodeClam().init({ clamdscan: config.avScanning.clamdscan })
+      } catch (error) {
+        throw ConfigurationError('Cannot retrieve Clam AV config', {
```
JR40159, line 24: Have you tested what happens if the Clam AV service isn't running? I assume this throw would be called. If so, the error message needs updating

backend/src/connectors/fileScanning/clamAv.ts
```diff
@@ -0,0 +1,88 @@
...
+          clamAvConfig: config.avScanning,
+        })
+      }
+    }
+  }
+
+  async scan(file: FileInterfaceDoc) {
+    const avStream = av.passthrough()
```
JR40159, line 32: It's a bit of a pain but to be safe, you should probably check that `av` has been initialised at the start of functions like this when using the AV scanner and throw an error recommending that `init()` should be called first if `av` has not been assigned.

backend/src/connectors/fileScanning/clamAv.ts
```diff
@@ -0,0 +1,88 @@
...
+    const s3Stream = (await getObjectStream(file.bucket, file.path)).Body as Readable
+    s3Stream.pipe(avStream)
+    log.info({ modelId: file.modelId, fileId: file._id, name: file.name }, 'Scan started.')
+    if (!file.avScan.find((result) => result.toolName === toolName)) {
+      const updatedAvScanArray = [...file.avScan, { toolName: toolName, state: ScanState.InProgress }]
+      file.avScan = updatedAvScanArray
+      await FileModel.updateOne(
```
JR40159, line 39: In my opinion, these scanners should not be responsible for storing the scan information in the DB, they should just be an interface with a particular scanning and this method should take in a file stream and return a scan result object. Ther service layer should be responsible for getting the file stream and processing the scan results.

backend/src/models/File.ts
```diff
@@ -51,11 +49,14 @@ const FileSchema = new Schema<FileInterface>(
     bucket: { type: String, required: true },
     path: { type: String, required: true },
 
-    avScan: {
-      state: { type: String, enum: Object.values(ScanState), default: 'notScanned' },
-      isInfected: { type: Boolean },
-      viruses: [{ type: String }],
-    },
+    avScan: [
```
JR40159, line 52: Don't forget, this will require a migration. I don't think a migration was done when this field was originally added so this migration will have to account for when this field doesn't exist

JR40159 (changes requested):

backend/src/connectors/fileScanning/clamAv.ts
ARADDCC002: This is a bit tricky as it's async and the service that calls these connectors already returns the response to the user. We'd need to call the scanners, wait for the results, store them, _then_ return a 200 response to the user from the backend.

JR40159: Just because we want a process to run asynchronously doesn't mean all the implementation needs to be in a single method. Use a `.then()` on the asynchronous call to provide the implementation for how we will process the results once the scanners are complete.

backend/src/models/File.ts
JR40159: From looking at this diff it looks like the previous type of `avScan` was an object and now `avScan` is an array? That's why I thought this would require a migration

backend/src/services/mirroredModel.ts
```diff
@@ -351,9 +352,16 @@ async function checkReleaseFiles(user: UserInterface, modelId: string, semvers:
     for (const file of files) {
       if (!file.avScan) {
         scanErrors.missingScan.push({ name: file.name, id: file.id })
-      } else if (file.avScan.state !== ScanState.Complete) {
+      } else if (
+        file.avScan.some(
+          (scanResult) =>
+            scanResult.state === ScanState.NotScanned ||
```
JR40159, line 358: Can this not be `scanResult.state !== ScanState.Complete`?

JR40159 (changes requested):

JR40159 (dismissed):
When setting the scanners connectors to an empty array, the information still appears in the UI?

infrastructure/helm/bailo/templates/bailo/bailo.configmap.yaml
```diff
@@ -193,6 +192,10 @@ data:
         audit: {
           kind: '{{ .Values.connectors.audit.kind }}',
         },
+
+        fileScanners: {
+          kinds: '{{ .Values.connectors.fileScanners.kinds }}',
```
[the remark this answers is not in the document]
ARADDCC002: This is just a reference though isn't it? I don't think we define if it's an array or not here

ARADDCC012 (approved):

PR #1563, opened 2024-10-14
Added templating and access requests buttons to the new model create screen

ghost:
image

### DIFF
--- a/frontend/src/entry/CreateEntry.tsx
+++ b/frontend/src/entry/CreateEntry.tsx
@@ -107,7 +107,7 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
     }
   }
 
-  const privateLabel = () => {
+  const privateLabel = useMemo(() => {
     return (
       <Stack direction='row' justifyContent='center' alignItems='center' spacing={1}>
         <Lock />
@@ -119,9 +119,9 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
         </Stack>
       </Stack>
     )
-  }
+  }, [createEntryKind])
 
-  const publicLabel = () => {
+  const publicLabel = useMemo(() => {
     return (
       <Stack direction='row' justifyContent='center' alignItems='center' spacing={1}>
         <LockOpen />
@@ -133,43 +133,35 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
         </Stack>
       </Stack>
     )
-  }
+  }, [createEntryKind])
 
-  const TemplateLabel = ({ createEntryKind }) => {
-    const CreateTemplateLabel = useMemo(() => {
-      return (
-        <Stack direction='row' justifyContent='center' alignItems='center' spacing={1}>
-          <FolderCopy />
-          <Stack sx={{ my: 1 }}>
-            <Typography fontWeight='bold'>Templating</Typography>
-            <Typography variant='caption'>
-              {`Allow this to be used as a template for another ${EntryKindLabel[createEntryKind]}`}
-            </Typography>
-          </Stack>
+  const allowTemplatingLabel = useMemo(() => {
+    return (
+      <Stack direction='row' justifyContent='center' alignItems='center' spacing={1}>
+        <FolderCopy />
+        <Stack sx={{ my: 1 }}>
+          <Typography fontWeight='bold'>Templating</Typography>
+          <Typography variant='caption'>
+            {`Allow this to be used as a template for another ${EntryKindLabel[createEntryKind]}`}
+          </Typography>
         </Stack>
-      )
-    }, [createEntryKind])
+      </Stack>
+    )
+  }, [createEntryKind])
 
-    return CreateTemplateLabel
-  }
-
-  const UngovernedAccessLabel = ({ createEntryKind }) => {
-    const createUngovernedAccessLabel = useMemo(() => {
-      return (
-        <Stack direction='row' justifyContent='center' alignItems='center' spacing={1}>
-          <Gavel />
-          <Stack sx={{ my: 1 }}>
-            <Typography fontWeight='bold'>Ungoverned Access Requests</Typography>
-            <Typography variant='caption'>
-              {`Allow users to request access without the need for authorisation from ${EntryKindLabel[createEntryKind]} owners`}
-            </Typography>
-          </Stack>
+  const ungovernedAccessLabel = useMemo(() => {
+    return (
+      <Stack direction='row' justifyContent='center' alignItems='center' spacing={1}>
+        <Gavel />
+        <Stack sx={{ my: 1 }}>
+          <Typography fontWeight='bold'>Ungoverned Access Requests</Typography>
+          <Typography variant='caption'>
+            {`Allow users to request access without the need for authorisation from ${EntryKindLabel[createEntryKind]} owners`}
+          </Typography>
         </Stack>
-      )
-    }, [createEntryKind])
-
-    return createUngovernedAccessLabel
-  }
+      </Stack>
+    )
+  }, [createEntryKind])
 
   if (isCurrentUserError) {
     return <MessageAlert message={isCurrentUserError.info.message} severity='error' />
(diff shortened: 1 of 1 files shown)


frontend/src/entry/CreateEntry.tsx
```diff
@@ -60,6 +62,8 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
   const [collaborators, setCollaborators] = useState<CollaboratorEntry[]>(
     currentUser ? [{ entity: `${EntityKind.USER}:${currentUser?.dn}`, roles: ['owner'] }] : [],
   )
+  const [ungovernedAccess, setungovernedAccess] = useState<boolean>(false)
```
ARADDCC002, line 65: `boolean` and `string` can be inferred from their defaults so we don't need to declare the types, so we can just do `useState(false)`.

frontend/src/entry/CreateEntry.tsx
```diff
@@ -174,24 +194,36 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
...
-                  data-test='publicButtonSelector'
-                />
-                <FormControlLabel
-                  value='private'
-                  control={<Radio />}
-                  label={privateLabel()}
-                  data-test='privateButtonSelector'
-                />
-              </RadioGroup>
+              <Box sx={{ display: 'flex', flexDirection: 'column', gap: 1 }}>
```
ARADDCC002, line 197: We don't need to use this `<Box/>` component here

frontend/src/entry/CreateEntry.tsx
```diff
@@ -174,24 +194,36 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
...
+                    label={publicLabel()}
+                    data-test='publicButtonSelector'
+                  />
+                  <FormControlLabel
+                    value='private'
+                    control={<Radio />}
+                    label={privateLabel()}
+                    data-test='privateButtonSelector'
+                  />
+                  <FormControlLabel
```
ARADDCC002, line 215: The option to allow for templating should appear at the bottom of the expansion menu under or above the choice to allow for ungoverned access. Although I can see why you'd put it under "Access control", this section is accessing the model card page - easy mistake to make though!

frontend/src/entry/CreateEntry.tsx
```diff
@@ -224,6 +256,16 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
                     ]}
                   />
                 </Box>
+                <FormControlLabel
```
ARADDCC002, line 259: I would put the "allow templating" option here, and I'd suggest adding a sub-heading of "Additional settings", a bit like how the user selector appears under the heading of "Manage access list".

To easily space out different components you can use the MUI `<Stack />` component, which works like so:

```
<Stack spacing={2}>
  <Typography>This is some text!</Typography>
  <Typography>This is some more text!</Typography>
</Stack>
```

ARADDCC002 (changes requested):

frontend/src/entry/CreateEntry.tsx
```diff
@@ -130,6 +136,34 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
     )
   }
 
+  const templateLabel = () => {
```
ARADDCC002, line 139: `const templateLabel = useCallback(() => {`

frontend/src/entry/CreateEntry.tsx
```diff
@@ -130,6 +136,34 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
...
+          <Typography fontWeight='bold'>Templating</Typography>
+          <Typography variant='caption'>
+            {`Allow this to be used as a template for another ${EntryKindLabel[createEntryKind]}`}
+          </Typography>
+        </Stack>
+      </Stack>
+    )
+  }
+
+  const ungovAccessLabel = () => {
```
ARADDCC002, line 153: `const ungovAccessLabel = useCallback(() => {`

I'd also consider renaming this to `ungovernedAccessLabel` as well

ARADDCC002 (changes requested):

frontend/src/entry/CreateEntry.tsx
```diff
@@ -224,6 +258,32 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
...
+                    control={
+                      <Switch
+                        onChange={(e) => setungovernedAccess(e.target.checked)}
+                        checked={ungovernedAccess}
+                        size='small'
+                      />
+                    }
+                  />
+                  <FormControlLabel
+                    value='false'
```
ARADDCC012, line 276: `value` can be removed

ARADDCC012 (changes requested):

ARADDCC002 (approved):

frontend/src/entry/CreateEntry.tsx
```diff
@@ -130,6 +135,42 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
     )
   }
 
+  const TemplateLabel = ({ createEntryKind }) => {
```
ARADDCC012, line 138: This should just be (I suggest changing the name to be clear that it in itself isn't a template):
```
const allowTemplatingLabel = useMemo(() => {
      return (
        <Stack direction='row' justifyContent='center' alignItems='center' spacing={1}>
          <FolderCopy />
          <Stack sx={{ my: 1 }}>
            <Typography fontWeight='bold'>Templating</Typography>
              {`Allow this to be used as a template for another ${EntryKindLabel[createEntryKind]}`}
            </Typography>
          </Stack>
        </Stack>
      )
    }, [createEntryKind])
```

It's important to note that hooks in react have to be called in the same order on every render cycle, so they should always be defined at the base of the component.

I suggest having a quick read on how memo hooks work in particular, but in short this is for performance reasons. https://react.dev/reference/react/useMemo

frontend/src/entry/CreateEntry.tsx
```diff
@@ -130,6 +135,42 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
...
+            </Typography>
+          </Stack>
+        </Stack>
+      )
+    }, [createEntryKind])
+
+    return CreateTemplateLabel
+  }
+
+  const UngovernedAccessLabel = ({ createEntryKind }) => {
```
ARADDCC012, line 156: Same here, this should be:
```
    const ungovernedAccessLabel = useMemo(() => {
      return (
        <Stack direction='row' justifyContent='center' alignItems='center' spacing={1}>
          <Gavel />
          <Stack sx={{ my: 1 }}>
            <Typography fontWeight='bold'>Ungoverned Access Requests</Typography>
              {`Allow users to request access without the need for authorisation from ${EntryKindLabel[createEntryKind]} owners`}
            </Typography>
          </Stack>
        </Stack>
      )
    }, [createEntryKind])
```

frontend/src/entry/CreateEntry.tsx
```diff
@@ -224,6 +265,31 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
...
+                  </Typography>
+                  <FormControlLabel
+                    control={
+                      <Switch
+                        onChange={(e) => setungovernedAccess(e.target.checked)}
+                        checked={ungovernedAccess}
+                        size='small'
+                      />
+                    }
+                    label={UngovernedAccessLabel({ createEntryKind })}
```
ARADDCC012, line 280: This should be `label={ungovernedAccessLabel}`

frontend/src/entry/CreateEntry.tsx
```diff
@@ -224,6 +265,31 @@ export default function CreateEntry({ createEntryKind, onBackClick }: CreateEntr
...
+                  />
+                  <FormControlLabel
+                    control={
+                      <Switch
+                        onChange={(event) => setAllowTemplating(event.target.checked)}
+                        checked={allowTemplating}
+                        size='small'
+                      />
+                    }
+                    label={TemplateLabel({ createEntryKind })}
```
ARADDCC012, line 290: This should be `label={allowTemplatingLabel}`

ARADDCC012 (changes requested):

ARADDCC012 (approved):

PR #1566, opened 2024-10-15
Feature/bai 1466 record who exports a model

JR40159:


backend/src/services/smtp/smtp.ts
```diff
@@ -193,7 +193,7 @@ async function sendEmail(email: Mail.Options) {
     log.info({ messageId: info.messageId }, 'Email sent')
   } catch (err) {
     const content = { to: email.to, subject: email.subject, text: email.text }
-    log.warn(content, `Unable to send email`)
+    log.warn({ content, err }, `Unable to send email`)
```
ARADDCC012, line 196: Super minor, but as this will show in the logs can we call this `error` instead of `err`?

ARADDCC012 (approved):

PR #1571, opened 2024-10-16
Fix overflow wrap on model card summary

ghost:


frontend/src/common/MarkdownDisplay.tsx
```diff
@@ -17,31 +18,34 @@ export default function MarkdownDisplay({ children }: MarkdownDisplayProps) {
...
-          component: Typography,
-          props: {
-            gutterBottom: true,
-            variant: 'caption',
-            paragraph: true,
-          },
+          component: (props: TypographyProps) => (
+            <Typography
+              gutterBottom
+              paragraph
```
ARADDCC012, line 39: This isn't your doing, but this shouldn't be a paragraph. This prop can be removed

frontend/src/common/MarkdownDisplay.tsx
```diff
@@ -17,31 +18,34 @@ export default function MarkdownDisplay({ children }: MarkdownDisplayProps) {
...
+              variant='caption'
+              component='h2'
```
ARADDCC012, line 41: This should be using `component='h4'`

frontend/src/common/MarkdownDisplay.tsx
```diff
@@ -17,31 +18,34 @@ export default function MarkdownDisplay({ children }: MarkdownDisplayProps) {
...
-          props: {
-            gutterBottom: true,
-            variant: 'caption',
-            paragraph: true,
-          },
+          component: (props: TypographyProps) => (
+            <Typography
+              gutterBottom
+              paragraph
+              variant='caption'
```
ARADDCC012, line 40: We should fix this while we're here. A heading should not be using `caption`. I suggest changing the variants on all headings as follows:
h2 -> variant='h5'
h3 -> variant='h6'
h4 -> variant='subtitle1'

frontend/src/common/MarkdownDisplay.tsx
```diff
@@ -17,31 +18,34 @@ export default function MarkdownDisplay({ children }: MarkdownDisplayProps) {
...
+              sx={{ wordWrap: 'break-word' }}
+              {...props}
+            />
+          ),
         },
         p: {
-          component: Typography,
-          props: { paragraph: true },
+          component: (props: TypographyProps) => <Typography sx={{ wordWrap: 'break-word' }} {...props} />,
         },
```
ARADDCC012, line 49: To fix the wrapping for a single paragraph you'll need to add the following to override the styling of `span`s:
```
        span: {
          component: (props: TypographyProps) => (
            <Typography component='span' sx={{ wordWrap: 'break-word' }} {...props} />
          ),
        },
```

ARADDCC012 (approved):

PR #1582, opened 2024-10-23
Bai 1111 allow semver matching and latest semver for releases

IR96334:
WIP PR for advanced semver range queries as stated in https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004


backend/src/services/release.ts
```diff
@@ -295,6 +331,133 @@ export async function getReleaseBySemver(user: UserInterface, modelId: string, s
...
+  }
+
+  //4
+  const lowerSemverObj = semverStringToObject(lowerSemverTrimmed)
+  let upperSemverObj
+  if (upperSemverTrimmed) {
+    upperSemverObj = semverStringToObject(upperSemverTrimmed)
+  }
+
+  if (inclusivity)
```
JR40159, line 381: This function is a lot more simpler and readable, good job!
To improve upon this a bit more I would move all the logic needed to convert the `semverRangeStandardised` content into a form that can be used to generate this mongo query into a single function. At the moment this function contains multiple variable variables and function calls to convert the various semver formats when all you really need at this point is just the lowerSemverObj and upperSemverObj. Also your exclusivity logic doesn't currently work as it doesn't allow for half the semver query being inclusive.

Once you've got your `semverRangeStandardised` and checked it's not null you could capture all the logic up to this point of the mongo query in a single function, something like:

```
function parseSemverQuery(semverRangeStandardised: string): {
  upper: { major: string; minor: string; patch: string; inclusive: boolean }
  lower: { major: string; minor: string; patch: string; inclusive: boolean }
} {
  //parse semverRangeStandardised 
}
```
Then to reduce duplication in the inclusivity logic I would use if statements to set either `$gte` or `$gt` and `$lse` or `$ls` rather than repeat all the other bits of the mongo query in all the conditions. Eg:
```
if (semverQuery.upper.inclusive) {
  lowerComparator = '$gte' 
} else {
  lowerComparator = '$gt' 
}
if (semverQuery.lower.inclusive) {
  uppercomparator = '$lse' 
} else {
  uppercomparator = '$ls' 
}
```
Then you should be able to use this two variable in the relevant places in your single mongo query object Eg `'semver.major': { [lowerComparator]: semverQuery.lower.major }`

JR40159 (changes requested):

backend/src/services/release.ts
```diff
@@ -295,6 +331,133 @@
...
+  let upperSemverTrimmed, inclusivity
+  if (upperSemver) {
+    if (upperSemver.includes('=')) {
+      //remove and extract
+      upperSemverTrimmed = upperSemver.replace('<=', '')
+      inclusivity = true
+    } else {
+      //exclusive
+      //remain
+      upperSemverTrimmed = upperSemver.replace('<', '')
```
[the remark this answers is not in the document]
IR96334: Will only ever be one in upperSemver

backend/src/services/release.ts
```diff
@@ -295,6 +331,133 @@ export async function getReleaseBySemver(user: UserInterface, modelId: string, s
...
+            {
+              'semver.major': { $gte: lowerSemverObj.major },
+              'semver.minor': { $gt: lowerSemverObj.minor },
+            },
+          ],
+        },
+        {
+          $or: [
+            {
+              'semver.major': { $lte: upperSemverObj.major },
```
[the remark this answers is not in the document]
IR96334: Checks added for upperSemver, etc

backend/src/services/release.ts
IR96334: Changes to semverRangeStandardised added, also added dynamic query stuff for inclusivity.

backend/src/services/release.ts
```diff
@@ -295,6 +326,168 @@ export async function getReleaseBySemver(user: UserInterface, modelId: string, s
...
+    throw BadReq(`Semver range, '${querySemver}' is invalid `)
+  }
+
+  const [expressionA, expressionB] = semverRangeStandardised.split(' ')
+
+  let lowerSemver: string = '',
+    upperSemver: string = '',
+    lowerInclusivity = false,
+    upperInclusivity = false
+  if (expressionA.includes('>')) {
```
JR40159, line 342: When an if statement becomes this complex it's always worth seeing if it could be simplified.
Have you considered:
```
const lowerInclusivity = expressionA.includes('>=')
const upperInclusivity = expressionB.includes('<=') | expressionA.includes('<=')
const lowerSemver = expressionA.replace(/[<>=]/g, '')
const upperSemver = expressionB.replace(/[<>=]/g, '')

**Review by JR40159 — CHANGES_REQUESTED** [review]

File: backend/src/services/release.ts [hunk 5]
```diff
@@ -295,6 +326,168 @@ export async function getReleaseBySemver(user: UserInterface, modelId: string, s
...
+  } else {
+    if (expressionA.includes('=')) {
+      upperSemver = expressionA.replace('<=', '')
+      upperInclusivity = true
+    } else {
+      upperSemver = expressionA.replace('<', '')
+    }
+  }
+
+  let lowerSemverObj: { metadata?: string | undefined; major: number; minor: number; patch: number } = {
```
ARADDCC012, line 367: In a situation where two things are the exact same type, we should avoid redefining types. This allows us to have a single source of truth. If we ever updated that type for any reason, we'd only have to update it in one place and TS would guide us through updating all variables of that type.

In this scenario we should be using the `SemverObject` type you've already created in `backend/src/models/Release.ts` for both `lowerSemverObj` and `upperSemverObj`.

backend/src/services/release.ts
```diff
@@ -295,6 +326,168 @@ export async function getReleaseBySemver(user: UserInterface, modelId: string, s
...
+      major: NaN,
+      minor: NaN,
+      patch: NaN,
+    },
+    upperSemverObj: { metadata?: string | undefined; major: number; minor: number; patch: number } = {
```
ARADDCC012, line 372: Although JS allows us to define multiple variables on one line like you're doing here, I would strongly advise against it as it's simply less readable than defining multiple variables separately.

```
let lowerSemverObj: SemverObject = { ... }
let upperSemverObj: SemverObject = { ... }
```

backend/src/services/release.ts
```diff
@@ -237,9 +239,11 @@ export async function newReleaseComment(user: UserInterface, modelId: string, se
 export async function getModelReleases(
   user: UserInterface,
   modelId: string,
+  q?: string,
```
ARADDCC012, line 242: I'm never a fan of single character variable names like this. It's more confusing that simply giving it a proper name.

Consider calling this `querySemver` as you have later on.

backend/src/services/release.ts
```diff
@@ -237,9 +239,11 @@ export async function newReleaseComment(user: UserInterface, modelId: string, se
...
 ): Promise<Array<ReleaseDoc & { model: ModelInterface; files: FileInterface[] }>> {
+  const query = q === undefined ? { modelId } : getQuerySyntax(q, modelId)
```
ARADDCC012, line 244: This can be simplified to `const query = getQuerySyntax(q, modelId)` as you've already covered off the case of `q === undefined` inside of `getQuerySyntax`

backend/src/services/release.ts
```diff
@@ -237,9 +239,11 @@ export async function newReleaseComment(user: UserInterface, modelId: string, se
...
   const results = await Release.aggregate()
-    .match({ modelId })
+    .match(query!)
```
ARADDCC012, line 246: By using `!` we're telling TS that this can't possibly be `undefined`/`null`. This is very rarely the correct way to manage types and is likely indicative of a flaw in your `getQuerySyntax()` logic.

I'd suggest removing this assertion and seeing where `undefined`/`null` are possibilities in earlier logic. You should fix it there instead of asserting here.

backend/src/services/release.ts
```diff
@@ -248,7 +252,9 @@ export async function getModelReleases(
   const model = await getModelById(user, modelId)
 
   const auths = await authorisation.releases(user, model, results, ReleaseAction.View)
-  return results.filter((_, i) => auths[i].success)
+  return results
```
ARADDCC012, line 255: If we care about performance here, we can combine this `filter` and `map` into a single `reduce`:
```
return results.reduce<ReleaseInterface>((updatedResults, result, index) => {
  if (auths[index].success) {
    updatedResults.push({ ...result, semver: semverObjectToString(result.semver) })
  }
  return updatedResults
}, [])
```

backend/src/services/release.ts
```diff
@@ -276,11 +282,36 @@ export async function getReleasesForExport(user: UserInterface, modelId: string,
   return releases
 }
 
+export function semverStringToObject(semver: string) {
```
ARADDCC012, line 285: We can give this an explicit return type to ensure it's not broken by accident:
```
export function semverStringToObject(semver: string): SemverObject { ...
```

backend/src/services/release.ts
```diff
@@ -55,7 +56,8 @@ async function validateRelease(user: UserInterface, model: ModelDoc, release: Re
     const fileNames: Array<string> = []
 
     for (const fileId of release.fileIds) {
-      let file
+      let file: Document<unknown, any, FileInterface> &
```
ARADDCC012, line 59: This can be simplified to `let file: FileInterfaceDoc | undefined`

backend/src/services/release.ts
IR96334: Yeah, regex makes this a tonne easier, good idea

IR96334: Although, expressionA can be upperSemver, not just lowerSemver. I will a check to see whether upperSemver will be expressionA or B.

IR96334: Done, still required some if statements as expressionA can be the upperSemver as the lower/upper parts of the variables state wether they are upper or lower bounds, not first or second in query

backend/src/services/release.ts
IR96334: Removed as was causing bugs and didn't seem strictly necessary.

backend/src/services/release.ts
IR96334: Fixed above

backend/src/services/release.ts
IR96334: True. Fixed

backend/src/services/release.ts
IR96334: Removed the check in getquerysyntax, dealing with it earlier makes more sense to me, but willing to change

backend/src/services/release.ts
IR96334: Turns out I didn't even need it anymore so just removed

backend/src/services/release.ts
IR96334: Added, but wasn't sure on how to implement the updatedResults as I can't push to it as its not an array, etc

backend/src/services/release.ts
IR96334: Added

backend/src/services/release.ts
IR96334: Don't know why its saying I've done this, but I have changed anyway!

backend/src/routes/v2/release/getReleases.ts
```diff
@@ -17,6 +17,9 @@ export const getReleasesSchema = z.object({
       required_error: 'Must specify model id as URL parameter',
     }),
   }),
+  query: z.object({
+    q: z.string().optional(),
```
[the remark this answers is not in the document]
IR96334: Done

backend/src/services/release.ts
```diff
@@ -295,6 +335,142 @@ export async function getReleaseBySemver(user: UserInterface, modelId: string, s
...
+        'semver.minor':
+          | {
+              $gt: number
+            }
+          | { $lt: number }
+      }
+  )[]
+}
+
+export function getQuerySyntax(querySemver: string, modelID: string) {
```
JR40159, line 421: We shouldn't need to export this? To make it clear what this function is doing, could we change the name to something like `convertSemverQueryToMongoQuery()`

backend/src/services/release.ts
```diff
@@ -295,6 +335,142 @@ export async function getReleaseBySemver(user: UserInterface, modelId: string, s
   return release
 }
 
+function getSemverQueryBounds(querySemver: string) {
```
JR40159, line 338: Could we change this to something like `parseSemverQuery()`?

JR40159 (changes requested):

backend/src/services/release.ts
```diff
@@ -295,6 +335,142 @@ export async function getReleaseBySemver(user: UserInterface, modelId: string, s
...
+  const semverRangeStandardised = semver.validRange(querySemver, { includePrerelease: false })
+
+  if (!semverRangeStandardised) {
+    throw BadReq(`Semver range, '${querySemver}' is invalid.`)
```
JR40159, line 342: In general with errors, we try and avoid injecting variables into the messages. Instead you can pass in a context object containing any relevant data. Eg `throw BadReq(`Semver range is invalid.`, {semverQuery: querySemver})`

backend/src/services/release.ts
```diff
@@ -295,6 +335,142 @@ export async function getReleaseBySemver(user: UserInterface, modelId: string, s
...
+  }
+
+  const [expressionA, expressionB] = semverRangeStandardised.split(' ')
+
+  let lowerInclusivity, lowerSemver, upperInclusivity, upperSemver
+
+  //If lower semver
+  if (expressionA.includes('>')) {
+    lowerInclusivity = expressionA.includes('>=')
+    lowerSemver = expressionA.replace(/[<>=]/g, '')
```
JR40159, line 352: To avoid having multiple variables for the upper and lower semvers and another set of if statements after this one, could you include `.replace(/[<>=]/g, '')` at the start of the function `semverStringToObject(0`? It would also mean you'd only need to have this replace in one place

JR40159 (changes requested):

backend/test/services/release.spec.ts
```diff
@@ -342,18 +344,100 @@ describe('services > release', () => {
...
     expect(releaseModelMocks.append.mock.calls.at(0)).toMatchSnapshot()
   })
 
+  //test good - give good semver range with lower and upper bounds, returns valid query
+  //test fail - invalid range
+  test('getQuerySyntax > good'),
+    async () => {
+      const testQuery = getQuerySyntax('1.0.x', 'testModelID')
+      expect(testQuery).toBe({
+        modelId: 'testModelID',
```
JR40159, line 372: For large expects like this one, I would use snapshots.

backend/test/services/release.spec.ts
```diff
@@ -342,18 +344,100 @@ describe('services > release', () => {
...
 
     expect(releaseModelMocks.match.mock.calls.at(0)).toMatchSnapshot()
     expect(releaseModelMocks.sort.mock.calls.at(0)).toMatchSnapshot()
     expect(releaseModelMocks.lookup.mock.calls.at(0)).toMatchSnapshot()
     expect(releaseModelMocks.append.mock.calls.at(0)).toMatchSnapshot()
   })
 
+  //test good - give good semver range with lower and upper bounds, returns valid query
+  //test fail - invalid range
+  test('getQuerySyntax > good'),
```
JR40159, line 368: I think `getQuerySyntax` shouldn't be exported so here you would need to test calling `getModelReleases with a semverQuery` and verify that the match is what you expect.

JR40159 (changes requested):

backend/src/services/release.ts
IR96334: I have changed the name. I have exported as I test it in release.spec.ts test file.

backend/src/services/release.ts
IR96334: Changed

backend/test/services/release.spec.ts
JR40159, line 368: I don't think your new tests are correctly running. The whole test including the async functions needs to be included in the call to `test()`. As such this line should be `  test('getQuerySyntax > good',` and 432 should have the closing bracket.

backend/src/services/release.ts
IR96334: Changed

backend/test/services/release.spec.ts
JR40159: Make sure you look at the test coverage to check if there's anywhere you've missed.

JR40159 (changes requested):

backend/src/services/release.ts
IR96334: Done. I still believe I need all 3 replaces as I don't know whether expressionA is lower or upper.

backend/src/services/release.ts
```diff
@@ -56,8 +56,7 @@ async function validateRelease(user: UserInterface, model: ModelDoc, release: Re
     const fileNames: Array<string> = []
 
     for (const fileId of release.fileIds) {
-      let file: Document<unknown, any, FileInterface> &
-        Omit<FileInterface & Required<{ _id: Schema.Types.ObjectId }>, never>
+      let file: FileInterface | (undefined & Omit<FileInterface & Required<{ _id: Schema.Types.ObjectId }>, never>)
```
ARADDCC012, line 59: I'm not sure what's happened here again. `let file: FileInterfaceDoc | undefined` should be the correct type

backend/src/services/release.ts
```diff
@@ -330,50 +339,29 @@ function getSemverQueryBounds(querySemver: string) {
...
-      major: NaN,
-      minor: NaN,
-      patch: NaN,
-    }
+  if (expressionB) {
+    upperInclusivity = expressionB.includes('<=')
+    upperSemver = expressionB.replace(/[<=]/g, '')
+  }
+
+  let lowerSemverObj, upperSemverObj
```
ARADDCC012, line 364: There should both be typed. Also defining multiple values in a single line is something we should probably avoid.

backend/src/services/release.ts
```diff
@@ -330,50 +339,29 @@ function getSemverQueryBounds(querySemver: string) {
...
+    throw BadReq(`Semver range, '${querySemver}' is invalid.`)
   }
 
   const [expressionA, expressionB] = semverRangeStandardised.split(' ')
 
-  let lowerSemver: string = '',
-    upperSemver: string = '',
-    lowerInclusivity = false,
-    upperInclusivity = false
+  let lowerInclusivity, lowerSemver, upperInclusivity, upperSemver
```
ARADDCC012, line 347: These are all `any`, they should be typed. If this was a JS project I wouldn't have a problem with defining multiple vars on a single line like this, but in TS we should either be providing a default value or explicitly typing it as `MyType | undefined`. As a result it's much more readable to define them on separate lines.
```
let var1
let var2
// etc
```

backend/src/services/release.ts
```diff
@@ -276,11 +291,36 @@ export async function getReleasesForExport(user: UserInterface, modelId: string,
...
+  const minorNum: number = Number(minor)
+  const patchNum: number = Number(patch)
+  return { major: majorNum, minor: minorNum, patch: patchNum, ...(metadata && { metadata }) }
+}
+
+export function semverObjectToString(semver: SemverObject): string {
+  if (!semver) {
+    return ''
+  }
+  let metadata: string
```
ARADDCC012, line 309: This type isn't strictly true as it's `undefined` until you assign a value. Instead I'd suggest `let metadata = ''`

backend/src/services/release.ts
IR96334: Hmmm... fixed

backend/src/services/release.ts
IR96334: Yeah removed the typing because I wanted it to be undefined, forgetting that I can give the possible type of undefined!

backend/src/services/release.ts
IR96334: Changed

backend/src/services/release.ts
IR96334: Changed

backend/test/services/release.spec.ts
IR96334: Done

backend/test/services/release.spec.ts
IR96334: Done

IR96334: Done

JR40159 (approved):

PR #1585, opened 2024-10-23
Bugfix/bai 1475 python docs no longer work locally

JR40159:


JR40159:
> Would removing the static docs break our GitHub pages documentation.
> 
> https://gchq.github.io/Bailo/docs/python/index.html
> 
> Might need a pipeline to create the static docs?

Nope, we build the docs as part of our Github workflow: https://github.com/gchq/Bailo/blob/main/.github/workflows/web.yml#L46

ARADDCC012 (approved):

PE39806 (changes requested):
When running on my local machine, I still face `Cannot GET /docs/python/index.html` when trying to access the local Python docs. When execing into the backend docker image, I can see that `/app/python-docs/dirhtml` does not exist but `/app/python-docs/_build/dirhtml` does. This requires fixing.

PR #1591, opened 2024-10-28
Bugfix/bai 1464 deleted access request still appearing on the "your reviews" page

PE39806:
Fix for bug: a deleted access request is still offered to the MSRO or MTR for approval in the "Your Reviews" page, but then fails to open when selected, causing the UI to hang.

### DIFF
--- a/backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
+++ b/backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
@@ -3,19 +3,15 @@ import ReviewModel from '../models/Review.js'
 import { removeAccessRequestReview } from '../services/review.js'
 
 export async function up() {
-  // get the access requests
   const accessRequests = await AccessRequestModel.find({ deleted: false })
-  // get all reviews
   const reviews = await ReviewModel.find({})
-  // Remove access request reviews where the access request is already deleted
   for (const review of reviews) {
     const reviewAccessRequestId = review.get('accessRequestId')
-    if (reviewAccessRequestId !== undefined) {
-      // accessRequestId field indicates this is an access request review
-      if (!accessRequests.some((accessRequest) => accessRequest.get('id') == reviewAccessRequestId)) {
-        // couldn't find a corresponding access request meaning that it must have been deleted, so delete the review
-        removeAccessRequestReview(reviewAccessRequestId)
-      }
+    if (
+      reviewAccessRequestId &&
+      !accessRequests.some((accessRequest) => accessRequest.get('id') == reviewAccessRequestId)
+    ) {
+      removeAccessRequestReview(reviewAccessRequestId)
     }
   }
 }
--- a/backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
+++ b/backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
@@ -1,18 +1,10 @@
 import AccessRequestModel from '../models/AccessRequest.js'
-import ReviewModel from '../models/Review.js'
 import { removeAccessRequestReviews } from '../services/review.js'
 
 export async function up() {
-  const accessRequests = await AccessRequestModel.find({ deleted: false })
-  const reviews = await ReviewModel.find({})
-  for (const review of reviews) {
-    const reviewAccessRequestId = review.get('accessRequestId')
-    if (
-      reviewAccessRequestId &&
-      !accessRequests.some((accessRequest) => accessRequest.get('id') == reviewAccessRequestId)
-    ) {
-      await removeAccessRequestReviews(reviewAccessRequestId)
-    }
+  const deletedAccessRequests = await (AccessRequestModel as any).findDeleted()
+  for (const accessRequest of deletedAccessRequests) {
+    await removeAccessRequestReviews(accessRequest.get('id'))
   }
 }
 
--- a/backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
+++ b/backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
@@ -1,0 +1,25 @@
+import AccessRequestModel from '../models/AccessRequest.js'
+import ReviewModel from '../models/Review.js'
+import { removeAccessRequestReview } from '../services/review.js'
+
+export async function up() {
+  // get the access requests
+  const accessRequests = await AccessRequestModel.find({ deleted: false })
+  // get all reviews
+  const reviews = await ReviewModel.find({})
+  // Remove access request reviews where the access request is already deleted
+  for (const review of reviews) {
+    const reviewAccessRequestId = review.get('accessRequestId')
+    if (reviewAccessRequestId !== undefined) {
+      // accessRequestId field indicates this is an access request review
+      if (!accessRequests.some((accessRequest) => accessRequest.get('id') == reviewAccessRequestId)) {
+        // couldn't find a corresponding access request meaning that it must have been deleted, so delete the review
+        removeAccessRequestReview(reviewAccessRequestId)
+      }
+    }
+  }
+}
+
+export async function down() {
+  /* NOOP */
+}
--- a/backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
+++ b/backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
@@ -11,7 +11,7 @@ export async function up() {
       reviewAccessRequestId &&
       !accessRequests.some((accessRequest) => accessRequest.get('id') == reviewAccessRequestId)
     ) {
-      removeAccessRequestReviews(reviewAccessRequestId)
+      await removeAccessRequestReviews(reviewAccessRequestId)
     }
   }
 }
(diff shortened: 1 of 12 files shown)


backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
```diff
@@ -0,0 +1,25 @@
+import AccessRequestModel from '../models/AccessRequest.js'
+import ReviewModel from '../models/Review.js'
+import { removeAccessRequestReview } from '../services/review.js'
+
+export async function up() {
+  // get the access requests
+  const accessRequests = await AccessRequestModel.find({ deleted: false })
```
ARADDCC002, line 7: I'd remove the comments as the code is very readable without them

backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
```diff
@@ -0,0 +1,25 @@
...
+  // get all reviews
+  const reviews = await ReviewModel.find({})
+  // Remove access request reviews where the access request is already deleted
+  for (const review of reviews) {
+    const reviewAccessRequestId = review.get('accessRequestId')
+    if (reviewAccessRequestId !== undefined) {
+      // accessRequestId field indicates this is an access request review
+      if (!accessRequests.some((accessRequest) => accessRequest.get('id') == reviewAccessRequestId)) {
```
ARADDCC002, line 15: You can change this to

 `if (reviewAccessRequestId && !accessRequests.some((accessRequest) => accessRequest.get('id') == reviewAccessRequestId) { ... }`

Alternatively, you can keep them as separate conditions but throw some sort of error if you can't find `reviewAccessRequestId`

backend/src/services/accessRequest.ts
```diff
@@ -91,6 +91,12 @@ export async function removeAccessRequest(user: UserInterface, accessRequestId:
 
   await accessRequest.delete()
 
+  try {
+    await removeAccessRequestReview(accessRequestId)
+  } catch (error) {
+    log.warn(error, 'Error when creating Release Review Requests. Approval cannot be given to this Access Request')
```
ARADDCC002, line 97: Instead of using `log.error` or `log.warn` we have some utility functions that are used for throwing specific errors. They can be found in [here.](https://github.com/gchq/Bailo/blob/main/backend/src/utils/error.ts)

Here's an example of how to use one:

`if (!myValue) {
  throw BadReq('myValue was undefined!', myValue)
}`

backend/src/services/review.ts
```diff
@@ -141,6 +149,20 @@ export async function findReviewsForAccessRequests(accessRequestIds: string[]) {
...
 
+export async function findReviewForAccessRequest(accessRequestId: string) {
+  const review = await Review.findOne({ accessRequestId: accessRequestId })
+
+  if (!review) {
+    throw NotFound('The requested access request review was not found.', { accessRequestId })
+  }
+
+  if (!requiredRoles.accessRequest.includes(review.role)) {
+    log.warn('Permission error when sending getting review for Access Request.', { accessRequestId })
```
ARADDCC002, line 160: See previous comment about throwing errors

ARADDCC002 (dismissed):

backend/src/services/review.ts
```diff
@@ -141,6 +149,20 @@ export async function findReviewsForAccessRequests(accessRequestIds: string[]) {
...
 }
 
+export async function findReviewForAccessRequest(accessRequestId: string) {
+  const review = await Review.findOne({ accessRequestId: accessRequestId })
+
+  if (!review) {
+    throw NotFound('The requested access request review was not found.', { accessRequestId })
+  }
+
+  if (!requiredRoles.accessRequest.includes(review.role)) {
```
JR40159, line 159: I'm a bit confused by the need to do this check if this function just needs to find existing reviews that relate to a specific access request?

backend/src/services/review.ts
```diff
@@ -141,6 +149,20 @@ export async function findReviewsForAccessRequests(accessRequestIds: string[]) {
   return reviews.filter((review) => requiredRoles.accessRequest.includes(review.role))
 }
 
+export async function findReviewForAccessRequest(accessRequestId: string) {
+  const review = await Review.findOne({ accessRequestId: accessRequestId })
```
JR40159, line 153: This isn't a big issue right now, but preferably I think it would be good if this process of deleting reviews could handle having multiple reviews associated with a single access request.

backend/src/services/review.ts
```diff
@@ -89,6 +89,14 @@ export async function createAccessRequestReviews(model: ModelDoc, accessRequest:
   await Promise.all(createReviews)
 }
 
+export async function removeAccessRequestReview(accessRequestId: string) {
+  const accessRequestReview = await findReviewForAccessRequest(accessRequestId)
```
JR40159, line 93: Did you consider just using a single delete Mongo query rather than two separate Mongo queries?

backend/src/services/review.ts
```diff
@@ -89,6 +89,14 @@ export async function createAccessRequestReviews(model: ModelDoc, accessRequest:
...
+
+  await accessRequestReview.delete()
+
+  return { accessRequestId }
```
JR40159, line 97: I would probably return the result of the call to delete the document in Mongo rather than the access request ID that was provided via the method parameter.

JR40159 (changes requested):

backend/src/services/review.ts
PE39806: I took this from the existing functionality in the above `findReviewsForAccessRequests` function https://github.com/gchq/Bailo/blob/8f15cdfa16d6815642093db6145bd6748c8e2ed8/backend/src/services/review.ts#L149

backend/src/services/review.ts
PE39806: Good thinking. I will refactor the functions to do this.

PE39806: In changing from removing only a single review to removing multiple reviews for a single access request, it is not possible to do this as a single query. A `find`, followed by `deleteMany` will be required to handle multiple reviews.

backend/src/services/review.ts
```diff
@@ -89,6 +91,18 @@ export async function createAccessRequestReviews(model: ModelDoc, accessRequest:
   await Promise.all(createReviews)
 }
 
+export async function removeAccessRequestReviews(accessRequestId: string): Promise<DeleteResult> {
+  const deletions = await Review.deleteMany({ accessRequestId: accessRequestId })
+
+  if (deletions.deletedCount === 0) {
```
JR40159, line 97: I don't think the is necessarily an error. For example, the model might have been missing the relevant review roles when the access request was made so new reviews were made for that access request.

backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
```diff
@@ -0,0 +1,21 @@
...
+export async function up() {
+  const accessRequests = await AccessRequestModel.find({ deleted: false })
+  const reviews = await ReviewModel.find({})
+  for (const review of reviews) {
+    const reviewAccessRequestId = review.get('accessRequestId')
+    if (
+      reviewAccessRequestId &&
+      !accessRequests.some((accessRequest) => accessRequest.get('id') == reviewAccessRequestId)
+    ) {
+      removeAccessRequestReviews(reviewAccessRequestId)
```
JR40159, line 14: `await`?

backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
```diff
@@ -0,0 +1,21 @@
+import AccessRequestModel from '../models/AccessRequest.js'
+import ReviewModel from '../models/Review.js'
+import { removeAccessRequestReviews } from '../services/review.js'
+
+export async function up() {
+  const accessRequests = await AccessRequestModel.find({ deleted: false })
+  const reviews = await ReviewModel.find({})
```
JR40159, line 7: Why do you need to iterate though every review (there could be a lot)? Surely once you have all of the deleted access requests you just need to call `removeAccessRequestReviews()` for each of them with the `accessRequestID`.

JR40159 (changes requested):

backend/src/migrations/008_delete_reviews_with_removed_access_requests.ts
PE39806: The mongoose-delete plugin doesn't work nicely in typescript. Ideally, I'd make `AccessRequestModel` a `SoftDeleteModel` and use the `findDeleted` method but this doesn't seem to work as expected, even with `npm install --save @types/mongoose-delete` helping out. Instead, we have to iterate over all of the reviews and find which ones are "orphaned" by not having an associated access request, and removing them. It would be worth revisiting this at another time to get this missing functionality as it could be useful for future use.

PE39806: Using the not-best-practice `(AccessRequestModel as any).findDeleted()` has allowed for a workaround. This should really be sorted for any future work.

package.json
```diff
@@ -17,5 +17,8 @@
     "husky": "^9.1.6",
     "lint-staged": "^15.2.10",
     "prettier": "^3.3.3"
+  },
+  "dependencies": {
+    "@types/mongoose-delete": "^1.0.6"
```
JR40159, line 22: I don't think we need this as this type turned out to be useless?

JR40159 (changes requested):

JR40159 (approved):

PR #1612, opened 2024-11-07
BAI-1500 ModelScan REST API

PE39806:
https://github.com/protectai/modelscan is a Python tool to scan model artefacts. It is installed via Pip and then can be interacted with via the Python library or the CLI.
We need it to be accessible via a REST API.

The new application will be added here: https://github.com/gchq/Bailo/tree/main/lib


JR40159 (changes requested):
Is it worth having a few basic tests?

JR40159 (approved):

PR #1621, opened 2024-11-13
Dockerise our ModelScan application

PE39806:
Write a Dockerfile for the REST Model Scan application.

The Dockerfile should expose the relevant ports.

The Image should be added to the dev docker compose build.

### DIFF
--- a/lib/modelscan_api/Dockerfile.dev
+++ b/lib/modelscan_api/Dockerfile.dev
@@ -14,8 +14,6 @@ COPY ./requirements.txt /app/requirements.txt
 
 RUN pip install --no-cache-dir --upgrade -r /app/requirements.txt
 
-COPY ./bailo_modelscan_api /app/bailo_modelscan_api
-
 EXPOSE 80
 
-CMD ["fastapi", "run", "bailo_modelscan_api/main.py", "--port", "80"]
+CMD ["fastapi", "dev", "bailo_modelscan_api/main.py", "--port", "80", "--host", "0.0.0.0"]
--- a/lib/modelscan_api/Dockerfile.dev
+++ b/lib/modelscan_api/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM python:3.12-slim
+FROM python:3.12-slim-bullseye
 
 # Prevents Python from writing pyc files.
 ENV PYTHONDONTWRITEBYTECODE=1
--- a/lib/modelscan_api/bailo_modelscan_api/main.py
+++ b/lib/modelscan_api/bailo_modelscan_api/main.py
@@ -41,6 +41,15 @@ app = FastAPI(
 modelscan = ModelScan(settings=get_settings().modelscan_settings)
 
 
+@app.get("/health")
+def health_check() -> dict[str, str]:
+    """Minimal health check for the API endpoint.
+
+    :return: always `{"status": "healthy"}`
+    """
+    return {"status": "healthy"}
+
+
 @app.post(
     "/scan/file",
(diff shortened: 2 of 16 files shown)


JR40159 (changes requested):

JR40159 (changes requested):
It might be worth fixing the example return objects in the Swagger docs

lib/modelscan_api/Dockerfile.dev
```diff
@@ -0,0 +1,19 @@
+FROM python:3.12-slim
```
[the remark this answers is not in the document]
PE39806: I had not. It looks like `3.12-slim` resolves to `3.12-slim-bookend` which is one major Debian release ahead of `bullseye`. I'm happy to downgrade to `3.12-slim-bullseye` as there are no meaningful functionality changes.
Is there a reason we don't use the `slim` docker image for the docs build? If there is a known issue then I'd like to investigate that here too.

lib/modelscan_api/bailo_modelscan_api/main.py
```diff
@@ -101,6 +130,7 @@ def scan_file(in_file: UploadFile, background_tasks: BackgroundTasks) -> dict[st
     finally:
         try:
             # Clean up the downloaded file as a background task to allow returning sooner.
+            # If using a temporary dir then this would happen anyway, but if Settings' download_dir evaluates then this is required.
```
JR40159, line 133: The /tmp only gets cleaned on restarts?

lib/modelscan_api/bailo_modelscan_api/main.py
```diff
@@ -101,6 +130,7 @@ def scan_file(in_file: UploadFile, background_tasks: BackgroundTasks) -> dict[st
...
             logger.info("Cleaning up downloaded file.")
             background_tasks.add_task(Path.unlink, pathlib_path, missing_ok=True)
         except UnboundLocalError:
```
JR40159, line 136: When logging the exception could you include info about which file has caused the error?

lib/modelscan_api/bailo_modelscan_api/main.py
```diff
@@ -58,33 +77,43 @@ def scan_file(in_file: UploadFile, background_tasks: BackgroundTasks) -> dict[st
...
+                        out_file.write(content)
+            except OSError as exception:
+                logger.exception("Failed writing the file to the disk.")
+                raise HTTPException(
+                    status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+                    detail=f"An error occurred while trying to write the uploaded file to the disk: {exception}",
+                ) from exception
+
+            # Scan the uploaded file.
+            result = modelscan.scan(pathlib_path)
```
JR40159, line 113: An info log either side of this might be useful

JR40159 (changes requested):

lib/modelscan_api/bailo_modelscan_api/main.py
PE39806: Python's TemporaryDirectory creates a subdir of `/tmp` (e.g. `/tmp/zbc123xyz`) which is automatically cleaned up periodically https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory.cleanup

lib/modelscan_api/bailo_modelscan_api/main.py
PE39806: `logger.exception` automatically includes the raised exception details

JR40159 (approved):

PR #1624, opened 2024-11-13
Feature/bai 1485 manual user access

ARADDCC002:


frontend/src/entry/settings/EntryAccessInput.tsx
```diff
@@ -75,6 +93,23 @@ export default function EntryAccessInput({ value, onUpdate, entryKind, entryRole
     handleInputChange(event, value)
   }, 500)
 
+  const handleAddEntityManuallyOnClick = useCallback(async () => {
```
ARADDCC012, line 96: This doesn't need to be using `useCallback`

frontend/src/entry/settings/ManualEntryAccess.tsx
```diff
@@ -0,0 +1,67 @@
+import ExpandMoreIcon from '@mui/icons-material/ExpandMore'
+import { Accordion, AccordionDetails, AccordionSummary, Button, Stack, TextField, Typography } from '@mui/material'
+import { getUserInformation } from 'actions/user'
+import { useState } from 'react'
+import MessageAlert from 'src/MessageAlert'
+import { CollaboratorEntry, EntityKind } from 'types/types'
+import { getErrorMessage } from 'utils/fetcher'
+
+interface ManualEntryAccessProps {
+  accessList: CollaboratorEntry[]
```
ARADDCC012, line 10: This component doesn't have to care about the access list. It can probably have a single prop `onAddEntityManually: (entityName: string) => void`. We can let the main `EntryAccessInput` component manage the access list logic, though I also think the checks we're doing within `handleAddEntityManuallyOnClick` can and should be handled by the backend.

frontend/src/entry/settings/ManualEntryAccess.tsx
```diff
@@ -0,0 +1,67 @@
...
+  setAccessList: (accessList: CollaboratorEntry[]) => void
+}
+
+export default function ManualEntryAccess({ accessList, setAccessList }: ManualEntryAccessProps) {
```
ARADDCC012, line 14: Consider a name change to something like `ManualEntityInput` as this component isn't responsible for managing entry access in itself.

frontend/src/entry/settings/ManualEntryAccess.tsx
```diff
@@ -0,0 +1,67 @@
...
+        id='manual-user-add-header'
+      >
+        <Typography sx={{ mr: 1 }} component='caption'>
+          Trouble finding a user? Click here to add them manually
+        </Typography>
+      </AccordionSummary>
+      <AccordionDetails sx={{ p: 0 }}>
+        <Stack spacing={2} direction={{ xs: 'column', sm: 'row' }}>
+          <TextField
+            id='manual-entity-name-select'
```
ARADDCC012, line 51: Doesn't look like this `id` is needed

frontend/src/entry/settings/ManualEntryAccess.tsx
```diff
@@ -0,0 +1,67 @@
...
+            placeholder='Joe Bloggs'
```
ARADDCC012, line 52: Correct me if I'm wrong here, but aren't we expecting users to provide a dn here? I'm not sure this placeholder is useful if that's the case.

Depending on how intuitive it is to know what information should be provided by a user, we may benefit from using a help icon that explains what can be provided.

frontend/src/entry/settings/ManualEntryAccess.tsx
ARADDCC002: Good spot, I think this was left over from when I was using a FormControl to add a label to it .

frontend/actions/user.ts
```diff
@@ -49,9 +49,9 @@ interface UserInformationResponse {
   entity: UserInformation
 }
 
-export function useGetUserInformation(dn: string) {
+export function useGetUserInformation(dn: string, kind?: string) {
```
ARADDCC012, line 52: Can we give `kind` a more specific type? I'm assuming this can be `'user'` or `'group'` but I'm not sure just by looking at it.

backend/src/services/model.ts
```diff
@@ -319,6 +322,18 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
   if (modelDiff.settings?.mirror?.destinationModelId && modelDiff.settings?.mirror?.sourceModelId) {
     throw BadReq('You cannot select both mirror settings simultaneously.')
   }
+  if (modelDiff.collaborators) {
+    const invalidUsers = []
+    modelDiff.collaborators.forEach(async (collaborator) => {
```
ARADDCC012, line 327: We can't await the promises from this async callback when using `.forEach()` as it doesn't return anything. Instead we can use `.map()` like so:
```
await Promise.all(modelDiff.collaborators.map(async (collaborator) => { ... }))
```

backend/src/services/model.ts
```diff
@@ -319,6 +322,18 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
   if (modelDiff.settings?.mirror?.destinationModelId && modelDiff.settings?.mirror?.sourceModelId) {
     throw BadReq('You cannot select both mirror settings simultaneously.')
   }
+  if (modelDiff.collaborators) {
+    const invalidUsers = []
```
ARADDCC012, line 326: Can we type this array please? I also thing we should change this to store the `collaboration.entity` so that we can inform users of what has failed by returning exactly what they provided.
```
const invalidUser: CollaboratorEntry[] = []
```

backend/src/services/model.ts
```diff
@@ -319,6 +322,18 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
...
+    modelDiff.collaborators.forEach(async (collaborator) => {
+      const userInformation = await authentication.getUserInformation(collaborator.entity)
+      if (!userInformation) {
+        invalidUsers.push(userInformation)
```
ARADDCC012, line 330: As mentioned in another comment (https://github.com/gchq/Bailo/pull/1624/files#r1843611009) I think it's better if we store the entity string that's invalid, so we can return that to the user.
```
invalidEntities.push(collaborator.entity)
```

backend/src/services/model.ts
```diff
@@ -319,6 +322,18 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
...
+      }
+    })
+    if (invalidUsers.length > 0) {
+      throw BadReq('One or more user is invalid')
```
ARADDCC012, line 334: We could tell the user which users are invalid by doing something like:
```
// We probably need to have the same entry.kind formatting logic here to make this a nice error.
// We'll also need to replicate the plural functionality in the backend to allow this to work.
throw BadReq(
  `Failed to update ${`model.kind`}. Invalid ${invalidEntities.length === 1 ? 'entity' : 'entities'}: `${invalidEntities.join(', ')}`
  )
```

frontend/actions/user.ts
```diff
@@ -63,6 +63,12 @@ export function useGetUserInformation(dn: string) {
   }
 }
 
+export function getUserInformation(dn: string) {
```
ARADDCC012, line 66: This won't be needed if we let the backend handle all of the validation.

frontend/src/entry/settings/ManualEntryAccess.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export default function ManualEntryAccess({ accessList, setAccessList }: ManualEntryAccessProps) {
+  const [manualEntityName, setManualEntityName] = useState('')
+  const [errorMessage, setErrorMessage] = useState('')
+
+  const handleAddEntityManuallyOnClick = async () => {
+    setErrorMessage('')
+    if (manualEntityName !== undefined && manualEntityName !== '') {
+      if (accessList.find((collaborator) => collaborator.entity === `${EntityKind.USER}:${manualEntityName}`)) {
+        return setErrorMessage(`The requested user has already been added below.`)
```
ARADDCC012, line 22: The backend should handle this, not the frontend. An API user can bypass this check.

frontend/src/entry/settings/ManualEntryAccess.tsx
```diff
@@ -0,0 +1,67 @@
...
+      }
+      const response = await getUserInformation(manualEntityName)
```
ARADDCC012, line 24: This should be handled by the backend

frontend/src/entry/settings/ManualEntryAccess.tsx
```diff
@@ -0,0 +1,67 @@
...
+      if (!response.ok) {
+        return setErrorMessage(await getErrorMessage(response))
+      }
+      const updatedAccessList = [...accessList]
+      const newAccess = { entity: `${EntityKind.USER}:${manualEntityName}`, roles: [] }
+      updatedAccessList.push(newAccess)
+      setAccessList(updatedAccessList)
```
ARADDCC012, line 31: Consider condensing this like so:
```
setAccessList([...accessList, { entity: `${EntityKind.USER}:${manualEntityName}`, roles: [] }])
```

It turns 4 lines into 1 and I think it's also easier to read.

frontend/actions/user.ts
ARADDCC002: This is no longer needed

backend/src/services/model.ts
```diff
@@ -305,7 +305,10 @@ export async function updateModelCard(
   return revision
 }
 
-export type UpdateModelParams = Pick<ModelInterface, 'name' | 'teamId' | 'description' | 'visibility'> & {
+export type UpdateModelParams = Pick<
+  ModelInterface,
+  'name' | 'teamId' | 'description' | 'visibility' | 'collaborators'
```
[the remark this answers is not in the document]
ARADDCC002: I hadn't merged main in that's why, merged and fixed. :)

frontend/src/entry/settings/ManualEntryAccess.tsx
ARADDCC002: This is UI validation to make sure you're not adding the same user to the form below, I think it's still worthwhile having.

backend/src/services/model.ts
```diff
@@ -319,6 +322,18 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
   if (modelDiff.settings?.mirror?.destinationModelId && modelDiff.settings?.mirror?.sourceModelId) {
     throw BadReq('You cannot select both mirror settings simultaneously.')
   }
+  if (modelDiff.collaborators) {
```
JR40159, line 325: Can you pull this functionality out into its own function? It would be useful to reuse this logic in create model as well

backend/src/services/model.ts
JR40159: I'd prefer to keep the message simple and add this information into a context object as we do with other errors across the app.

backend/src/services/model.ts
JR40159, line 325: Don't forget to add tests for this! :)

backend/src/services/model.ts
```diff
@@ -319,6 +322,18 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
...
+    const invalidUsers = []
+    modelDiff.collaborators.forEach(async (collaborator) => {
+      const userInformation = await authentication.getUserInformation(collaborator.entity)
```
JR40159, line 328: I think this should throw an error if the user can't be found. It's a bit difficult as this uses our connectors and the implementation could be anything but hopefully the typing of `getUserInformation()` can enforce that it can't just return nothing

JR40159 (changes requested):

backend/src/services/model.ts
```diff
@@ -335,6 +342,21 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
   return model
 }
 
+async function checkCollaboratorAuthorisation(collaborators: CollaboratorEntry[]) {
```
ARADDCC012, line 345: Can we add a check in here to ensure there's not duplicate collaborators in the list? It's not technically in scope for this ticket, so I'm happy for it to be spun out into a new ticket if necessary.

frontend/src/entry/settings/ManualEntityInput.tsx
```diff
@@ -5,27 +5,21 @@ import { useState } from 'react'
...
 }
 
-export default function ManualEntryAccess({ accessList, setAccessList }: ManualEntryAccessProps) {
+export default function ManualEntityInput({ onAddEntityManually, errorMessage }: ManualEntityInputProps) {
   const [manualEntityName, setManualEntityName] = useState('')
-  const [errorMessage, setErrorMessage] = useState('')
 
   const { uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
 
   const handleAddEntityManuallyOnClick = async () => {
```
ARADDCC012, line 19: This doesn't need to be `async` anymore

frontend/src/entry/settings/EntryAccessInput.tsx
```diff
@@ -72,6 +73,18 @@ export default function EntryAccessInput({ value, onUpdate, entryKind, entryRole
     setUserListQuery(value)
   }, [])
 
+  const handleNewManualEntityEntry = useCallback(
```
ARADDCC012, line 76: This doesn't need to be using `useCallback`.

It also should have a name that matches the event it's linked to: `handleAddEntityManually`.

Side note, we should be careful about our use of the word `entry` as it has a specific meaning in the codebase.

frontend/src/entry/settings/EntryAccessInput.tsx
```diff
@@ -28,6 +28,7 @@ export default function EntryAccessInput({ value, onUpdate, entryKind, entryRole
   const [open, setOpen] = useState(false)
   const [accessList, setAccessList] = useState<CollaboratorEntry[]>(value)
   const [userListQuery, setUserListQuery] = useState('')
+  const [manualEntityEntryErrorMessage, setManualEntityEntryErrorMessage] = useState('')
```
ARADDCC012, line 31: This has nothing to do with entries. Consider renaming to `manualEntityInputErrorMessage`

backend/src/services/model.ts
```diff
@@ -335,6 +342,39 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
...
```
JR40159, line 345: Could you change the name of this to `validateCollaborators` as this function isn't related to authorisation?

backend/src/services/model.ts
```diff
@@ -335,6 +342,39 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
...
+  if (duplicateEntities.length > 0) {
+    throw BadReq(`The following duplicate collaborators have been found: ${duplicateEntities.join(', ')}`)
+  }
+  await Promise.all(
+    collaborators.map(async (collaborator) => {
+      if (collaborator.entity === '') {
+        throw BadReq('Collaborator name must be a valid string')
+      }
+      try {
+        await authentication.getUserInformation(collaborator.entity)
```
JR40159, line 370: Ideally it would be good to only do this check on just the collaborators that have been added to save having to make repeated external calls for entities that have already been validated. However, this would mean having to do a DB call to get the existing collaborators array so you could diff it with this one

JR40159 (changes requested):

frontend/src/entry/settings/ManualEntityInput.tsx
```diff
@@ -16,7 +16,8 @@ export default function ManualEntityInput({ onAddEntityManually, errorMessage }:
 
   const { uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
 
-  const handleAddEntityManuallyOnClick = async () => {
+  const handleAddEntityManuallyOnClick = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault()
```
ARADDCC012, line 20: It's not always clear what behaviour we're trying to prevent when doing this sort of thing. Can we add a brief comment to explain?

ARADDCC002: This stops the page from refreshing when you submit the form

backend/src/services/model.ts
```diff
@@ -335,6 +342,53 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
...
+    }
+    return acc
+  }, [])
+  await Promise.all(
+    newCollaborators.map(async (collaborator) => {
+      if (collaborator === '') {
+        throw BadReq('Collaborator name must be a valid string')
+      }
+      try {
+        await authentication.getUserInformation(collaborator)
```
JR40159, line 380: This check will only work for users. If the new UI functionality only works for users and not groups, I'm happy that checking the validity of groups is out of scope for this. However, the collaborator you are checking here must be a user.

backend/src/services/model.ts
```diff
@@ -335,6 +342,53 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
...
+      } catch (err) {
+        if (err instanceof Error && err.name === 'Not Found') {
+          throw NotFound(`Unable to find user ${collaborator}`, { err })
```
JR40159, line 383: I don't think there is any point in catching a not found, then throwing a not found. I thought we might want to catch a not found and then throw a bad request. If we do want to throw a not found here, I don't think this catch is required and whatever gets thrown by `getUserInformation()` should just propagate up.

JR40159 (changes requested):

backend/src/services/model.ts
```diff
@@ -335,6 +342,53 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
   return model
 }
 
+export async function checkCollaboratorAuthorisation(
```
JR40159, line 345: This function doesn't check authorisation, could we change the name to something like `validateCollaborators`

JR40159 (changes requested):

backend/src/services/model.ts
JR40159: Also, you shouldn't need to export it as it should only be used from within this file?

JR40159 (approved):

JR40159 (dismissed):
Wait for a release.

PR #1625, opened 2024-11-18
Updated backend test config

ARADDCC012:


backend/test/testUtils/setupTestConfig.ts
```diff
@@ -0,0 +1,3 @@
+import { vi } from 'vitest'
+
+vi.mock('../../src/utils/config.js')
```
ARADDCC012, line 3: Note to check whether this should be:
```
import config from '../../src/utils/__mocks__/config.js'

vi.mock('../../src/utils/config.js', () => config)
```

ARADDCC002 (approved):

PR #1628, opened 2024-11-22
Feature/bai 1459 create a new file scan connector for modelscan

PE39806:


ARADDCC002 (changes requested):

backend/src/connectors/fileScanning/modelScan.ts
```diff
@@ -0,0 +1,206 @@
...
+    {
+      description: string
+      operator: string
+      module: string
+      source: string
+      scanner: string
+      severity: string
+    },
+  ]
+  // TODO: currently unknown what this might look like
```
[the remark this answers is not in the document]
PE39806: A ticket has been created to add this typing and handle the associated error(s)

backend/config/docker_compose.cjs
```diff
@@ -65,11 +65,15 @@ module.exports = {
...
+
+    modelscan: {
+      host: 'modelscan',
+    },
   },
 
   connectors: {
     fileScanners: {
-      kinds: ['clamAV'],
+      kinds: [],
```
JR40159, line 76: Could we not have `['clamAV,''ModelScan']` for the docker compose config?

backend/src/connectors/fileScanning/modelScan.ts
```diff
@@ -0,0 +1,206 @@
...
+  errors: object[]
+}
+
+class NodeModelScanAPI {
```
JR40159, line 63: I'm not a huge fan of having multiple classes in one file and I do think we necessarily need to use a class here. Rather than using a class I would prefer we created a modelScan client in https://github.com/gchq/Bailo/tree/main/backend/src/clients . Most other clients we have follow a pretty similar pattern to this in wanting to define a simple set of methods that make http requests and return the data.

backend/src/connectors/fileScanning/modelScan.ts
```diff
@@ -0,0 +1,206 @@
...
+    super()
+  }
+
+  info() {
+    return [modelScanToolName]
+  }
+
+  async init() {
+    try {
+      av = await new NodeModelScanAPI().init(config.avScanning.modelscan)
```
JR40159, line 162: Following from previous comment, rather than initialising a new class that I don't think is needed, we could just make a call to the model scan info endpoint to check an instance is reachable.

JR40159 (changes requested):

backend/src/clients/modelScan.ts
JR40159, line 1: Could you write some unit tests for this new client please? We have unit tests for our existing clients if you need examples.

JR40159 (dismissed):

ARADDCC002 (approved):

PR #1633, opened 2024-11-27
Feature/bai 1461 manual rerequest filescanning

ARADDCC002:


PE39806:
The "Last ran at..." text does not appear when viruses are found, and it would be helpful to see the time for when a scan was last run to check whether re-running a scan has completed or not.
image

JR40159 (changes requested):

JR40159 (dismissed):
As mentioned in the ticket, this might require a migration as files that existing prior to adding the ability to scan them might be missing the `avScan` property. Could you check this please?

ARADDCC002:
> As mentioned in the ticket, this might require a migration as files that existing prior to adding the ability to scan them might be missing the `avScan` property. Could you check this please?

Does migration 009 not cover this use case?

PE39806:
Test what happens when a scanner is killed off part way through a scan - we want to be able to catch this and update the scan result to "could not complete" rather than indefinitely reporting "virus scan in progress".

backend/src/connectors/fileScanning/clamAv.ts
```diff
@@ -2,14 +2,15 @@ import NodeClam from 'clamscan'
...
+import { FileInterfaceDoc } from '../../models/File.js'
 import log from '../../services/log.js'
 import config from '../../utils/config.js'
 import { ConfigurationError } from '../../utils/error.js'
-import { BaseFileScanningConnector, FileScanResult } from './Base.js'
+import { BaseFileScanningConnector, FileScanResult, ScanState } from './Base.js'
 
 let av: NodeClam
 export const clamAvToolName = 'Clam AV'
+const maxInitRetries = 5
```
PE39806, line 13: I'd like the max retry count `maxInitRetries = 5` and delay between pings `10000` to be configurable like `retryDelayInMinutes` e.g. `connectors.fileScanners.maxInitRetries` and `connectors.fileScanners.initRetryDelay`.

backend/test/clients/modelScan.spec.ts
```diff
@@ -113,8 +113,8 @@ describe('clients > modelScan', () => {
     )
   })
 
-  test('scanStream > rejected', async () => {
-    fetchMock.default.mockRejectedValueOnce('Unable to communicate with the inferencing service.')
+  test('scanFile > rejected', async () => {
```
PE39806, line 116: Revert `scanFile` to `scanStream`

PE39806 (approved):

PR #1638, opened 2024-11-28
fix: enable arg & kwarg passing to superclass for sub agents

onecrazygenius:
# Affected core subsystem(s):
python library

# Description of change:
Updated tokenagent and pkiagent to accept args and kwargs for super.__init__(), which will allow using these classes with verify=False

### DIFF
--- a/lib/python/src/bailo/core/agent.py
+++ b/lib/python/src/bailo/core/agent.py
@@ -69,6 +69,8 @@ class PkiAgent(Agent):
         cert: str,
         key: str,
         auth: str,
+        *args,
+        **kwargs
     ):
         """Initiate an agent for PKI authentication.
 
@@ -76,7 +78,7 @@ class PkiAgent(Agent):
         :param key: Path to key file
         :param auth: Path to certificate authority file
         """
-        super().__init__(verify=auth)
+        super().__init__(verify=auth, *args, **kwargs)
 
         self.cert = cert
         self.key = key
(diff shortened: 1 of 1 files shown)


PE39806 (changes requested):

lib/python/src/bailo/core/agent.py
```diff
@@ -102,13 +104,16 @@ def __init__(
         self,
         access_key: str | None = None,
         secret_key: str | None = None,
+        *args,
```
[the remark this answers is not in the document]
onecrazygenius: will remove args from both class's per that expectation, understand that reasoning of signatures! just was a usecase where we need verify=False but wanted with TokenAgent. will make the changes and recommit w/ black checks

lib/python/src/bailo/core/agent.py
```diff
@@ -64,19 +64,14 @@ def put(self, *args, **kwargs):
 
 
 class PkiAgent(Agent):
-    def __init__(
-        self,
-        cert: str,
-        key: str,
-        auth: str,
-    ):
+    def __init__(self, cert: str, key: str, auth: str, **kwargs):
```
onecrazygenius, line 67: @PE39806 I have run black which is collapsing this line now. is this an issue?

onecrazygenius:
made changes to remove *args, but keep *kwarg forwarding to super.__init__

lib/python/src/bailo/core/agent.py
PE39806: That is fine. I think that, at some point, black's config has been modified and this is the first time this file has been changed since then.

PE39806 (approved):
This is all great. Thank you for implementing this change!

PR #1663, opened 2024-12-05
Stream files to the ModelScan API endpoint rather than hold entire files in memory

PE39806:


backend/src/clients/modelScan.ts
```diff
@@ -65,13 +66,21 @@ export async function getModelScanInfo() {
   return (await res.json()) as ModelScanInfoResponse
 }
 
-export async function scanFile(file: Blob, file_name: string) {
+export async function scanStream(stream: Readable, file_name: string, file_size: number) {
```
JR40159, line 69: Change snake case to camel case- this isn't python!

JR40159 (changes requested):

JR40159 (approved):

PR #1677, opened 2024-12-10
Started work on new bulk endpoint for fetching friendly names

IR96334:


JR40159 (changes requested):
It seems like quite a lot of overhead in the backend (adding a whole new endpoint method to the authentication connector) just to support getting friendly names for reactions? Would the frontend mapping the users to their user info by calling the existing endpoint for each of them be that much slower?

JR40159 (approved):

PR #1680, opened 2024-12-11
Feature/bai 1549 edit sceham details via UI

ARADDCC002:


frontend/src/common/EditableText.tsx
```diff
@@ -0,0 +1,59 @@
...
+  submitButtonText = 'Submit',
+}: EditableTextProps) {
+  const [isEditMode, setIsEditMode] = useState(false)
+  const [newValue, setNewValue] = useState(value)
+
+  const previousText = value
+
+  function handleCancelOnClick() {
+    setIsEditMode(false)
+    setNewValue(previousText)
```
ARADDCC012, line 25: `previousValue` isn't needed, you can do `setNewValue(value)`

frontend/src/schemas/SchemaList.tsx
```diff
@@ -133,57 +120,21 @@ export default function SchemaList({ schemaKind }: SchemaDisplayProps) {
...
+            key={schema.id}
+            schema={schema}
+            index={index}
+            schemasLength={schemas.length}
+            open={open}
+            anchorEl={anchorEl}
+            onMenuClose={handleCloseMenu}
+            onDeleteSchemaClick={handleDeleteSchema}
+            onOpenMenuClick={handleOpenMenu}
+            onPatchSchemaClick={handlePatchSchema}
```
ARADDCC012, line 133: In frontend land we usually use the terminology "edit" instead of put/patch. The component doesn't need to care about the request method

frontend/src/common/EditableText.tsx
```diff
@@ -1,34 +1,53 @@
...
-  }
+  }, [previousText])
 
-  function handleSubmit() {
+  const handleSubmit = useCallback(() => {
     setIsEditMode(false)
     onSubmit(newValue)
+  }, [newValue, onSubmit])
+
+  const submitButtons = () => {
```
ARADDCC012, line 38: I'd suggest using useMemo here so we don't have to call this as a function in our JSX below

frontend/src/common/EditableText.tsx
```diff
@@ -1,34 +1,53 @@
...
+    return (
+      <Stack direction='row' spacing={1} sx={{ py: 1 }} justifyContent='flex-end'>
+        <Button variant='contained' type='submit' size='small'>
+          {submitButtonText}
+        </Button>{' '}
```
ARADDCC012, line 43: Space character looks like it shouldn't be here

frontend/src/common/EditableText.tsx
```diff
@@ -1,34 +1,53 @@
...
+  multiline = false,
+  richText = false,
 }: EditableTextProps) {
   const [isEditMode, setIsEditMode] = useState(false)
   const [newValue, setNewValue] = useState(value)
 
   const previousText = value
 
-  function handleCancelOnClick() {
+  const handleCancelOnClick = useCallback(() => {
```
ARADDCC012, line 28: `useCallback` isn't needed here

frontend/src/common/EditableText.tsx
```diff
@@ -1,34 +1,53 @@
...
     setIsEditMode(false)
     setNewValue(previousText)
-  }
+  }, [previousText])
 
-  function handleSubmit() {
+  const handleSubmit = useCallback(() => {
```
ARADDCC012, line 33: `useCallback` isn't needed here

ARADDCC012 (approved):

PR #1682, opened 2024-12-13
a temporary fix for spacing alignment of av chips

ARADDCC002:


PE39806 (approved):

frontend/src/entry/model/releases/FileDownload.tsx
PE39806, line 1: I'm not sure exactly what has changed, but in both releases and release views long file names are no longer truncated and cause the avscanners to disappear
image
image

ARADDCC012 (approved):

frontend/src/entry/model/releases/FileDownload.tsx
PE39806: This is now fixed for >600px width windows, but for windows that are <=600px the name overflows. Ideally, we should fix this but I am aware that not many of our end users are expected to use such a narrow display format.
image

PR #1691, opened 2024-12-18
restructured the model mirror page and added a backend check for exis…

ARADDCC002:
…ting model card before allowing export


backend/config/default.cjs
```diff
@@ -191,7 +191,7 @@ module.exports = {
         enabled: false,
       },
       export: {
-        enabled: false,
+        enabled: true,
```
JR40159, line 194: I don't think this is required?

frontend/src/entry/model/mirroredModels/ExportModelAgreement.tsx
```diff
@@ -46,27 +46,25 @@ export default function ExportModelAgreement({ modelId }: ExportModelAgreementPr
...
-              </LoadingButton>
-            </Restricted>
-            <MessageAlert message={errorMessage} severity='error' />
-          </Stack>
-        </Box>
-      </Card>
-    </Container>
+    <>
+      <Typography variant='h6' component='h1' color='primary' align='center'>
+        Model Export Agreement
```
JR40159, line 51: I don't think it's super clear that this section of the page doesn't just contain the agreement but also the ability to actually kick off an export. Users will be looking for how they export a model, rather than where the agreement is. The agreement is just something they need to do before being able to do the actual thing they want which is exporting a model.

frontend/src/entry/model/mirroredModels/ExportSettings.tsx
JR40159, line 1: I can see why you've moved this to top of the page and and not hidden it by default but I'm not sure whether this will cause confusion with users thinking they need to re-populate this every time they export a model? It still doesn't feel quite right to me how we've combined the model setting required to export a model and the ability to actually do the model export.

JR40159 (changes requested):

JR40159 (approved):

PR #1692, opened 2024-12-18
Feature/import release documents from zip file

JR40159:


backend/src/services/mirroredModel.ts
```diff
@@ -206,7 +157,86 @@ export async function importModel(
   }
 }
 
-async function importModelFile(content: Response, importedPath: string, mirroredModelId: string, sourceModelId) {
+async function importDocuments(res: Response, mirroredModelId: string, sourceModelId: string, payloadUrl: string) {
+  const modelCards: ModelCardRevisionInterface[] = []
+  // TODO - Use this releases array for auditing purposes
```
JR40159, line 162: Yep, I think it would probably make sense to align the release info with the model card info and return the all the semvers that have been imported and the new releases?

backend/src/services/mirroredModel.ts
```diff
@@ -206,7 +157,86 @@ export async function importModel(
...
+    })
+  } catch (error) {
+    log.error({ error }, 'Unable to read zip file.')
+    throw InternalError('Unable to read zip file.', { mirroredModelId })
+  }
+
+  const modelCardJsonStrings: string[] = []
+  const releaseJsonStrings: string[] = []
+  Object.keys(zipContent).forEach(async function (key) {
+    // TODO - We may want to assign these regexes to readable variables (e.g. modelCardRegex, releaseRegex, fileRegex) to make this a bit more readable
```
JR40159, line 180: Yep!

backend/src/services/mirroredModel.ts
```diff
@@ -206,7 +157,86 @@ export async function importModel(
...
+    if (/^[0-9]+\.json$/.test(key)) {
+      modelCardJsonStrings.push(Buffer.from(zipContent[key]).toString('utf8'))
+    } else if (/^releases\/(.*)\.json$/.test(key)) {
+      releaseJsonStrings.push(Buffer.from(zipContent[key]).toString('utf8'))
+    } else if (/^files\/(.*)\.json$/.test(key)) {
+      // TODO - Handle file parsing
```
JR40159, line 186: To be done as part of the next subtask

backend/src/services/mirroredModel.ts
```diff
@@ -206,7 +157,86 @@ export async function importModel(
...
+    } else {
+      // TODO - Is this how we want to handle unrecognised zip content?
```
JR40159, line 188: I think so? Do you have an alternative in mind?

backend/src/services/mirroredModel.ts
```diff
@@ -206,7 +157,86 @@ export async function importModel(
...
+    const modelCard = parseModelCard(modelCardJson, mirroredModelId, sourceModelId)
+    modelCards.push(modelCard)
+  })
+
+  const newModelCards = (await Promise.all(modelCards.map((card) => saveImportedModelCard(card)))).filter(
+    (card): card is ModelCardRevisionInterface => !!card,
+  )
+
+  releaseJsonStrings.forEach((releaseJson) => {
+    // TODO - As this is a forEach loop it will not wait for this iteration to finish before handling the next. Check we are okay with this.
```
JR40159, line 203: I don't think we get any advantage here in using a `forEach` loop so maybe we should just go with the simpler for loop?

ARADDCC012: As `parseRelease` is no longer async this doesn't matter anymore

ARADDCC012 (approved):

PR #1729, opened 2025-01-14
updated landing app to use latest next and mui versions

ARADDCC002:


lib/landing/scripts/generate.js
```diff
@@ -76,7 +77,7 @@ async function run() {
   cp('-R', '../../frontend/src/docs', './src/')
   cp('../../frontend/src/Copyright.tsx', './src/')
   cp('../../frontend/src/Link.tsx', './src/')
-  cp('../../frontend/src/common/Title.tsx', './src/common/')
+  cp('-rf', '../../frontend/src/common/Title.tsx', './src/common/')
```
PE39806, line 80: I'd prefer to have the `r` capitalised to `-Rf` so that it matches the above `R`s.

lib/landing/scripts/generate.js
```diff
@@ -88,8 +89,9 @@ async function run() {
 
   cp('../../frontend/pages/docs/directory.tsx', './pages/docs')
 
+  // TODO fix python docs generation
```
PE39806, line 92: :(

lib/landing/src/docs/DocsWrapper.tsx
```diff
@@ -110,17 +117,17 @@ export default function DocsWrapper({ children }: DocsWrapperProps): ReactElemen
 
   return (
     <>
-      {/* Banner height + Toolbar height = 96px */}
       <Title text='Documentation' />
+      {/* Banner height + Toolbar height = 96px */}
```
PE39806, line 121: This will need tidying up

PE39806 (changes requested):

PE39806 (approved):

PR #1731, opened 2025-01-20
Feature/bai 1510 add a webhook event for updating a release and add webhook documentation to

PE39806:
As a user I want to create a webhook to trigger when a release is updated so that I can be notified about the event.


frontend/pages/docs/users/programmatically-using-bailo/webhooks.mdx
JR40159, line 1: Could you also include what information is included in the body of each webhook?

JR40159 (changes requested):

JR40159 (approved):

PR #1734, opened 2025-01-21
ui improvements to review listings

ARADDCC002:


PE39806 (approved):

frontend/src/entry/model/reviews/ReviewDisplay.tsx
```diff
@@ -11,40 +12,22 @@ export interface ReviewDisplayProps {
...
-        updatedApprovals.push({ ...reviewResponse, role: reviewResponse.role })
-      }
-      if (reviewResponse.decision === Decision.RequestChanges) {
-        updatedRequests.push({ ...reviewResponse, role: reviewResponse.role })
-      }
-    }),
-      setAcceptedReviewResponses(updatedApprovals)
-    setChangesRequestedReviewResponses(updatedRequests)
-  }, [reviewResponses])
+  const orderedReviewResponses = useMemo(() => reviewResponses.sort(sortByCreatedAtDescending) || [], [reviewResponses])
```
ARADDCC012, line 15: As `.sort()` mutates the existing array, I'd suggest using `reviewResponses.toSorted(sortByCreatedAtDescending)`. This works the same as `.sort()` but leaves the original array untouched which will avoid any unwanted behaviour in parent components.

ARADDCC012 (changes requested):

frontend/src/entry/model/reviews/ReviewDisplay.tsx
ARADDCC002: Didn't know about `toSorted` interesting!

ARADDCC012 (approved):

PR #1778, opened 2025-01-22
Bai 1544 show entry kind for all results in top right search

IR96334:
Adds chip denoting entry kind to top-right search results to help user differentiate

### DIFF
--- a/frontend/src/wrapper/EntrySearch.tsx
+++ b/frontend/src/wrapper/EntrySearch.tsx
@@ -1,12 +1,11 @@
 import SearchIcon from '@mui/icons-material/Search'
-import { Box, InputBase, List, ListItemButton, ListItemText, Popover, Stack } from '@mui/material'
+import { Box, InputBase, List, ListItemText, Popover, Stack } from '@mui/material'
 import { alpha, styled, useTheme } from '@mui/material/styles'
-import { useListModels } from 'actions/model'
+import { EntrySearchResult, useListModels } from 'actions/model'
 import { ChangeEvent, useMemo, useState } from 'react'
 import EmptyBlob from 'src/common/EmptyBlob'
 import Loading from 'src/common/Loading'
 import useDebounce from 'src/hooks/useDebounce'
-import Link from 'src/Link'
 import MessageAlert from 'src/MessageAlert'
 
 const Search = styled('div')(({ theme }) => ({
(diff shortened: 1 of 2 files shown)


frontend/src/wrapper/EntrySearch.tsx
```diff
@@ -53,28 +55,34 @@ export default function EntrySearch() {
...
+                  primary={entry.name}
+                  secondary={entry.description}
+                  slotProps={{
+                    primary: {
+                      style: {
+                        whiteSpace: 'nowrap',
+                        textOverflow: 'ellipsis',
+                        overflow: 'hidden',
+                        color: theme.palette.primary.main,
+                        maxWidth: '400px',
```
ARADDCC002, line 75: Because this is set to the same width as the surrounding box, really long names will appear under the entry kind chip. I'd suggest instead of settings this to a specific amount of pixels, giving it a `marginRight`. If you change `style` to `sx` on line 70 you can use `mr: 6`, which should give enough padding.

frontend/src/wrapper/EntrySearch.tsx
```diff
@@ -53,28 +55,34 @@ export default function EntrySearch() {
...
-                  },
 
-                  secondary: {
-                    style: { whiteSpace: 'nowrap', textOverflow: 'ellipsis', overflow: 'hidden' },
-                  },
-                }}
-              />
-            </ListItemButton>
+                    secondary: {
+                      style: { whiteSpace: 'nowrap', textOverflow: 'ellipsis', overflow: 'hidden', maxWidth: '400px' },
```
ARADDCC002, line 80: We can use `sx` here instead of `style`.

frontend/src/wrapper/EntrySearch.tsx
```diff
@@ -53,28 +55,34 @@ export default function EntrySearch() {
...
+              disablePadding
+              secondaryAction={<Chip label={toTitleCase(EntryKindLabel[entry.kind])} size='small' />}
+            >
+              <ListItemButton>
+                <ListItemText
+                  primary={entry.name}
+                  secondary={entry.description}
+                  slotProps={{
+                    primary: {
+                      style: {
```
ARADDCC002, line 70: We can use `sx` here instead of `style`.

ARADDCC002 (changes requested):

frontend/src/wrapper/EntrySearch.tsx
ARADDCC002: Also, see my other comment about changing the `maxWidth` property to be margin instead

ARADDCC002 (approved):

PR #1783, opened 2025-01-24
fixed some text wrapping issues in side nav and file download component

ARADDCC002:


PE39806:
The only problem I can see is if there is a `NavMenuItem` with a very long title, then we may lose some displayed data e.g. here the file names for a Release are lost. If the window is made even smaller and/or if the `Toolbar` is collapsed then this isn't a problem.
image

PE39806 (approved):

ARADDCC002:
> The only problem I can see is if there is a `NavMenuItem` with a very long title, then we may lose some displayed data e.g. here the file names for a Release are lost. If the window is made even smaller and/or if the `Toolbar` is collapsed then this isn't a problem. image

The expanded menu isn't designed to be used at the same time as the main application, it's purely for navigational purposes, so I'm happy that when it's expanded the rest of the app might not be fully usable depending on screen resolution.

PR #1792, opened 2025-01-28
Dev/update ts 5.7.3

ARADDCC002:


frontend/Dockerfile
```diff
@@ -12,6 +12,8 @@ RUN --mount=type=cache,target=/cache/npm npm ci --cache=/cache/npm
 
 COPY . .
 
+RUN npm install sharp
```
PE39806, line 15: Does this need versioning? Also, if added here then it will likely need adding to [Dockerfile.standalone](https://github.com/gchq/Bailo/blob/main/Dockerfile.standalone)

PE39806: Sharp is an optional dependency which causes some issues because of the way it is resolved in `package.lock`, hence being removed from there and `package`.
TODO: add version "^0.33.5" and also add to [Dockerfile.standalone](https://github.com/gchq/Bailo/blob/main/Dockerfile.standalone)

PE39806 (approved):

frontend/eslint.config.mjs
```diff
@@ -0,0 +1,72 @@
...
+        {
+          argsIgnorePattern: '^_',
+          varsIgnorePattern: '^_',
+          caughtErrorsIgnorePattern: '^_',
+        },
+      ],
+
+      '@typescript-eslint/no-extra-semi': 'off',
+      '@typescript-eslint/no-explicit-any': 'off',
+      '@typescript-eslint/no-require-imports': 'off',
```
ARADDCC012, line 56: These two rules didn't exist in the old config. I'm not keen on disabling them across the entire frontend, so if they're needed in specific cases I'd inline disable and add a comment to explain why if we can.
```
      '@typescript-eslint/no-require-imports': 'off',
      '@typescript-eslint/no-unused-expressions': 'off',
```

frontend/eslint.config.mjs
```diff
@@ -0,0 +1,72 @@
...
+      '@typescript-eslint/no-unused-expressions': 'off',
+
+      'react/jsx-newline': [
+        1,
+        {
+          prevent: true,
+        },
+      ],
+
+      'cypress/no-async-tests': 'off',
```
ARADDCC012, line 66: Our Cypress suite has its own eslint config. This should go in there.

On that note, I don't see an updated version of our Cypress eslint config in this PR.

ARADDCC012 (changes requested):

frontend/cypress/.eslintrc.cjs
```diff
@@ -5,6 +5,7 @@ module.exports = {
   plugins: ['cypress'],
   extends: ['plugin:cypress/recommended'],
   rules: {
+    'cypress/no-async-tests': 'off',
```
ARADDCC012, line 8: Sorry I should've looked into this last time I flagged this rule. According to the docs this is included in `plugin:cypress/recommended`, so I'm not keen on turning it off. It also doesn't look like we have any async cypress tests, so I think we can safely remove this line.

frontend/cypress/.eslintrc.cjs
ARADDCC012, line 1: I would prefer to keep all of our eslint configs aligned. Can we update this one to the new format also?

ARADDCC012 (approved):

PR #1796, opened 2025-01-28
Bai 1456 fix model and datacard tab loading

IR96334:


frontend/pages/data-card/[dataCardId].tsx
```diff
@@ -52,16 +56,19 @@ export default function DataCard() {
...
-          subheading={`ID: ${dataCard.id}`}
-          tabs={tabs}
-          requiredUrlParams={{ dataCardId: dataCard.id }}
-          titleToCopy={dataCard.name}
-          subheadingToCopy={dataCard.id}
-        />
+      {!dataCard || isDataCardLoading || isCurrentUserLoading || isUiConfigLoading ? (
+        <Loading />
+      ) : (
+        dataCard && (
```
Comment (ARADDCC012) [inline frontend/pages/data-card/[dataCardId].tsx:62]: Nesting additional conditional logic within a ternary is generally considered something to avoid, simply for the fact that it becomes difficult to read

Comment (ARADDCC012) [inline frontend/pages/data-card/[dataCardId].tsx:62]: Nesting additional conditional logic within a ternary is generally considered something to avoid, simply for the fact that it becomes difficult to read.

In this case it's unnecessary to include the `dataCard &&` condition due to the fact that if the first check (which includes the condition !dataCard) returns false, then dataCard must be truthy (i.e. not `undefined` in this case).

Also, and this is my fault, the loading condition for data cards is simpler than models. That's my mistake for not clarifying in the ticket.

In short, you can do the following:
```
      {!dataCard || isDataCardLoading ? (
        <Loading />
      ) : (
        <PageWithTabs
          title={dataCard.name}
          subheading={`ID: ${dataCard.id}`}
          tabs={tabs}
          requiredUrlParams={{ dataCardId: dataCard.id }}
          titleToCopy={dataCard.name}
          subheadingToCopy={dataCard.id}
        />
      )}
```

frontend/pages/model/[modelId].tsx
```diff
@@ -97,21 +97,24 @@ export default function Model() {
...
-          actionButtonOnClick={requestAccess}
-          requiredUrlParams={{ modelId: model.id }}
-          titleToCopy={model.name}
-          subheadingToCopy={model.id}
-          sourceModelId={model.settings.mirror?.sourceModelId}
-        />
+      {!model || isModelLoading || isCurrentUserLoading || isUiConfigLoading ? (
+        <Loading />
+      ) : (
+        model && (
```
Comment (ARADDCC012) [inline frontend/pages/model/[modelId].tsx:103]: Same here, we don't need this second condition to check the model exists. If it fails the first check then we know the model exists.
```
      {!model || isModelLoading || isCurrentUserLoading || isUiConfigLoading ? (
        <Loading />
      ) : (
        <PageWithTabs
          title={model.name}
          subheading={`ID: ${model.id}`}
          additionalInfo={model.description}
          tabs={tabs}
          displayActionButton={model.card !== undefined}
          actionButtonTitle='Request access'
          actionButtonOnClick={requestAccess}
          requiredUrlParams={{ modelId: model.id }}
          titleToCopy={model.name}
          subheadingToCopy={model.id}
          sourceModelId={model.settings.mirror?.sourceModelId}
        />
      )}
```

frontend/pages/data-card/[dataCardId].tsx
```diff
@@ -18,6 +20,8 @@ export default function DataCard() {
     isModelLoading: isDataCardLoading,
     isModelError: isDataCardError,
   } = useGetModel(dataCardId, EntryKind.DATA_CARD)
+  const { isCurrentUserLoading } = useGetCurrentUser()
```
Comment (ARADDCC012) [inline frontend/pages/data-card/[dataCardId].tsx:23]: These two hooks aren't needed for data cards. Apologies for the ticket being unclear.

ARADDCC012 (changes requested):

frontend/pages/data-card/[dataCardId].tsx
Comment (IR96334) [reply frontend/pages/data-card/[dataCardId].tsx:62]: Yeah, makes sense as the and operator is doing the same thing as my ternary operator so its redundant.

frontend/pages/model/[modelId].tsx
Comment (IR96334) [reply frontend/pages/model/[modelId].tsx:103]: Fixed

frontend/pages/data-card/[dataCardId].tsx
Comment (IR96334) [reply frontend/pages/data-card/[dataCardId].tsx]: Fixed

ARADDCC012 (approved):

Issue #1794, opened 2025-01-28
Add Python 3.13 support


PE39806:
As per https://devguide.python.org/versions/, Python 3.13 is the latest stable release which we need to support. For the Python Bailo client, we should migrate to fully supporting python 3.13 and handle any packages that require changes.
Currently, we only have proper support for up-to Python 3.11 so this will be 2 major releases (3.12 & 3.13) that require supporting, though python is usually very good at maintaining backwards compatibility in its core libraries. It will likely be linting packages like black and pytest which will have the most notable changes.

One big challenge will be maintaining Python 3.9 support, especially as some libraries like Sphinx have already dropped support for them. For some, like Sphinx, this is manageable as Sphinx is only needed when generating docs so is not a core requirement. It is likely worth splitting the existing optional requirements into more than just test e.g. test, docs and dev. This will mean that the core package can continue to support Python 3.9-3.13 while using packages that have dropped support for older Python versions.

The Modelscan REST API makes use of the modelscan package which has a maximum supported version of python 3.12 so this will need upgrading once modelscan adds 3.13 support. This will be covered separately at another time.

PR #1802, opened 2025-01-31
Bump python package versions to support new minimum versions

PE39806:
Also bump the python docs and modelscan API requirements where applicable.


PE39806:
This eases migration for #1794 by increasing the minimum version of dependencies to 3.9 rather than 3.8

JR40159 (approved):

PR #1804, opened 2025-01-31
Upgrade from debian bullseye to bookworm

PE39806:
As per https://www.debian.org/releases/, switch to the latest version of Debian.


ARADDCC002 (dismissed):

PE39806:
I'm holding off on merging this for now as we've had some reports from users of issues when building this.

PE39806:
> I'm holding off on merging this for now as we've had some reports from users of issues when building this.

This seems to now be resolved.

PR #1886, opened 2025-02-04
BAI-1351 : Only create new model cards if the UI detects a change in the model card

IR96334:


frontend/src/entry/overview/FormEditPage.tsx
```diff
@@ -146,6 +154,9 @@ export default function FormEditPage({ entry, readOnly = false }: FormEditPagePr
                     onClick={() => {
                       handleActionButtonClose()
                       setIsEdit(!isEdit)
+                      setOldSchema((splitSchema) => {
```
ARADDCC012, line 157: I don't think this is doing what you're expecting it to do. This is taking the existing value of `oldSchema` and setting it as the updated value via `Object.assign()`, which simply creates a shallow copy (any nested objects will be not be copies, but instead the same objects will be used in both old and new root objects).

When passing a function as a parameter to a setter function, the parameter (in this case you've called that `splitSchema`, this is not the `splitSchema` state variable you've just used the same name) is the current value of that state variable. So your logic is saying "take the existing value and use that at the new value, but create a copy of the root object", which isn't actually resulting in any meaningful change.

What you actually want to do is copy the value of the state variable `splitSchema` and use that to update the value of `oldSchema`, along with creating a deep copy of the original `splitSchema` so that any nested objects are also copies.

```
// You'll need to import this to allow us to use cloneDeep
import * as _ from 'lodash-es'

setOldSchema(_.cloneDeep(splitSchema))
```

ARADDCC012 (changes requested):

frontend/src/entry/overview/FormEditPage.tsx
IR96334: Done

ARADDCC012 (approved):

ARADDCC012 (changes requested):
Update RJSF dependencies

ARADDCC012 (approved):

PR #1887, opened 2025-02-04
fix bad clamav filesystem permissions

PE39806:


infrastructure/helm/bailo/values.yaml
ARADDCC013, line 1: image: clamav/clamav:latest_base # https://docs.clamav.net/manual/Installing/Docker.html#the-official-images-on-docker-hub
require support for unpriv access to run in EKS and OpenShift. Please keep the base image.

ARADDCC013 (approved):

PR #1892, opened 2025-02-05
added new optional configurable organisation and state fields for ent…

ARADDCC002:
…ries


backend/src/routes/v2/model/patchModel.ts
```diff
@@ -12,7 +12,10 @@ import { parse } from '../../../utils/validate.js'
 export const patchModelSchema = z.object({
   body: z.object({
     name: z.string().optional().openapi({ example: 'Yolo v4' }),
+
     kind: z.nativeEnum(EntryKind).optional().openapi({ example: EntryKind.Model }),
+    organisation: z.string().optional().openapi({ example: 'My Organisation' }),
```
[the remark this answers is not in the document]
JR40159: Also this implementation does mean that the options for both are only enforced/available via the UI not the API or python client? Enforcing that the user is always restricted to a set of options for both will ensure consistency

JR40159 (changes requested):

backend/src/routes/v2/model/patchModel.ts
```diff
@@ -7,12 +7,16 @@ import audit from '../../../connectors/audit/index.js'
...
 import { modelInterfaceSchema, registerPath } from '../../../services/specification.js'
+import config from '../../../utils/config.js'
 import { parse } from '../../../utils/validate.js'
 
 export const patchModelSchema = z.object({
   body: z.object({
     name: z.string().optional().openapi({ example: 'Yolo v4' }),
+
     kind: z.nativeEnum(EntryKind).optional().openapi({ example: EntryKind.Model }),
+    organisation: z.enum(config.ui.modelDetails.organisations as [string, ...string[]]).optional(),
```
JR40159, line 18: Could we not just update the type of the config to avoid having to cast?

JR40159 (changes requested):

frontend/src/entry/EntryStateInput.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const handleChange = (event: SelectChangeEvent) => {
+    onChange(event.target.value)
+  }
+
+  const stateOptions = useMemo(
+    () =>
+      uiConfig
+        ? [
+            <MenuItem value={''} key='unset'>
```
JR40159, line 26: Consistency- here you use the term 'unset' and for organisation you use 'no organisation'

frontend/src/entry/EntryOrganisationInput.tsx
```diff
@@ -0,0 +1,58 @@
...
+  if (!uiConfig || isUiConfigLoading) {
+    return <Loading />
+  }
+
+  if (uiConfig.modelDetails.organisations.length === 0) {
+    return <></>
+  }
+
+  return (
+    <LabelledInput fullWidth label='Organisation (optional)' htmlFor={htmlId}>
```
JR40159, line 52: Don't most web forms just rely on the * notation rather than explicitly saying 'optional'?

JR40159 (changes requested):

JR40159 (changes requested):
I think this probably should include how this new information is displayed in the marketplace?

frontend/src/entry/EntryStateInput.tsx
ARADDCC012: My vote would be the go-to generic of "None" for both

ARADDCC012: My vote would be "None" for both

PR #1905, opened 2025-02-13
Added overflow and ellipsis to titles in release view and marketplace

IR96334:


PE39806:
This works when viewing in a larger window, but the Title and ID still overflow when the window width is reduced.
image
image

A long Release name also overflows on a small-width window, but that's not part of this ticket's scope.

PE39806:
With this change, there's still a problem with the name causing a horizontal overflow when the window width is <1000px.

IR96334:
> With this change, there's still a problem with the name causing a horizontal overflow when the window width is <1000px.

Yeah, I will add some breakpoints for xs, maybe in the stack around the title, id, and desc

PE39806 (approved):
This fixes the original problems found with text overflowing. I have subsequently found more places that text overflows, but they're out of scope for this ticket :)

ARADDCC002 (approved):

PR #1914, opened 2025-02-18
Add model files tab

IR96334:


frontend/src/entry/model/Files.tsx
```diff
@@ -0,0 +1,36 @@
...
+      ) : (
+        <EmptyBlob text={`No files found for model ${model.name}`} />
+      ),
+    [entryFiles, model.id, model.name],
+  )
+
+  if (isEntryFilesError) {
+    return <MessageAlert message={isEntryFilesError.info.message} severity='error' />
+  }
+
```
ARADDCC002, line 29: I know we do what you've done in various places in the app, but I think it's best we move to checking/returning the loading state outside of the render. 

Under all of our `if (is...Error) {` checks, we should do:

```
if (isEntryfilesLoading) {
  return <Loading />
}
```

Ideally we need to change this for all our pages/components but that's future work!

IR96334: Changed

JR40159 (commented):
I'm not a big fan of the 'Associated Releases' button, it seems rather large and clunky. Could we have something a bit more subtle that sits inline beside the file name? Also it would be good if there was a way of telling if a file isn't attached to any releases without having to press a button?

JR40159:
> I'm not a big fan of the 'Associated Releases' button, it seems rather large and clunky. Could we have something a bit more subtle that sits inline beside the file name? Also it would be good if there was a way of telling if a file isn't attached to any releases without having to press a button?

Actually, I can see why changing the information dependant on if a file is attached to any releases is non-trivial. Happy for that to be out of scope for this

frontend/src/entry/model/Files.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <EmptyBlob text={`No files found for model ${model.name}`} />
+      ),
+    [entryFiles.length, model.id, model.name, sortEntryFiles],
+  )
+
+  if (isEntryFilesError) {
+    return <MessageAlert message={isEntryFilesError.info.message} severity='error' />
+  }
+
+  if (isEntryFilesLoading) return <Loading />
```
ARADDCC002, line 52: We should wrap this if statement in braces

frontend/src/entry/model/Files.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  return (
+    <>
+      <Container sx={{ my: 2 }}>
+        <Stack direction={{ xs: 'column' }} spacing={2} justifyContent='center' alignItems='center'>
```
ARADDCC002, line 57: I think you'll need to change this to `direction={{ xs: 'column', sm: 'row'}}`.

frontend/src/entry/model/Releases.tsx
```diff
@@ -50,6 +50,8 @@ export default function Releases({ model, currentUserRoles, readOnly = false }:
     return <MessageAlert message={isReleasesError.info.message} severity='error' />
   }
 
+  if (isReleasesLoading) return <Loading />
```
ARADDCC002, line 53: Same as above comments about braces

ARADDCC002 (changes requested):

ARADDCC002:
You're currently using the `<ReleaseDisplay />` component  inside your list which is fine if a file is only used in a small number of releases, but I wonder if it will look a bit big and cluttered if files are used in lots of releases.

Instead of displaying the release display we normally see in the releases tab, can we make a smaller list item that just displays:

- Semver (and latest label!)
- Date created

frontend/src/entry/model/Files.tsx
IR96334: Do you think it would be okay to just have the files displayed in a list for now? As opposed to a matrix.

IR96334:
> You're currently using the `<ReleaseDisplay />` component inside your list which is fine if a file is only used in a small number of releases, but I wonder if it will look a bit big and cluttered if files are used in lots of releases.
> 
> Instead of displaying the release display we normally see in the releases tab, can we make a smaller list item that just displays:
> 
> * Semver (and latest label!)
> * Date created



> You're currently using the `<ReleaseDisplay />` component inside your list which is fine if a file is only used in a small number of releases, but I wonder if it will look a bit big and cluttered if files are used in lots of releases.
> 
> Instead of displaying the release display we normally see in the releases tab, can we make a smaller list item that just displays:
> 
> * Semver (and latest label!)
> * Date created

screenshot of associated releases list

I have had a look at what it might look like with multiple releases, I feel like this isn't too harsh on the eyes and is a nice way to re-use existing functionality. What do you think?

frontend/src/common/AssociatedReleasesDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+    [latestRelease, model, sortAssociatedReleases],
+  )
+
+  useEffect(() => {
+    if (model && releases.length > 0) {
+      setLatestRelease(releases[0].semver)
+    }
+  }, [model, releases])
+
+  if (isReleasesError) return <MessageAlert message={isReleasesError.info.message} severity='error' />
```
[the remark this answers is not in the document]
IR96334: Done

frontend/src/common/AssociatedReleasesDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  if (isModelError) return <MessageAlert message={isModelError.info.message} severity='error' />
+
+  if (isReleasesLoading) return <Loading />
```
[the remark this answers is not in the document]
IR96334: Done

frontend/src/entry/model/Files.tsx
IR96334: Done

frontend/src/entry/model/Releases.tsx
IR96334: Done

ARADDCC002 (changes requested):

frontend/src/common/AssociatedReleasesDialog.tsx
```diff
@@ -0,0 +1,96 @@
+import { Dialog, DialogContent, DialogTitle, List, ListItem, ListItemText, Stack, Typography } from '@mui/material'
```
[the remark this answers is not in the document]
IR96334: Done

ARADDCC002 (approved):

frontend/src/entry/model/Files.tsx
```diff
@@ -0,0 +1,65 @@
...
+type FilesProps = {
+  model: EntryInterface
+}
+
+export default function Files({ model }: FilesProps) {
+  const { entryFiles, isEntryFilesLoading, isEntryFilesError } = useGetModelFiles(model.id)
+
+  const sortEntryFiles: Array<FileInterface> = useMemo(
+    () =>
+      entryFiles.sort((a, b) => {
```
ARADDCC012, line 19: Can we pull this sorting function out into a utility function? We already have a sorting function in `frontend/utils/arrayUtils.ts`, so you can use that as guidance

frontend/src/entry/model/Files.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { EntryInterface, FileInterface } from 'types/types'
+
+type FilesProps = {
+  model: EntryInterface
+}
+
+export default function Files({ model }: FilesProps) {
+  const { entryFiles, isEntryFilesLoading, isEntryFilesError } = useGetModelFiles(model.id)
+
+  const sortEntryFiles: Array<FileInterface> = useMemo(
```
ARADDCC012, line 17: Minor, but `sortEntryFiles` sounds like you're calling a function to sort the files. It's a variable that stores the sorted files, so consider renaming to `sortedEntryFiles`

frontend/src/entry/model/Files.tsx
```diff
@@ -0,0 +1,65 @@
...
+    [entryFiles],
+  )
+
+  const entryFilesList = useMemo(
+    () =>
+      entryFiles.length ? (
+        sortEntryFiles.map((file) => (
+          <Card key={file._id} sx={{ width: '100%' }}>
+            <Stack spacing={1} p={2}>
+              <FileDownload key={file.name} file={file} modelId={model.id} hideAssociatedReleases={false} />
```
ARADDCC012, line 38: `key` and `hideAssociatedReleases` aren't needed

frontend/src/entry/model/Releases.tsx
```diff
@@ -50,6 +50,10 @@ export default function Releases({ model, currentUserRoles, readOnly = false }:
     return <MessageAlert message={isReleasesError.info.message} severity='error' />
   }
 
+  if (isReleasesLoading) {
```
ARADDCC012, line 53: I don't see why we need to hide the draft button when the releases are loading. Can we revert this back to where it was?

frontend/src/entry/model/releases/AssociatedReleasesDialog.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const { releases, isReleasesLoading, isReleasesError } = useGetReleasesForModelId(modelId)
+  const { model, isModelLoading, isModelError } = useGetModel(modelId, 'model')
+  const [latestRelease, setLatestRelease] = useState('')
+
+  const associatedReleases: Array<ReleaseInterface> = useMemo(
+    () => releases.filter((release) => isFileInterface(file) && release.fileIds.includes(file._id)),
+    [file, releases],
+  )
+
+  const sortAssociatedReleases: Array<ReleaseInterface> = associatedReleases.sort((a, b) => {
```
ARADDCC012, line 29: This could use the utility function I've suggested you create in `frontend/utils/arrayUtils.ts`.

We can also combine it with the above `associatedReleases` logic, so we don't have multiple arrays to deal with.
```
  const associatedReleases: Array<ReleaseInterface> = useMemo(
    () => releases.filter((release) => isFileInterface(file) && release.fileIds.includes(file._id)).sort(sortByCreatedAtDescending),
    [file, releases],
  )
```

frontend/src/entry/model/releases/AssociatedReleasesDialog.tsx
```diff
@@ -0,0 +1,96 @@
...
+      return -1
+    }
+    if (a.createdAt < b.createdAt) {
+      return 1
+    } else {
+      return 0
+    }
+  })
+
+  const isLatestRelease = useCallback(
```
ARADDCC012, line 40: This should be a `useMemo` instead of a `useCallback`

frontend/src/entry/model/releases/AssociatedReleasesDialog.tsx
```diff
@@ -0,0 +1,96 @@
...
+      if (latestRelease === semver) {
+        return <Typography color='secondary'>(Latest)</Typography>
+      }
+    },
+    [latestRelease],
+  )
+
+  useEffect(() => {
+    if (model && releases.length > 0) {
+      setLatestRelease(releases[0].semver)
```
ARADDCC012, line 51: We can't assume that the order of releases will result in the latest being the first one. I'd instead use your sorted array and look at the last one as we know that will be the latest release.

frontend/src/entry/model/releases/AssociatedReleasesDialog.tsx
```diff
@@ -0,0 +1,96 @@
...
+  if (isReleasesError) {
+    return <MessageAlert message={isReleasesError.info.message} severity='error' />
+  }
+
+  if (isModelError) {
+    return <MessageAlert message={isModelError.info.message} severity='error' />
+  }
+
+  if (isModelLoading || isReleasesLoading) {
+    return <Loading />
```
ARADDCC012, line 85: I'd suggest moving this into the `DialogContent` otherwise this will appear in random places in the UI, depending on where `AssociatedReleasesDialog` is called

frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -22,8 +26,9 @@ interface ChipDetails {
   icon: ReactElement
 }
 
-export default function FileDownload({ modelId, file }: FileDownloadProps) {
+export default function FileDownload({ modelId, file, hideAssociatedReleases = true }: FileDownloadProps) {
```
ARADDCC012, line 29: It's standard practice to default boolean props to `false` if they're optional. I'd suggest following convention and adjusting the props where this component is called.

frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -167,25 +172,56 @@ export default function FileDownload({ modelId, file }: FileDownloadProps) {
...
+                  {` ${formatDateString(file.createdAt.toString())}`}
+                </Typography>
+              </Typography>
             </Stack>
-          )}
-          <Typography variant='caption'>{prettyBytes(file.size)}</Typography>
+          </Stack>
         </Stack>
       )}
+      <AssociatedReleasesDialog
```
ARADDCC012, line 219: You don't need the closing tag if you aren't providing any `children` to a component.
```
      <AssociatedReleasesDialog
        modelId={modelId}
        open={associatedReleasesOpen}
        onClose={() => setAssociatedReleasesOpen(false)}
        file={file}
      />
```

frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -167,25 +172,56 @@ export default function FileDownload({ modelId, file }: FileDownloadProps) {
...
+                    {file.name}
+                  </Typography>
+                </Link>
+              </Tooltip>
+            </Stack>
+            <Stack alignItems={{ sm: 'center' }} direction={{ sm: 'column', md: 'row' }} spacing={2}>
+              {!hideAssociatedReleases && (
+                <Button
+                  size='small'
+                  startIcon={<MenuIcon />}
```
ARADDCC012, line 190: This isn't a menu so consider removing the menu icon

frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -167,25 +172,56 @@ export default function FileDownload({ modelId, file }: FileDownloadProps) {
...
+                  variant='contained'
+                  onClick={() => setAssociatedReleasesOpen(true)}
+                  sx={{ minWidth: '200px' }}
```
ARADDCC012, line 193: The button will automatically size itself based on the text inside, so this shouldn't be needed

PR #1915, opened 2025-02-19
Migrate scans to have their own mongo collection

PE39806:
Currently scans are stored here: https://github.com/gchq/Bailo/blob/main/backend/src/models/File.ts#L22

This ticket is to create a new Mongo collection to store scans relate to files or Docker Images. This will include adding a scan type to the scan document and writing a migration script to migrate all existing file scans to the new location.

The file scanning code in the file service will need to be moved into its own service and refactored to work with this new collection.


backend/src/models/Scan.ts
```diff
@@ -0,0 +1,110 @@
...
+// This interface stores information about the properties on the base object.
+// It should be used for plain object representations, e.g. for sending to the
+// client.
+export interface ScanInterface {
+  _id: ObjectId
+
+  // file or image
+  artefactType: ArtefactTypeKeys
+  // for files
+  fileId?: string
```
ARADDCC002, line 15: I'm not a big fan of these conditional properties being optional depending on the value of `artefactType`. I think it's better if we do a sort of discriminating union, this means that `fileId` will only be needed when `artefactType = 'file'` for example.

```
type ArtefactDetails = | {
    artefactKind: "file",
    fileId: string,
} | {
    artefactKind: "image",
    imageName: string
}

type Artefact = {
    otherProperties: string
}

type ArtefactWithDetails = Artefact & ArtefactDetails

// This compiles
const artefact: ArtefactWithDetails = {
    artefactKind: "image",
    imageName: "my-file.txt",
    otherProperties: "this is another property!"
}

// This does NOT compile
const invalidArtefact: ArtefactWithDetails = {
    kind: "image",
    fileId: "my-file.txt",
    otherProperties: "this is another property!"
}
```

backend/src/models/Scan.ts
```diff
@@ -0,0 +1,110 @@
...
+  state: ScanStateKeys
+  isInfected?: boolean
+  viruses?: string[]
+  lastRunAt: Date
+
+  createdAt: Date
+  updatedAt: Date
+}
+
+export const ArtefactType = {
```
ARADDCC002, line 32: We often replace `type` with `kind`, just to avoid confusing it with TS types. So I would replace this with `ArtefactKind`

backend/src/services/file.ts
```diff
@@ -148,8 +147,15 @@ export async function getFilesByIds(user: UserInterface, modelId: string, fileId
     throw NotFound(`The requested files were not found.`, { fileIds: notFoundFileIds })
   }
 
+  const fileAvScans = await ScanModel.find({ fileId: { $in: fileIds } })
```
JR40159, line 150: Mongo can do this for you as part of an aggregate query. Eg: https://github.com/gchq/Bailo/blob/887208c0c1340ab410fd34140d714433dfae15ed/backend/src/services/review.ts#L38

backend/src/models/Scan.ts
```diff
@@ -0,0 +1,110 @@
+import { model, ObjectId, Schema } from 'mongoose'
+import MongooseDelete, { SoftDeleteDocument } from 'mongoose-delete'
+
+import { ScanState, ScanStateKeys } from '../connectors/fileScanning/Base.js'
+
+// This interface stores information about the properties on the base object.
+// It should be used for plain object representations, e.g. for sending to the
+// client.
+export interface ScanInterface {
```
JR40159, line 9: Duplication https://github.com/gchq/Bailo/blob/887208c0c1340ab410fd34140d714433dfae15ed/backend/src/connectors/fileScanning/Base.ts#L2

backend/src/models/File.ts
```diff
@@ -41,16 +41,23 @@ const FileSchema = new Schema<FileInterfaceDoc>(
...
+        {
+          toolName: { type: String },
+          scannerVersion: { type: String },
+          state: { type: String, enum: Object.values(ScanState) },
+          isInfected: { type: Boolean },
+          viruses: [{ type: String }],
+          lastRunAt: { type: Schema.Types.Date },
+        },
+      ],
+      required: false,
```
JR40159, line 55: I'd prefer to just remove this property from the file model

JR40159 (changes requested):

backend/src/models/Scan.ts
```diff
@@ -0,0 +1,110 @@
...
+  artefactType: ArtefactTypeKeys
+  // for files
+  fileId?: string
+  // for images
+  // map this to registry GET /v2/<name>/manifests/<reference>
+  repositoryName?: string
+  imageDigest?: string
+
+  toolName: string
+  scannerVersion?: string
```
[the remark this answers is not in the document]
PE39806: No, `scannerVersion` was made an optional field https://github.com/gchq/Bailo/blob/fac7e811fbc8949cd7769206c788fbf95fe44823/backend/src/connectors/fileScanning/Base.ts#L4

backend/src/models/Scan.ts
```diff
@@ -0,0 +1,110 @@
...
+// client.
+export interface ScanInterface {
+  _id: ObjectId
+
+  // file or image
+  artefactType: ArtefactTypeKeys
+  // for files
+  fileId?: string
+  // for images
+  // map this to registry GET /v2/<name>/manifests/<reference>
```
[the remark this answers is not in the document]
PE39806: I've added a comment explaining that ImageRef currently does not suit the requirements of AV Scans as it uses Tags which can be overwritten rather than Digests which are immutable. A future piece of work should change this.

backend/src/models/Scan.ts
```diff
@@ -0,0 +1,70 @@
+import { model, ObjectId, Schema } from 'mongoose'
+import MongooseDelete, { SoftDeleteDocument } from 'mongoose-delete'
+
+import { FileScanResult, ScanState } from '../connectors/fileScanning/Base.js'
+
+// This interface stores information about the properties on the base object.
```
ARADDCC002, line 6: I don't think we need this comment

backend/src/models/Scan.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+  | {
+      artefactKind: 'image'
+      repositoryName: string
+      // use Digest as image Tags can be overwritten but digests are immutable
+      imageDigest: string
+      // TODO: ultimately use backend/src/models/Release.ts:ImageRef, but ImageRef needs converting to use Digest rather than Tag first
+    }
+
+// The doc type includes all values in the plain interface, as well as all the
```
ARADDCC002, line 35: I think we can remove this comment too

frontend/types/types.ts
```diff
@@ -92,6 +92,13 @@ export interface FileInterface {
 }
 
 export interface AvScanResult {
+  artefactKind: ArtefactKindKeys
```
ARADDCC002, line 95: I think we should also add the discriminating union here so that we don't accidentally type files with properties that only used by images

ARADDCC002 (changes requested):

backend/src/models/Scan.ts
JR40159: What was the outcome of this?

JR40159 (changes requested):
Requested update on previous comment

backend/src/models/Scan.ts
PE39806: ScanInterface now extends FileScanResult
https://github.com/gchq/Bailo/pull/1915/files#diff-8441c4cb326c81f6a2a291a15b57ebca906af0cd70cc89cc8575915a46f77d75R6-R11

JR40159 (changes requested):
This file https://github.com/gchq/Bailo/blob/60a39eb6a516b74550fcac123876cdb36f1f251b/backend/src/routes/v2/model/file/getFiles.ts still requires updating and files are now missing av scans

JR40159:
> This file https://github.com/gchq/Bailo/blob/60a39eb6a516b74550fcac123876cdb36f1f251b/backend/src/routes/v2/model/file/getFiles.ts still requires updating and files are now missing av scans

Types still aren't correct in both  https://github.com/gchq/Bailo/blob/60a39eb6a516b74550fcac123876cdb36f1f251b/backend/src/routes/v2/model/file/getFiles.ts and https://github.com/gchq/Bailo/blob/60a39eb6a516b74550fcac123876cdb36f1f251b/backend/src/routes/v2/model/file/postSimpleUpload.ts

The return interfaces and the interfaces used in the Open API spec are still incorrect.

PR #1919, opened 2025-02-24
Soft deletion script for models

ARADDCC002:


JR40159 (changes requested):
The only bits I can spot that are missing are deleting webhooks and inferences

JR40159 (approved):

PR #1922, opened 2025-02-25
added a reviewing filter to the marketplace

ARADDCC002:


PE39806 (changes requested):
I don't like that selecting the filters "Reviewing as MSRO" and "Reviewing as MTR" returns models via a logical `or`, but selecting "Mine" and either/both MSRO/MTR only returns models with the respective MSRO/MTR filter(s) but not a logical `or` with Mine (essentially Mine is ignored). This should be made to be consistent so that selecting Mine and either/both MSRO/MTR returns the logical `or` of all of the filters.

PE39806 (approved):

PR #1976, opened 2025-03-05
Bugxfix/error handling

JR40159:


PE39806 (changes requested):
Replace `ResponseException` on https://github.com/gchq/Bailo/blob/bugxfix/error-handling/lib/python/tests/test_release.py#L71 with `BailoException` to fix the broken Python integration test. This is because we now return an explicit Bailo error rather than a generic error. You will also need to remove `ResponseException` from that file's imports.

PE39806 (approved):

PR #1978, opened 2025-03-07
Added virtualised list to marketplace for quicker model rendering

ARADDCC002:


frontend/src/marketplace/EntryList.tsx
```diff
@@ -22,49 +23,69 @@ export default function EntryList({
...
+
+  Row.displayName = 'row'
 
   if (entriesErrorMessage) return <MessageAlert message={entriesErrorMessage} severity='error' />
 
   return (
     <>
       {entries.length === 0 && <EmptyBlob data-test='emptyEntryListBlob' text='No entries here' />}
-      {entriesDisplay}
+      <FixedSizeList
```
ARADDCC012, line 80: From past experience setting `overscanCount` to something like 5 or 10 helps to reduce stutter when scrolling very quickly

frontend/src/marketplace/EntryList.tsx
```diff
@@ -22,49 +23,69 @@ export default function EntryList({
...
-              variant='h5'
-              sx={{ fontWeight: '500', textDecoration: 'none', color: theme.palette.primary.main }}
+  const Row = memo(({ data, index, style }: { data: EntrySearchResult[]; index: number; style: CSSProperties }) => {
+    const entry = data[index]
+
+    return (
+      <Box
+        justifyContent='flex-start'
+        alignItems='center'
+        sx={{ backgroundColor: index % 2 ? theme.palette.container.main : 'none', p: 2, margin: 'auto', ...style }}
```
ARADDCC012, line 33: Now that we've got a scroll bar to the right and have also changed how the container for each list item looks, I'd be quite keen to reduce/remove the horizontal padding we have on this list and the tabs above it.

frontend/src/marketplace/EntryList.tsx
```diff
@@ -22,49 +23,69 @@ export default function EntryList({
...
+              </Typography>
+            </Link>
+          </Stack>
+          <Typography
+            variant='body1'
+            sx={{ marginBottom: 2, whiteSpace: 'nowrap', textOverflow: 'ellipsis', overflow: 'hidden' }}
+          >
+            {entry.description}
+          </Typography>
+          <div>
```
ARADDCC012, line 56: Is this `div` needed?

frontend/src/marketplace/EntryList.tsx
```diff
@@ -22,49 +23,69 @@ export default function EntryList({
...
-      <Fragment key={entry.id}>
-        <Stack direction='row'>
-          <Link
-            sx={{ textDecoration: 'none', whiteSpace: 'nowrap', textOverflow: 'ellipsis', overflow: 'hidden' }}
-            href={`${entry.kind}/${entry.id}`}
-          >
-            <Typography
-              variant='h5'
-              sx={{ fontWeight: '500', textDecoration: 'none', color: theme.palette.primary.main }}
+  const Row = memo(({ data, index, style }: { data: EntrySearchResult[]; index: number; style: CSSProperties }) => {
```
ARADDCC012, line 26: If you want to use `memo()`, which is a HOC (Higher Order Component), you should split this out into its own component. 

That being said, I don't really see the performance benefit of memoizing something that's going to be unrendered and rerendered as the user scrolls.

IMO this should be pulled out into a new `EntryListRow` component and we can do away with the `memo`. I do think it's worth checking that there isn't a degradation of performance if we do this, but I wouldn't expect there to be any noticeable change either way. Therefore I'd vote for simplifying by removing the `memo`.

ARADDCC012 (changes requested):

frontend/src/marketplace/EntryList.tsx
ARADDCC002: It stops the chips in the `<ChipSelector />` component from also stacking incorrectly, however I feel we should probably add a stack inside `<ChipSelector />` to stop this from happening.

ARADDCC002: Nevermind just realised that a `<Stack />` inside the chip component won't work either!

ARADDCC012 (approved):

ARADDCC012 (approved):

PR #1996, opened 2025-03-12
Fix bad 015_migrate_avscan_to_own_model.ts logic

PE39806:


backend/src/migrations/015_migrate_avscan_to_own_model.ts
```diff
@@ -2,16 +2,15 @@ import FileModel from '../models/File.js'
...
-    { avScan: { $exists: true }, 'avScan.toolName': { $exists: false } },
-    { $set: { 'avScan.toolName': 'Unknown Scanner' } },
-  )
   // convert avScan from being stored in File to a new Scan Document
   const files = await FileModel.find({})
   for (const file of files) {
     if (file.get('avScan') !== undefined) {
       for (const avResult of file.get('avScan')) {
+        // toolName was originally not a required field so may not exist
+        if (!Object.prototype.hasOwnProperty.call(avResult, 'toolName')) {
```
ARADDCC012, line 11: You should be able to write this as
```
if (!avResult.hasOwnProperty('toolName'))
```

JR40159 (approved):

PR #2034, opened 2025-03-17
Update Scan errors

PE39806:
E.g. https://github.com/gchq/Bailo/blob/main/backend/src/connectors/fileScanning/clamAv.ts#L76

Create a function in the base scanner to unify error log and creation of the error object for all scanners. The function should take in all information needed to create a scan result and log that information.

Update uses of `file._id` to `file.id` so that the object ID is output in string form.

### DIFF
--- a/backend/src/connectors/fileScanning/Base.ts
+++ b/backend/src/connectors/fileScanning/Base.ts
@@ -14,25 +14,31 @@ export const ScanState = {
 } as const
 export type ScanStateKeys = (typeof ScanState)[keyof typeof ScanState]
 
-export abstract class BaseFileScanningConnector {
-  abstract readonly toolName: string
+export type FileScanningConnectorInfo = Pick<FileScanResult, 'toolName' | 'scannerVersion'>
 
-  abstract info(): string[]
+export abstract class BaseFileScanningConnector {
+  abstract info(): Promise<FileScanningConnectorInfo>
   abstract scan(file: FileInterface): Promise<FileScanResult[]>
 
-  getScannerVersion(): Promise<string | void> {
-    // default to promise undefined, but allow child classes to override this
-    return Promise.resolve(undefined)
-  }
-
-  async scanError(error: unknown, file: FileInterface): Promise<FileScanResult[]> {
-    const scannerVersion = await this.getScannerVersion()
-    log.error({ error, modelId: file.modelId, fileId: file._id.toString(), name: file.name }, 'Scan errored.')
+  async scanError(
+    error: unknown | undefined = undefined,
+    file: FileInterface | undefined = undefined,
+    errorMessage: string = 'Scan errored.',
+    extraErrorObj: object = {},
+  ): Promise<FileScanResult[]> {
+    const scannerInfo = await this.info()
+    log.error(
+      {
+        // conditional spreading operator so that objects are only used if they exist
+        ...(error !== undefined ? { error } : {}),
+        ...(file !== undefined ? { modelId: file.modelId, fileId: file._id.toString(), name: file.name } : {}),
+        ...extraErrorObj,
+      },
+      errorMessage,
+    )
     return [
       {
-        toolName: this.toolName,
-        // include scannerVersion if it is defined
-        ...(scannerVersion !== undefined ? { scannerVersion } : {}),
+        ...scannerInfo,
         state: ScanState.Error,
         lastRunAt: new Date(),
       },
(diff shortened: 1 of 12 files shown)


backend/src/migrations/015_migrate_avscan_to_own_model.ts
```diff
@@ -14,7 +14,7 @@ export async function up() {
         // create new Scan Document
         const newScan = new ScanModel({
           artefactKind: ArtefactKind.File,
-          fileId: file._id,
+          fileId: file._id.toString(),
```
JR40159, line 17: You're updating an existing migration that might of already been run. This will lead to an inconsistency in data?

backend/src/connectors/fileScanning/Base.ts
```diff
@@ -14,6 +15,27 @@ export const ScanState = {
 export type ScanStateKeys = (typeof ScanState)[keyof typeof ScanState]
 
 export abstract class BaseFileScanningConnector {
+  abstract readonly toolName: string
+
   abstract info(): string[]
   abstract scan(file: FileInterface): Promise<FileScanResult[]>
+
+  getScannerVersion(): Promise<string | void> {
```
JR40159, line 23: I think we should enforce that every scanner should be able to return a scanner version. As well as being a reasonable requirement, this would also simplify the implementation in multiple places

JR40159: To simplify things even further, you could update the scanner info to be an object that contains the tool name and version.

backend/src/connectors/fileScanning/Base.ts
```diff
@@ -14,6 +15,27 @@ export const ScanState = {
...
+    // default to promise undefined, but allow child classes to override this
+    return Promise.resolve(undefined)
+  }
+
+  async scanError(error: unknown, file: FileInterface): Promise<FileScanResult[]> {
+    const scannerVersion = await this.getScannerVersion()
+    log.error({ error, modelId: file.modelId, fileId: file._id.toString(), name: file.name }, 'Scan errored.')
```
JR40159, line 30: This log should include both the tool name

backend/src/connectors/fileScanning/Base.ts
```diff
@@ -14,6 +15,27 @@ export const ScanState = {
...
+    return [
+      {
+        toolName: this.toolName,
+        // include scannerVersion if it is defined
+        ...(scannerVersion !== undefined ? { scannerVersion } : {}),
```
JR40159, line 35: Seems inconsistent, where you've used 'Unknown' before.

backend/src/connectors/fileScanning/modelScan.ts
```diff
@@ -43,16 +47,12 @@ export class ModelScanFileScanningConnector extends BaseFileScanningConnector {
...
     this.init()
-
-    let modelscanVersion: string | undefined = undefined
-    try {
-      modelscanVersion = (await getModelScanInfo()).modelscanVersion
-    } catch (error) {
-      log.error('Could not run ModelScan as it is not running', error)
+    const scannerVersion = await this.getScannerVersion()
+    if (scannerVersion === undefined) {
+      log.error('Could not use ModelScan as it is not running')
```
JR40159, line 52: Should you not be calling the new function here?

backend/src/connectors/fileScanning/modelScan.ts
```diff
@@ -43,16 +47,12 @@ export class ModelScanFileScanningConnector extends BaseFileScanningConnector {
 
   async scan(file: FileInterfaceDoc): Promise<FileScanResult[]> {
     this.init()
```
JR40159, line 49: Why do we initialise model scan on every scan? Isn't this something we can do once in the constructor?

backend/src/connectors/fileScanning/clamAv.ts
```diff
@@ -42,46 +43,44 @@ export class ClamAvFileScanningConnector extends BaseFileScanningConnector {
     }
   }
 
+  async getScannerVersion() {
+    if (!av) return
```
JR40159, line 47: Should this ever happen? Isn't this an error case?

JR40159 (changes requested):

backend/src/connectors/fileScanning/modelScan.ts
PE39806: I think this is a check for if the API is still accessible, and if not then it could just be a temporary problem (e.g. the k8s pod is being replaced with a newer version)

JR40159 (approved):

PR #2035, opened 2025-03-17
Dev/fix virtualised marketplace list issues

ARADDCC002:


frontend/src/marketplace/EntryList.tsx
```diff
@@ -32,23 +43,36 @@ export default function EntryList({
       onSelectedChipsChange={onSelectedChipsChange}
       data={data}
       index={index}
-      style={style}
+      style={{ padding: '20px', ...style }}
     />
   )
 
+  const columnWidths = entries.map((entry) => (entry.tags.length === 0 ? 100 : 140))
```
ARADDCC012, line 50: Can we pull this outside of Row and into the main component so we can `useMemo` it?

frontend/src/marketplace/EntryList.tsx
```diff
@@ -32,23 +43,36 @@ export default function EntryList({
       onSelectedChipsChange={onSelectedChipsChange}
       data={data}
       index={index}
-      style={style}
+      style={{ padding: '20px', ...style }}
```
ARADDCC012, line 46: Can we use `theme.spacing` for this padding?

frontend/src/marketplace/EntryList.tsx
```diff
@@ -32,23 +43,36 @@ export default function EntryList({
...
+  const columnWidths = entries.map((entry) => (entry.tags.length === 0 ? 100 : 140))
+
+  const getItemSize = (index: number) => columnWidths[index]
+
+  if (entries.length === 0) {
+    return (
+      <EmptyBlob
+        data-test='emptyEntryListBlob'
+        text='No items here'
+        style={{ height: windowHeight - 230, paddingTop: 40 }}
```
ARADDCC012, line 59: Can we use `theme.spacing` for the `paddingTop`?

frontend/src/marketplace/EntryList.tsx
```diff
@@ -1,6 +1,6 @@
 import { EntrySearchResult } from 'actions/model'
-import { CSSProperties } from 'react'
-import { FixedSizeList } from 'react-window'
+import { CSSProperties, useLayoutEffect, useState } from 'react'
+import { VariableSizeList as List } from 'react-window'
```
ARADDCC012, line 3: I'm not a fan of renaming the component like this. At a glance it's not obvious that it's not a MUI `List`

ARADDCC012 (changes requested):

frontend/src/marketplace/EntryList.tsx
ARADDCC002: This is how it's done in all the docs, but I agree it's confusing, so reverted.

frontend/src/marketplace/EntryList.tsx
ARADDCC002: I've added it as a useMemo, but it's already outside of `Row`

ARADDCC012 (approved):

PR #2053, opened 2025-03-18
Fixed bug with userdisplay and latestrelease

IR96334:


frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -203,12 +225,15 @@ export default function FileDownload({ modelId, file, showAssociatedReleases = f
           </Stack>
           <Stack direction={{ sm: 'column', md: 'row' }} spacing={2} alignItems='center' justifyContent='space-between'>
             <Stack direction='row'>
-              <Typography textOverflow='ellipsis' overflow='hidden' variant='caption' sx={{ mb: 2 }}>
-                Added by {<UserDisplay dn={file.createdAt.toString()} />} on
-                <Typography textOverflow='ellipsis' overflow='hidden' variant='caption' fontWeight='bold'>
-                  {` ${formatDateString(file.createdAt.toString())}`}
+              {sortedAssociatedReleases.length > 0 && (
+                <Typography textOverflow='ellipsis' overflow='hidden' variant='caption' sx={{ mb: 2 }}>
+                  Added by <UserDisplay dn={sortedAssociatedReleases[sortedAssociatedReleases.length - 1].createdBy} />{' '}
```
JR40159, line 230: "Added by" about a file is different to who created one of the associated releases. Also how would this work if a file wasn't associated to any releases?

JR40159 (changes requested):

frontend/src/entry/model/releases/FileDownload.tsx
IR96334: To my knowledge the user who uploaded the file is only stored in the release. This means if a file in the future was able to be uploaded separately to a release, I don't think you could display the release of the uploader until the file store holds the name of the uploader.

JR40159: So we don't store who uploaded a file, but we do store who created a release but those are different things. As a result I don't think we can have "Added by" on a file because we don't have that information.

frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -201,23 +222,19 @@ export default function FileDownload({ modelId, file, showAssociatedReleases = f
...
-            <Stack direction='row'>
-              <Typography textOverflow='ellipsis' overflow='hidden' variant='caption' sx={{ mb: 2 }}>
-                Added by {<UserDisplay dn={file.createdAt.toString()} />} on
-                <Typography textOverflow='ellipsis' overflow='hidden' variant='caption' fontWeight='bold'>
-                  {` ${formatDateString(file.createdAt.toString())}`}
-                </Typography>
-              </Typography>
-            </Stack>
-          </Stack>
+          <Typography variant='caption' sx={{ mb: 2 }}>
```
ARADDCC002, line 225: We don't need this padding at the bottom

ARADDCC002:
Moved to #2060

PR #2054, opened 2025-03-18
Feature/bai 1634 select existing files releases

ARADDCC002:


PE39806 (changes requested):
The new button doesn't seem to be enabled until a file has already been uploaded in a release, and the formatting when uploading new files leaves *much* to be desired...
image

frontend/src/entry/model/releases/ExistingFileSelector.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  return (
+    <>
+      <Button
+        disabled={!files || files.length === 0}
+        variant='outlined'
+        style={{ width: '100%' }}
+        onClick={() => setIsDialogOpen(true)}
+      >
+        Select existing file
```
PE39806, line 119: Update this to `Select existing files` (plural)

frontend/types/types.ts
```diff
@@ -91,7 +91,10 @@ export interface FileInterface {
   updatedAt: Date
 }
 
+export type FileWithScanResultsInterface = FileInterface & { avScan: ScanResultInterface[]; id: string }
+
 export interface ScanResultInterface {
+  _id: string
```
PE39806, line 97: I'm not sure this is a string - `_id` in the backend will be a Mongoose/MongoDB `ObjectID` so this might be an Object rather than a string.

PE39806 (changes requested):
There's some inconsistent behaviour when adding multiple files with the same name to a single release. Previously, you would get an error like `Bad Request: Releases cannot have multiple files with the same name.  The duplicates in this file are: '["safe_model.h5","unsafe_model.h5"]'` but with the following steps the error doesn't occur:

1. Create a new release with a new file attached
2. Create a second release with the exact same file attached (via "attach new files" - not "select existing file")
3. Create a third release and add both files via select existing file
4. On selecting "create release" this will fail with the above error message as expected
5. Select "attach new files" and re-upload the same file as before
6. On selecting "create release" this will not fail and a new release will be created with either: none of the duplicate-named file(s), or just the uploaded file(s)

I think the order that the files are sent from the frontend to backend determines which of those two options happens. When the existing files are sent first, then no files are selected, but if the new files are sent first then they are uploaded and attached to the release.

frontend/src/entry/model/releases/ExistingFileSelector.tsx
```diff
@@ -0,0 +1,132 @@
...
+    const newCheckedFiles = checkedFiles.filter((checkedFile) => file._id !== checkedFile._id)
+    if (currentIndex === -1) {
+      newCheckedFiles.push(file)
+      setCheckedFiles(newCheckedFiles)
+    } else {
+      setCheckedFiles(newCheckedFiles)
+    }
+  }
+
+  const fileList = () => {
```
ARADDCC012, line 61: This should be a `useMemo` instead of a function

frontend/src/entry/model/releases/ExistingFileSelector.tsx
```diff
@@ -0,0 +1,132 @@
...
+  if (isFilesLoading) {
+    return <Loading />
+  }
+
+  return (
+    <>
+      <Button
+        disabled={!files || files.length === 0}
+        variant='outlined'
+        style={{ width: '100%' }}
```
ARADDCC012, line 116: Consider using `sx` instead of `style`

frontend/src/entry/model/releases/ExistingFileSelector.tsx
```diff
@@ -0,0 +1,132 @@
...
+        onClick={() => setIsDialogOpen(true)}
+      >
+        Select existing files
+      </Button>
+      <Dialog open={isDialogOpen} onClose={() => setIsDialogOpen(false)}>
```
ARADDCC012, line 121: Consider pulling this dialog out into a separate component

frontend/src/entry/model/releases/ExistingFileSelector.tsx
```diff
@@ -0,0 +1,132 @@
...
+        <DialogTitle>Select an existing file for {model.name}</DialogTitle>
+        <DialogContent>
+          <List>{fileList()}</List>
```
ARADDCC012, line 124: Consider putting `<List></List>` inside the `fileList` `useMemo`, so we don't render an empty list if there are no files.

frontend/src/entry/model/releases/ExistingFileSelector.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  if (isFilesLoading) {
+    return <Loading />
+  }
+
+  return (
+    <>
+      <Button
+        disabled={!files || files.length === 0}
```
ARADDCC012, line 114: Can we add a tooltip to explain why this button is disabled? E.g. "No existing files". Just be aware that you'll need to wrap this button with a `div` to allow putting a tooltip on a disabled element.

ARADDCC012 (changes requested):

frontend/src/entry/model/releases/ExistingFileSelector.tsx
ARADDCC002: I'm not sure how much value this adds as if we separate them, either the parent component is just a button that opens a dialog with no real logic, or if we instead pass in `fileList` as a prop for the content of the Dialog, we may as well make a reusable Dialog component instead.

frontend/src/entry/model/releases/ExistingFileSelector.tsx
ARADDCC002: Decided that having the dialog show the empty blog was more consistent with other parts of the app so I've removed the `disabled` prop entirely.

PE39806 (changes requested):
Creating a Release with at least one file, then creating a new Release and selecting "Select existing files", then ticking at least one file, then selecting "Add files", then re-selecting "Select existing files", then (with no changes to the chosen files) selecting "Add files" will remove all files, including any uploaded via "Attach new files". It is then not possible to re-upload files via "Attach new files", but it is possible to add files via "Select existing files" again and to upload new files (which were not previously uploaded) via "Attach new files".

I would expect selecting "Select existing files" followed by no changes to the list then "Add files" to not remove all files currently in the Release.

PE39806 (approved):

PR #2061, opened 2025-03-19
changed where we cal the rescanning update text so it is not called w…

ARADDCC002:
…hen request fails


frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -75,7 +75,6 @@ export default function FileDownload({ modelId, file, showAssociatedReleases = f
 
   const handleRerunFileScanOnClick = useCallback(async () => {
     const res = await rerunFileScan(modelId, (file as FileInterface)._id)
```
JR40159, line 77: I think the root of the issue is that this request is currently just fire and forget. To make it more useful to the frontend and other clients, we should update it to return the update scan results. The backend already updates the scan results to be in progress but doesn't then return them

JR40159 (changes requested):

JR40159 (approved):

JR40159 (approved):

PR #2063, opened 2025-03-20
Bai 1633 add the ability to delete file

IR96334:


ARADDCC002 (changes requested):

frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -216,38 +269,69 @@ export default function FileDownload({ modelId, file, showAssociatedReleases = f
       {isFileInterface(file) && (
         <Stack>
           <Stack direction={{ sm: 'column', md: 'row' }} spacing={2} alignItems='center' justifyContent='space-between'>
-            <Stack alignItems={{ sm: 'center' }}>
+            <Stack direction={{ sm: 'column', md: 'row' }} spacing={2} alignItems={{ sm: 'center', md: 'flex-end' }}>
```
ARADDCC002, line 272: change `alignItems` to be just `'center'`

PE39806 (changes requested):
The ... by each File in the Release page has no options, and the ... by each File in the Release tag has only "Rerun File Scan" option. Either these both need removing and should be reverted to having just the rescan button, or they should always have the associatedReleases, deleteFile and rescanFile options as in the Files tab. @ARADDCC002 do you have any preference between those options?

image
image

frontend/src/entry/model/releases/FileDownload.tsx
IR96334: I have the flex-end, as when in the wider view the size appears slightly misaligned from the title. It is a small thing, but I feel it makes it look more symmetrical

IR96334:
> The ... by each File in the Release page has no options, and the ... by each File in the Release tag has only "Rerun File Scan" option. Either these both need removing and should be reverted to having just the rescan button, or they should always have the associatedReleases, deleteFile and rescanFile options as in the Files tab. @ARADDCC002 do you have any preference between those options?
> 
> image image

I had talked about this with @ARADDCC002 and we have made the options in the menu purposefully consistent with the just having the menu, no matter how many options were enabled. Although the editablerelease is a mistake, and will be fixed.

frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -216,38 +269,69 @@ export default function FileDownload({ modelId, file, showAssociatedReleases = f
...
             </Stack>
           </Stack>
-          <Typography variant='caption'>
-            Uploaded on
-            <span style={{ fontWeight: 'bold' }}>{` ${formatDateString(file.createdAt.toString())}`}</span>
-          </Typography>
+          <Stack direction={{ sm: 'column', md: 'row' }} spacing={2} alignItems='center' justifyContent='space-between'>
+            <Typography variant='caption'>
+              Uploaded on
+              <span style={{ fontWeight: 'bold' }}>{` ${formatDateString(file.createdAt.toString())}`}</span>
```
PE39806, line 332: I would prefer `formatDateTimeString` here as it will help to make files more distinctive. If I were to make two releases on the same day with the same file name then this change would make it possible to tell them apart.

PE39806 (approved):

PR #2075, opened 2025-03-31
fixed an issue where items in marketplace issue were not resizing pro…

ARADDCC002:
…perly when filters were updated

### DIFF
--- a/frontend/pages/index.tsx
+++ b/frontend/pages/index.tsx
@@ -264,7 +264,7 @@ export default function Marketplace() {
                   <EntryList
                     entries={models}
                     entriesErrorMessage={isModelsError ? isModelsError.info.message : ''}
-                    selectedChips={[...selectedLibraries, ...selectedRoles, ...selectedTask]}
+                    selectedChips={selectedLibraries}
                     onSelectedChipsChange={handleLibrariesOnChange}
                   />
                 </div>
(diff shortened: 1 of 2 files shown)


frontend/pages/index.tsx
```diff
@@ -264,7 +264,7 @@ export default function Marketplace() {
                   <EntryList
                     entries={models}
                     entriesErrorMessage={isModelsError ? isModelsError.info.message : ''}
-                    selectedChips={selectedLibraries}
+                    selectedChips={[...selectedLibraries, ...selectedRoles, ...selectedTask]}
```
ARADDCC012, line 267: 1. This list should be stable (useMemo), so that it's not recreated on every render cycle.
2. The `EntryList`s for both models and data cards should handle `selectedChips` and `onSelectedChipsChange` the same way.
3. With this change `selectedChips` and `onSelectedChipsChange` are out of sync. If we do want to include roles and tasks as `selectedChips` then we also need to update `onSelectedChipsChange` to do the following:

    * Not call `handleLibrariesOnChange`
    * Call a new function that splits out the updated selected chips into libraries, task and roles and calls `handleLibrariesOnChange`, `handleTaskOnChange` and `handleSelectedRolesOnChange` respectively.

ARADDCC012 (changes requested):

PE39806 (dismissed):

ARADDCC012 (approved):

PR #2116, opened 2025-04-01
fixed an issue where role filters were appling to all entries

ARADDCC002:


backend/src/services/model.ts
```diff
@@ -155,7 +155,10 @@ export async function searchModels(
     } else {
       query.collaborators = {
         $elemMatch: {
-          ...(filters.length > 0 && { roles: { $elemMatch: { $in: filters } } }),
+          ...(filters.length > 0 && {
```
JR40159, line 158: This is already in an if statement that is `if (filters.length > 0) {`. This duplicates that condition?

backend/src/services/model.ts
```diff
@@ -155,7 +155,10 @@ export async function searchModels(
...
+            roles: { $elemMatch: { $in: filters } },
+            entity: toEntity(EntityKind.USER, user.dn),
```
JR40159, line 160: This won't work for groups, I think it should be the same as the line above `entity: { $in: await authentication.getEntities(user)`

JR40159 (changes requested):

JR40159 (approved):

PR #2122, opened 2025-04-03
Bai 1641 sort files tab by different fields

IR96334:


frontend/src/entry/model/Files.tsx
```diff
@@ -14,13 +16,58 @@ type FilesProps = {
 
 export default function Files({ model }: FilesProps) {
   const { entryFiles, isEntryFilesLoading, isEntryFilesError } = useGetModelFiles(model.id)
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null)
+  const menuOpen = Boolean(anchorEl)
+  const [orderByValue, setOrderByValue] = useState('createdAt')
+  const [orderByButtonTitle, setOrderByButtonTitle] = useState('Order by')
+  const [ASCorDESC, setASCorDESC] = useState('DESC')
```
ARADDCC002, line 23: I think we should type this:

```
export const SortingDirection= {
  ASC: 'ascending',
  DESC: 'descending',
} as const
export type SortingDirectionKeys = (typeof SortingDirection)[keyof typeof SortingDirection]
```

That way, we can use it like so:

 `const [ascOrDesc, setAscOrDesc] = useState<SortingDirectionKeys>(SortingDirection.DESC)`

In the above I also changed `ASCorDESC` to `ascOrDesc` as the casing is more consistent

frontend/src/entry/model/Files.tsx
```diff
@@ -14,13 +16,58 @@ type FilesProps = {
...
+      if (menuOption === orderByValue) {
+        return true
+      } else {
+        return false
+      }
+    },
+    [orderByValue],
+  )
+
+  function handleMenuButtonClick(event: React.MouseEvent<HTMLButtonElement>) {
```
ARADDCC002, line 57: Change to 

`(event: MouseEvent<HTMLButtonElement>)`

And import `MouseEvent` from `react`

frontend/src/entry/model/Files.tsx
```diff
@@ -48,8 +95,149 @@ export default function Files({ model }: FilesProps) {
...
     <>
       <Container sx={{ my: 2 }}>
-        <Stack direction={{ xs: 'column' }} spacing={2} justifyContent='center' alignItems='center'>
-          {entryFilesList}
+        <Stack spacing={2}>
+          <Box display='flex'>
+            <Box ml='auto'>
+              <Stack direction={'row'}>
+                <Button
+                  variant='text'
```
ARADDCC002, line 103: `text` is the default value for the `variant` prop so it can be removed

frontend/src/entry/model/Files.tsx
```diff
@@ -48,8 +95,149 @@ export default function Files({ model }: FilesProps) {
...
+                  onClick={handleMenuButtonClick}
+                  endIcon={anchorEl ? <ExpandLess /> : <ExpandMore />}
+                >
+                  <Stack sx={{ minWidth: '150px' }} direction={'row'} spacing={2} justifyContent={'space-evenly'}>
+                    {ASCorDESC === 'ASC' ? (
```
ARADDCC002, line 108: After doing the type change for `SortingDirection` then you should be able to do:

`ascOrDesc === SortingDirection.ASC`

ARADDCC002 (changes requested):

frontend/src/entry/model/Files.tsx
```diff
@@ -14,13 +16,84 @@ type FilesProps = {
...
+  }
+
+  const orderByMenuListItems = useCallback(
+    (value, title) => (
+      <MenuItem
+        onClick={() => {
+          setOrderByValue(value)
+          setOrderByButtonTitle(title)
+        }}
+        sx={{ paddingX: '8px' }}
```
ARADDCC002, line 72: `paddingX` can be shortened to `px`

and of defining pixels, we can just use numbers, like: `px: 2` (obviously change the number to match what you need! Essentially, each number corresponds to a set number of pixels specified in our app (although we could well be using the default value). For example, if `1` corresponds to `5px` then `2` would be `10px` etc.

frontend/src/entry/model/Files.tsx
```diff
@@ -48,8 +121,93 @@ export default function Files({ model }: FilesProps) {
...
+              >
+                {orderByMenuListItems('name', 'Name')}
+                {orderByMenuListItems('createdAt', 'Date uploaded')}
+                {orderByMenuListItems('updatedAt', 'Date updated')}
+                <Divider />
+                <MenuItem
+                  onClick={() => {
+                    setASCorDESC('ASC')
+                  }}
+                  sx={{ paddingX: '8px' }}
```
ARADDCC002, line 159: See earlier comment about setting to `px: <number>`

frontend/src/entry/model/Files.tsx
```diff
@@ -48,8 +121,93 @@ export default function Files({ model }: FilesProps) {
...
+                    <Grid2 size={8}>
+                      <ListItemText>Ascending</ListItemText>
+                    </Grid2>
+                  </Grid2>
+                </MenuItem>
+                <MenuItem
+                  onClick={() => {
+                    setASCorDESC('DESC')
+                  }}
+                  sx={{ paddingX: '8px' }}
```
ARADDCC002, line 184: See earlier comment

ARADDCC002 (changes requested):

ARADDCC002 (changes requested):

PR #2126, opened 2025-04-04
added the ability to add tags to a file via the file management tab

ARADDCC002:


ARADDCC002:
Appears to be an issue whereby adding a new tag does not call the mutate function properly, and occasionally it does not re-render the file tag properly.

backend/src/routes/v2/model/file/patchFile.ts
```diff
@@ -0,0 +1,65 @@
...
+  params: z.object({
+    modelId: z.string({
+      required_error: 'Must specify model id as param',
+    }),
+    fileId: z.string({
+      required_error: 'Must specify file id as param',
+    }),
+  }),
+  body: z.object({
+    metadata: z.object({
```
JR40159, line 22: I think this should be the same as the query content when uploading a file.

frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -259,6 +264,14 @@ export default function FileDownload({
     )
   }, [anchorElScan, chipDisplay, file, openScan])
 
+  const handleFileTagSelectorOnChange = useCallback(
```
JR40159, line 267: This doesn't seem to reliably work when testing locally?

frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -343,11 +360,27 @@ export default function FileDownload({
...
-              <span style={{ fontWeight: 'bold' }}>{` ${formatDateTimeString(file.createdAt.toString())}`}</span>
-            </Typography>
+          <Stack spacing={1} direction='row' alignItems='center'>
+            <Button
+              sx={{ width: 'fit-content' }}
+              size='small'
+              startIcon={<LocalOffer />}
+              onClick={(event) => setAnchorElFileTag(event.currentTarget)}
+            >
+              Apply file tags
```
JR40159, line 370: This could be hidden dependant on if the user has permission to do this

frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -343,11 +360,27 @@ export default function FileDownload({
...
+            </Button>
+            {file.tags.length === 0 && <Typography variant='caption'>None applied</Typography>}
+            <>
+              {file.tags.map((fileTag) => {
```
JR40159, line 374: Having a second row of information for every file could be expensive when a model has a lot of files. By the new 'order by' button we could have a show/hide tags so the user has the option to make the list more concise

JR40159: It would be good to make the tags clickable like in the marketplace but happy for that to be done as part of another PR

JR40159 (changes requested):

backend/src/services/file.ts
```diff
@@ -304,3 +304,30 @@ export async function saveImportedFile(file: FileInterface) {
...
+) {
+  const file = await getFileById(user, fileId)
+  if (!file) {
+    throw BadReq('Cannot find requested file', { modelId: modelId, fileId: fileId })
+  }
+  const model = await getModelById(user, modelId)
+  if (!model) {
+    throw BadReq('Cannot find requested model', { modelId: modelId })
+  }
+  const patchFileAuth = await authorisation.file(user, model, file, FileAction.Update)
```
JR40159, line 322: It doesn't look like our authorisation has any check for updating files

JR40159 (changes requested):

frontend/src/entry/model/ModelFileManagement.tsx
```diff
@@ -240,17 +272,17 @@ export default function Files({ model }: FilesProps) {
...
             enabled).
           </Typography>
-          <Stack width='100%' direction='row' justifyContent='space-between' sx={{ px: 0.5 }}>
+          <Stack width='100%' direction={{ sm: 'column', md: 'row' }} justifyContent='space-between' sx={{ px: 0.5 }}>
             <Restricted action='createRelease' fallback={<Button disabled>Add new files</Button>}>
               <>
                 <label htmlFor='add-files-button'>
-                  <LoadingButton loading={isFilesUploading} fullWidth component='span' variant='outlined'>
+                  <LoadingButton loading={isFilesUploading} component='span' variant='outlined'>
                     Add new files
```
JR40159, line 286: Can users add tags when uploading files via this method?

frontend/src/entry/model/releases/FileDownload.tsx
```diff
@@ -343,11 +390,46 @@ export default function FileDownload({
...
+            {!hideTags && (
+              <>
+                <Restricted action='editEntry' fallback={<></>}>
+                  <Button
+                    sx={{ width: 'fit-content' }}
+                    size='small'
+                    startIcon={<LocalOffer />}
+                    onClick={(event) => setAnchorElFileTag(event.currentTarget)}
+                  >
+                    Apply file tags
```
JR40159, line 403: Can this option be hidden if the user doesn't have permission?

JR40159 (changes requested):

backend/src/routes/v2/model/file/patchFile.ts
```diff
@@ -0,0 +1,63 @@
...
+export const patchFileSchema = z.object({
+  params: z.object({
+    modelId: z.string({
+      required_error: 'Must specify model id as param',
+    }),
+    fileId: z.string({
+      required_error: 'Must specify file id as param',
+    }),
+  }),
+  body: z.object({
```
JR40159, line 21: Can this also include all the other fields that can be included in the post simple upload query params?

JR40159 (changes requested):

frontend/src/entry/model/releases/FileDownload.tsx
ARADDCC002: It should already be hidden using the `<Restricted />` component.

backend/src/routes/v2/model/file/patchFile.ts
ARADDCC002: I think `metadata` is fine, but do we want to be able to change `mime` and `name` too?

ARADDCC002: I've updated the API to accept those params, but left the service layer as-is for now until we determine which we definitely want to change.

JR40159 (approved):

JR40159 (approved):

PR #2140, opened 2025-04-10
Added  help icon describing different searches

IR96334:


frontend/pages/index.tsx
```diff
@@ -168,32 +170,35 @@ export default function Marketplace() {
...
-                endAdornment={
-                  <InputAdornment position='end'>
-                    <IconButton color='secondary' type='submit' sx={{ p: '10px' }} aria-label='filter'>
-                      <SearchIcon />
-                    </IconButton>
-                  </InputAdornment>
-                }
-              />
-            </FormControl>
+            <Stack direction={'row'} spacing={0.5}>
```
ARADDCC002, line 173: As direction prop takes in a string you can just do `direction='row'`

frontend/pages/index.tsx
```diff
@@ -168,32 +170,35 @@ export default function Marketplace() {
...
+              <FormControl
+                sx={{
+                  display: 'flex',
+                  alignItems: 'center',
+                  maxWidth: '400px',
+                  marginBottom: 3,
```
ARADDCC002, line 179: `marginBottom: 3` -> `mb: 3`

frontend/pages/index.tsx
```diff
@@ -168,32 +170,35 @@ export default function Marketplace() {
...
+                <InputLabel htmlFor='entry-filter-input'>Search</InputLabel>
+                <FilledInput
+                  sx={{ flex: 1, backgroundColor: theme.palette.background.paper, borderRadius: 2 }}
+                  id='entry-filter-input'
+                  value={filter}
+                  disableUnderline
+                  onChange={handleFilterChange}
+                  endAdornment={
+                    <InputAdornment position='end'>
+                      <IconButton color='secondary' type='submit' sx={{ p: '10px' }} aria-label='filter'>
```
ARADDCC002, line 193: In cases where we're using such a smaller amount of padding/margin it's better to use the theme spacing if possible. `p:3` for example. Although due to base value probably not being divisible by 10 you might not achieve the same padding - so if it _needs_ to be exactly `10px` then ignore this comment

frontend/pages/index.tsx
```diff
@@ -168,32 +170,35 @@ export default function Marketplace() {
...
+                        <SearchIcon />
+                      </IconButton>
+                    </InputAdornment>
+                  }
+                />
+              </FormControl>
+              <HelpDialog title={'Search Info'} content={<SearchInfo />} />
```
ARADDCC002, line 200: `title='Search Info'`

ARADDCC002 (changes requested):

Issue #1982, opened 2025-03-10
Model Discovery across multiple Instances


ARADDCC002:
## Overview

Having the ability to discover and share models across different running instances of Bailo will help boost collaboration between different groups or organisations. Essentially, when setting up an instance of Bailo, admins will be given a choice to configure a list of additional instances of Bailo to share models with, which can be viewed on the Marketplace assuming that end users turn on the appropriate filter.

For example, if Instance A holds 10 models, and instance B holds 5, then if the user decides to turn the instance sharing filter on, they will give 15 models when they query for all available models.

## Configuration

Each additional instance will have a number of properties that will need to be configured:

- Remote registry 
- Base URL
- Location/Label
- User credentials
- Proxy address
- Proxy auth

Using this information for any additional instances, we should be able to retrieve a list of models based on the query that's sent.

## UI Changes

- Add a new filter to the left-hand side that toggles on model sharing. 
- Add a new label/tag to any models that appear in the marketplace so that it's clear if the model is from the instance the user is using, or one of any additional instances. 

## Server changes

- Update any model query API end points with a new filter that allows users to also query linked instances.
- New property on ModelSearchResult called "location" to store where the model came from (if it came from an additional instance.

## Things to consider

- Models hosted on other instances will only be read-only to begin with, unless clicking a remote hosted model takes you to that instance of Bailo instead.
- Consider what happens if two linked instances of Bailo are running different versions.
- Cache other instance results to reduce API delays due to expanded search.
- Handle duplicate names - decide if we want to:
  - Show duplicates
  - Remove duplicates
  - Prioritise local instance duplicates over remote

## Future improvements

It's likely that we won't add all features on day one. In the future, models hosted on remote instances should be able to be transferred from one instance to another. From a user perpsective, this should just be a single button press on the UI, but this will need a big backend change.

PR #2142, opened 2025-04-11
Initial connectivity between peers

GCHQDeveloper36:
Delivers towards #1982 

- Adds a general system healthcheck endpoint, with placeholder response for now. In the future this could be developed further to return a more accurate healthcheck payload.
- Adds initial peer configuration structure
- Adds a basic 'peer status' endpoint which combines the config, and the healthcheck endpoints to validate connectivity.

~~Also restructures the docker compose environment to allow for multiple instances in the future.~~ moved to #2238 

To test this in a rudimentary way today, modify the `docker_compose.cjs` file to include a federation block:

```typescript

  federation: {
    id: 'bailo-dev',
    state: 'enabled',
    peers: {
      remote1: {
        state: 'enabled',
        baseUrl: 'http://backend:3001',
        label: 'Remote 1',
        kind: 'bailo',
      },
    },
  },
```

If you would like to test the proxy config, add a new service to `docker-compose.yml` as follows:

```yaml
  proxy:
    image: ubuntu/squid:5.2-22.04_beta
    networks:
      internal:
      default:
    ports:
      - 3128:3128
```

and then modify the app config similar to:

```typescript
httpClient: {
  ...
  proxy: 'http://proxy:3128',
  ...
}
```


backend/src/services/federation.ts
```diff
@@ -0,0 +1,45 @@
...
+function ensureFederationNotDisabled(connector: BasePeerConnector) {
+  if (FederationState.DISABLED == connector.getConfiguredState()) {
+    throw ConfigurationError(`Federation is currently disabled; will not fetch peer ${connector.getId()} status`)
+  }
+}
+
+export async function getAllPeerStatus(): Promise<Map<string, PeerConfigStatus>> {
+  ensureFederationNotGloballyDisabled()
+  const peerStatus = new Map<string, PeerConfigStatus>()
+  const promises: Array<Promise<any>> = []
```
ARADDCC002, line 23: Is there a reason that this is set to `any`? I could be wrong but my understanding is that it's an array of whatever `getPeerStatus` returns, which I think is `SystemStatus`?

backend/src/types/types.ts
```diff
@@ -55,6 +60,38 @@ export interface AccessRequestUserPermissions {
...
+  label: string
+  kind: PeerKindKeys
+  proxy: string
+  httpConfig: ProxyAgentOptions
+}
+
+export type SystemStatus = {
+  code: number
+  ping?: string
+  federation?: {
```
ARADDCC002, line 83: Very very nit-picky of me, but can we split `federation` into its own type? Purely because the rest of these seem to be split up.

docker-compose.prod.yml
```diff
@@ -0,0 +1 @@
+name: bailo-prod
```
ARADDCC002, line 1: I'm not sure if you've intentionally renamed this file or not, but some of the builds are failing because they're looking for `docker-compose-prod.yml` not `docker-compose.prod.yml`

backend/src/utils/config.ts
```diff
@@ -26,6 +28,12 @@ export interface Config {
     jwks: string
   }
 
+  httpClient: {
+    proxy: string
+    noProxy: string[]
+    defaultOpts: AgentOptions
```
ARADDCC002, line 34: A lot of unit tests are failing because they can't find this value, you should be able to resolve it by adding it to `backend/src/utils/__mocks__/config.ts`

ARADDCC002 (changes requested):

backend/src/routes/v2/system/status.ts
```diff
@@ -0,0 +1,46 @@
...
+  async (req: Request, res: Response<SystemStatus>) => {
+    const _ = parse(req, getSystemStatusSchema)
+
+    const federation = {
+      state: config.federation.state,
+      id: config.federation.id,
+    }
+
+    return res.json({
+      code: 200,
```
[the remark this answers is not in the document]
GCHQDeveloper36: Intent behind this is for a possible Bailo-specific return code in the future - there's still the overarching HTTP response code, but these codes could in the future indicate some other status. Given we're not actually using it at the moment I could remove this field if preferred for clarity.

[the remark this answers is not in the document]
ARADDCC002: I can see where you're coming from. I think it's best for now to remove it though as it's not needed, and my worry about preemptively adding code might mean that it hangs around indefinitely if we never implement it in the future.

backend/src/connectors/peer/index.ts
JR40159, line 1: Did you consider using a similar pattern to the file scanners where the index returns a 'wrapper' class? Leveraging a class design here would eliminate the need to repeatedly check `peersConfigured` because that could be done when instantiating the wrapper class. I think a wrapper class could provide a simpler and more useful interface for the service layer of the application enabling simpler calls like `peers.modelSearch(...)`. 

Whilst still extending the same base class as the peers, the wrapper methods could also accept an optional property of an array of peer IDs to dictate which peers the actions should be carried out on if not all. This would eliminate the need to have multiple functions for the same action like you've got for getting a single status vs all the statuses.

backend/src/connectors/peer/index.ts
```diff
@@ -0,0 +1,80 @@
...
+      return new BailoPeerConnector(id, cfg)
+    default:
+      throw ConfigurationError(`'${cfg.kind}' not supported.`, {
+        validKinds: Object.values(PeerKind),
+      })
+  }
+}
+
+export function getPeerIds(): Array<string> {
+  if (!peersConfigured) {
```
JR40159, line 60: Using a class would remove the need for this

backend/src/connectors/peer/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    setupPeers()
+  }
+  return peerIds
+}
+
+export function getPeerConnectors(): Map<string, BasePeerConnector> {
+  if (!peersConfigured) {
```
JR40159, line 67: Using a class would remove the need for this

backend/src/connectors/peer/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    setupPeers()
+  }
+  return peers
+}
+
+export default function getPeerConnector(id: string): BasePeerConnector {
+  if (!peersConfigured) {
```
JR40159, line 74: Using a class would remove the need for this

backend/src/connectors/peer/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    keys.push('state')
+  }
+  if (keys.length > 0) {
+    throw ConfigurationError(`Missing configuration for peer`, {
+      missingConfig: keys,
+    })
+  }
+}
+
+function setupPeers(): void {
```
JR40159, line 35: This could be done to enable the instantiation of a wrapper class.

backend/src/routes/v2/system/status.ts
JR40159, line 1: Any thoughts on using the UI config for this? I think that UI config it a misleading name actually what it returns is system configuration that needs to be externally available.

backend/src/services/federation.ts
```diff
@@ -0,0 +1,58 @@
+import { BasePeerConnector } from '../connectors/peer/base.js'
+import getPeerConnector from '../connectors/peer/index.js'
+import peer, { getPeerIds } from '../connectors/peer/index.js'
+import { FederationState, PeerConfigStatus, SystemStatus } from '../types/types.js'
+import config from '../utils/config.js'
+import { ConfigurationError } from '../utils/error.js'
+
+function isGloballyDisabled() {
+  return FederationState.DISABLED == config.federation.state
```
JR40159, line 9: Not `===`?

backend/src/services/federation.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function isDisabledForPeer(connector: BasePeerConnector) {
+  return FederationState.DISABLED == connector.getConfiguredState()
```
JR40159, line 13: Not `===`?

backend/src/services/federation.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export async function getAllPeerStatus(): Promise<Map<string, PeerConfigStatus>> {
+  // Early return if globally disabled
+  if (isGloballyDisabled()) {
```
JR40159, line 18: Rather than do this check on every action, you could just do this once when instantiating the peers?

backend/src/services/federation.ts
```diff
@@ -0,0 +1,58 @@
...
+  federation: {
+    state: FederationState.DISABLED,
+  },
+}
+
+export async function getPeerStatus(id: string): Promise<PeerConfigStatus> {
+  const connector = getPeerConnector(id)
+  const config = connector.getConfig()
+  let status: SystemStatus
+  if (isGloballyDisabled() || isDisabledForPeer(connector)) {
```
JR40159, line 48: Rather than do this check on every action, you could just do this once when instantiating the peers?

backend/src/services/http.ts
```diff
@@ -1,7 +1,56 @@
...
+
+const defaultOpts = config.httpClient?.defaultOpts
+
 // This function has the same syntax as 'https.Agent', but is centralised throughout
 // the application in case it needs to be altered.
-export function getHttpsAgent(config?: https.AgentOptions) {
-  return new https.Agent({ ...config })
+export function getHttpsAgent(opts?: https.AgentOptions) {
+  return new ProxyAgent({
+    getProxyForUrl,
```
JR40159, line 17: This is specifically for the connections between peers so I don't think it make sense to call this every time we want a https agent anywhere in our application. Instead I think this should be an optional parameter that is passed in like `opts`

backend/src/services/http.ts
```diff
@@ -1,7 +1,56 @@
...
+
+export const remotePeerIdHeader = 'x-bailo-remote-id'
+
+/**
+ * Internal callback to support different proxy configurations for each remote peer
+ * @param url of the request
+ * @param req that contains a header referencing the peer ID
+ * @returns empty string if no proxy should be used, or the configured proxy for the request
+ */
+function getProxyForUrl(url: string, req: http.ClientRequest): string {
```
JR40159, line 31: Similar to previous comment in this being specific for peer connections. I would prefer this logic be contained in a 'helper' method in the base peer class

JR40159 (changes requested):

backend/src/connectors/peer/base.ts
```diff
@@ -0,0 +1,36 @@
+import { FederationStateKeys, RemoteFederationConfig, SystemStatus } from '../../types/types.js'
+
+export abstract class BasePeerConnector {
+  id: string
+  config: RemoteFederationConfig
+  constructor(id: string, config: RemoteFederationConfig) {
```
JR40159, line 6: Any thoughts on if this should actually check if this config is valid by attempting to check the status of the peer at this point? Should that then prevent Bailo from starting up as invalid configuration has been provided?

JR40159 (changes requested):

backend/src/connectors/peer/base.ts
GCHQDeveloper36: I wouldn't want it to block the startup of the app if a peer was temporarily unavailable or otherwise misconfigured. There's definitely something there around initial status gathering for the app, but I don't think that should be taking place as a blocking action.

backend/src/services/http.ts
GCHQDeveloper36: I've refactored this in the latest revision so that the app will, by default, pick up environment variables if defined for 'app-wide' proxy config, but will be overridden per-peer if configured.

JR40159 (changes requested):

backend/src/utils/config.ts
```diff
@@ -47,6 +53,12 @@ export interface Config {
     }
   }
 
+  federation: {
+    state: FederationStateKeys
+    id: string
+    peers: Map<string, RemoteFederationConfig>
```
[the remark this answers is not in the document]
GCHQDeveloper36: I don't necessarily disagree but am reluctant to make this change just yet. There's almost certainly going to be other changes/improvements to config which will probably overlap with this, I'd like to see the code in a more realistic setting first before figuring out how to improve it further. I can think of a few cases where we'd like to rely on a locally provided ID rather than the remote's, and one that is more meaningful than a UUID. Can we see how this works in the next phases?

GCHQDeveloper36:
Rebased and most comments now addressed.

JR40159 (changes requested):

JR40159 (approved):

JR40159 (changes requested):
Failing E2E tests caused by:
```
backend-1  | error - (src/routes/middleware/expressErrorHandler.ts:1): Unknown error format was found, returning generic error to user. (err=UnknownZodTypeError {
backend-1  |   message: 'Unknown zod object type, please specify `type` and other OpenAPI props using `ZodSchema.openapi`.',
backend-1  |   data: {
backend-1  |     currentSchema: {
backend-1  |       valueType: [ZodObject],
backend-1  |       keyType: [ZodString],
backend-1  |       typeName: 'ZodMap'
backend-1  |     },
backend-1  |     schemaName: undefined
backend-1  |   }
backend-1  | })
backend-1  | debug - 500 GET /api/v2/specification 51.927ms
```

JR40159:
I'm happy for this PR to be merged once the failing test is fixed.

GCHQDeveloper36:
> I'm happy for this PR to be merged once the failing test is fixed.

Cheers - update pushed which fixed this locally. If the tests now pass would be great to get this finally merged in :-)

ARADDCC002 (approved):

PR #2157, opened 2025-04-22
Fix bugs and refactor file scanners

JR40159:
- Fix async bug in delayed retry logic.
- Fix async bug in initialisation of scanners.
- Remove duplication by moving retry logic into wrapper also simplifies init logic for each scanner.
- Store scanner version on init- remove dependency on scanners after initialisation.

### DIFF
--- a/backend/src/connectors/fileScanning/clamAv.ts
+++ b/backend/src/connectors/fileScanning/clamAv.ts
@@ -48,7 +48,10 @@ export class ClamAvFileScanningConnector extends BaseFileScanningConnector {
         },
       ]
     } catch (error) {
-      return this.scanError(`Unable to scan file. ${this.toolName} errored.`, { error, file })
+      return this.scanError(`This file could not be scanned due to an error caused by ${this.toolName}`, {
+        error,
+        file,
+      })
     }
   }
 }
--- a/backend/src/connectors/fileScanning/clamAv.ts
+++ b/backend/src/connectors/fileScanning/clamAv.ts
@@ -5,60 +5,35 @@ import { getObjectStream } from '../../clients/s3.js'
 import { FileInterfaceDoc } from '../../models/File.js'
 import log from '../../services/log.js'
 import config from '../../utils/config.js'
-import { ConfigurationError } from '../../utils/error.js'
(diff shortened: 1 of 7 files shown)


backend/src/connectors/fileScanning/clamAv.ts
```diff
@@ -73,7 +48,7 @@ export class ClamAvFileScanningConnector extends BaseFileScanningConnector {
         },
       ]
     } catch (error) {
-      return this.scanError(error, file)
+      return this.scanError(`Unable to scan file. ${this.toolName} errored.`, { error, file })
```
ARADDCC002, line 51: Maybe reword to `This file could not be scanned due to an error caused by <tool name>`

backend/src/connectors/fileScanning/modelScan.ts
```diff
@@ -4,59 +4,35 @@ import { getModelScanInfo, scanStream } from '../../clients/modelScan.js'
...
+      return await this.scanError('Could not use ModelScan as it is not running.')
     }
 
     const s3Stream = (await getObjectStream(file.bucket, file.path)).Body as Readable
     try {
       const scanResults = await scanStream(s3Stream, file.name, file.size)
 
       if (scanResults.errors.length !== 0) {
-        return this.scanError(scanResults.errors, file)
+        return this.scanError(`Unable to scan file. ${this.toolName} errored.`, { errors: scanResults.errors, file })
```
ARADDCC002, line 35: See previous comment

backend/src/connectors/fileScanning/modelScan.ts
```diff
@@ -81,7 +57,7 @@ export class ModelScanFileScanningConnector extends BaseFileScanningConnector {
         },
       ]
     } catch (error) {
-      return this.scanError(error, file)
+      return this.scanError(`Unable to scan file. ${this.toolName} errored.`, { error, file })
```
ARADDCC002, line 60: See previous comment

ARADDCC002 (changes requested):

ARADDCC002 (approved):

Issue #1891, opened 2025-02-05
Sort files in a release


PE39806:
Currently, the order of the files returned in a model's release is in a random order (due to JS arrays not guaranteeing order). This looks strange in the frontend as a fresh GET will shuffle the order, so should be replaced with a fixed order e.g. by datetime uploaded.

PE39806:
This was implemented in #2167

PR #2167, opened 2025-04-28
Add pagination to File Management tab

ARADDCC002:
Added a new `<Paginate />` component that provides a pagination for any array of objects.


PE39806:
Functionally, this all looks good. My only query would be whether it's possible to have the file scan result chip show the correct result immediately? Currently "Virus scan results could not be found" is shown for a single frame before being updated to the correct value.

ARADDCC002:
> Functionally, this all looks good. My only query would be whether it's possible to have the file scan result chip show the correct result immediately? Currently "Virus scan results could not be found" is shown for a single frame before being updated to the correct value.

Unfortunately, the list re-renders each item so that's why you're seeing it behave that way. If we render everything then we lose the benefit of pagination.

What we could do is change how it loads, and perhaps use a skeleton or something whilst we wait for things to fetch. I'll need to see where it's best to put that as I don't know if that logic should sit in the `<Paginate />` component, or the children.

PE39806:
This work also happens to close of #1891

frontend/src/common/Paginate.tsx
```diff
@@ -0,0 +1,245 @@
...
+  TextField,
+} from '@mui/material'
+import { isArray } from 'lodash-es'
+import { MouseEvent, ReactElement, useCallback, useEffect, useMemo, useState } from 'react'
+import EmptyBlob from 'src/common/EmptyBlob'
+import { FileInterface } from 'types/types'
+import { sortByCreatedAtDescending } from 'utils/arrayUtils'
+
+interface PaginateProps {
+  list: any[]
```
PE39806, line 21: Can we use generics instead of `any`? e.g. `list: T[]`, `searchFilterProperty?: keyof T extends string ? keyof T : string`, `children: ({ data, index }: { data: T; index: number }) => ReactElement` and `export default funciton Paginate<t>({`

frontend/src/common/Paginate.tsx
```diff
@@ -0,0 +1,245 @@
...
+  children,
+}: PaginateProps) {
+  const [page, setPage] = useState(1)
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null)
+  const menuOpen = Boolean(anchorEl)
+  const [orderByValue, setOrderByValue] = useState('createdAt')
+  const [orderByButtonTitle, setOrderByButtonTitle] = useState('Order by')
+  const [ascOrDesc, setAscOrDesc] = useState<SortingDirectionKeys>(SortingDirection.DESC)
+  const [searchFilter, setSearchFilter] = useState('')
+  const [filteredList, setFilteredList] = useState(list.sort(sortByCreatedAtDescending))
```
PE39806, line 59: Is it okay to use `.sort` within `useState`? The array will be mutated in place which I think can cause some synchronisation issues when within `useState`

frontend/src/common/Paginate.tsx
```diff
@@ -0,0 +1,245 @@
...
+    [filteredList, pageSize],
+  )
+
+  const handlePageOnChange = (_event: React.ChangeEvent<unknown>, value: number) => {
+    setPage(value)
+  }
+
+  const sortFilesByValue = useMemo(
+    () => (a: FileInterface, b: FileInterface) => {
+      if (ascOrDesc === SortingDirection.DESC) {
```
PE39806, line 76: This should be `ascOrDesc === SortingDirection.ASC` as I would expect ascending order to be A, B, C whereas this will return C, B, A.

frontend/src/common/Paginate.tsx
```diff
@@ -0,0 +1,245 @@
...
+        return a[orderByValue] < b[orderByValue] ? -1 : 1
+      }
+      return a[orderByValue] > b[orderByValue] ? -1 : 1
+    },
+    [ascOrDesc, orderByValue],
+  )
+
+  const checkMenuOption = useCallback(
+    (menuOption: string) => {
+      if (menuOption === orderByValue) {
```
PE39806, line 86: This callback could be simplified to `menuOption === orderByValue`

frontend/src/common/Paginate.tsx
```diff
@@ -0,0 +1,245 @@
...
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null)
+  const menuOpen = Boolean(anchorEl)
+  const [orderByValue, setOrderByValue] = useState('createdAt')
+  const [orderByButtonTitle, setOrderByButtonTitle] = useState('Order by')
+  const [ascOrDesc, setAscOrDesc] = useState<SortingDirectionKeys>(SortingDirection.DESC)
+  const [searchFilter, setSearchFilter] = useState('')
+  const [filteredList, setFilteredList] = useState(list.sort(sortByCreatedAtDescending))
+
+  useEffect(() => {
+    setFilteredList(list.filter((item) => item[searchFilterProperty].includes(searchFilter)))
```
PE39806, line 62: This filter is currently case sensitive but the marketplace and menu bar searches are not. It could be made case insensitive e.g. `item[searchFilterProperty].toLowerCase().includes(searchFilter.toLowerCase())`

frontend/src/common/Paginate.tsx
```diff
@@ -0,0 +1,245 @@
...
+    return sortingProperties.map((sortingProperty) => {
+      return orderByMenuListItems(sortingProperty)
+    })
+  }, [sortingProperties, orderByMenuListItems])
+
+  const listDisplay = useMemo(() => {
+    const sortedList = filteredList.sort(sortFilesByValue).slice((page - 1) * pageSize, page * pageSize)
+    if (isArray(sortedList)) {
+      return sortedList.map((item, index) => (
+        <div key={item} style={{ width: '100%' }}>
```
PE39806, line 192: Could replace `style` with `sx`

PE39806 (changes requested):
This is functionally excellent. I think the only change I would like to see would be resetting to page 1 when changing the search filter (and possibly the sorting order) as otherwise it's possible to, for example, be on page 5 but search only returns 1 page of items, so then the page up/down arrows don't work as expected.

frontend/src/entry/model/files/FileDownload.tsx
JR40159, line 1: Can we rename this file to something more accurate?

frontend/src/entry/model/Files.tsx
```diff
@@ -246,7 +112,7 @@ export default function Files({ model }: FilesProps) {
             that are no longer needed, and also manually retrigger anti-virus scanning (if anti-virus scanning is
             enabled).
           </Typography>
-          <Stack width='100%' direction='row' justifyContent='space-between' sx={{ px: 0.5 }}>
+          <Stack width='100%' direction='row' justifyContent='flex-end' sx={{ px: 0.5 }}>
             <Restricted action='createRelease' fallback={<Button disabled>Add new files</Button>}>
```
JR40159, line 116: Seems strange that we have this button on the right of the page?

frontend/src/common/Paginate.tsx
```diff
@@ -0,0 +1,245 @@
...
+    <>
+      <Stack direction={{ xs: 'column', sm: 'row' }} spacing={1} justifyContent='space-between' sx={{ width: '100%' }}>
+        <TextField
+          size='small'
+          placeholder={searchPlaceholderText}
+          value={searchFilter}
+          onChange={(e) => setSearchFilter(e.target.value)}
+          sx={{ maxWidth: '200px' }}
+        />
+        <Pagination count={pageCount} page={page} onChange={handlePageOnChange} />
```
JR40159, line 213: Could we also include this at the bottom of the page?

JR40159 (changes requested):

frontend/src/common/Paginate.tsx
ARADDCC002: `sx` is a MUI only prop which can't be used with `<div />`

frontend/src/common/Paginate.tsx
ARADDCC002: I think this can be safely removed as the list should update with the `useEffect` call at the top of the component anyway.

frontend/src/entry/model/Files.tsx
ARADDCC002: This is to keep it consistent with the other tabs. I'm happy to move it, but we'd need to move all of them and I feel like that's out of scope for this ticket.

JR40159 (approved):

PR #2168, opened 2025-04-29
Upload a Docker image to S3 from the backend

PE39806:
This is part of a larger piece of work to add docker images to the model mirroring functionality.

This work is to:
- GET an image tag's manifest `${registryUrl}/v2/${imageName}/manifests/${tag}`
- parse that to get the relevant layers
- GET each layer `${registryUrl}/v2/${imageName}/blobs/{layerId}`
- compress the layer(s)
- stream the compressed layer(s) to `config.modelMirror.export.bucket`


PE39806:
> I've found a few issues when processing bigger docker images e.g. `pytorch/pytorch:2.7.0-cuda12.8-cudnn9-devel` (8GB) which is using a lot of RAM. The algorithm needs adapting as the layer streams are fetched asynchronously but gzip is synchronous (one layer after another) so there is a bottleneck.

This has been resolved by processing the layers one at a time, and making sure to use `pipe` rather than `buffer`.

backend/src/services/exportRegistry.ts
JR40159, line 1: Add some tests

JR40159 (changes requested):

backend/src/clients/registry.ts
```diff
@@ -22,7 +22,7 @@ const agent = getHttpsAgent({
   rejectUnauthorized: !config.registry.connection.insecure,
 })
 
-async function registryRequest(token: string, endpoint: string) {
+async function registryRequest(token: string, endpoint: string, returnStream: boolean = false) {
```
JR40159, line 25: Looks like by default this function is now returning an `any` type. Could you look at using generics to conditionally type this function?

JR40159 (changes requested):

JR40159 (approved):

backend/src/clients/registry.ts
PE39806: Whe function has always returned `an`y which is because of the `await res.json()` being the returned value. I don't think I can use generics in this case because we don't know which API endpoint is being requested, but the functions that call/wrap `registryRequest` do have typeguards to validate that the response is okay.

PR #2211, opened 2025-05-02
Feature/bai 1717 add org and state to model display

ARADDCC002:


PE39806:
Could we trigger the model page to auto-update the Organisation and/or Model State when they are changed in the Settings tab?

JR40159 (changes requested):
I think we need to use colour spacing and sizing to make it clearer that the org/state and collaborator information is separate to the main title, ID and description. For example, the 'View Collaborators' button is using the same colour scheme as the title, ID and description and the state/org is using it's own generic black text.

JR40159 (approved):

PR #2220, opened 2025-05-08
Fixed delete with no auth error

IR96334:


backend/src/connectors/authorisation/base.ts
```diff
@@ -271,7 +271,10 @@ export class BasicAuthorisationConnector {
         ) {
           return {
             success: false,
-            info: 'You do not have permission to upload a file.',
+            info:
+              action === FileAction.Delete
```
ARADDCC002, line 275: I'm not sure I like using a ternary for this as it limits this auth check to just two actions. If we're going to include both the `Delete` and `Upload` auth file checks in this one call I think we should use a switch case instead as we might need to add more checks in the future,

JR40159: Taking a step back from this- maybe this function should behave in a similar way to the `releases` function? As with releases, the permissions a user has to perform file related actions are decided by what permissions a user has on a model. It also might be a good idea to refactor the model permissions `update` and `write` to be more clear. Perhaps refactor `update` to allow a user to update any part of a model- eg edit the model card, create a release, upload a file etc. and and `modifySettings` to allow a user to change a model's settings. Basically, have `update` to give collaborator permissions and `modifySettings` to give admin level permissions.

PR #2235, opened 2025-05-21
improvements to file download 404 handling

ARADDCC002:


backend/src/clients/s3.ts
```diff
@@ -46,25 +47,30 @@ export async function putObjectStream(
...
-      fileSize = progress.loaded
-    }
-  })
-
-  const s3Response = await upload.done()
-  log.debug(s3Response, 'Object upload complete')
 
-  return {
-    fileSize,
+  try {
```
JR40159, line 51: It's probably worth also including the instantiation of the Upload class in this try block

JR40159 (changes requested):

JR40159 (approved):

PE39806 (approved):

PR #2239, opened 2025-05-23
WIP: HuggingFace search

GCHQDeveloper36:
Heavy WIP. Builds off #2142 to deliver basic search over another peer for models.


GCHQDeveloper36:
Squashed & rebased with latest main (v2.9.0)

backend/src/connectors/peer/base.ts
```diff
@@ -56,4 +70,17 @@ export abstract class BasePeerConnector {
...
+
+  /**
+   * Namespace cache by user
+   *
+   * @param user to extract the DN from
+   * @param key to use as the remainder of the key (such as a query string)
+   * @returns a user-specific cache key
+   */
+  buildCacheKey(user: UserInterface, key: string) {
+    return `${user.dn}+${key}`
```
PE39806, line 84: I'm not super keen on using a raw `user.dn` as that could include unexpected characters in it depending on things like region and language set. I would be happier if we base-64 encode it first e.g. `Buffer.from(user.dn).toString('base64')`. This would still be user specific.

backend/src/connectors/peer/wrapper.ts
```diff
@@ -34,4 +36,28 @@ export class PeerConnectorWrapper {
...
+      throw InternalError('Invalid peer IDs provided to wrapper')
+    }
+    for (const id of peersToQuery) {
+      const peer = this.peers.get(id)
+      if (!peer) {
+        throw InternalError(`Peer connector not found: ${id}`)
+      }
+      const queryResponse = await peer.queryModels(opts, user)
+      results.push(...queryResponse)
+    }
```
PE39806, line 59: Sequential is okay if we only have a few peers to query but we could future proof this by changing it to user `Promise.all` e.g.
```
await Promise.all(
    peersToQuery.map(id => this.peers.get(id)?.queryModels(opts, user) ?? Promise.resolve([]))
  )
```
I'm not quite sure how the best way to make this work with `throw` is, however, so if it's too clunky then it's not worth rewriting for now.

backend/src/connectors/peer/huggingface.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (cacheResult) {
+      log.debug(`Cache hit - returning cached result for ${cacheKey}`)
+      return cacheResult
+    }
+    log.debug(`Cache miss - executing query for ${cacheKey}`)
+
+    // No cache result - execute query, cache response if appropriate
+    const models = new Array<ModelSearchResult>()
+    // Don't allow a whole search
+    if (opts.query.length < 5) {
```
PE39806, line 78: I don't like this returning "silently" - can we throw an error instead? This way the frontend can handle the error and let the user know why the result is not what they might otherwise be expecting.

backend/src/connectors/peer/base.ts
PE39806: Also, this is only per-user and not also per-peer, so caches with multiple peers night have collisions.
```suggestion
    return `${this.id}:${Buffer.from(user.dn).toString('base64')}+${key}`
```

PE39806 (dismissed):
This is looking like excellent functionality and I know will be very useful. I've found some minor issues, but overall this is looking really clean and well thought out.

backend/src/connectors/peer/wrapper.ts
GCHQDeveloper36: Updated this and some of the other code in this class to be more parallelised. If we accept that the IDs are validated above this code, then we can just assert we have a peer and ignore the current check.

Issue #2286, opened 2025-06-02
`updateFile` ignores some optional params


PE39806:
The code at https://github.com/gchq/Bailo/blob/cc91d3c2576ef416ec2ac4538e0812995a785fa4/backend/src/services/file.ts#L319-L346 does not process the optional `name` and `mime` parts of the `patchFileParams` param, meaning that they are ignored. This is not expected behaviour and the function should be updated to fix this.

PR #2306, opened 2025-06-04
rewrite updateFile to process all patchFileParams and update tests

PE39806:
Fix for #2286

### DIFF
--- a/backend/src/services/file.ts
+++ b/backend/src/services/file.ts
@@ -337,10 +337,6 @@ export async function updateFile(
     throw BadReq('Cannot find requested model', { modelId: modelId })
   }
 
-  if (!Object.keys(patchFileParams).every((k) => ['tags', 'name', 'mime'].includes(k))) {
-    throw BadReq('Invalid patch parameter specific', { modelId: modelId, fileId: fileId })
-  }
-
   const patchFileAuth = await authorisation.file(user, model, file, FileAction.Update)
   if (!patchFileAuth.success) {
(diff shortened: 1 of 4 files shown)


backend/src/services/file.ts
```diff
@@ -322,25 +323,37 @@ export async function updateFile(
...
   const patchFileAuth = await authorisation.file(user, model, file, FileAction.Update)
   if (!patchFileAuth.success) {
     throw Forbidden(patchFileAuth.info, { userDn: user.dn, modelId, file })
   }
 
-  if (patchFileParams.tags) {
-    const updatedFile = await FileModel.findOneAndUpdate({ _id: fileId }, { $set: { tags: patchFileParams.tags } })
-    if (!updatedFile) {
-      throw BadReq('There was a problem updating the file', { modelId: modelId, fileId: fileId })
+  for (const [patchFileParamKey, patchFileParamValue] of Object.entries(patchFileParams)) {
```
JR40159, line 345: This seems overly complex in comparison to other examples eg https://github.com/gchq/Bailo/blob/main/backend/src/services/webhook.ts#L47

JR40159 (changes requested):

backend/src/services/file.ts
JR40159: Also I've noticed another error here: https://github.com/gchq/Bailo/blob/main/backend/src/routes/v2/model/file/patchFile.ts#L25 please could you remove the default value

JR40159 (approved):

PR #2331, opened 2025-06-16
Integrate image import/export into Model Mirroring 

PE39806:


backend/src/routes/v2/model/postRequestImport.ts
```diff
@@ -51,7 +53,7 @@ export const postRequestImportFromS3 = [
   async (req: Request, res: Response<PostRequestImportResponse>) => {
     req.audit = AuditInfo.CreateImport
     const {
-      body: { payloadUrl, sourceModelId, mirroredModelId, exporter, importKind, filePath },
+      body: { payloadUrl, sourceModelId, mirroredModelId, exporter, importKind, filePath, imageName, imageTag },
```
JR40159, line 56: Could we combine these two properties into a single property of `name:tag`?

backend/src/services/mirroredModel.ts
```diff
@@ -372,9 +374,7 @@ export async function importModel(
 
   let res: Response
   try {
-    res = await fetch(payloadUrl, {
-      agent: getHttpsAgent(),
```
JR40159, line 376: ?

backend/src/services/mirroredModel.ts
```diff
@@ -402,7 +402,34 @@ export async function importModel(
...
+      return {
+        mirroredModel,
+        importResult: {
+          ...result,
+        },
+      }
+    }
+    case ImportKind.Image: {
+      log.info({ mirroredModelId, payloadUrl }, 'Importing image data.')
+      if (!fileId) {
```
JR40159, line 415: `fileId` required for importing an image?

backend/src/services/mirroredModel.ts
```diff
@@ -341,6 +337,10 @@ export type FileImportInformation = {
   sourcePath: string
   newPath: string
 }
+export type ImageImportInformation = {
+  sourcePath: string
+  image: { modelId: string; imageName: string; imageTag: string }
+}
 
 export async function importModel(
```
JR40159, line 345: For readability, could you move this function back to the top of the file beneath `exportModel()`

backend/src/services/mirroredModel.ts
```diff
@@ -493,13 +525,27 @@ async function importDocuments(
     }),
   )
 
+  // Parse image documents.
```
JR40159, line 528: If we can remove the image content being added to the mongo documents zip file, then the modifications to this function will also be able to be removed

backend/src/services/mirroredModel.ts
```diff
@@ -879,6 +964,39 @@ async function addReleaseToZip(
         },
       )
     }
+
+    if (Array.isArray(release.images)) {
+      for (const image of release.images) {
+        zip.append(JSON.stringify({ modelId: model.id, ...image.toObject() }), {
```
JR40159, line 970: I don't see the point of adding the image information as separate documents to the mongo documents zip here. This is duplication of information that will already be containing within the release documents in the zip file?

backend/src/services/mirroredModel.ts
```diff
@@ -879,6 +964,39 @@ async function addReleaseToZip(
...
+        })
+        const imageLogData = {
+          releaseId: release.id,
+          sourceModelId: model.id,
+          imageName: image.name,
+          imageTag: image.tag,
+          imageId: image._id,
+        }
+
+        // setup gzip prior to calling addRegistryImageToZip to allow the stream to drain, otherwise the stream will get stuck
```
JR40159, line 981: Could the following be simplified into a single call to a refactored `exportCompressedRegistryImage`. As `gzipStream` isn't shared, why can't it be contained within the functionality of `exportCompressedRegistryImage`? This would mean that function would be responsible for creating the gzip, setting up the S3 upload of the gzip, adding content to the gzip and awaiting the completion of the S3 upload

JR40159 (changes requested):

backend/src/services/mirroredModel.ts
PE39806: oops

backend/src/services/mirroredModel.ts
PE39806: The problem is that `gzipStream` needs to be passed to an appropriate function (e.g. `s3Upload` here or `uploadToExportS3Location` in the `streamDockerRegistryToS3` script) otherwise `gzipStream` fills up and doesn't drain, causing the code to hang indefinitely.

PE39806: This isn't possible because `gzipStream` needs to be passed to the `s3Upload`/`streamDockerRegistryToS3` functions before calling `exportCompressedRegistryImage`.

PR #2337, opened 2025-06-20
Add model collaborators get & set methods to python client

PE39806:
As a Bailo user, I would like to be able to programmatically view and add roles for specific people to a model so that I can control who uses a model without using the UI.

### DIFF
--- a/backend/src/routes/v2/model/getModelsSearch.ts
+++ b/backend/src/routes/v2/model/getModelsSearch.ts
@@ -4,7 +4,8 @@ import { z } from 'zod'
 
 import { AuditInfo } from '../../../connectors/audit/Base.js'
 import audit from '../../../connectors/audit/index.js'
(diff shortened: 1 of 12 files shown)


backend/src/routes/v2/model/getModelsSearch.ts
```diff
@@ -94,9 +97,12 @@ export const getModelsSearch = [
       tags: model.card?.metadata?.overview?.tags || [],
       kind: model.kind,
       organisation: model.organisation,
+      state: model.state,
+      collaborators: model.collaborators,
       createdAt: model.createdAt,
       updatedAt: model.updatedAt,
     }))
+    log.debug('Search models', models)
```
JR40159, line 105: Necessary log?

JR40159 (changes requested):

JR40159 (approved):

PR #2344, opened 2025-06-27
Add review role to schemas

ARADDCC002:
Review roles can now be dynamically added via the UI. The previously hard-coded roles have been removed in most of the application.


GCHQDeveloper36:
In the current version of this PR, is creating a new entry (model) and selecting a schema that has some assigned review roles, meant to add those roles to the entry access list? From some local testing it seems I can create the review roles, and relate those to a schema, but that new data doesn't seem to have an impact on release or access request reviews yet.

lib/python/src/bailo/helper/entry.py
```diff
@@ -122,15 +122,6 @@ def get_roles(self):
 
         return res["roles"]
 
-    def get_user_roles(self):
```
PE39806, line 125: Update `lib/python/CHANGELOG.md` with this removal (and other python changes)

backend/src/services/review.ts
```diff
@@ -233,3 +263,14 @@ export async function findReviewRoles(): Promise<ReviewRoleInterface[]> {
...
 }
+
+export async function addDefaultReviewRoles() {
+  for (const reviewRole of config.defaultReviewRoles) {
+    log.info({ name: reviewRole.name }, `Ensuring review role ${reviewRole.name} exists`)
+    const reviewRoleSchema = new ReviewRoleModel({
+      ...reviewRole,
+    })
+    await ReviewRoleModel.deleteOne({ id: reviewRole.id })
+    await reviewRoleSchema.save()
```
PE39806, line 274: Can this be simplified to `await ReviewRoleModel.findOneAndUpdate({ id: reviewRole.id }, reviewRole, { upsert: true})` to make this atomic and reduce total writes to mongo?

backend/src/services/review.ts
```diff
@@ -6,22 +6,25 @@ import { CollaboratorEntry, ModelDoc, ModelInterface } from '../models/Model.js'
...
 import log from './log.js'
 import { getModelById } from './model.js'
 import { requestReviewForAccessRequest, requestReviewForRelease } from './smtp/smtp.js'
 
-// This should be replaced by using the dynamic schema
-const requiredRoles = {
-  release: ['mtr', 'msro'],
-  accessRequest: ['msro'],
+export interface DefaultReviewRole {
+  id: unknown | Uint8Array<ArrayBufferLike>
```
PE39806, line 20: Is `unknown` correct? The type of `id` for `DefaultSchema` is `string` https://github.com/gchq/Bailo/blob/21b79c25b0f533b030cd83700f1e335765141bf2/backend/src/services/schema.ts#L15

PE39806 (changes requested):

backend/src/services/review.ts
```diff
@@ -73,8 +91,20 @@ export async function createReleaseReviews(model: ModelDoc, release: ReleaseDoc)
   await Promise.all(createReviews)
 }
 
+// Update access request to use schema roles
```
JRB66955, line 94: ditto

backend/src/services/model.ts
```diff
@@ -411,6 +412,44 @@ export async function createModelCardFromSchema(
...
+              entity: defaultEntity,
+              roles: [...updatedCollaborators[existingUpdatedUser].roles, reviewRole.short],
+            }
+          } else {
+            updatedCollaborators.push({ entity: defaultEntity, roles: [reviewRole.short] })
+          }
+        }
+      }
+    }
+  }
```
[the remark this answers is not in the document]
GCHQDeveloper36: FWIW - I'm going to take a look at my own version of this code next week to see if there's possible overlap there

JR40159 (changes requested):

JR40159 (changes requested):
When a user selects a schema we should somehow let them know if there are any default collaborators

backend/src/services/roles.ts
```diff
@@ -0,0 +1,56 @@
...
+  let reviewRolesForSchema: string[] = []
+  if (modelId) {
+    const model = await ModelModel.findOne({ id: modelId })
+    if (!model) {
+      throw BadReq('Could not find requested model', { modelId })
+    }
+    if (!model.card) {
+      throw BadReq('Model does not have a schema assigned to it', { modelId })
+    }
+    const modelSchema = await SchemaModel.findOne({ id: model.card.schemaId })
```
JR40159, line 18: This will break if the user goes to a page like the model access prior to selecting a schema

JR40159 (changes requested):

backend/src/services/roles.ts
ARADDCC002: I think this is now resolved?

backend/src/routes/v2/model/roles/getAllModelReviewRoles.ts
```diff
@@ -7,20 +7,19 @@ interface GetModelCurrentUserRolesResponse {
   roles: Array<Role>
 }
 
+// TODO check this is still being used
```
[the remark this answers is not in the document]
ARADDCC002: Decided to remove as they're not used and users can use the existing model roles end point to get the information

PE39806 (dismissed):

JR40159 (changes requested):
One overall comment- did you consider for any roles defined in the schema, not actually saving them on the model collaborators that are using that schema? But rather any time the model is retrieved from the database, dynamically appending the roles that are defined in the schema. This would mean that if the roles on a schema are updated (any of CRUD) then all relevant models would be automatically updated.

backend/src/services/review.ts
```diff
@@ -212,7 +239,7 @@ export async function createReviewRole(user: UserInterface, newReviewRole: Revie
     ...newReviewRole,
   })
 
-  const auth = await authorisation.reviewRole(user, reviewRole, ReviewRoleAction.Create)
+  const auth = await authorisation.reviewRole(user, reviewRole['_id'] as string, ReviewRoleAction.Create)
```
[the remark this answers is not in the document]
ARADDCC002: It doesn't recognise `reviewRole._id`

frontend/pages/reviewRoles/view.tsx
```diff
@@ -1,28 +1,29 @@
...
+  const [errorMessage, setErrorMessage] = useState('')
   const { currentUser, isCurrentUserLoading, isCurrentUserError } = useGetCurrentUser()
 
   const listRoles = useMemo(
     () =>
       reviewRoles.map((reviewRole, index) => (
         <SimpleListItemButton
           selected={selectedRole === index}
-          key={reviewRole.id}
+          key={reviewRole._id}
```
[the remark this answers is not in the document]
ARADDCC002: Do you mean why can't we use `id`? If so, it doesn't exist anymore as it was just the same as `shortName`.

ARADDCC002:
> If my instance of Bailo already has some review roles and I switch to your branch I get the following error when the backend tries to start:
> 
> ```
> backend-1  | info  - (src/services/schema.ts:1): Ensuring schema minimal-access-request-general-v10 exists (reference=minimal-access-request-general-v10)
> backend-1  | /app/node_modules/mongodb/src/operations/insert.ts:88
> backend-1  |       throw new MongoServerError(res.writeErrors[0]);
> backend-1  |             ^
> backend-1  | 
> backend-1  | 
> backend-1  | MongoServerError: E11000 duplicate key error collection: bailo.v2_review_roles index: id_1 dup key: { id: null }
> backend-1  |     at InsertOneOperation.execute (/app/node_modules/mongodb/src/operations/insert.ts:88:13)
> backend-1  |     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
> backend-1  |     at async tryOperation (/app/node_modules/mongodb/src/operations/execute_operation.ts:283:14)
> backend-1  |     at async executeOperation (/app/node_modules/mongodb/src/operations/execute_operation.ts:115:12)
> backend-1  |     at async Collection.insertOne (/app/node_modules/mongodb/src/collection.ts:286:12) {
> backend-1  |   errorLabelSet: Set(0) {},
> backend-1  |   errorResponse: {
> backend-1  |     index: 0,
> backend-1  |     code: 11000,
> backend-1  |     errmsg: 'E11000 duplicate key error collection: bailo.v2_review_roles index: id_1 dup key: { id: null }',
> backend-1  |     keyPattern: { id: 1 },
> backend-1  |     keyValue: { id: null }
> backend-1  |   },
> backend-1  |   index: 0,
> backend-1  |   code: 11000,
> backend-1  |   keyPattern: { id: 1 },
> backend-1  |   keyValue: { id: null }
> backend-1  | }
> backend-1  | 
> backend-1  | Node.js v24.4.1
> ```

Have you pulled this branch before? It's possible that you already have some `ReviewRoles` added to your database, so I'd recommend removing the collection and starting again.

backend/src/services/model.ts
```diff
@@ -416,6 +417,26 @@ export async function createModelCardFromSchema(
     throw BadReq('Cannot create a new Card using a hidden schema.', { schemaId, kind: schema.kind })
   }
 
+  // Check schema review roles for any default entities
```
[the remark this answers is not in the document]
ARADDCC002: Not entirely sure how this can be effectively tested, as the function we'd be testing doesn't actually return the `model`, it returns the model card revision. We'd just be mocking the call to the database which I'm not sure is a helpful test.

backend/src/services/schema.ts
```diff
@@ -95,46 +101,112 @@ export async function updateSchema(user: UserInterface, schemaId: string, diff:
...
+    const models = await ModelModel.find({ 'card.schemaId': schemaId })
+    for (const model of models) {
+      // Remove any roles from model collaborators that have been removed from the schema
+      for (const collaborator of model.collaborators) {
+        collaborator.roles = collaborator.roles.filter((role) => !removedRoles.includes(role))
+      }
+      // Update add users/roles based on new defaultEntities
+      const updatedCollaborators: CollaboratorEntry[] = [...model.collaborators]
+      for (const reviewRoleDiff of diff.reviewRoles) {
+        const reviewRole = await ReviewRoleModel.findOne({ shortName: reviewRoleDiff })
```
PE39806, line 124: Running a DB query in a nested `for` seems inefficient. This could be replaced with
```
// outside of for loops
const reviewRoles = await ReviewRoleModel.find({ shortName: { $in: diff.reviewRoles } })
const roleMap = new Map(reviewRoles.map(role => [role.shortName, role]))
// deeply nested for loop
for (const reviewRoleDiff of diff.reviewRoles) {
  const reviewRole = roleMap.get(reviewRoleDiff)
}
```
which would be less deeply nested in the `for` loops, resulting in fewer DB queries.

backend/src/services/review.ts
```diff
@@ -227,7 +254,75 @@ export async function createReviewRole(user: UserInterface, newReviewRole: Revie
...
+    }
+    const schemas = await SchemaModel.find({ id: schemaIds })
+    if (!schemas || schemas.length === 0) {
+      throw BadReq('Unable to find schemas', { schemaIds })
+    }
+    if (schemas.length > 0) {
+      const schemaRoles: string[] = []
+      schemas.forEach((schema) => schemaRoles.push(...schema.reviewRoles))
+      const uniqueRoles = [...new Set(schemaRoles)]
+      reviewRoles = await ReviewRoleModel.find({ shortName: uniqueRoles })
```
PE39806, line 274: Since the return type is `Promise<ReviewRoleInterface[]>` rather than a `Document`, can this be `lean`?

JRB66955 (approved):

PE39806 (dismissed):

JR40159 (dismissed):

JR40159 (approved):

PR #2426, opened 2025-07-03
Soft delete reviews & responses when a release is deleted

GCHQDeveloper36:
Currently, when a release is deleted from a model entry, the reviews & responses are left in the system. In particular, this leads to the reviews page breaking since it can't find the related semver/model combination.

This change finds the reviews associated with a release, and any responses associated with that review, and soft deletes them before continuing to soft delete the release.

There would be some benefit to wrapping these in a transaction in the future once #1237 is merged.

### DIFF
--- a/backend/src/migrations/019_2426_remove_orphaned_reviews_responses.ts
+++ b/backend/src/migrations/019_2426_remove_orphaned_reviews_responses.ts
@@ -1,0 +1,41 @@
+import ReleaseModel from '../models/Release.js'
+import { ReviewDoc } from '../models/Review.js'
+import { UserInterface } from '../models/User.js'
+import { removeResponses } from '../services/response.js'
+import { removeReleaseReviews } from '../services/review.js'
+
+/**
+ * This isn't really used but needed for consistency
+ */
+const migrationUser: UserInterface = {
+  dn: 'migration',
+}
+
+export async function up() {
+  // Find all releases have already been marked as deleted
+  const deletedReleases = await ReleaseModel.find({ deleted: true }).lean()
+  const deletedReviews: ReviewDoc[] = []
+
+  // For each deleted release, deleted the reviews associated with it
+  for (const { modelId, semver } of deletedReleases) {
+    const reviews = await removeReleaseReviews(migrationUser, modelId, semver)
+    deletedReviews.push(...reviews)
+  }
+
+  const reviewIds = deletedReviews.map((r) => r.id)
+  // For each deleted review, delete the responses associated with it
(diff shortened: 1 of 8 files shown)


ARADDCC002:
I'll get time to properly review this next week, but I wonder if this needs a migration script to cover any preexisting orphaned reviews and responses? 

We've had an issue in the past where the side bar navigation component occasionally doesn't render, and I wonder if it's because of dodgy reviews (we fetch user reviews to determine the number for that icon over the `/reviews` button to display how many active reviews you have), so there's a very slim (but possible!) chance it might help resolve that.

GCHQDeveloper36:
> I'll get time to properly review this next week, but I wonder if this needs a migration script to cover any preexisting orphaned reviews and responses?
> 
> We've had an issue in the past where the side bar navigation component occasionally doesn't render, and I wonder if it's because of dodgy reviews (we fetch user reviews to determine the number for that icon over the `/reviews` button to display how many active reviews you have), so there's a very slim (but possible!) chance it might help resolve that.

I've added a migration script and moved the code around a little. Would probably benefit from running this on a few dev instances to make sure it's picking up the right releases/reviews/responses.

backend/src/services/response.ts
```diff
@@ -61,6 +62,22 @@ export async function updateResponse(user: UserInterface, responseId: string, co
   return response
 }
 
+export async function removeResponses(user: UserInterface, parentIds: string[]) {
```
ARADDCC002, line 65: `user` doesn't look like it's being used

backend/src/migrations/019_2426_remove_orphaned_reviews_responses.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * This isn't really used but needed for consistency
+ */
+const migrationUser: UserInterface = {
+  dn: 'migration',
+}
+
+export async function up() {
+  // Find all releases have already been marked as deleted
+  const deletedReleases = await ReleaseModel.find({ deleted: true }).lean()
```
ARADDCC002, line 16: We should remove `lean` as it's the semver is failing validation when this object is returned as a POJO. We also need to consider access request reviews as well as releases

backend/src/services/review.ts
```diff
@@ -113,6 +113,31 @@ export async function removeAccessRequestReviews(accessRequestId: string) {
   return deletions
 }
 
+export async function removeReleaseReviews(user: UserInterface, modelId: string, semver: string): Promise<ReviewDoc[]> {
```
JR40159, line 116: Could you include some unit tests to cover this?

PR #2436, opened 2025-07-14
Replace unzip buffer with streams

PE39806:
##### Checklist

- Import: Model/document import logic
- Tests: Unit tests & mocks for streamed `.zip` processing

### Description of change

Refactor the model/document import function to support `.zip` files of greater size by streaming and processing entries one at a time instead of reading the full archive into memory. This new implementation uses streamed unzipping via the `unzipper` library (rather than `fflate`), parses and stores each JSON file as it is read, reduces memory usage and allows for better handling of large archives.

This PR also updated the existing tests to use `unzipper` rather than `fflate`.


PE39806:
I'd suggest toggling "hide whitespace" when reviewing as a few changes are just big `try catch` blocks which only affect indentation.

PR #2445, opened 2025-07-15
Build modelscan production image differently

GCHQDeveloper36:
FastApi 0.116.0 was released which contained a dependency uplift on fastapi-cli 0.0.8, itself with a new dependency on fastapi-cloud-cli that was released publicly in the last couple of weeks. Our default Dockerfile for modelscan includes an innocuous `--upgrade` when doing a pip install with the requirements file which unfortunately pulled these new versions in. This broke when running the modelscan image as a non-root user and with a read-only filesystem, as the cloud CLI dependency has code that checks write-ability in a new config location.

Even though this might be fixed in 0.1.4, since we don't need the cloud CLI, I've updated the requirements.txt file to use a new extras declaration that excludes that dependency by default. I've also removed the upgrade command for the requirements pip install to avoid similar issues in the future.

Alongside this, I needed to bring the final image more up-to-date to resolve several CVEs that were getting flagged by container scanners. I did try switching to distroless, but it's limited to Python 3.11 by default and still uses Debian under the covers, but without the ability to update packages.

In the end I made the switch to just regular Ubuntu that we can update ourselves as required, with only a minimal increase in image size.

To run as a non-root, read-only (except /tmp) user:
`docker run --user 65532:65532 --read-only --mount=type=tmpfs,target=/tmp --rm bailo-modelscan:latest`

You can run an example container scan with a command like:
`docker run --pull=always -v /tmp/trivy-cache:/root/.cache/ -v /var/run/docker.sock:/var/run/docker.sock aquasec/trivy:latest image bailo-modelscan:latest --scanners vuln`


lib/modelscan_api/Dockerfile
```diff
@@ -1,19 +1,39 @@
-FROM python:3.12.11-slim-bookworm
+# syntax=docker/dockerfile:1
+ARG PYTHON_VERSION=3.12 \
+    PYTHONPATH=/venv/lib/python${PYTHON_VERSION}/site-packages \
+    PY_IMG=python:${PYTHON_VERSION}-slim-bookworm \
```
JR40159, line 4: I think this is being covered in a separate PR, but I'm not a massive fan of allowing builds to override the base image as I think it adds complexity around knowing which images are supported and safe to use.

JR40159 (changes requested):

lib/modelscan_api/Dockerfile
JR40159: Perhaps it would be better to remove this change from this PR and keep it confined to the other PR.

GCHQDeveloper36: Done, removed (just left pythonpath in)

PR #2446, opened 2025-07-15
Add image arguments for Dockerfiles

GCHQDeveloper36:
Fairly simple change to add in build-args that can be overridden when necessary to use a different base image.

### DIFF
--- a/Dockerfile.standalone
+++ b/Dockerfile.standalone
@@ -1,4 +1,8 @@
-FROM node:24-alpine AS backend
+# syntax=docker/dockerfile:1
+ARG NODE_IMG=node:24.4.0-alpine \
+    BASE_IMG=ubuntu:jammy
+
+FROM ${NODE_IMG} AS backend
 RUN apk add --no-cache libc6-compat
 RUN apk update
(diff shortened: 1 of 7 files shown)


PE39806:
I've done some testing with [dependabot cli](https://github.com/dependabot/cli) and it looks like dependabot will no longer pick up on new versions if supplied via `ARG ... FROM ${ARG}` rather than directly via `FROM`. I'd like to keep dependabot (or a similar tool) doing something rather than having to manually stay on top of these dependencies.

Perhaps we can create a Dockerfile with the fixed versions e.g.
```
FROM python:3.13.2-bookworm AS sphinx-docs
FROM node:24.4.0-alpine AS installer
...
```
That way dependabot will pick up on that and we can either manually update the rest or possibly create a GitHub Action to update the PR(s) to bump the args.

What are your thoughts? If you know of any alternatives to dependabot that will pick up on the `ARG ... FROM ${ARG}` change then I would happily switch to using that.

GCHQDeveloper36:
Closing as build args aren't supported by dependabot.

PR #2447, opened 2025-07-15
Collapse Dockerfile.dev into single Dockerfile per service

GCHQDeveloper36:
Updates backend/frontend/modelscan services to remove the Dockerfile.dev files. This was fairly simple for modelscan, but a little more convoluted for backend/frontend. Next step after this PR & #2445, #2446 will be to add in a common base image for alpine to avoid duplicating the updates etc.


docker-compose-prod.yml
```diff
@@ -38,7 +38,6 @@ services:
     image: bailo_modelscan:latest
     build:
       context: ./lib/modelscan_api
-      dockerfile: ./Dockerfile
```
JR40159, line 41: Out of interest, is the reason for not needing to provide a target here because of the order of the stages defined in the `Dockerfile`? As result of the production page being the final stage does that become the default? If so, would production builds be more like the ones defined in this docker compose file be more efficient by specifying a target stage rather than relying on the default?

JR40159 (approved):

docker-compose-prod.yml
GCHQDeveloper36: Without a target specified, it will basically target the final stage in the Dockerfile by default. If we wanted to be explicit we could still target the 'prod' stage?

Issue #2453, opened 2025-07-18
Improve Python Docstrings


PE39806:
### Problem

A number of the Python docstrings which generate our [Python Sphinx docs](https://gchq.github.io/Bailo/docs/python/index.html) are out of date, missing parameters or non-existent.

### Proposed Solution

Update the docstrings to add and improve the Python documentation. This will be done over several separate PRs, with Open Source contributions being very welcome.

- [ ] As part of this, we should switch the `backend/Dockerfile`  (and `backend/Dockerfile.dev` until #2447 is merged) to pull in the latest _local_ version of the python bailo client `/lib/python` rather than building the most recently published version on PyPI. This means that any forks that modify the python client will include their own version of the docs, rather than our version.

### Alternatives

n/a

### What is the priority of your issue?

Low

PR #2454, opened 2025-07-18
Make docker-compose backend build docs for the local python version

PE39806:
Contributes towards #2453.
Backend docs now pull in the latest local version of the python bailo client `/lib/python` rather than building the most recently published version on PyPI. This means that any fork that modififies the python client will include their own version of the docs for their own Bailo instance, rather than the latest version on PyPI.

Note that this _does not_ change the docs built as part of the GitHub Action as the Action pulls the latest version of the bailo package from PyPI https://github.com/gchq/Bailo/blob/dev/build-python-docs-using-local-files/.github/workflows/web.yml#L44. This is intentional as we want the public docs for Open Source to match up with the current published version of bailo on PyPI.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -11,8 +11,9 @@ RUN apt-get update && apt-get install -y pandoc && rm -rf /var/lib/apt/lists/*
 WORKDIR /app/docs
 
 COPY docs .
+COPY --from=python . /app/python
 
-RUN --mount=type=cache,target=/root/.cache/pip pip install --upgrade bailo -r requirements.txt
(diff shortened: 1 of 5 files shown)


PE39806:
Because the `matrix` has changed (as below), the "Required" builds will require updating with the new build before this can be merged.

https://github.com/gchq/Bailo/pull/2454/files#diff-53168f9f47f224a6a8aff835f5d9972c5b946de67cbcac6dc7330c3a1e47c02aR22

JR40159 (approved):

PR #2488, opened 2025-08-01
Improve API `POST` `/api/v2/model/{modelId}/export/s3` endpoint time

PE39806:
The endpoint was set up to wait for each artefact to be uploaded to S3 before completing. This is problematic when there are a large number of artefacts being uploaded as the connection may be open for a long time (and therefore risks being cut off).
This PR changes the functionality to no longer wait for each release artefact to be uploaded before returning the response.

### DIFF
--- a/backend/src/services/mirroredModel.ts
+++ b/backend/src/services/mirroredModel.ts
@@ -733,20 +733,15 @@ async function addReleasesToZip(
   return zip
 }
 
-async function addReleaseToZip(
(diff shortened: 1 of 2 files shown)


ARADDCC002 (approved):

JR40159 (changes requested):
Need to safely handle errors thrown in asynchronous code.

JR40159 (approved):

PR #2489, opened 2025-08-01
fixed a bug where failed scans would display as passed

ARADDCC002:


frontend/src/entry/model/files/FileDisplay.tsx
```diff
@@ -144,20 +144,23 @@ export default function FileDisplay({
   const updateChipDetails = useCallback(() => {
     if (!isFileInterface(file) || file.avScan === undefined) {
```
JR40159, line 145: This looks like a strange way to structure this. Why not use a single if statement with `elseif`s with a more appropriate default in the `else` block`? Having the default as 'everything is good' seems like a risky choice.

JR40159 (changes requested):

JRB66955 (approved):

JR40159 (approved):

PR #2490, opened 2025-08-01
added a check for the file display menu to only appear if it has options

ARADDCC002:


frontend/src/entry/model/files/FileDisplay.tsx
```diff
@@ -324,6 +324,10 @@ export default function FileDisplay({
     )
   }
 
+  const showMenu = () => {
```
JR40159, line 327: Why's this needed?

JR40159 (changes requested):

frontend/src/entry/model/files/FileDisplay.tsx
ARADDCC002: The menu is made up of up to three options `Associated releases`, `Delete file` & `Rerun file scan`. These options are dynamic and dependent on which component the `FileDisplay` component is called in, and if the file scanners are active. If all of these options are false, the menu will have no options, but the menu icon will still appear.

JR40159 (approved):

PR #2548, opened 2025-08-05
Implemented collaborator role in review role

ARADDCC002:


lib/python/src/bailo/core/enums.py
```diff
@@ -30,8 +30,6 @@ class Role(ValuedEnum):
     """A reviewing role."""
 
     OWNER = "owner"
-    MODEL_TECHNICAL_REVIEWER = "mtr"
-    MODEL_SENIOR_RESPONSIBLE_OFFICER = "msro"
```
PE39806, line 34: Update the python changelog with this please.

PE39806 (changes requested):

Issue #2449, opened 2025-07-16
Improve stream handling


PE39806:
### Problem

Some of the way that we handle streams is inconsistent and not best practice.

### Proposed Solution

We should be making sure to wrap streams in `try catch` blocks where the `catch` calls `stream.destroy()` (or equivalent) to prevent having orphaned events. Where possible, we should try to use (Node streams)[https://nodejs.org/api/stream.html] rather than (WHATWG streams)[https://nodejs.org/api/webstreams.html#web-streams-api] for consistency, although this is not a hard and fast rule.
A chunk of this work was addressed in https://github.com/gchq/Bailo/pull/2436 but that PR ballooned in scope, and was fundamentally flawed due to trying to stream extraction of `.zip` files. It is suggested to look into this to find possible solutions to some of the above requirements.

### Alternatives

We could prefer (WHATWG streams)[https://streams.spec.whatwg.org/] but I've found that these a little bit more cumbersome. I'm happy to have a healthy discussion around this if others disagree, and I think there will always be certain cases where WHATWG streams may be better suited than Node streams (and vice versa).

### What is the priority of your issue?

Low

PR #2550, opened 2025-08-06
Replace mirrored model zip with targz

### LINKED ISSUE #2448: Replace mirrored model `.zip` with `.tar.gz`
### Problem

As per [No Streaming Unzip API](https://maxim-ui.readthedocs.io/en/latest/node_modules/yauzl/README.html#no-streaming-unzip-api), `.zip` files have a central directory at the end of the `.zip` file. This means they are not able to be properly streamed which is a risk when handling large files.

### Proposed Solution

Reimplementing some of the work in https://github.com/gchq/Bailo/pull/2436, the `.zip` archive creation and extraction should be replaced with `.tar.gz` as this has sequential files which are also compressed, meaning that it works well with streaming. This is already used in our code e.g. https://github.com/gchq/Bailo/blob/main/backend/src/services/mirroredModel.ts#L277-L281 so should be simple enough to reuse, although some refactoring is likely required.
I'd like to keep the changes in https://github.com/gchq/Bailo/pull/2436/files#diff-9443efb258e4232777a51344c4cdd7c7f989e28fc77f84b1bd6ee470b8dae333R473-R584 but make them applicable to `.tar.gz` files.

### Alternatives

7z, DAR and CPIO also theoretically support streaming but TAR is already in our code base so we should use that.

Low

PE39806:
Closes #2448 and works towards #2449
As well as replacing the use of `.zip` with `.tar.gz`, this also refactors `backend/src/services/mirroredModel.ts` into several files to make it less tightly coupled and much easier to test.


backend/src/services/importers/documentImporter.ts
```diff
@@ -0,0 +1,149 @@
...
+      {
+        name: entry.name,
+        type: entry.type,
+        size: entry.size,
+        importId,
+      },
+      'Processing un-tarred entry',
+    )
+
+    if (entry.type === 'file') {
```
ARADDCC002, line 53: I think this is left over from old code, but seems a good time to replace with `ImportKind.File`

backend/src/services/importers/imageImporter.ts
```diff
@@ -0,0 +1,130 @@
...
+        {
+          name: entry.name,
+          type: entry.type,
+          size: entry.size,
+          importId,
+        },
+        'Processing un-tarred entry',
+      )
+
+      if (entry.type === 'file') {
```
ARADDCC002, line 53: `ImportKind.File`

backend/src/services/importers/imageImporter.ts
```diff
@@ -0,0 +1,130 @@
...
+        // Process file
+        if (manifestRegex.test(entry.name)) {
+          // manifest.json must be uploaded after the other layers otherwise the registry will error as the referenced layers won't yet exist
+          log.debug({ importId }, 'Extracting un-tarred manifest')
+          manifestBody = await json(stream)
+
+          next()
+        } else if (blobRegex.test(entry.name)) {
+          // convert filename to digest format
+          const layerDigest = `${entry.name.replace(/^(blobs\/sha256\/)/, 'sha256:')}`
```
ARADDCC002, line 63: Is there any chance we might want to use anything other than `sha256`, and if so, should we make this configurable somehow?

ARADDCC002 (changes requested):

backend/src/services/importers/imageImporter.ts
PE39806: No - Docker digests are always a sha256 sum https://docs.docker.com/dhi/core-concepts/digests/#what-are-docker-image-digests

backend/src/services/importers/documentImporter.ts
PE39806: I don't agree in this context - `entry` is type `Headers` from the `tar-stream` package (as per https://www.npmjs.com/package/tar-stream#headers) so I don't think we should be mixing this with our own types.

backend/src/services/mirroredModel.ts
JR40159, line 1: It would be worth having a mirrored model directory in the services directory that contains the mirroredModel.ts and well as the new importers and parsers directories.

ARADDCC002 (approved):

PR #2556, opened 2025-08-07
improved review role display. readded missing role name display code

ARADDCC002:


backend/config/default.cjs
```diff
@@ -248,7 +248,7 @@ module.exports = {
     },
 
     roleDisplayNames: {
-      owner: 'Owner',
```
JR40159, line 251: Owner is still default

infrastructure/helm/bailo/values.yaml
```diff
@@ -257,7 +257,7 @@ config:
       states: []
 
     roleDisplayNames:
-      owner: 'Owner'
```
JR40159, line 260: Default is still Owner

JR40159 (changes requested):

PR #2573, opened 2025-08-19
Move review count logic from the frontend into the backend

JRB66955:


backend/test/services/review.spec.ts
```diff
@@ -218,14 +218,14 @@ describe('services > review', () => {
   const user: any = { dn: 'test' }
 
   test('findReviews > all reviews for user', async () => {
-    await findReviews(user, true)
+    await findReviews(user, true, false)
 
     expect(reviewModelMock.match.mock.calls.at(0)).toMatchSnapshot()
     expect(reviewModelMock.match.mock.calls.at(1)).toMatchSnapshot()
   })
```
PE39806, line 226: Please add a test where the param `open` in `findReviews` is `true`

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -84,21 +84,14 @@ export default function SideNavigation({
...
 }: SideNavigationProps) {
   const [reviewCount, setReviewCount] = useState(0)
-  const { reviews, isReviewsLoading, isReviewsError } = useGetReviewRequestsForUser()
+  const { reviews, isReviewsLoading, isReviewsError } = useGetReviewRequestsForUser(true)
   const { responses, isResponsesLoading, isResponsesError } = useGetUserResponses()
 
   useEffect(() => {
     async function fetchReviewCount() {
       onResetErrorMessage()
       if (reviews) {
```
PE39806, line 93: I'd prefer this being `if (Array.isArray(reviews))` as it does type and property checks as well.

backend/src/services/review.ts
```diff
@@ -53,6 +55,17 @@ export async function findReviews(
...
+    const responses = await getResponsesByParentIds(parentIds)
+    reviews = reviews.filter(
+      (review) =>
+        !responses.find(
+          (response) => response.entity === `user:${user.dn}` && response.parentId.toString() === review._id.toString(),
+        ),
+    )
+  }
+
   return reviews.filter((_, i) => auths[i].success)
```
PE39806, line 69: I think that this won't work as intended as if `open` is `true` then `reviews = reviews.filter(...)` is called _after_ getting the `auths` so we cannot then `filter` on the `auths` indices. I think swapping the `const auths = ...` and `if (open) {...` blocks around will resolve this.

ARADDCC002 (dismissed):

PE39806 (approved):

PR #2577, opened 2025-08-22
Feature/bai 1695 schema diff

ARADDCC002:


frontend/src/schemas/SchemaCompare.tsx
```diff
@@ -0,0 +1,73 @@
...
+import ReactDiffViewer, { DiffMethod } from 'react-diff-viewer-continued'
+import Loading from 'src/common/Loading'
+import MessageAlert from 'src/MessageAlert'
+import { SchemaInterface } from 'types/types'
+
+export default function SchemaCompare() {
+  const { schemas, isSchemasLoading, isSchemasError } = useGetSchemas()
+
+  const [beforeSchema, setBeforeSchema] = useState<SchemaInterface | null>()
+  const [afterSchema, setAfterSchema] = useState<SchemaInterface | null>()
```
PE39806, line 14: Does it matter that these will initially be `undefined`?
This could be changed to `const [beforeSchema, setBeforeSchema] = useState<SchemaInterface | null>(null)`, for example, to always be initialised as `null`.

frontend/src/schemas/SchemaCompare.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <Stack direction='row' spacing={12} justifyContent='center' alignItems='center'>
+          <Autocomplete
+            disablePortal
+            options={schemas}
+            fullWidth
+            size='small'
+            getOptionDisabled={(option) => option.id === afterSchema?.id}
+            getOptionLabel={(option: SchemaInterface) => option.name}
+            sx={{ width: 300 }}
+            renderInput={(params) => <TextField {...params} label='Schema ' />}
```
PE39806, line 44: Could: change this `label` to `Old Schema`.

frontend/src/schemas/SchemaCompare.tsx
```diff
@@ -0,0 +1,73 @@
...
+          <Forward color='primary' fontSize='large' />
+          <Autocomplete
+            disablePortal
+            options={schemas}
+            fullWidth
+            size='small'
+            getOptionDisabled={(option) => option.id === beforeSchema?.id}
+            getOptionLabel={(option: SchemaInterface) => option.name}
+            sx={{ width: 300 }}
+            renderInput={(params) => <TextField {...params} label='Schema' />}
```
PE39806, line 56: Could: change this `label` to `New Schema`.

frontend/src/schemas/SchemaCompare.tsx
```diff
@@ -0,0 +1,73 @@
...
+            options={schemas}
+            fullWidth
+            size='small'
+            getOptionDisabled={(option) => option.id === afterSchema?.id}
+            getOptionLabel={(option: SchemaInterface) => option.name}
+            sx={{ width: 300 }}
+            renderInput={(params) => <TextField {...params} label='Schema ' />}
+            onChange={handleBeforeSchemaChange}
+          />
+          <Forward color='primary' fontSize='large' />
```
PE39806, line 47: Add an `aria-label` e.g. `"Compare arrow"`.

frontend/src/schemas/SchemaCompare.tsx
```diff
@@ -0,0 +1,73 @@
...
+            getOptionLabel={(option: SchemaInterface) => option.name}
+            sx={{ width: 300 }}
+            renderInput={(params) => <TextField {...params} label='Schema' />}
+            onChange={handleAfterSchemaChange}
+          />
+        </Stack>
+        {beforeSchema && afterSchema ? (
+          <ReactDiffViewer
+            oldValue={JSON.stringify(beforeSchema, null, 2)}
+            newValue={JSON.stringify(afterSchema, null, 2)}
```
PE39806, line 63: I think that these will be re-run on every render. Can `useMemo` be utilised to avoid this?

PE39806 (changes requested):

PE39806 (approved):

PR #2631, opened 2025-09-02
Feature/bai 1696 schema migration plan UI

ARADDCC002:


PE39806 (changes requested):

frontend/src/MuiForms/QuestionViewer.tsx
```diff
@@ -0,0 +1,39 @@
...
+  formContext?: FormContextType
+  schema?: any
+}
+
+export default function QuestionViewer({ label, id, schema, formContext }: QuestionViewProps) {
+  if (!formContext) {
+    return <MessageAlert message='Unable to render widget due to missing context' severity='error' />
+  }
+
+  const schemaPath = `${formContext.rootSection}.${id?.replace('root_', '').replaceAll('_', '.')}`
```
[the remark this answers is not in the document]
ARADDCC002: Return an error message rather than null to make it more obvious if there's a problem. Realistically, `id` should _never_ be null/undefined, but you're rest that it's best to check.

frontend/src/schemas/SchemaMigrator.tsx
```diff
@@ -69,12 +69,18 @@ export default function SchemaMigrator({ sourceSchema, targetSchema }: SchemaMig
...
     return questionMigrations.map((migrationAction) => {
-      if (migrationAction.kind === MigrationKind.DELETE) {
+      if (migrationAction.kind === MigrationKind.MOVE) {
         return (
           <Stack direction='row' spacing={1} alignItems='center' key={migrationAction.id}>
             <Tooltip title='Remove action'>
-              <IconButton onClick={() => handleRemoveActionItem(migrationAction)}>
+              <IconButton
+                onClick={() => handleRemoveActionItem(migrationAction)}
+                aria-label={`Remove action ${migrationAction.id}`}
```
PE39806, line 82:
```suggestion
                aria-label={`Move action ${migrationAction.id}`}
```

ARADDCC002: This is to remove an action from the migration plan

PE39806 (dismissed):

frontend/src/schemas/SchemaMigrator.tsx
```diff
@@ -0,0 +1,321 @@
...
+    for (const step of sourceSteps) {
+      step.steps = sourceSteps
+    }
+    setSplitSourceSchema({ reference: sourceSchema.id, steps: sourceSteps })
+
+    for (const step of targetSteps) {
+      step.steps = targetSteps
+    }
+    setSplitTargetSchema({ reference: targetSchema.id, steps: targetSteps })
+  }, [sourceSchema, targetSchema])
```
JR40159, line 62: This might be more a general React convention question but it seems weird to me that here we handle what happens if all the props from the parent component change- ie basically all the content in this component needs updating. Would you not just move the `useEffect` at the parent level?

frontend/src/schemas/SchemaMigrator.tsx
```diff
@@ -0,0 +1,321 @@
...
+    setIsTargetSchemaActive(false)
+  }
+
+  const handleSelectTargetQuestion = () => {
+    setIsSourceSchemaActive(false)
+    setIsTargetSchemaActive(true)
+  }
+
+  const handleSourceQuestionOnClick = (selection: QuestionSelection) => {
+    setErrorText('')
```
JR40159, line 155: To save this line being duplicated, could you just do this once when the add new action button is handled?

JR40159 (changes requested):

frontend/src/schemas/SchemaMigrationSelector.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { SyntheticEvent, useCallback, useState } from 'react'
+import Loading from 'src/common/Loading'
+import MessageAlert from 'src/MessageAlert'
+import SchemaMigrator from 'src/schemas/SchemaMigrator'
+import { SchemaInterface } from 'types/types'
+
+export default function SchemaMigrationSelector() {
+  const { schemas, isSchemasLoading, isSchemasError } = useGetSchemas()
+
+  const [sourceSchema, setSourceSchema] = useState<SchemaInterface | null>()
```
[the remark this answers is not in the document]
ARADDCC002: I've set it to be either defined or undefined. Unfortunately, because the `<Autocomplete />` component that's used to select them can be null, we do need to factor this in.

frontend/src/schemas/SchemaMigrator.tsx
ARADDCC002: We want to remove it when the user selects a new question, as it might be misleading if the error is still there.

frontend/src/schemas/SchemaMigrationSelector.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  const handleTargetSchemaChange = useCallback(
+    (_event: SyntheticEvent, newValue: SchemaInterface | undefined | null) => {
+      setErrorText('')
+      return newValue ? setTargetSchema(newValue) : setTargetSchema(undefined)
+    },
+    [],
+  )
+
+  useEffect(() => {
```
JR40159, line 40: If the whole point of this `useEffect` is to keep these four states in-line with each other, maybe they should be combined so we have two states instead of four?

frontend/src/Form/JsonSchemaViewer.tsx
```diff
@@ -0,0 +1,136 @@
...
+} from 'src/Form/FormTemplates'
+import ValidationErrorIcon from 'src/Form/ValidationErrorIcon'
+import Nothing from 'src/MuiForms/Nothing'
+import QuestionViewer from 'src/MuiForms/QuestionViewer'
+import { SplitSchemaNoRender } from 'types/types'
+import { setStepState } from 'utils/formUtils'
+
+export interface QuestionSelection {
+  path: string
+  schema: any
```
JR40159, line 20: Could we use a type from `@rjsf/utils` here?

frontend/src/MuiForms/QuestionViewer.tsx
```diff
@@ -0,0 +1,44 @@
...
+import dayjs from 'dayjs'
+import customParseFormat from 'dayjs/plugin/customParseFormat'
+import MessageAlert from 'src/MessageAlert'
+dayjs.extend(customParseFormat)
+
+interface QuestionViewProps {
+  label?: string
+  id?: string
+  formContext?: FormContextType
+  schema?: any
```
JR40159, line 14: Could this also be typed?

JR40159 (changes requested):

JR40159 (approved):

Issue #2659, opened 2025-09-04
Kubernetes secrets example config


S010MON:
### To Reproduce

1. Start the application and login
2. Create a minikube cluster that is able to call the Bailo CR
3. Upload an image to the Bailo CR (creating a model, model-card, and artefact)
4. Create a personal access token
5. Select `Kubernetes Secret`
6. Run through config steps 1 and 2 on the k8s cluster (create secret and apply)
7. Update the `manifest.yaml` file for the deployment and `kubectl apply -f manifest.yaml`

### Expected Behavior:
Image pulls successfully

### Actual Behavior: 
When creating the secret the secret name is created using the `token.description` in step 1, however the `configFileName` is subsequently used as the  reference for the `imagePullSecrets` leading to any images failing to pull when adding this to the k8s manifest.

### Provide environment information

```bash
Identified on https://github.com/gchq/Bailo/commit/ab0d3a40c46a83b6dd3d6366abfaad02f31b0127
```

### What is the priority of your issue?

Low

PR #2660, opened 2025-09-04
Gh 2659 kubernetes secrets example config

S010MON:
This PR fixes a bug in the auto-generated example config that Bailo creates when creating a personal access token as a Kubernetes Secret.

When creating the secret the secret name is created using the `token.description` in step 1, however the `configFileName` is subsequently used as the  reference for the `imagePullSecrets` leading to any images failing to pull when adding this to the k8s manifest.

Being a mainly backend dev, please thoroughly check this PR to ensure that it meets your standards!

### DIFF
--- a/backend/src/migrations/026_refactor_tags_to_entry.ts
+++ b/backend/src/migrations/026_refactor_tags_to_entry.ts
@@ -1,0 +1,28 @@
+import ModelModel from '../models/Model.js'
+
+export async function up() {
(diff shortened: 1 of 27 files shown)


S010MON:
closes #2659

JR40159 (dismissed):

JR40159 (approved):

ARADDCC002:
@S010MON This needs to be updated with the latest changes in main and I think is failing on linting. If you're able to update these we can get it merged in.

PR #2681, opened 2025-09-10
Added schema migration API and integrated existing UI

ARADDCC002:


backend/src/services/schemaMigration.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Forbidden } from '../utils/error.js'
+import { handleDuplicateKeys } from '../utils/mongo.js'
+
+export async function createSchemaMigrationPlan(
+  user: UserInterface,
+  schemaMigration: Partial<SchemaMigrationInterface>,
+) {
+  const schemaMigrationDoc = new SchemaMigration(schemaMigration)
+
+  const auth = await authorisation.schemaMigration(user, schemaMigrationDoc, SchemaAction.Create)
```
PE39806, line 14:
```suggestion
  const auth = await authorisation.schemaMigration(user, schemaMigrationDoc, SchemaMigrationAction.Create)
```

backend/src/routes/v2/schema/getSchemaMigration.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const getSchemaMigrationsSchema = z.object({
+  params: z.object({
+    name: z.string().optional().openapi({ example: 'My Schema Migration' }),
+  }),
+})
+
+registerPath({
+  method: 'get',
+  path: '/api/v2/schema-migrations',
```
PE39806, line 20:
```suggestion
  path: '/api/v2/schema-migration',
```
Filename is singular, and `POST` endpoint is also singular.

backend/src/routes/v2/schema/getSchemaMigration.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import { AuditInfo } from '../../../connectors/audit/Base.js'
+import audit from '../../../connectors/audit/index.js'
+import { SchemaMigrationInterface } from '../../../models/SchemaMigration.js'
+import { searchSchemaMigrations } from '../../../services/schemaMigration.js'
+import { registerPath, schemaMigrationInterfaceSchema } from '../../../services/specification.js'
+import { parse } from '../../../utils/validate.js'
+
+export const getSchemaMigrationsSchema = z.object({
+  params: z.object({
```
PE39806, line 13: This could be `query` rather than `params` but I'm open to either.

frontend/types/types.ts
```diff
@@ -652,3 +652,18 @@ export interface QuestionMigration {
...
+
+export interface CombinedSchema {
+  schema: SchemaInterface
+  splitSchema: SplitSchemaNoRender
+}
+
+export interface SchemaMigrationInterface {
+  name: string
+  description?: string
+  questionMigrationss: QuestionMigration[]
```
PE39806, line 664:
```suggestion
  questionMigrations: QuestionMigration[]
```

frontend/src/schemas/SchemaMigrationList.tsx
```diff
@@ -0,0 +1,71 @@
...
+            return { key: migration.name, ...migration }
+          })}
+          emptyListText={'No schema migration plans available'}
+          sortingProperties={[
+            { value: 'createdAt', title: 'Date uploaded', iconKind: 'date' },
+            { value: 'updatedAt', title: 'Date updated', iconKind: 'date' },
+          ]}
+          searchPlaceholderText='Search by name'
+          defaultSortProperty='createdAt'
+          searchFilterProperty='name'
```
PE39806, line 64: This might become slow if a system ends up with many migrations. Can we filter in the backend instead?

ARADDCC002: Unfortunately due to how the component works, it can only do client-side filtering, as it takes an existing array of data. Thankfully, I don't see the list being big enough anyway.

JR40159 (changes requested):

backend/src/routes/v2/schema/getSchemaMigrations.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+  },
+})
+
+interface GetSchemaMigrationsResponse {
+  schemaMigrations: SchemaMigrationInterface[]
+}
+
+export const getSchemaMigrations = [
+  bodyParser.json(),
```
JR40159, line 43: No longer needed

backend/src/routes/v2/schema/postSchemaMigration.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+  },
+})
+
+interface PostSchemaMigrationPlanResponse {
+  schemaMigration: SchemaMigrationInterface
+}
+
+export const postSchemaMigration = [
+  bodyParser.json(),
```
JR40159, line 58: No longer needed

JR40159 (changes requested):

JR40159 (approved):

PR #2691, opened 2025-09-15
added queing logic

ARADDCC002:


backend/src/services/mirroredModel/mirroredModel.ts
```diff
@@ -383,19 +386,9 @@ export async function uploadReleaseFiles(
...
-          modelId: model.id,
-          releaseSemver: release.semver,
-          fileId: file.id,
-          mirroredModelId,
-        },
-        'Error when uploading Release File to S3.',
-      )
-    }
+      ),
+    )
```
PE39806, line 390: The `try/catch` functionality can be reinstated by utilising `queue.add(...).catch((error) => {log.error(...)}`. This would allow safe error reporting in an async manner.

backend/src/services/mirroredModel/mirroredModel.ts
```diff
@@ -483,7 +477,9 @@ async function addReleaseToTarball(
   }
 
   // Fire-and-forget upload of artefacts so that the endpoint is able to return without awaiting lots of uploads
-  uploadReleaseFiles(user, model, release, files, mirroredModelId)
+  log.debug({ semver: release.semver }, 'Adding files to be exported to queue')
+  await uploadReleaseFiles(user, model, release, files, mirroredModelId, queue)
+  log.debug({ semver: release.semver }, 'Finished adding files to be exported to queue')
   uploadReleaseImages(user, model, release, mirroredModelId)
```
PE39806, line 483: Could images also utilise a queue? Likely the same one as the bottleneck is the s3 upload.

PR #2706, opened 2025-09-24
increase backend limit to pull more than 100 images

JRB66955:


PE39806 (commented):
I was originally anticipating this only being used for the `GET /v2/_catalog` call but your work is much better as it also covers pagination for `GET /v2/<name>/tags/list`.
It looks like you're correctly implementing [the spec](https://github.com/distribution/distribution/blob/4ead5a108593cec429ec57d0eb698d5825e037d0/docs/content/spec/api.md#pagination), but could you add an optional `bool` param which would allow for _not_ making repeated requests for the subsequent `Lik`s? (e.g. `traverseLinks: bool = true`)? I can't imagine there are many times we would need this but it might be nice to have.

PE39806 (dismissed):

backend/src/clients/registry.ts
```diff
@@ -44,67 +44,94 @@ async function registryRequest(
...
-    throw InternalError('Unable to communicate with the registry.', { err })
-  }
-
-  const headersObject = res.headers ? Object.fromEntries(res.headers) : {}
-  const contentType = res.headers.get('content-type') || ''
 
-  let body: unknown
+  const allRepositories: string[] = []
+  let paginateParameter = ''
+  let headersObject: {
```
JR40159, line 53: Could this be done better?

JR40159 (changes requested):

backend/src/clients/registry.ts
```diff
@@ -44,67 +44,94 @@ async function registryRequest(
   returnRawBody: boolean = false,
   extraFetchOptions: Partial<Omit<RequestInit, 'headers' | 'dispatcher' | 'signal'>> = {},
   extraHeaders: HeadersInit = {},
+  traverseLinks: boolean = true,
 ): Promise<RegistryRequestResult> {
```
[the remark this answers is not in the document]
JR40159: Don't do - would lead to a much larger refactor

backend/src/clients/registry.ts
JRB66955: I don't like this block of variable definitions either, but I've not thought of a nicer way? The loop requires them to be defined upfront to allow them to be (re-)assigned for each loop/request. I've shuffled some of the variables around to more explicitly outline their scope at least.

You could condense the actual definition to something like ` let link: string | null, res: Response, body: any, stream: ReadableStream | Readable | undefined` on a single line, but I don't like doing that either.

PE39806 (approved):
I'm happy with this but check with @JR40159 before merging.

PR #2707, opened 2025-09-24
Queue Scans to avoid timeout Error when handling many large files

PE39806:
Add `BaseQueueFileScanningConnector` for queued scans and update `ClamAvFileScanningConnector` and `ModelScanFileScanningConnector` to use this.

Modified files currently have no tests hence no additional test coverage.

### DIFF
--- a/backend/config/default.cjs
+++ b/backend/config/default.cjs
@@ -184,11 +184,13 @@ module.exports = {
 
   avScanning: {
     clamdscan: {
+      concurrency: 4,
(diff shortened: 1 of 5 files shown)


JR40159 (changes requested):
What's the reason for not just having a single queue in the wrapper? That would have the additional protection from the backend from being overwhelmed by running different scanners simultaneously as well as eliminating the need for another base class.

JR40159 (approved):

PR #2714, opened 2025-09-26
Move review logic to backend

JR40159:


ARADDCC002 (dismissed):

backend/src/services/review.ts
```diff
@@ -61,6 +61,16 @@ export async function findReviews(
     )
   }
 
+  if (open === false) {
+    const parentIds = reviews.map((review) => review._id)
+    const responses = await getResponsesByParentIds(parentIds)
+    reviews = reviews.filter((review) =>
+      responses.find(
+        (response) => response.entity === `user:${user.dn}` && response.parentId.toString() === review._id.toString(),
```
JR40159, line 69: Should this filter be updated so that it looks to see if anyone has responded to the review, not the specific user?

frontend/utils/fetcher.ts
```diff
@@ -14,26 +14,29 @@ type ErrorResponse = {
   error: Error
 }
 
-export const fetcherHeaders = async (url: string) => {
-  const res = await fetch(url, { method: 'head' })
+export const fetcher = async <TMethod extends 'head' | 'get'>(
```
JR40159, line 17: Simplify to Boolean here

backend/src/services/review.ts
JR40159: Could this then be used for the review status of releases and access requests?

PR #2774, opened 2025-10-06
Implements resumable downloads

### LINKED ISSUE #2585: Resumable downloads
### Problem

Content range headers are not supported when downloading files.

### Proposed Solution

Add support for fetching parts of files

### Alternatives

N/A

Medium

GCHQDeveloper36:
Will close #2585.

I've done some initial testing with curl & Chrome in a couple of scenarios:

- Create a 1GB random file `dd if=/dev/urandom of=random.bin bs=1G count=1`
- Generate a hash `sha256sum random.bin > random.bin.sha256`
- Upload through UI to Bailo
- Download file as normal - verify hash is as expected
- Using curl, make a HEAD request, validate ETag and range headers are provided `curl -I localhost:8080/api/v2/model/....`
- Using curl, make a GET request but provide a Range header e.g. `curl -H 'Range: bytes=0-100' localhost:8080... -o part1.bin`
- Do the same, but the remaining bytes: `curl -H 'Range: bytes=101-' localhost:8080... -o part2.bin`
- Concat & hash the resulting file: `cat part1.bin part2.bin > complete.bin`, `sha256sum complete.bin`
- This should match the original hash

With Chrome, particularly with a 1GB file, simply pausing and resuming works as expected. Additionally, if you start a download, and trigger a service restart (e.g. `docker compose restart backend`, the download will be interrupted. Once it's finished restarting, you can resume as normal. This requires the ETag header, otherwise it will start downloading the whole file from scratch again.

### DIFF
--- a/backend/src/routes/v2/model/file/getDownloadFile.ts
+++ b/backend/src/routes/v2/model/file/getDownloadFile.ts
@@ -7,54 +7,90 @@ import { z } from 'zod'
 
 import { AuditInfo } from '../../../../connectors/audit/Base.js'
 import audit from '../../../../connectors/audit/index.js'
-import { FileInterface, FileWithScanResultsInterface } from '../../../../models/File.js'
+import { FileWithScanResultsInterface } from '../../../../models/File.js'
 import { downloadFile, getFileById } from '../../../../services/file.js'
+import log from '../../../../services/log.js'
 import { getFileByReleaseFileName } from '../../../../services/release.js'
 import { registerPath } from '../../../../services/specification.js'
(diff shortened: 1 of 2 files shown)


PE39806 (dismissed):
Amazing work, thank you. I'd like to see unit tests added but I also can see that this is all working so am happy with it as is.

GCHQDeveloper36:
Went through all the API docs for that endpoint and documented it as extensively as I can. I also added support for 304 (not modified) which might be useful in some cases, and some general additional error handling (happy to take feedback on that). If you just want to stick with the API docs addition for now then I can revert some of the newer changes.

PR #2776, opened 2025-10-07
Revert "Image export prefetch registry layers"

JRB66955:
Reverts gchq/Bailo#2713


PE39806 (changes requested):
TLDR: only revert the changes to `backend/src/services/mirroredModel/mirroredModel.ts` and the associated test file (but even then keep the `import ... from 'node:stream'` changes).

PR #2796, opened 2025-10-14
Implement schema migration API

ARADDCC002:


backend/src/services/schemaMigration.ts
```diff
@@ -1,15 +1,20 @@
...
+import { BadReq, Forbidden } from '../utils/error.js'
+import { convertStringToId } from '../utils/id.js'
 import { handleDuplicateKeys } from '../utils/mongo.js'
 
 export async function createSchemaMigrationPlan(
   user: UserInterface,
   schemaMigration: Partial<SchemaMigrationInterface>,
 ) {
-  const schemaMigrationDoc = new SchemaMigration(schemaMigration)
+  const migrationId = convertStringToId(schemaMigration.name || '')
```
PE39806: I don't think `migrationId` should default to `''`.

frontend/src/schemas/SchemaMigrationList.tsx
```diff
@@ -13,13 +14,23 @@ export default function SchemaMigrationList() {
...
         </Typography>
       </Stack>
+      <Stack direction='row' alignItems='center'>
+        <Typography fontWeight='bold' color='primary'>
+          {data[index].id}
+        </Typography>
+        <CopyToClipboardButton
+          textToCopy={data[index].id}
+          notificationText='Copied access request ID to clipboard'
+          ariaLabel='copy access request ID to clipboard'
```
PE39806:
```suggestion
          notificationText='Copied migration ID to clipboard'
          ariaLabel='copy migration ID to clipboard'
```

PE39806 (changes requested):

IR96334:
I would suggest limiting the title length of the "source question" and "destination question" buttons.
Seems a bit unnecessary to display the entire question, especially when the questions are 100 characters plus.

frontend/src/schemas/SchemaMigrationList.tsx
ARADDCC002: Copy and paste at it again...

backend/src/utils/object.ts
```diff
@@ -13,3 +13,12 @@ export function deepFreeze(object) {
...
 }
+
+export function getPropValue(sourceObject: any, dotNotationPath: string) {
+  let returnData = sourceObject
+
+  dotNotationPath.split('.').forEach((subPath) => {
+    returnData = returnData[subPath] || undefined
+  })
+  return returnData
+}
```
PE39806: I've played with this some more and it doesn't differentiate between null and undefined. Here's a possible fix which also avoids `any`:
```suggestion
export function getPropValue<T = unknown>(
  source: unknown,
  path: string
): T | undefined {
  const trimmedPath = path.trim()
  if (!trimmedPath) {
    return source as T
  }

  return trimmedPath.split('.').reduce<any>((acc, key) => {
    return acc != null ? acc[key] : undefined
  }, source)
}
```

frontend/src/entry/overview/MigrationListDialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+type MigrationListDialogProps = {
+  open: boolean
+  migrations: SchemaMigrationInterface[]
+  onCancel: () => void
+  entry: EntryInterface
+}
+
+export default function MigrationListDialog({ open, migrations, onCancel, entry }: MigrationListDialogProps) {
+  const [selectedMigrationPlan, setSelectMigrationPlan] = useState<string | undefined>()
```
PE39806:
```suggestion
  const [selectedMigrationPlan, setSelectMigrationPlan] = useState<SchemaMigrationInterface['id'] | undefined>()
```

PE39806 (approved):

PE39806 (approved):

PR #2798, opened 2025-10-14
Move mirror metadata to `tar.gz` file

### LINKED ISSUE #2449: Improve stream handling

PE39806:
Replace the existing upload metadata with a JSON entry containing the metadata in the existing `tar.gz` functionality. The JSON entry _must_ be the first entry added to the `tar.gz` as `tar.gz` files preserve ordering, meaning that the first entry in is the first entry out (allowing reading the metadata JSON entry before any other files).

This allows for more flexibility with custom metadata and allows for greater flexibility with character sets by allowing developers to specify encoding.

As well as improving metadata handling, this unifies all imports and exports to be `tar.gz` files which reduces some branching conditions.

Also closes #2449

### DIFF
--- a/backend/src/services/mirroredModel/exporters/base.ts
+++ b/backend/src/services/mirroredModel/exporters/base.ts
@@ -12,6 +12,17 @@ import { BadReq, Forbidden, InternalError } from '../../../utils/error.js'
 import { MirrorLogData } from '../mirroredModel.js'
 import { finaliseTarGzUpload, initialiseTarGzUpload } from '../tarball.js'
 
+/**
+ * Wraps any thrown error with additional execution context information.
+ *
+ * - If `err` is a BailoError, appends method identification to the message and rethrows it.
+ * - If not, wraps it into an `InternalError` with method details and relevant `BaseExporter` log data.
+ *
+ * @param err - The caught error or thrown value.
+ * @param context - Method decorator context containing method metadata.
+ * @param self - The calling `BaseExporter` instance.
+ * @throws InternalError | BailoError
+ */
 function wrapErrorContext(err: unknown, context: ClassMethodDecoratorContext, self: BaseExporter): never {
   if (isBailoError(err)) {
     err.message += `\nMethod \`${String(self.constructor.name)}.${String(context.name)}\` failure.`
@@ -23,6 +34,15 @@ function wrapErrorContext(err: unknown, context: ClassMethodDecoratorContext, se
   })
 }
 
+/**
+ * Method decorator.
+ * Ensures that the decorated method can only run after `init()` has been successfully invoked.
+ *
+ * @typeParam M - Method type extending a function bound to `BaseExporter`.
+ * @param value - The original method implementation.
+ * @param context - Method decorator metadata.
+ * @throws InternalError - If called before the instance is initialised.
+ */
 export function requiresInit<M extends (this: BaseExporter, ...args: any[]) => any>(
   value: M,
   context: ClassMethodDecoratorContext,
@@ -38,6 +58,15 @@ export function requiresInit<M extends (this: BaseExporter, ...args: any[]) => a
   } as M
 }
 
+/**
+ * Method decorator.
+ * Invokes `checkAuths()` before executing the decorated method.
+ * If `checkAuths()` fails, wraps the error with additional method context information.
+ *
+ * @typeParam M - Method type extending a function bound to `BaseExporter`.
+ * @param value - The original method implementation.
+ * @param context - Method decorator metadata.
+ */
 export function checkAuths<M extends (this: BaseExporter, ...args: any[]) => any>(
   value: M,
   context: ClassMethodDecoratorContext,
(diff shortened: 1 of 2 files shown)


backend/src/services/mirroredModel/tarball.ts
```diff
@@ -0,0 +1,213 @@
...
+import {
+  ExportMetadata,
+  FileImportInformation,
+  ImageImportInformation,
+  ImportKind,
+  MongoDocumentImportInformation,
+} from './mirroredModel.js'
+import { uploadToS3 } from './s3.js'
+
+export function createTarGzStreams() {
```
JR40159: `Export` for internal function?

backend/test/routes/schema/getSchema.spec.ts
```diff
@@ -8,6 +8,7 @@ import { testModelSchema } from '../../testUtils/testModels.js'
 vi.mock('../../../src/utils/user.js')
 vi.mock('../../../src/connectors/audit/index.js')
 vi.mock('../../../src/connectors/authorisation/index.js')
+vi.mock('../../../src/services/mirroredModel/tarball.ts', () => ({}))
```
JR40159: Why are changes like this one needed?

PE39806: `backend/src/services/mirroredModel/tarball.ts` imports `backend/src/services/mirroredModel/importers/documentsImporter.ts` which has ``const modelCardRegex = new RegExp(String.raw`^${config.modelMirror.contentDirectory}/[0-9]+\.json$`)``. The `config` key isn't mocked so the import errors, resulting in the test file failing. `tarball.ts` imports several files which all have this same problem so it is easier to mock `tarball.ts` than each sub-import. The extra `() => ({})` is required to stop the additional imports from being attempted.

backend/src/services/mirroredModel/tarball.ts
```diff
@@ -0,0 +1,207 @@
...
+  // It is safer to have an extra PassThrough for handling backpressure and explicitly closing on error(s)
+  const uploadStream = new PassThrough()
+  tarStream.pipe(gzipStream).pipe(uploadStream)
+
+  const uploadPromise = uploadToS3(fileName, uploadStream, logData)
+
+  try {
+    const metadataBuffer = Buffer.from(JSON.stringify(metadata), 'utf8')
+    log.debug(
+      { metadata, name: config.modelMirror.metadataFile, size: metadataBuffer.length },
```
JR40159: `logData`?

backend/src/services/mirroredModel/tarball.ts
```diff
@@ -0,0 +1,207 @@
...
+  tarStream.finalize()
+  await uploadPromise
+}
+
+type TarEntry =
+  | { type: 'text'; filename: string; content: string }
+  | { type: 'stream'; filename: string; stream: Readable; size?: number }
+export async function addEntryToTarGzUpload(tarStream: Pack, entry: TarEntry, logData?: Record<string, unknown>) {
+  const entryName = `${config.modelMirror.contentDirectory}/${entry.filename}`
+  log.debug({ entryName, entry }, 'Adding entry to tarball.')
```
JR40159: `logData`?

backend/src/services/mirroredModel/mirroredModel.ts
```diff
@@ -309,253 +201,82 @@ export async function exportCompressedRegistryImage(
...
-            releaseId: release.id,
-            fileId: file.id,
-          },
-        ),
-      )
-      .catch((error) =>
+    // Not `await`ed for fire-and-forget approach
+    exportQueue
+      .add(async () => {
+        const fileExporter = new FileExporter(user, model, file, logData)
```
JR40159: Could we use the standard pattern of calling the `init()` on the constructor in a single line like 
https://github.com/gchq/Bailo/blob/main/backend/src/connectors/fileScanning/clamAv.ts#L36

JR40159 (dismissed):

PR #2801, opened 2025-10-15
Update compose to use published images

GCHQDeveloper36:
Updated the base compose file to use the published modelscan image since most of the time that should be the default. If you're developing the modelscan container, you can just add `-f compose.modelscan.yaml` to tell compose it should build and use that image.

By default, docker compose up/build/pull will use the published modelscan image, and the 'local' dev containers for frontend/backend.

If you add `-f compose.modelscan.yaml` it will also use the 'local' dev container for modelscan.

I also updated the prod compose file to reference the (current) published images for frontend/backend.

This also fixes the current main revision where if you do `compose pull` it will not find the bailo containers (understandably), and should instead say 'skipped'.

### DIFF
--- a/compose.modelscan.yaml
+++ b/compose.modelscan.yaml
@@ -1,0 +1,11 @@
+name: bailo-prod
+
+services:
+  modelscan:
+    volumes:
+      - ./lib/modelscan_api/bailo_modelscan_api:/app/bailo_modelscan_api
+    image: bailo_modelscan:${INSTANCE_NAME}
+    pull_policy: build
+    build:
+      context: ./lib/modelscan_api
+      target: dev
--- a/compose.override.yaml
+++ b/compose.override.yaml
@@ -26,18 +26,16 @@ services:
       - ./backend/src:/app/src
     build:
       target: dev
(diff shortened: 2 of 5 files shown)


PE39806 (changes requested):
Since the current `compose.prod.yaml` builds the images differently to `compose.override.yaml` (the multistage build has a different layer target), I think we need another compose file for the pulled images as otherwise this PR would remove that true-to-prod "local" build which is useful for testing. Perhaps a `compose.release.yaml` would be a good place to have the pulled images?

I'm not 100% sure on these thoughts so any discussion is very welcome.

PE39806:
Another thought - we still need some way to test functionality via GitHub Actions which currently uses the local build for all images in `compose.prod.yaml`. This functionality needs to be kept for things like the cypress tests, or the GitHub Actions will require updating to still use the locally built version rather than pulling from GHCR.

JR40159 (changes requested):
It would be good to discuss the pros/cons of this. Using a local build for the docker compose files can be very useful for local development.

PR #2822, opened 2025-10-28
Update Image Import to not repeatedly check user

PE39806:
Avoid `checkUserAuth` for image import, similar to file import, so that proc users can be used for image imports.

### DIFF
--- a/backend/src/services/registry.ts
+++ b/backend/src/services/registry.ts
@@ -1,15 +1,4 @@
-import { Readable } from 'node:stream'
-
-import {
-  doesLayerExist,
-  getImageTagManifest,
-  getRegistryLayerStream,
-  initialiseUpload,
-  listImageTags,
-  listModelRepos,
-  putManifest,
-  uploadLayerMonolithic,
-} from '../clients/registry.js'
+import { getImageTagManifest, getRegistryLayerStream, listImageTags, listModelRepos } from '../clients/registry.js'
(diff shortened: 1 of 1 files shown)


backend/src/services/mirroredModel/importers/image.ts
```diff
@@ -108,15 +131,7 @@ export class ImageImporter extends BaseImporter {
               },
               'Putting image blob.',
             )
-            await putImageBlob(
```
ARADDCC002: Are we using this function anywhere else? Locally I can't see it being used other than here. If we no longer need it we need to consider removing it unless this is likely to be used in the future.

backend/src/services/mirroredModel/importers/image.ts
```diff
@@ -143,10 +158,18 @@ export class ImageImporter extends BaseImporter {
   async finishListener(resolve: (reason?: ImageMirrorInformation) => void, reject: (reason?: unknown) => void) {
     log.debug({ ...this.logData }, 'Uploading manifest.')
     if (hasKeysOfType<{ mediaType: 'string' }>(this.manifestBody, { mediaType: 'string' })) {
-      await putImageManifest(
```
ARADDCC002: Same as before, do we need to remove this function

ARADDCC002 (changes requested):

PR #2882, opened 2025-11-06
Fix an issue where documents with no deleted property were not being retrieved from DB

ARADDCC002:


JRB66955 (dismissed):

backend/src/models/plugins/softDeletePlugin.ts
```diff
@@ -45,7 +45,7 @@ export function softDeletionPlugin(schema: Schema) {
       this.where('deleted').equals(this['_conditions'].deleted)
       next()
     } else {
-      this.where('deleted').equals(false)
+      this.or([{ deleted: null }, { deleted: false }])
```
PE39806:
```suggestion
      this.or([{ deleted: { $exists: false } }, { deleted: false }])
```

PE39806 (changes requested):
Improve backwards compatibility

PE39806 (approved):

PR #2888, opened 2025-11-07
New endpoint to allow admins to get all access requests

JRB66955:
<del> TODO: </del>

<del> - work out api routing issue with middleware </del>
<del> - tests tbd </del>

### DIFF
--- a/backend/src/connectors/authorisation/base.ts
+++ b/backend/src/connectors/authorisation/base.ts
@@ -288,6 +288,7 @@ export class BasicAuthorisationConnector {
         }
 
         if (
+          !isNamed &&
           !(await this.model(user, model, ModelAction.View)).success &&
           ([AccessRequestAction.View] as AccessRequestActionKeys[]).includes(action)
         ) {


backend/src/routes/v2/model/accessRequest/getAccessRequests.ts
JR40159: Could we refactor this to be a search endpoint to be consistent with the model search endpoint?

backend/test/services/accessRequest.spec.ts
```diff
@@ -135,6 +164,46 @@ describe('services > accessRequest', () => {
...
+    const accessRequests = await findAccessRequest({} as any, [], '', [], false)
+    expect(accessRequests).toMatchSnapshot()
+  })
+
+  test('findAccessRequest > all filters', async () => {
+    mockAuthentication.hasRole.mockReturnValueOnce(false)
+    accessRequestModelMocks.find.mockResolvedValue([])
+    modelModelMocks.aggregate.mockResolvedValueOnce([])
+
+    const accessRequests = await findAccessRequest({} as any, ['modelId'], 'schemaId', ['filter'], false)
```
IR96334: maybe include 'mine' in filter to increase coverage

backend/src/services/accessRequest.ts
```diff
@@ -131,6 +134,84 @@ export async function getAccessRequestById(user: UserInterface, accessRequestId:
...
+  }
+
+  const query: any = {}
+
+  if (modelId.length) {
+    query.modelId = { $all: modelId }
+  }
+
+  if (schemaId) {
+    query.schemaId = { $all: schemaId }
```
PE39806: I'm confused by this use of `$all` as `schemaId` is a `string`.

backend/src/services/accessRequest.ts
```diff
@@ -131,6 +134,84 @@ export async function getAccessRequestById(user: UserInterface, accessRequestId:
...
+        userDn: user.dn,
+        requiredRole: Roles.Admin,
+      })
+    }
+  }
+
+  const query: any = {}
+
+  if (modelId.length) {
+    query.modelId = { $all: modelId }
```
PE39806: Jus to check, is this definitely `$all` rather than `$in`?

PE39806 (dismissed):

backend/src/services/accessRequest.ts
```diff
@@ -131,6 +133,71 @@ export async function getAccessRequestById(user: UserInterface, accessRequestId:
...
+    {
+      $unwind: '$model',
+    },
+  ]
+
+  const results = await AccessRequestModel.aggregate(stages)
+  //Auth already checked, so just need to check if they require admin access
+  if (adminAccess) {
+    return results
+  }
```
GB907762: Because you've grouped by model id, using your aggregate query, you can't return results directly as you'll get something like this. 
```
{
  "accessRequests": [
    {
        _id: 1
        accessRequests: {
             ...
        }
        ....
```
You can get around this by ignoring the last 3 pipeline stages.

backend/src/services/accessRequest.ts
JRB66955: Updated to $in now

backend/src/connectors/authorisation/base.ts
```diff
@@ -287,6 +287,17 @@ export class BasicAuthorisationConnector {
           return { success: false, info: 'You cannot change an access request you do not own.', id: request.id }
```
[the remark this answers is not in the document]
JR40159: False alarm, a problem with my local instance

JR40159 (changes requested):

PR #2891, opened 2025-11-07
Dev/refactor model search

### LINKED ISSUE #2803: Unexpected search results based on tokenisation
### To Reproduce

1. Create a model with name example/mybluemodel-123
2. Create a model with a completely different name, add a model card, add some text including 'mybluemodel' somewhere.
3. Search from the marketplace for 'example' - see 123 model listed
4. Search for 'blue' - no model listed
5. Search for 'mybluemodel' - both are listed, typically with the second model listed first.

### Current vs. Expected behavior

Users expected to see partial text matches in their results. Additionally, ranking of results would be expected to be based on the following priority:

1. model name
2. model description
3. model card

### Provide environment information

```bash
Bailo v2.8.1
```

### Additional context

_No response_

High

JR40159:
Closes #2803


ARADDCC002 (approved):

backend/src/services/model.ts
```diff
@@ -263,19 +256,45 @@ async function searchLocalModels(user: UserInterface, opts: EntrySearchOptionsPa
...
+        score: { $meta: 'textScore' },
+      }),
+    )
   }
 
-  const results = await cursor
+  // Remove duplicate items
+  const seen = new Set()
+
+  results = results.filter((item) => {
```
JR40159: Move the logic that preserves the array uniqueness into the if statement for the full text search

PE39806 (approved):

PR #2925, opened 2025-11-21
Enable delete entry functionality

PE39806:
- [x] Implement Soft Delete model code 
- [x] Enable delete entry button
- [x] Hook up to the v2 delete entry API
- [x] Address outstanding TODO in `frontend/src/entry/settings/DangerZone.tsx`
- [x] Add unit tests
- [x] Update Python client
- [x] Add integration tests

### DIFF
--- a/frontend/src/entry/settings/DangerZone.tsx
+++ b/frontend/src/entry/settings/DangerZone.tsx
@@ -78,7 +78,13 @@ export default function DangerZone({ entry }: DangerZoneProps) {
           <Button onClick={() => setOpenConfirm(false)} disabled={loading}>
             Cancel
           </Button>
-          <Button color='error' variant='contained' loading={loading} disabled={confirmInput.trim() !== entry.name}>
+          <Button
+            color='error'
+            variant='contained'
+            onClick={handleDeleteEntry}
+            loading={loading}
+            disabled={confirmInput.trim() !== entry.name}
+          >
             Delete
           </Button>
         </DialogActions>


frontend/src/entry/settings/DangerZone.tsx
```diff
@@ -9,22 +14,74 @@ type DangerZoneProps = {
 
 export default function DangerZone({ entry }: DangerZoneProps) {
```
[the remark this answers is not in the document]
ARADDCC002: Likely a mistake - mirrored models should also definitely be able to be deleted. There's nothing stopping someone from creating a new one, and just changing the `targetId` in the source model if they require a new mirrored model for some reason.

backend/src/models/plugins/softDeletePlugin.ts
```diff
@@ -25,6 +25,82 @@ export function softDeletionPlugin(schema: Schema) {
...
+    const update: Record<string, any> = {
+      deleted: true,
+      deletedAt: new Date().toISOString(),
+    }
+    if (user) {
+      update.deletedBy = user
+    }
+
+    const options = session ? { session } : {}
+    return this.updateOne(filter, update, options)
```
ARADDCC002: Is there a reason this is on two lines and not just:

`return this.updateOne(filter, update, { session })`

frontend/src/entry/settings/DangerZone.tsx
```diff
@@ -9,22 +14,75 @@ type DangerZoneProps = {
...
+      sendNotification({
+        variant: 'success',
+        msg: `${toTitleCase(EntryKindLabel[entry.kind])} deleted`,
+        anchorOrigin: { horizontal: 'center', vertical: 'bottom' },
+      })
+      router.push('/')
+    }
+
+    setLoading(false)
+    setOpenConfirm(false)
```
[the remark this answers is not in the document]
IR96334: If passes: I don't know if it matters as you are routed to different page.
If fails: Would you want to stay on dialog with the error message?

frontend/src/entry/settings/DangerZone.tsx
```diff
@@ -9,22 +14,75 @@ type DangerZoneProps = {
...
+            type='submit'
+            onClick={handleDeleteEntry}
+            loading={loading}
+            disabled={confirmInput.trim() !== entry.name}
+          >
+            Delete
+          </Button>
+        </DialogActions>
+      </Dialog>
+      <MessageAlert message={errorMessage} severity='error' />
```
IR96334: Do we need error here and in dialog?

frontend/src/entry/settings/DangerZone.tsx
```diff
@@ -9,22 +14,75 @@ type DangerZoneProps = {
...
+
+    if (!response.ok) {
+      setErrorMessage(await getErrorMessage(response))
+    } else {
+      sendNotification({
+        variant: 'success',
+        msg: `${toTitleCase(EntryKindLabel[entry.kind])} deleted`,
+        anchorOrigin: { horizontal: 'center', vertical: 'bottom' },
+      })
+      router.push('/')
```
ARADDCC002: At this point I'd expect to be routed to the marketplace, so I'm not sure why this isn't working as I'd expect `('/')` to work

ARADDCC002 (dismissed):

frontend/src/entry/settings/DangerZone.tsx
PE39806: Strange, I've not managed to run into this. What did the model look like, and what other things were on the model to cause this?

IR96334 (changes requested):

ARADDCC002 (approved):

PR #2933, opened 2025-11-25
Tag selector fixes

IR96334:


frontend/src/MuiForms/TagSelector.tsx
```diff
@@ -77,16 +79,27 @@ export default function TagSelector({ onChange, value, label, formContext, requi
...
-                  ))}
-                </Box>
+                {value.length === 0 ? (
+                  <Typography
+                    sx={{
+                      fontStyle: 'italic',
+                      color: theme.palette.customTextInput.main,
+                    }}
+                  >
+                    No tags
```
ARADDCC002: I've changed my mind on this... because this _could_ be used in the future by the JSON schema it might be worth replacing this with:

`formContext && formContext.emptyPlaceholderText ? formContext.emptyPlaceholderText : 'No tags'`

That way, if we want to use it for forms we can set the text that way, as the question its used for might not be called "tags" but still use this component.

ARADDCC002 (approved):

PR #2978, opened 2025-12-01
Flaky A11y Fix

JRB66955:


ARADDCC002 (changes requested):
Instead of getting by `h1` we should supply a `data-test` property to the text we want to search for.

PR #2987, opened 2025-12-01
Show unreachable, enabled peers

IR96334:


frontend/pages/index.tsx
```diff
@@ -196,11 +196,11 @@ export default function Marketplace() {
     [roleOptions],
   )
 
-  const reachablePeerList: { key: string; value: boolean }[] = useMemo(() => {
+  const unreachablePeerList: string[] = useMemo(() => {
     if (!peers) return []
     return Array.from(peers.entries())
-      .filter(([, value]) => isEnabled(value))
-      .map(([key, value]) => ({ key, value: isReachable(value) }))
+      .filter(([, value]) => isEnabled(value) && !isReachable(value))
```
ARADDCC002: minor, but I think we should change it to `[_key, value]`. Seems more readable to me personally.

PR #3341, opened 2026-03-11
Run GitHub Actions on all PRs

PE39806:
- Add manual step to all relevant GitHub Actions
- Enforce GH Actions on all PRs (not just to `main`)


GB907762 (commented):
This is a nice idea. Although, I'm not convinced with running actions where the target branch is not `main`? 

I think if you're merging into a feature branch these should run at the point changes are committed to `main`?

PR #3371, opened 2026-03-24
Remove duplicate state and Improve EntryAccessInput

JR40159:
1. Remove duplication in the use of a separate state for default entities
2. Remove duplication and re-use entity display from model access also in the default collaborators for review roles.
3. Improve entity autocomplete to clear where an entity is selected.
4. Add missing swagger docs to get model roles endpoint using `z.infer` to avoid duplicating type and Zod specification.
5. Fix incorrect type in the frontend for Roles and remove duplicate name.
-----
6. Update Autocomplete to use multiple mode to fix issue with input not clearing
7. Isolate state inside the edit component so that the draft form is captured
8. Fix broken cancel broken on create review role
9. Reduce duplication of handler functions
10. Improve use of generics


PE39806 (dismissed):

GB907762 (changes requested):

GB907762 (dismissed):

ON93754 (dismissed):

IR96334 (dismissed):

ARADDCC002 (approved):

ARADDCC002 (dismissed):
Adding a new user to the model access list seems to remove all roles from current users.

IR96334 (approved):

ON93754 (approved):

PR #3360, opened 2026-03-20
Add multi-platform image scan support

GB907762:
A multi-platform image is an image with a single invocation but multiple targets. 

See here: https://docs.docker.com/build/building/multi-platform/

PR #3410, opened 2026-04-01
Delete registry images with manifest lists

PE39806:
## Changes

Backend

- Add support for soft deleting images which have manifest lists/fat manifests/multi-platform manifests
- Copies in some changes from #3404 for testing and enabling functionality
- Some typing improvements

Frontend

- Add delete image button and popup
- Add associated releases for image button and popup
- Improve delete image and file `Button` to use `loading` property to prevent multiple clicks while deletion is in progress.

## Testing snippets

To _stream_ a multiplatform image (e.g. alpine) to the Bailo registry:
```bash
IMAGE_NAME="alpine"
IMAGE_TAG="3.23.3"
MODEL_ID=<your-model-id>
docker buildx imagetools create --tag "localhost:8080/$MODEL_ID/$IMAGE_NAME:$IMAGE_TAG" "$IMAGE_NAME:$IMAGE_TAG"
```
Note that this will use a considerable number of API calls/requests so be aware of things like https://docs.docker.com/docker-hub/usage/

## TODO

- [x] Align registry methods with #3360
- [x] Unit tests
- [x] Consolidate image rename to not be big `if/else`
- [x] Optimise/reduce number of registry requests


backend/src/services/registry.ts
```diff
@@ -376,9 +401,168 @@ export async function renameImage(user: UserInterface, source: ImageTagRef, dest
...
+          { repository: source.repository, name: source.name },
+          { repository: destination.repository, name: destination.name },
+          layer.digest,
+        )
+      }
+
+      // A temp tag is required because PUT-by-digest fails (the JSON is re-serialised, changing its canonical bytes)
+      // so we create the digest via a tag and then delete the tag (leaving the new digest)
+      const tmpTag = `__bailo_tmp_${crypto.randomUUID()}__`
+      // NOTE: Temp tags may remain if process crashes before cleanup.
```
JR40159, line 450: try-catch-**finally**?

JR40159 (changes requested):

backend/src/services/registry.ts
JR40159: Just seen you're already doing this :) Do we need a different temp tag for each manifest or could we reuse one?

PE39806: Each manifest is unique so needs their own unique temporary tag.

backend/src/clients/registry.ts
JR40159, line 1: Should the removal of the images from releases be wrapped in a transaction?

JR40159 (changes requested):

JR40159 (approved):

JRB66955 (approved):

PR #3454, opened 2026-04-07
Don't cache file download requests

JR40159:


GB907762 (changes requested):

GB907762 (approved):

backend/src/routes/v2/model/file/getDownloadFile.ts
```diff
@@ -18,7 +18,7 @@ import { parseRangeHeaders } from '../../../../utils/range.js'
 import { parse } from '../../../../utils/validate.js'
 
 // Default cache response header
-const cacheControl = 'public, max-age=604800, immutable'
+const cacheControl = 'no-store'
```
PE39806: Prefer:
```suggestion
const cacheControl = 'private, no-cache, must-revalidate'
```
- `private` prevents shared caches (CDNs, proxies) from storing the response, but allows the user’s browser to cache (something the user can clear if necessary)
- `no-cache` allows the browser to store the file but requires it to revalidate with the server (via ETag/If-None-Match) before reuse, preventing stale/cached downloads
- `must-revalidate` instructs caches that once the response is stale, they must successfully revalidate with the origin before serving it (avoiding use of outdated content)

PE39806 (changes requested):

backend/src/routes/v2/model/file/getDownloadFile.ts
JR40159: - private prevents shared caches (CDNs, proxies) from storing the response, **but allows the user’s browser to cache (something the user can clear if necessary)**. - Caching by the browser is the things that's causing issues.
- no-cache allows the browser to store the file but requires it to **revalidate with the server (via ETag/If-None-Match)** before reuse, preventing stale/cached downloads - The issue whereby the file hasn't changed by the authorisation has would still exist
- must-revalidate instructs caches that once the response is **stale**, they must successfully revalidate with the origin before serving it (avoiding use of outdated content) - I assume this would need another header to define when a response is stale

PE39806 (approved):

PR #3456, opened 2026-04-08
Update AdditionalInformation to have expandable descriptions

JR40159:


GB907762 (changes requested):
Mirrored models have an additional information property. This is a separate component and so will not inherit this functionality. If possible can we reuse this component in that form.

JRB66955:
I'm finding if the description is _extremely long_ (i.e. ~900-1000chars) for a given modelcard schema I get a 404 when trying to create a model - I don't think this PR is causing this issue but worth investigating is this is easily fixable.

PE39806:
> I'm finding if the description is _extremely long_ (i.e. ~900-1000chars) for a given modelcard schema I get a 404 when trying to create a model - I don't think this PR is causing this issue but worth investigating is this is easily fixable.

I've been having this to but it doesn't seem to be from this PR. When looking at the frontend logs I see a hydration error identical to https://github.com/emotion-js/emotion/issues/3308#issuecomment-2775761085 which is perplexing. We have https://github.com/gchq/Bailo/blob/897a941d6fb2e36f0dd88b13f81f4be2981817e3/frontend/utils/createEmotionCache.ts#L3-L5 which should prevent this, but clearly something else is going on.

ARADDCC002 (dismissed):

frontend/src/common/ExpandableTypography.tsx
```diff
@@ -0,0 +1,24 @@
...
+  maxLength?: number
+}
+
+export default function ExpandableTypography({ children: text, maxLength = 100, ...props }: ExpandableTypographyProps) {
+  const [expanded, setExpanded] = useState(false)
+
+  if (text.length > maxLength) {
+    return (
+      <Box sx={{ mb: 1 }}>
+        <Typography {...props}>{expanded ? text : `${text.slice(0, maxLength)}...`}</Typography>
```
GB907762: This can sometimes be flaky. For example the slice could split after a space resulting in `some text           ...`.

Using the `text-overlflow: elipsis` property would assure this is rendered correctly.

You may decide this is minor and not worth worrying about.

frontend/src/MuiForms/AdditionalInformation.tsx
```diff
@@ -60,15 +61,19 @@ export default function AdditionalInformation({
           {required && <span style={{ color: theme.palette.error.main }}>{' *'}</span>}
         </Typography>
         {description && editMode && (
-          <Typography variant='caption' color='textSecondary' fontWeight='bold'>
-            {description}
-          </Typography>
+          <Box sx={{ mb: 1 }}>
+            <ExpandableTypography>{description}</ExpandableTypography>
+          </Box>
         )}
```
GB907762: I don't think this box is needed
```suggestion
        {description && editMode && <ExpandableTypography>{description}</ExpandableTypography>}
```

GB907762 (changes requested):

GB907762 (dismissed):

GB907762 (approved):

ARADDCC002 (approved):

PR #3462, opened 2026-04-10
Create API endpoint for historic model volume metric

PE39806:


JRB66955 (dismissed):

backend/src/services/metrics.ts
```diff
@@ -0,0 +1,75 @@
...
+  const match: Record<string, unknown> = {
+    createdAt: { $gte: start, $lte: end },
+  }
+
+  if (organisation) {
+    match.organisation = organisation
+  }
+
+  const pipeline: PipelineStage[] = [
+    { $match: match },
```
ON93754, line 31: More of a general question but do we need to explicitly handle soft-deleted models or is that automatically handled by the soft delete plugin we use ?

backend/src/services/metrics.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    {
+      $group: {
+        _id: {
+          $dateTrunc: {
+            date: '$createdAt',
+            unit: period,
+            timezone,
```
GB907762: This needs to be sanitised. If it's a random string an `InternalError` will be thrown. This is due to `_id` being resolved as `null`.

backend/src/services/metrics.ts
```diff
@@ -0,0 +1,75 @@
...
+    periodStart: Date
+    periodEnd: Date
+    count: number
+  }>(pipeline)
+
+  return results.map((r) => ({
+    periodStart: r.periodStart.toISOString(),
+    periodEnd: r.periodEnd.toISOString(),
+    count: r.count,
+  }))
```
GB907762: This is a nit pick. So feel free to disregard. But this might help maintainers in future.

```suggestion
  const dataPoints = await ModelModel.aggregate<{
    periodStart: Date
    periodEnd: Date
    count: number
  }>(pipeline)

  return dataPoints.map((dataPoint) => ({
    periodStart: dataPoint.periodStart.toISOString(),
    periodEnd: dataPoint.periodEnd.toISOString(),
    count: dataPoint.count,
  }))
```

backend/src/routes/v2/metrics/getModelVolume.ts
```diff
@@ -0,0 +1,93 @@
...
+    const {
+      query: { period, startDate: start_date, endDate: end_date, timezone, organisation },
+    } = parse(req, getModelVolumeSchema)
+
+    if (!(await authentication.hasRole(req.user, Roles.Admin))) {
+      throw Forbidden('You do not have the required role.', {
+        userDn: req.user.dn,
+        requiredRole: Roles.Admin,
+      })
+    }
```
GB907762: Should this be at route level. My feeling is that this should be defined within a method in the authorisation connector? It's likely going to be reused.

backend/src/services/metrics.ts
```diff
@@ -0,0 +1,75 @@
...
+import { ModelVolumeDataPointSchema } from '../routes/v2/metrics/getModelVolume.js'
+
+export const ModelVolumePeriodEnum = z.enum(['day', 'week', 'month', 'quarter', 'year'])
+type ModelVolumePeriod = z.infer<typeof ModelVolumePeriodEnum>
+type ModelVolumeDataPoint = z.infer<typeof ModelVolumeDataPointSchema>
+
+export async function calculateModelVolume(
+  period: ModelVolumePeriod,
+  startDate: string,
+  endDate: string,
```
GB907762: Could we make these optional. There could be a case where you want all models up to a certain date? For example all of the models.

GB907762 (changes requested):

backend/src/services/metrics.ts
PE39806: This is handled by https://github.com/gchq/Bailo/blob/1f5006e8c3f6f4b09be81328822f2207d4e0d352/backend/src/models/plugins/softDeletePlugin.ts#L119-L122

backend/src/services/metrics.ts
PE39806: There's no easy way to use MongoDB's parser so instead I will create a specific `try/catch` case for timezone errors.

GB907762 (approved):

JRB66955 (approved):

PR #3474, opened 2026-04-14
Bai 2331 multiarch UI

IR96334:
Also, contains bugfix for v3 api calls


frontend/src/entry/model/registry/VulnerabilityResult.tsx
```diff
@@ -1,59 +1,115 @@
...
+  return scanResults.sort((scanResult) =>
+    [
+      SeverityLevel.CRITICAL,
+      SeverityLevel.HIGH,
+      SeverityLevel.MEDIUM,
+      SeverityLevel.LOW,
+      SeverityLevel.UNKNOWN,
+    ].findIndex((severity) => scanResult.severityCounts[severity] > 0),
+  )[scanResults.length - 1].platform
+}
```
PE39806: `sort` takes two parameters https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#parameters

frontend/src/entry/model/registry/ModelImageTagDisplay.tsx
```diff
@@ -47,13 +55,12 @@ export default function ModelImageTagDisplay({ modelImage, tag, mutate }: ModelI
...
-      const tagResults = modelImage.scanSummaries.find((tagResult) => tagResult.tag === imageTag)
-
+      const tagResult = modelImage.scanSummaries.filter((tagResult) => tagResult.tag === imageTag)
       return (
         <VulnerabilityResult
-          results={tagResults}
+          scanResults={tagResult}
           warningOnly
-          detailedViewUrl={`/model/${modelImage.repository}/registry/${modelImage.name}/${imageTag}`}
+          detailedViewUrlPrefix={`/model/${modelImage.repository}/registry/${encodeURIComponent(modelImage.name)}/${encodeURIComponent(tagResult[0].tag)}/`}
```
PE39806: Is it always safe to assume `tagResult[0]` exists?
```suggestion
      const tagResult = modelImage.scanSummaries.filter((tagResult) => tagResult.tag === imageTag)
      if (!tagResults.length) {
        return null
      }
      
return (
<VulnerabilityResult
          scanResults={tagResult}
warningOnly
          detailedViewUrlPrefix={`/model/${modelImage.repository}/registry/${encodeURIComponent(modelImage.name)}/${encodeURIComponent(tagResult[0].tag)}/`}
```

frontend/src/entry/model/registry/ModelImageTagDisplay.tsx
```diff
@@ -11,6 +11,14 @@ import MessageAlert from 'src/MessageAlert'
 import { ArtefactKind, ModelImagesWithOptionalScanResults } from 'types/types'
 import { getErrorMessage } from 'utils/fetcher'
 
+function checkIfMultiPlatform(modelImage: ModelImagesWithOptionalScanResults, tag: string) {
+  const correctTag = modelImage.scanSummaries.find((scan) => scan.tag === tag)
+  if (correctTag) {
+    return !!correctTag.platform
```
PE39806, line 46: This will return `true` if the image only has one platform. Is this correct behaviour, or should this be changed to check for multiple platforms?

frontend/src/entry/model/registry/VulnerabilityResult.tsx
```diff
@@ -1,59 +1,115 @@
...
   warningOnly = false,
-  detailedViewUrl = '',
+  detailedViewUrlPrefix = '',
 }: VulnerabilityResultProps) {
   const theme = useTheme()
   const [anchorElScan, setAnchorElScan] = useState<HTMLElement | null>(null)
   const [labelText, setLabelText] = useState('')
   const [warningChipColour, setWarningChipColour] = useState('unset')
   const [warningChipHoverColour, setWarningChipHoverColour] = useState('unset')
+  let platforms: (string | undefined)[] | undefined
```
PE39806: Can `platforms` be memoised to prevent recalculating on every render?

frontend/src/entry/model/registry/VulnerabilityResult.tsx
```diff
@@ -163,7 +219,10 @@ export default function VulnerabilityResult({
           />
         )}
         <Popover
-          hidden={results.scanResults && results.scanResults.some((res) => res.state === ScanState.InProgress)}
+          //check why this is .some as it was never array before, right?
```
PE39806: ?

frontend/actions/artefactScanning.ts
```diff
@@ -37,19 +37,20 @@ export function rerunImageArtefactScan(modelId: string, name: string, tag: strin
   })
 }
 
-export function useGetImageScanResults(modelId: string, name: string, tag: string) {
+export function useGetImageScanResults(modelId: string, name: string, tag: string, digest: string) {
   const encodedModelId = encodeURIComponent(modelId)
```
PE39806: We don't need to encode `modelId` as we do not do this for other endpoints.

PE39806 (changes requested):

frontend/src/entry/model/registry/ModelImageTagDisplay.tsx
IR96334: Yes. Currently the platform property is only returned if it is multi-platform

frontend/src/entry/model/registry/VulnerabilityResult.tsx
IR96334: As this value will never change, I have decided against memoisation and just made it a function-scoped variable in parent component and passed it as prop.

frontend/actions/artefactScanning.ts
```diff
@@ -28,28 +28,27 @@ export function rerunArtefactScan(modelId: string, artefactId: string) {
...
   const encodedName = encodeURIComponent(name)
   const encodedTag = encodeURIComponent(tag)
+  const encodedDigest = encodeURIComponent(digest)
   const { data, isLoading, error, mutate } = useSWR<
     {
       imageBreakdown: ImageTagResult
     },
     ErrorInfo
-  >(`/api/v2/model/${encodedModelId}/image/${encodedName}/${encodedTag}`, fetcher)
+  >(`/api/v3/model/${modelId}/image/${encodedName}/${encodedTag}/${encodedDigest}`, fetcher)
```
GB907762: Throughout the app we have this antipattern. Sometimes swr will attempt to fetch based on the router even when the query has not resolved. This is more prevalent in this hook since the endpoint returns a large response body. 

This means that the browser will fetch `/api/v3/model/undefined/image/undefined/undefined/undefined`, causing an `InternalServer` error. Instead could we use this:

```suggestion
export function useGetImageScanResults(modelId: string, name: string, tag: string, digest: string, isReady?: boolean) {
  const encodedName = encodeURIComponent(name)
  const encodedTag = encodeURIComponent(tag)
  const encodedDigest = encodeURIComponent(digest)
  const { data, isLoading, error, mutate } = useSWR<
    {
      imageBreakdown: ImageTagResult
    },
    ErrorInfo
  >(isReady ? `/api/v3/model/${modelId}/image/${encodedName}/${encodedTag}/${encodedDigest}` : null, fetcher)

  return {
    mutateImage: mutate,
    image: data?.imageBreakdown,
    isImageLoading: isLoading,
    isImageError: error,
  }
}
```

And then use like this.

```typescript
  const {
    image: modelImage,
    isImageLoading,
    isImageError,
    mutateImage,
  } = useGetImageScanResults(modelId as string, name as string, tag as string, digest as string, router.isReady)
```

frontend/src/entry/model/registry/ModelImageTagDisplay.tsx
```diff
@@ -70,12 +113,21 @@ export default function ModelImageTagDisplay({ modelImage, tag, mutate }: ModelI
...
+        <Stack spacing={2} direction='row' alignItems='center'>
           <Box width='fit-content'>
             <CodeLine
               line={`docker pull ${uiConfig ? uiConfig.registry.host : 'unknownhost'}/${modelImage.repository}/${modelImage.name}:${tag}`}
             />
           </Box>
+          {checkIfMultiPlatform(modelImage, tag) && (
+            <Tooltip
+              title={
+                'A multi-platform image is a single registry entry (image:tag) that contains multiple copies of the same image, but for different OS and architecture combinations. e.g. linux/amd64, windows/amd64, linux/arm64, etc.'
```
GB907762: I like the tooltip.

Usually a title is in 'active voice' (Directing the user) or a noun phrase rather than a description. Would something like this fit better?

```suggestion
                'A single image tag that automatically provides the correct architecture for the host machine'
```

frontend/src/entry/model/registry/VulnerabilityResult.tsx
```diff
@@ -163,7 +204,9 @@ export default function VulnerabilityResult({
           />
         )}
         <Popover
-          hidden={results.scanResults && results.scanResults.some((res) => res.state === ScanState.InProgress)}
+          hidden={
+            currentResult.scanResults && currentResult.scanResults.some((res) => res.state === ScanState.InProgress)
```
GB907762, line 235: Each of the scan states are binded to a specific image platform.

If I select a platform that is `InProgress` it locks me from being able to view the popover.
Equally this happens with `error` states

frontend/src/entry/model/registry/ModelImageTagDisplay.tsx
```diff
@@ -8,9 +8,42 @@ import CodeLine from 'src/entry/model/registry/CodeLine'
...
+        break
+      } else if (a.severityCounts[severities[i]] < b.severityCounts[severities[i]]) {
+        res = 1
+        break
+      }
+    }
+    return res
+  })
+  return sortedResults
+}
```
GB907762, line 41: Something like this might be easier to read?

Also note that `Array.prototype.sort` sorts the array inplace instead of creating a copy so the assignment is redundant. Use `Array.prototype.toSorted` to create a clone.

```suggestion
function sortPlatformByVulnerability(scanResults: ImageTagResult[]): ImageTagResult[] {
  const getHighestSeverityLevel = (scanResult) =>
    severities.findIndex((severity) => scanResult.severityCounts[severity] > 0)

  return scanResults.toSorted(
    (scanResultA, scanResultB) => getHighestSeverityLevel(scanResultA) - getHighestSeverityLevel(scanResultB),
  )
}
```

frontend/pages/model/[modelId]/registry/[name]/[tag]/[digest].tsx
```diff
@@ -246,7 +251,7 @@ export default function ImageTagInformation() {
 
   return (
     <>
-      <Title text={name && tag ? `${name}:${tag}` : 'Loading...'} />
+      <Title text={name && modelImage.tag ? `${name}:${modelImage.tag}` : 'Loading...'} />
```
Comment (PE39806) [inline frontend/pages/model/[modelId]/registry/[name]/[tag]/[digest].tsx]: `modelImage` may be `undefined` while loading
```suggestion
      <Title text={name && modelImage?.tag ? `${name}:${modelImage.tag}` : 'Loading...'} />
```

frontend/src/entry/model/registry/ModelImageTagDisplay.tsx
```diff
@@ -8,9 +8,42 @@ import CodeLine from 'src/entry/model/registry/CodeLine'
...
+const severities = [
+  SeverityLevel.CRITICAL,
+  SeverityLevel.HIGH,
+  SeverityLevel.MEDIUM,
+  SeverityLevel.LOW,
+  SeverityLevel.UNKNOWN,
+]
+
+function sortPlatformByVulnerability(scanResults: ImageTagResult[]): ImageTagResult[] {
+  const sortedResults = scanResults.sort((a, b) => {
```
PE39806: > The `sort()` method of [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array) instances sorts the elements of an array [in place](https://en.wikipedia.org/wiki/In-place_algorithm) [...]
> To sort the elements in an array without mutating the original array, use [`toSorted()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted).

_https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort_

I think that we should switch to `toSorted` as `scanResults` comes from props/state and I'd be concerned that this may introduce strange side effects for things like react memoisation.

frontend/src/entry/model/registry/VulnerabilityResult.tsx
```diff
@@ -1,59 +1,100 @@
...
-  detailedViewUrl = '',
+  detailedViewUrlPrefix = '',
 }: VulnerabilityResultProps) {
   const theme = useTheme()
   const [anchorElScan, setAnchorElScan] = useState<HTMLElement | null>(null)
   const [labelText, setLabelText] = useState('')
   const [warningChipColour, setWarningChipColour] = useState('unset')
   const [warningChipHoverColour, setWarningChipHoverColour] = useState('unset')
 
+  const multiarchitecture = (!!platforms && !!scanResults) ?? false
```
PE39806: `(!!platforms && !!scanResults)` already returns a boolean so you don't need ` ?? false`
```suggestion
  const multiarchitecture = !!platforms && !!scanResults
```

frontend/src/entry/model/registry/VulnerabilityResult.tsx
```diff
@@ -1,59 +1,100 @@
...
+    scanResults ? selectedCurrentResult(multiarchitecture, currentPlatform, scanResults) : undefined,
+  )
+
+  const handlePlatformChange = (event: SelectChangeEvent) => {
+    if (scanResults) {
+      const res = scanResults.find((result) => result.platform === event.target.value)
+      setCurrentResult(res)
+      setCurrentPlatform(event.target.value as string)
+    }
+  }
```
PE39806: If `scanResults` updates (e.g. SWR refetch) then the state will not update. Prefer something like:
```suggestion
  const currentResult = useMemo(() => {
    if (!scanResults) return undefined
    return selectedCurrentResult(multiarchitecture, currentPlatform, scanResults)
  }, [scanResults, multiarchitecture, currentPlatform])


  const handlePlatformChange = (event: SelectChangeEvent) => {
    setCurrentPlatform(event.target.value as string)
  }
```

frontend/src/entry/model/registry/VulnerabilityResult.tsx
```diff
@@ -1,59 +1,100 @@
...
+  if (multiarchitecture && currentPlatform) {
+    return scanResults.find((result) => result.platform === currentPlatform)
+  }
+
+  return scanResults[0]
+}
+
+type VulnerabilityResultProps = {
+  scanResults: ImageTagResult[] | undefined
+  platforms?: (string | undefined)[]
```
PE39806: Prefer 
```suggestion
  platforms?: string[]
```
I think that this can be achieved by replacing `filter()` in https://github.com/gchq/Bailo/blob/13975cbc1a9a38596ec108ee80f0842e24ba0d24/frontend/src/entry/model/registry/ModelImageTagDisplay.tsx#L92 with `.filter((platform): platform is string => !!platform && platform !== 'unknown/unknown')`

PE39806 (changes requested):

frontend/src/entry/model/registry/VulnerabilityResult.tsx
IR96334: Found a way to get rid of platforms anyway. 🤷

frontend/src/entry/model/registry/ModelImageTagDisplay.tsx
IR96334: Can't seem to get it to work after testing, will keep previous implementation for now

frontend/src/entry/model/registry/VulnerabilityResult.tsx
```diff
@@ -1,107 +1,188 @@
...
   warningOnly = false,
-  detailedViewUrl = '',
+  detailedViewUrlPrefix = '',
 }: VulnerabilityResultProps) {
   const theme = useTheme()
   const [anchorElScan, setAnchorElScan] = useState<HTMLElement | null>(null)
   const [labelText, setLabelText] = useState('')
+  const [chipIcon, setChipIcon] = useState()
   const [warningChipColour, setWarningChipColour] = useState('unset')
   const [warningChipHoverColour, setWarningChipHoverColour] = useState('unset')
```
PE39806, line 55: Could these props and subsequent `useEffectEvent` & `useEffect` be converted into a single `useMemo`?

frontend/src/entry/model/registry/VulnerabilityResult.tsx
```diff
@@ -130,40 +213,27 @@ export default function VulnerabilityResult({
...
+            color: 'white',
+            '&:hover': { backgroundColor: warningChipHoverColour },
+          }}
+          label={labelText}
+          onClick={(e) => setAnchorElScan(e.currentTarget)}
+        />
         <Popover
-          hidden={results.scanResults && results.scanResults.some((res) => res.state === ScanState.InProgress)}
+          hidden={
+            currentResult.scanResults && currentResult.scanResults.every((res) => res.state === ScanState.InProgress)
```
PE39806: I think this should be changed back to `.some()` as any `ScanState.InProgress` means that the scan is still in progress
```suggestion
            currentResult.scanResults && currentResult.scanResults.some((res) => res.state === ScanState.InProgress)
```

frontend/src/entry/model/registry/VulnerabilityResult.tsx
```diff
@@ -1,107 +1,188 @@
...
+  detailedViewUrlPrefix = '',
 }: VulnerabilityResultProps) {
   const theme = useTheme()
   const [anchorElScan, setAnchorElScan] = useState<HTMLElement | null>(null)
   const [labelText, setLabelText] = useState('')
+  const [chipIcon, setChipIcon] = useState()
   const [warningChipColour, setWarningChipColour] = useState('unset')
   const [warningChipHoverColour, setWarningChipHoverColour] = useState('unset')
 
+  const multiarchitecture = !!scanResults && !!scanResults[0].platform
```
PE39806: This will fail if `scanResults` is an empty array. Maybe guard with an early
```typescript
if (!scanResults || scanResults.length === 0) {
  return <Chip label="Not scanned" color="warning" icon={<Warning />} />
}
```

PE39806 (changes requested):

frontend/src/entry/model/registry/VulnerabilityResult.tsx
IR96334: Moved this to prop, added check to modelimagetagdisplay, where it can only be executed AFTER checking the array has length > 0.

GB907762 (approved):

JRB66955 (approved):

PE39806 (approved):

PR #3477, opened 2026-04-16
Add Model Export action to Tokens

PE39806:
Enables using the token to trigger the model export endpoint.
Also fixes a bug where files were being repeatedly exported when multiple releases were selected.

### DIFF
--- a/backend/src/services/mirroredModel/exporters/documents.ts
+++ b/backend/src/services/mirroredModel/exporters/documents.ts
@@ -61,8 +61,7 @@ export class DocumentsExporter extends BaseExporter {
         }
       }
 
-      const scannersInfo = scanners.scannersInfo() ?? []
-      if (Array.isArray(scannersInfo) && scannersInfo.length > 0) {
+      if (scanners.scannersInfo().length > 0) {
         const scanErrors: {
           missingScan: Array<{ name: string; id: string }>
           incompleteScan: Array<{ name: string; id: string }>
--- a/backend/src/services/mirroredModel/exporters/documents.ts
+++ b/backend/src/services/mirroredModel/exporters/documents.ts
@@ -61,7 +61,7 @@ export class DocumentsExporter extends BaseExporter {
         }
       }
 
-      if (scanners.scannersInfo()) {
+      if (scanners.scannersInfo().length > 0) {
         const scanErrors: {
           missingScan: Array<{ name: string; id: string }>
           incompleteScan: Array<{ name: string; id: string }>
--- a/backend/src/services/mirroredModel/exporters/documents.ts
+++ b/backend/src/services/mirroredModel/exporters/documents.ts
@@ -61,7 +61,8 @@ export class DocumentsExporter extends BaseExporter {
         }
       }
 
-      if (scanners.scannersInfo().length > 0) {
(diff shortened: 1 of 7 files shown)


JR40159 (dismissed):

backend/src/services/mirroredModel/exporters/documents.ts
```diff
@@ -61,7 +61,8 @@ export class DocumentsExporter extends BaseExporter {
         }
       }
 
-      if (scanners.scannersInfo()) {
+      const scannersInfo = scanners.scannersInfo() ?? []
+      if (Array.isArray(scannersInfo) && scannersInfo.length > 0) {
```
GB907762: Could you clarify in what instance this would not be an array? In other instances within the service layer there's no null coalescence? 
```suggestion
      const scannersInfo = scanners.scannersInfo()
      if (scannersInfo.length > 0) {
```

backend/src/services/mirroredModel/exporters/file.ts
```diff
@@ -32,7 +32,8 @@ export class FileExporter extends BaseExporter {
       })
     }
 
-    if (scanners.scannersInfo().length > 0) {
+    const scannersInfo = scanners.scannersInfo() ?? []
+    if (Array.isArray(scannersInfo) && scannersInfo.length > 0) {
```
GB907762: Same here?

```suggestion
    const scannersInfo = scanners.scannersInfo()
    if (scannersInfo.length > 0) {
```

backend/src/services/mirroredModel/mirroredModel.ts
```diff
@@ -73,6 +80,11 @@ export async function exportModel(
     'Request checks complete, starting export.',
   )
 
+  const files = documentsExporter.getFiles()
+  if (!files) {
+    throw InternalError('DocumentsExporter returned no files after init()', { exportId })
+  }
```
GB907762, line 87: Why do you need to block an export if there's no files being exported? 
Or is this to do with releases not matching the object store?

Either way, could you change this error. It's not very clear to me since it doesn't abstract the terminology in the code.

```suggestion
  const files = documentsExporter.getFiles()
  if (!files) {
    throw BadReq('A model cannot be exported with no files, in releases.', { exportId })
  }
```

GB907762 (dismissed):

backend/src/services/mirroredModel/exporters/documents.ts
PE39806: Oops, this is a left over artefact from when I was rewriting some tests and got confused why their behaviour was being strange, which turned out to be because of `scannersMocks.default.scannersInfo.mockReturnValue(false)` (rather than `[]`). This test has been corrected so I will remove this redundant check.

backend/src/services/mirroredModel/mirroredModel.ts
PE39806: This is only for TypeScript type narrowing (to remove the `!` in the previous `documentsExporter.getFiles()!`) and would never actually be reachable. I'll add a comment explaining this. It relies on `[]` being truthy.

JRB66955 (approved):

ON93754 (approved):

PR #3479, opened 2026-04-20
Feature/bai 2358 create point in time metrics endpoint

ON93754:


backend/config/default.cjs
```diff
@@ -283,6 +283,10 @@ module.exports = {
       maxInitRetries: 5,
       initRetryDelay: 5000,
     },
+
+    metrics: {
+      kind: 'simple',
+    },
```
JRB66955, line 289: This will need adding to the `backend/src/utils/_mocks_/config.ts` as well, to fix the test issues.

JRB66955 (dismissed):

backend/package-lock.json
PE39806, line 1: There are no changes to `backend/package.json` so this change should also be reverted.

backend/src/routes/v2/metrics/getOverviewMetrics.ts
```diff
@@ -0,0 +1,83 @@
...
+  responses: {
+    200: {
+      description: 'Current snapshot of system metrics.',
+      content: {
+        'application/json': {
+          schema: z.object({
+            usersTotal: z.number(),
+            activeUsers: z.number(),
+            storageUsedGb: z.number(),
+          }),
```
PE39806: This doesn't match up with the actual output. Should it be `GetOverviewMetricsResponseSchema`?

backend/src/routes/v2/metrics/getPolicyMetrics.ts
```diff
@@ -0,0 +1,78 @@
...
+  responses: {
+    200: {
+      description: 'Current snapshot of system metrics.',
+      content: {
+        'application/json': {
+          schema: z.object({
+            usersTotal: z.number(),
+            activeUsers: z.number(),
+            storageUsedGb: z.number(),
+          }),
```
PE39806: As above but for `GetPolicyMetricsResponseSchema`.

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,421 @@
...
+}
+
+/**
+ * Counts distinct models referenced in another collection
+ * (e.g. releases or access requests).
+ *
+ * If an organisation filter is provided, a lookup is performed to ensure
+ * only models belonging to that organisation are counted.
+ */
+async function countDistinctModelsWithRelation(collection: any, filter: ModelFilter): Promise<number> {
```
PE39806: Prefer `import { Model } from 'mongoose'` with `Model<any>`
```suggestion
async function countDistinctModelsWithRelation(collection: Model<any>, filter: ModelFilter): Promise<number> {
```

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,421 @@
...
+
+type SchemaRoleMap = {
+  schemaRoleMap: Record<string, string[]>
+  defaultRoles: string[]
+}
+
+/**
+ * Builds a lookup structure describing which review roles apply to which schemas.
+ */
+async function buildSchemaRoleMap(): Promise<SchemaRoleMap> {
```
PE39806: This could benefit from utilising NodeCache.

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,421 @@
...
+}
+
+/**
+ * Calculates the full set of overview metrics either globally
+ * or scoped to a specific organisation.
+ */
+async function calculateOverviewMetricsForOrg(org?: string): Promise<BaseMetrics> {
+  const modelFilter: ModelFilter = org ? { organisation: org } : {}
+
+  const totalUsersPromise = org ? Promise.resolve(-1) : calculateTotalUsers()
```
PE39806: I think there should be some explanation/warning about `-1` as it is not immediately intuitive.

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,421 @@
...
+
+  if (filter.organisation) {
+    pipeline.push(...buildOrganisationLookupStages(filter.organisation))
+  }
+
+  pipeline.push({ $group: { _id: '$modelId' } }, { $count: 'count' })
+
+  const result = await collection.aggregate(pipeline)
+
+  return result[0]?.count ?? 0
```
PE39806: I think this can be optimised a bit
```suggestion
  // Faster approach for no organisation filter
  if (!filter.organisation) {
    // may need to change this field depending on the collection type
    const ids = await collection.distinct('modelId')
    return ids.length
  }

  // Organisation requires joining models
  const pipeline: PipelineStage[] = [
    ...buildOrganisationLookupStages(filter.organisation),
    { $group: { _id: '$modelId' } },
    { $count: 'count' },
  ]

  const result = await collection.aggregate(pipeline)

  return result[0]?.count ?? 0
```
It might also be worth indexing both `Release` and `AccessRequest` on `{ modelId: 1 }` to speed up search over their larger collections.

This could be even further optimised by storing the organisation of the model directly on the Releases and AccessRequests but this would add other complexity so is unnecessary for now.

PE39806 (changes requested):

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,524 @@
...
+  org?: string,
+): Promise<PolicyMetricsResult> {
+  const filter: any = {}
+
+  // Only undefined means global
+  if (org !== undefined) {
+    filter.organisation = org
+  }
+
+  const models = await ModelModel.find(filter).select('id organisation card collaborators').lean()
```
PE39806: This could be a huge number of models which risks excessive memory usage. Maybe look into using mongoose cursor streaming to maintain constant memory for larger datasets e.g.
```typescript
const cursor = ModelModel.find(filter)
  .select('id organisation card collaborators')
  .lean()
  .cursor()

for await (const model of cursor) {
  // existing evaluation logic
}
```

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,524 @@
...
+    bucket: ModelVolumeBucket,
+    startDate: string | Date,
+    endDate: string | Date,
+    timezone?: string,
+  ): Promise<GetModelVolumeResponse> {
+    const start = new Date(startDate)
+    const end = new Date(endDate)
+
+    try {
+      const [{ alignedStart }] = await ModelModel.aggregate<{
```
PE39806: Will this error if `aggregate` returns an empty array?

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,524 @@
...
+    buildModelMatchStage(filter),
+    {
+      $group: {
+        _id: '$card.schemaId',
+        count: { $sum: 1 },
+      },
+    },
+  ]
+
+  const modelCounts = await ModelModel.aggregate(pipeline)
```
PE39806: Prefer
```suggestion
  type SchemaAgg = { _id: string | null; count: number }
  const modelCounts = await ModelModel.aggregate<SchemaAgg>(pipeline)
```
and then you can safely remove the subsequent `any`s

PE39806 (changes requested):

backend/src/connectors/metrics/simple.ts
ON93754: Good spot

backend/src/connectors/metrics/base.ts
```diff
@@ -0,0 +1,32 @@
...
+      'A corporation': 2,
+    },
+  }),
+})
+
+export const ModelVolumeBucketEnum = z.enum(['day', 'week', 'month', 'quarter', 'year'])
+export type ModelVolumeBucket = z.infer<typeof ModelVolumeBucketEnum>
+export type ModelVolumeDataPoint = z.infer<typeof ModelVolumeDataPointSchema>
+
+export abstract class BaseMetricsConnector {
```
JR40159: Did you consider making this a normal class with implementation rather than an abstract one akin to the authorisation connector. This way it'll be easier for deployments to use the majority of the base implementation while overriding some bits

backend/src/routes/v3/metrics/getEntryVolume.ts
JR40159, line 1: Can we move this to be under a `v3` folder?

backend/src/routes/v3/metrics/getUsageMetrics.ts
JR40159, line 1: Can we move this to be under a `v3` folder?

backend/src/routes/v3/metrics/getComplianceMetrics.ts
JR40159, line 1: Can we move this to be under a `v3` folder?

backend/src/routes/v3/metrics/getModelVolume.ts
```diff
@@ -73,23 +53,31 @@ registerPath({
   },
 })
 
-type GetModelVolumeResponses = z.infer<typeof GetModelVolumeResponseSchema>
+export type GetModelVolumeResponse = z.infer<typeof GetModelVolumeResponseSchema>
```
JR40159: Does this need exporting?

backend/src/routes/v3/metrics/getModelVolume.ts
```diff
@@ -73,23 +53,31 @@ registerPath({
...
 
 export const getModelVolume = [
-  async (req: Request, res: Response<GetModelVolumeResponses>): Promise<void> => {
+  async (req: Request, res: Response<GetModelVolumeResponse>): Promise<void> => {
     req.audit = AuditInfo.ViewMetric
+
+    if (!(await authentication.hasRole(req.user, Roles.Admin))) {
```
JR40159: Can this be done in the service layer like the rest of the application, maybe in `calculateModelVolume`

backend/src/routes/v3/metrics/getUsageMetrics.ts
```diff
@@ -0,0 +1,89 @@
...
+      content: {
+        'application/json': {
+          schema: GetOverviewMetricsResponseSchema,
+        },
+      },
+    },
+  },
+})
+
+export type SchemaInfo = z.infer<typeof SchemaInfoSchema>
```
JR40159, line 62: Perhaps all these types should be defined in the base connector as they're used in the connection implementation

backend/src/routes/v3/metrics/getUsageMetrics.ts
```diff
@@ -0,0 +1,89 @@
...
+export type StateInfo = z.infer<typeof StateInfoSchema>
+export type BaseMetrics = z.infer<typeof BaseMetricsSchema>
+export type OrganisationMetrics = z.infer<typeof OrganisationMetricsSchema>
+export type GetOverviewMetricsResponse = z.infer<typeof GetOverviewMetricsResponseSchema>
+
+export const getOverviewMetrics = [
+  async (req: Request, res: Response<GetOverviewMetricsResponse>): Promise<void> => {
+    req.audit = AuditInfo.ViewMetric
+
+    if (!(await authentication.hasRole(req.user, Roles.Admin))) {
```
JR40159: Move to service layer

backend/src/routes/v3/metrics/getComplianceMetrics.ts
```diff
@@ -0,0 +1,88 @@
...
+      content: {
+        'application/json': {
+          schema: GetPolicyMetricsResponseSchema,
+        },
+      },
+    },
+  },
+})
+
+export type RoleSummary = z.infer<typeof RoleSummarySchema>
```
JR40159: Move type definitions to where they're used

backend/src/routes/v3/metrics/getComplianceMetrics.ts
```diff
@@ -0,0 +1,88 @@
...
+export type ModelRoleMetrics = z.infer<typeof ModelRoleMetricsSchema>
+export type PolicyBaseMetrics = z.infer<typeof PolicyBaseMetricsSchema>
+export type PolicyOrganisationMetrics = z.infer<typeof PolicyOrganisationMetricsSchema>
+export type GetPolicyMetricsResponse = z.infer<typeof GetPolicyMetricsResponseSchema>
+
+export const getPolicyMetrics = [
+  async (req: Request, res: Response<GetPolicyMetricsResponse>): Promise<void> => {
+    req.audit = AuditInfo.ViewMetric
+
+    if (!(await authentication.hasRole(req.user, Roles.Admin))) {
```
JR40159: Move to service layer

backend/src/routes/v3/metrics/getComplianceMetrics.ts
```diff
@@ -0,0 +1,88 @@
...
+import { Roles } from '../../../connectors/authentication/constants.js'
+import authentication from '../../../connectors/authentication/index.js'
+import { z } from '../../../lib/zod.js'
+import { calculatePolicyMetrics } from '../../../services/metrics.js'
+import { registerPath } from '../../../services/specification.js'
+import { Forbidden } from '../../../utils/error.js'
+import { parse } from '../../../utils/validate.js'
+
+export const getOverviewMetricsSchema = z.object({
+  query: z.object({}).strict(),
```
JR40159, line 11: ?

backend/src/services/metrics.ts
```diff
@@ -1,86 +1,24 @@
...
+import { getMetricsConnector } from '../connectors/metrics/index.js'
+import { GetModelVolumeResponse } from '../routes/v2/metrics/getModelVolume.js'
+import { GetOverviewMetricsResponse } from '../routes/v2/metrics/getOverviewMetrics.js'
+import { GetPolicyMetricsResponse } from '../routes/v2/metrics/getPolicyMetrics.js'
 
-import { z } from '../lib/zod.js'
-import ModelModel from '../models/Model.js'
-import { ModelVolumeDataPointSchema } from '../routes/v2/metrics/getModelVolume.js'
-import { BadReq } from '../utils/error.js'
+const metrics = getMetricsConnector()
```
JR40159: This can be removed by updating the import to `import metrics from '../connectors/metrics/index.js'
`

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+import { searchSchemas } from '../../services/schema.js'
+import { SchemaKind } from '../../types/enums.js'
+import { BadReq } from '../../utils/error.js'
+import { BaseMetricsConnector, ModelVolumeBucket, ModelVolumeDataPoint } from './base.js'
+import { addBucket, buildModelMatchStage, ModelFilter, SchemaRoleMap } from './metricUtils.js'
+
+const SCHEMA_ROLE_CACHE_TTL = 5 * 60 // 5 minutes
+const SCHEMA_ROLE_CACHE_KEY = 'schemaRoleMap'
+
+const schemaRoleCache = new NodeCache({
```
JR40159: Should we cache all the metrics?

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+  type SchemaAgg = { _id: string | null; count: number }
+  const modelCounts = await ModelModel.aggregate<SchemaAgg>(pipeline)
+
+  // Build lookup of schemaId to count, ignoring models without a schema
+  const countMap = new Map<string, number>(
+    modelCounts.filter((row: any) => row._id).map((row: any) => [row._id, row.count]),
+  )
+
+  // Return all schemas with their usage count
+  return schemas.map((schema) => ({
```
JR40159: Could this be done in the original mongo query by specifying which fields you want to be returned and by using an aggregate query to lookup and count the models that have that schema ID?

backend/src/connectors/metrics/base.ts
```diff
@@ -0,0 +1,32 @@
...
+  abstract calculateOverviewMetrics(): Promise<GetOverviewMetricsResponse>
```
JR40159: I thinking we need to either think of a better name than 'overview' metrics and/or include a doc string here describing what kind of metrics this function should return?

backend/src/connectors/metrics/base.ts
```diff
@@ -0,0 +1,32 @@
...
+  abstract calculatePolicyMetrics(): Promise<GetPolicyMetricsResponse>
```
JR40159: `calculateComplianceMetrics` might be better

JR40159 (changes requested):

backend/src/services/metrics.ts
```diff
@@ -1,86 +1,24 @@
...
-    }>(pipeline)
-
-    const formattedDataPoints = dataPoints.map((dataPoint) => ({
-      periodStart: dataPoint.periodStart.toISOString(),
-      periodEnd: dataPoint.periodEnd.toISOString(),
-      count: dataPoint.count,
-    }))
-
-    return { startDate: start.toISOString(), endDate: end.toISOString(), dataPoints: formattedDataPoints }
-  } catch (err: any) {
```
JR40159, line 79: If this is for a Mongo error the `isMongoServerError` typeguard might be helpful

JR40159 (changes requested):

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+      },
+    },
+  ]
+
+  type SchemaAgg = { _id: string | null; count: number }
+  const modelCounts = await ModelModel.aggregate<SchemaAgg>(pipeline)
+
+  // Build lookup of schemaId to count, ignoring models without a schema
+  const countMap = new Map<string, number>(
+    modelCounts.filter((row: any) => row._id).map((row: any) => [row._id, row.count]),
```
PE39806: Above typing removes need for `any`s.
```suggestion
    const countMap = new Map<string, number>(
    modelCounts
      .filter((row): row is { _id: string; count: number } => row._id !== null)
      .map((row) => [row._id, row.count]),
  )
```

backend/src/routes.ts
```diff
@@ -257,13 +259,16 @@ server.put('/api/v2/review/role/:shortName', ...putReviewRole)
 
 server.get('/api/v2/models/tags', getPopularTags)
 
-server.get('/api/v2/metrics/modelVolume', requireRole(Roles.Admin), ...getModelVolume)
+server.get('/api/v3/metrics', ...getOverviewMetrics)
+server.get('/api/v3/metrics/policy', ...getPolicyMetrics)
+server.get('/api/v3/metrics/modelVolume', ...getModelVolume)
```
PE39806: Why remove `requireRole(Roles.Admin)` from here and do auth in the endpoint? If this is the design pattern going forward then `backend/src/routes/middleware/requireRoles.ts` should be removed as it is unused.

PE39806 (changes requested):

backend/src/routes/v3/metrics/getModelVolume.ts
ON93754: Yes - it is used in `base.ts`

backend/src/routes.ts
ON93754: A few of us discussed it and came to the conclusion it wasn't that useful as handling the auth within each endpoint gives more flexibility to do custom checks for an individual endpoint, without adding more overall complexity in the app.

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+ */
+async function getOrganisationIds(): Promise<string[]> {
+  return await ModelModel.distinct('organisation')
+}
+
+export class SimpleMetricsConnector extends BaseMetricsConnector {
+  async calculateOverviewMetrics(): Promise<GetOverviewMetricsResponse> {
+    const organisationIds = await getOrganisationIds()
+
+    const globalPromise = calculateOverviewMetricsForOrg()
```
JR40159: The variable name `globalPromise` seems a bit vague

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+    schemaName: schema.name,
+    count: countMap.get(schema.id) ?? 0,
+  }))
+}
+
+/**
+ * Calculates the full set of overview metrics either globally
+ * or scoped to a specific organisation.
+ */
+async function calculateOverviewMetricsForOrg(org?: string): Promise<BaseMetrics> {
```
JR40159: Since the `org` parameter is optional and the overview metrics are not always calculated for an org, I'd suggest this function should be called ` calculateOverviewMetrics`

backend/src/services/metrics.ts
JR40159, line 1: Since all the logic is contained within the connectors, a service layer for metrics seems a bit pointless

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+    models: modelsResult,
+  }
+}
+
+/**
+ * Retrieves the list of distinct organisation identifiers
+ * for which models currently exist.
+ */
+async function getOrganisationIds(): Promise<string[]> {
+  return await ModelModel.distinct('organisation')
```
JR40159: Rather than using a Mongo query for this, did you consider just getting this list of orgs from the config?

backend/src/connectors/metrics/simple.ts
JR40159: It would make more sense to be if the type `BaseMetrics` was called `OverviewMetrics`? Then `calculateOverviewMetricsForOrg` would return  `OverviewMetrics` and we wouldn't be reusing the `base` term in this type and the name of the 'base' class

backend/src/services/metrics.ts
ON93754: Good spot

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+    const byOrganisationPromise = Promise.all(
+      organisationIds.map(async (org) => ({
+        organisation: org && org.trim() !== '' ? org : 'unset',
+        ...(await calculateOverviewMetricsForOrg(org)),
+      })),
+    )
+
+    const [global, byOrganisation] = await Promise.all([globalPromise, byOrganisationPromise])
+
+    return {
```
JR40159: Does the backend need to return both the grouped and ungrouped set of the same metrics? Could the endpoint have an optional `groupBy` parameter and the frontend can use this parameter to pass in `organisation`. Then I think it might be easier on the client-side to combine the data together. I can understand avoiding wanting to filter stuff in the frontend where possible but I don't think combining data would be that expensive

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+}
+
+/**
+ * Calculates the full set of overview metrics either globally
+ * or scoped to a specific organisation.
+ */
+async function calculateOverviewMetricsForOrg(org?: string): Promise<BaseMetrics> {
+  const modelFilter: ModelFilter = org === undefined ? {} : { organisation: org }
+
+  const totalUsersPromise = org === undefined ? calculateTotalUsers() : Promise.resolve<number | undefined>(undefined)
```
JR40159: So this metric is only available if not grouping by org?

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+    count: countMap.get(schema.id) ?? 0,
+  }))
+}
+
+/**
+ * Calculates the full set of overview metrics either globally
+ * or scoped to a specific organisation.
+ */
+async function calculateOverviewMetricsForOrg(org?: string): Promise<BaseMetrics> {
+  const modelFilter: ModelFilter = org === undefined ? {} : { organisation: org }
```
JR40159: It could be worth defining a `modelMatch` variable here as well and typing the parameter of each of the following functions to either take in a `ModelFilter` or a ModelMatch` depending on if they use an aggregate query or not. Currently it feels like this passes the filter to all the functions but then some of them then need to convert it to a match

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+  ]
+
+  const result = await collection.aggregate(pipeline)
+  return result[0]?.count ?? 0
+}
+
+/**
+ * Returns the number of models grouped by lifecycle state.
+ */
+async function calculateModelsByState(filter: ModelFilter): Promise<StateInfo[]> {
```
JR40159: Would `getModelCountByState` be accurate? Using a more specific term than calculate might be helpful

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+    {
+      $group: {
+        _id: '$card.schemaId',
+        count: { $sum: 1 },
+      },
+    },
+  ]
+
+  type SchemaAgg = { _id: string | null; count: number }
+  const modelCounts = await ModelModel.aggregate<SchemaAgg>(pipeline)
```
JR40159: This doesn't seem to work for me locally

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+  return modelsByState.map((row) => ({
+    state: row._id && row._id.trim() !== '' ? row._id : 'none',
+    count: row.count,
+  }))
+}
+
+/**
+ * Calculates how many models use each schema.
+ */
+async function calculateSchemaBreakdown(filter: ModelFilter): Promise<SchemaInfo[]> {
```
JR40159: There are multiple map/filter calls in this function. Could you use mongo to group the models by schemaID, count the number in each group and lookup the schema name for each schema ID?

backend/src/connectors/metrics/simple.ts
JR40159: Same for `calculatePolicyMetrics`

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+
+    return {
+      global,
+      byOrganisation,
+    }
+  }
+  async calculatePolicyMetrics(): Promise<GetPolicyMetricsResponse> {
+    const { schemaRoleMap, defaultRoles, roleMeta } = await buildSchemaRoleMap()
+
+    const global = await calculateMissingModelRolesForOrg(schemaRoleMap, defaultRoles, roleMeta)
```
JR40159: How come you awaited `global` and `byOrganisation` unlike `calculateOverviewMetrics`?

backend/src/connectors/metrics/simple.ts
JR40159: > Since the `org` parameter is optional and the overview metrics are not always calculated for an org, I'd suggest this function should be called ` calculateOverviewMetrics`

and for `calculateMissingModelRolesForOrg`

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+ * Calculates which models are missing required review roles, either globally
+ * or scoped to a specific organisation.
+ */
+async function calculateMissingModelRolesForOrg(
+  schemaRoleMap: Record<string, string[]>,
+  defaultRoles: string[],
+  roleMeta: Record<string, { roleId: string; roleName: string }>,
+  org?: string,
+): Promise<PolicyMetricsResult> {
+  const filter: any = {}
```
JR40159: Type properly rather than use `any`

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+  }
+}
+
+export async function buildSchemaRoleMap(): Promise<SchemaRoleMap> {
+  const cached = schemaRoleCache.get<SchemaRoleMap>(SCHEMA_ROLE_CACHE_KEY)
+  if (cached !== undefined) {
+    return cached
+  }
+
+  const value = await buildSchemaRoleMapInternal()
```
JR40159: I don't think there's much value in having `buildSchemaRoleMapInternal`

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+ */
+function streamModelsForRoleEvaluation(filter: Record<string, any>) {
+  return ModelModel.find(filter).select('id organisation card collaborators').lean().cursor()
+}
+
+/**
+ * Calculates which models are missing required review roles, either globally
+ * or scoped to a specific organisation.
+ */
+async function calculateMissingModelRolesForOrg(
```
JR40159: Did you consider doing a mongo query for each review role to get a list (or count) of all models that are using the any schema that review role is attached to and does not include that role in the collaborators. I think this would significantly reduce the logic required in this function

backend/src/connectors/metrics/simple.ts
```diff
@@ -0,0 +1,526 @@
...
+    )
+
+    return {
+      global,
+      byOrganisation,
+    }
+  }
+
+  async calculateModelVolume(
+    bucket: ModelVolumeBucket,
```
JR40159: Could we replace the term `bucket` with `interval`?

JR40159 (changes requested):

backend/src/connectors/metrics/simple.ts
ON93754: Another point discussed - perhaps just doing the queries by organisation, then doing a count up to get by global, returning that. this could potentially reduce the need to do the query by each individual org + the global

backend/src/routes/v3/metrics/getComplianceMetrics.ts
ON93754: Avoided this as it then gets messy/introduces circular dependencies - I found it was easiest to keep the types all in here then they could easily be imported by the utils file as well as base

backend/src/routes/v3/metrics/getUsageMetrics.ts
ON93754: Similar to above - i did originally have them closer to their implementation but this gave circular import headaches when trying to use in both the base and metricUtils files - in the end it seemed most simple to keep it here

JR40159 (dismissed):

backend/src/connectors/metrics/base.ts
```diff
@@ -0,0 +1,633 @@
...
+   * Calculates metrics about model usage over time.
+   */
+  async calculateModelVolume(
+    user: UserInterface,
+    interval: ModelVolumeInterval,
+    startDate: string | Date,
+    endDate: string | Date,
+    timezone?: string,
+  ): Promise<GetModelVolumeResponse> {
+    checkUserIsAuthorised(user)
```
PE39806:
```suggestion
    await checkUserIsAuthorised(user)
```

backend/src/connectors/metrics/base.ts
```diff
@@ -0,0 +1,633 @@
...
+   */
+  async getOrganisationIds(): Promise<string[]> {
+    return [...this.organisations, '']
+  }
+
+  /**
+   * Gets metrics around general model usage within Bailo.
+   */
+  async getUsageMetrics(user: UserInterface): Promise<GetUsageMetricsResponse> {
+    const cached = getCached<GetUsageMetricsResponse>(USAGE_METRICS_CACHE_KEY)
```
PE39806: Since `calculateUsageMetrics` uses `user`, the cache should also include the user as, for example, if we end up with different metrics having different roles then different users should not be sharing a single cache.

backend/src/connectors/metrics/base.ts
```diff
@@ -0,0 +1,633 @@
...
+
+const METRICS_CACHE_TTL = 5 * 60 // 5 minutes
+const USAGE_METRICS_CACHE_KEY = 'usageMetrics'
+const COMPLIANCE_METRICS_CACHE_KEY = 'complianceMetrics'
+
+const metricsCache = new NodeCache({
+  stdTTL: METRICS_CACHE_TTL,
+  checkperiod: METRICS_CACHE_TTL,
+  useClones: false,
+})
```
PE39806, line 40: Probably a good idea to use `maxKeys` otherwise the backend could run out of memory from an attempted DoS via the metrics endpoints.

backend/src/connectors/metrics/base.ts
```diff
@@ -0,0 +1,633 @@
...
+    {
+      $match: {
+        kind: SchemaKind.Model,
+        hidden: false,
+      },
+    },
+    // Lookup models that reference this schema
+    {
+      $lookup: {
+        from: 'models',
```
PE39806:
```suggestion
        from: 'v2_models',
```

PE39806 (changes requested):

JR40159 (dismissed):

IR96334 (dismissed):

JR40159 (dismissed):

IR96334 (dismissed):

ARADDCC002:
I have two models, one with no state set, and one set to "Development", but the API is only returning 2 models with no state set via the metrics usage end point

PE39806:
> I have two models, one with no state set, and one set to "Development", but the API is only returning 2 models with no state set via the metrics usage end point

I am unable to reproduce this. Is it possible that the query was cached for this result? If the endpoint was previously hit with both models having state unset then `NodeCache({ stdTTL: METRICS_CACHE_TTL // = 5 * 60 i.e. 5 minutes` will have used the old result.

PE39806 (dismissed):

ON93754:
I also was not able to reproduce that issue mentioned - i think it could have been a stale cache

PE39806 (approved):

JR40159 (approved):

PR #3492, opened 2026-04-22
Improve scanning code and do not await scans on upload

JR40159:


backend/src/models/File.ts
```diff
@@ -29,7 +30,7 @@ export interface FileInterface {
 // object from Mongoose it should use this interface
 export type FileInterfaceDoc = FileInterface & SoftDeleteDocument
 // `id` is used by the python API so we need to keep this to prevent a breaking change
-export type FileWithScanResultsInterface = FileInterface & { scanResults: ScanInterface[]; id: string }
+export type FileWithScanResultsInterface = FileInterface & { scanResults: ScanInterface[] }
```
PE39806: I think that it would be better practice to not include `id` in the MongoDB interface `type`s we have, and to instead use `HydratedDocument` as in the pending mongoose v9 PR https://github.com/gchq/Bailo/pull/3267/changes#diff-cc84e303ea94ac5fe04e9839aff43c8c03defe192ca9d08287a33db250a8dbc4 because `id` is a property coming from mongoose, so is better handled by using the proper mongoose type.

backend/src/connectors/artefactScanning/wrapper.ts
```diff
@@ -56,57 +51,38 @@ export class ArtefactScanningWrapper {
...
+  hasScannerForArtefactKind(kind: ArtefactKindKeys) {
+    const info = this.scannersInfo()
+    return info.some((scannerInfo) => scannerInfo.artefactKind === kind)
   }
 
-  async startScans(artefact: ArtefactInterface) {
+  async startScans({
+    file,
+    layerRef,
+  }: { file: FileInterface; layerRef?: never } | { file?: never; layerRef: LayerRefInterface }) {
```
GB907762, line 62: See here: https://www.typescriptlang.org/docs/handbook/basic-types.html#never

This is a bad type

```suggestion
  async startScans({ file, layerRef }: { file?: FileInterface; layerRef?: LayerRefInterface }) {
```

backend/src/services/scan.ts
```diff
@@ -73,38 +68,50 @@ export async function updateArtefactScanWithResults(
   }
 }
 
-async function runScans(
-  scannersInfo: ArtefactScanningConnectorInfo[],
-  scanIdentifier: ArtefactScanIdentifier,
-  artefact: ArtefactInterface,
-) {
+type RunScansParams = { file: FileInterface; layerRef?: never } | { file?: never; layerRef: LayerRefInterface }
```
GB907762, line 71: This is the same type as the comment above. Could you put this in the `types.ts` file or import from here?


```suggestion
type RunScansParams = { file?: FileInterface; layerRef?: LayerRefInterface }
```

backend/src/connectors/artefactScanning/wrapper.ts
```diff
@@ -56,57 +51,38 @@ export class ArtefactScanningWrapper {
...
-                  },
-                  'Scan started.',
-                )
-                break
-              default:
-                throw InternalError('Incompatible artefact type')
-            }
-
-            return scanner.scan(artefact)
+          if (file && scanner.artefactType === 'file') {
```
GB907762:
```suggestion
          if (file && scanner.artefactType === ArtefactKind.FILE) {
```

backend/src/connectors/artefactScanning/wrapper.ts
```diff
@@ -56,57 +51,38 @@ export class ArtefactScanningWrapper {
...
+              {
+                modelId: file.modelId,
+                fileId: file._id.toString(),
+                name: file.name,
+                toolName: scanner.toolName,
+              },
+              'Requesting scan to be queued',
+            )
+            return scanner.scan(file)
+          } else if (layerRef && scanner.artefactType === 'image') {
```
GB907762:
```suggestion
          } else if (layerRef && scanner.artefactType === ArtefactKind.IMAGE) {
```

backend/src/services/file.ts
```diff
@@ -174,7 +179,12 @@ export async function finishUploadMultipartFile(
 
   file.save()
 
-  return await scanFile(file)
+  runScans({ file }).catch((error) => {
+    log.error({ file, error }, 'Unable to set failure state after failing to run file scans. Safely aborted.')
+  })
+
+  const ret: FileWithScanResultsInterface = Object.assign(file, { scanResults: [] })
+  return ret
```
GB907762, line 187: `ret` is unclear naming and `Object.assign` doesn't create a clone. I'd prefer the following

```suggestion
  return {...file, scanResults:[]}
```

backend/src/services/scan.ts
```diff
@@ -171,13 +167,14 @@ export async function rerunFileScan(user: UserInterface, modelId: string, fileId
     throw Forbidden(auth.info, { userDn: user.dn })
   }
 
-  const scannersInfo = scanners.scannersInfo()
-  throwIfNoScanners(scannersInfo, ArtefactKind.FILE)
+  if (!scanners.hasScannerForArtefactKind('file')) {
```
GB907762:
```suggestion
  if (!scanners.hasScannerForArtefactKind(ArtefactKind.FILE)) {
```

backend/src/services/scan.ts
```diff
@@ -203,6 +200,10 @@ export async function rerunImageScan(user: UserInterface, modelId: string, image
     throw Forbidden(auth.info, { userDn: user.dn })
   }
 
+  if (!scanners.hasScannerForArtefactKind('image')) {
```
GB907762:
```suggestion
  if (!scanners.hasScannerForArtefactKind(ArtefactKind.IMAGE)) {
```

GB907762 (changes requested):

backend/src/connectors/artefactScanning/wrapper.ts
JR40159: Your suggestion would allow for both a file and an image to passed in which would not be good. The point of this type is to only allow one of the parameters to be passed in and is pretty standard practice for union types of this kind

backend/src/services/file.ts
JR40159: Agreed, this came from the existing code but definitely could be improved

backend/src/connectors/artefactScanning/wrapper.ts
GB907762:
```suggestion
  }: { file: FileInterface } | { layerRef: LayerRefInterface }) {
```
Yep good shout!

backend/src/connectors/artefactScanning/wrapper.ts
```diff
@@ -56,57 +51,38 @@ export class ArtefactScanningWrapper {
...
+          } else if (layerRef && scanner.artefactType === ArtefactKind.IMAGE) {
+            log.info(
+              {
+                ...layerRef,
+                toolName: scanner.toolName,
+              },
+              'Requesting scan to be queued',
+            )
+            return scanner.scan(layerRef)
           }
```
PE39806, line 86: Should have some `else` condition to cover a bad `scanner.artefactType` and/or neither `file` & `layerRef` being passed.

backend/src/services/file.ts
```diff
@@ -64,7 +64,12 @@ export async function uploadFile(
 
   await file.save()
 
-  return await scanFile(file)
+  runScans({ file }).catch((error) => {
+    log.error({ file, error }, 'Unable to set failure state after failing to run file scans. Safely aborted.')
+  })
+
+  const ret: FileWithScanResultsInterface = Object.assign(file, { scanResults: [] })
```
PE39806, line 71: This changes the API result, which could be a breaking change. I think that setting `scanResults` to something like `[{ state: ArtefactScanState.InProgress }]` (adapting the array size to match the relevant scanner type count) would "revert" behaviour to not be breaking.

backend/src/services/file.ts
```diff
@@ -174,7 +179,12 @@ export async function finishUploadMultipartFile(
 
   file.save()
 
-  return await scanFile(file)
+  runScans({ file }).catch((error) => {
+    log.error({ file, error }, 'Unable to set failure state after failing to run file scans. Safely aborted.')
+  })
+
+  const ret: FileWithScanResultsInterface = Object.assign(file, { scanResults: [] })
```
PE39806, line 186: As above.

backend/src/connectors/artefactScanning/wrapper.ts
JR40159: The updated type should prevent that from happening, but I'll add it in for belt and braces

backend/src/services/file.ts
JR40159: Would returning an empty array on upload be a breaking change? I'm not sure returning `[{ state: ArtefactScanState.InProgress }]` due to the asynchronous nature we can't say for sure if any scans have been started

GB907762 (approved):

PE39806 (approved):

PR #3498, opened 2026-04-24
Enforce all tags to be lowercase

JR40159:
- [x] Add migration script

### DIFF
--- a/backend/src/migrations/032_lowercase_tags.ts
+++ b/backend/src/migrations/032_lowercase_tags.ts
@@ -2,53 +2,26 @@ import FileModel from '../models/File.js'
 import Model from '../models/Model.js'
 
 export async function up() {
-  await Model.updateMany(
+  const query = { tags: { $exists: true, $type: 'array' }, 'tags.0': { $exists: true } }
+  const lowercaseTagsPipeline = [
     {
-      tags: { $exists: true, $type: 'array' },
-      'tags.0': { $exists: true },
-    },
-    [
-      {
-        $set: {
-          tags: {
-            $setUnion: [
-              {
-                $map: {
-                  input: '$tags',
-                  as: 'tag',
-                  in: { $toLower: '$$tag' },
-                },
+      $set: {
+        tags: {
+          $setUnion: [
+            {
+              $map: {
+                input: '$tags',
+                as: 'tag',
+                in: { $toLower: '$$tag' },
               },
-            ],
-          },
+            },
+          ],
         },
       },
-    ],
-  )
-
-  await FileModel.updateMany(
-    {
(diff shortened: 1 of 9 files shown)


PE39806 (changes requested):
Also apply to:

- [x] `backend/src/routes/v2/model/modelcard/putModelCard.ts`
- [x] `backend/src/routes/v2/model/tags/getPopularTags.ts`

GB907762 (changes requested):
Does this need a migration, and relevant tests?

backend/src/routes/v2/model/file/patchFile.ts
```diff
@@ -18,7 +18,7 @@ export const patchFileSchema = z.object({
     }),
   }),
   body: z.object({
-    tags: z.array(z.string()).optional(),
+    tags: z.array(z.string().toLowerCase()).optional(),
```
PE39806, line 21: If we want to trim whitespace as well, we can do that here too
```suggestion
    tags: z.array(z.string().trim().toLowerCase()).optional(),
```
and the same for the below zod schemas.

backend/src/migrations/032_lowercase_tags.ts
```diff
@@ -0,0 +1,54 @@
...
+                  as: 'tag',
+                  in: { $toLower: '$$tag' },
+                },
+              },
+            ],
+          },
+        },
+      },
+    ],
+  )
```
PE39806: The query and pipelines are identical. Prefer
```typescript
const query = ...
const lowercaseTagsPipeline = ...

await Model.updateMany(query, lowercaseTagsPipeline)
await FileModel.updateMany(query, lowercaseTagsPipeline)
```

PE39806 (changes requested):

ARADDCC002:
Noticed these are all backend changes, how does the frontend handle the same word with different casing? If it already validates it correctly that's fine but I'd like to make sure we're doing both client and server validation

JR40159:
> Noticed these are all backend changes, how does the frontend handle the same word with different casing? If it already validates it correctly that's fine but I'd like to make sure we're doing both client and server validation

When I checked, the frontend wouldn't allow duplicates

backend/src/routes/v2/model/file/patchFile.ts
JR40159: I think that is out of scope for this and we should tick to just doing this change

ARADDCC002 (approved):

PE39806 (approved):

JRB66955 (approved):

GB907762 (approved):

PR #3499, opened 2026-04-24
Improve tag selector

JR40159:


frontend/src/common/MultiFileInputFileDisplay.tsx
```diff
@@ -1,11 +1,9 @@
-import { LocalOffer } from '@mui/icons-material'
-import { Button, Chip, Grid, TextField, Tooltip, Typography } from '@mui/material'
+import { Chip, Grid, TextField, Tooltip, Typography } from '@mui/material'
 import { useGetModelFiles } from 'actions/entry'
 import { patchFile } from 'actions/file'
 import prettyBytes from 'pretty-bytes'
 import { ChangeEvent, useState } from 'react'
-import Restricted from 'src/common/Restricted'
-import EntryTagSelector from 'src/entry/model/releases/EntryTagSelector'
+import EntryTagSelector from 'src/common/TagSelector'
```
GB907762: By convention components are imported as the same keyword, to limit confusion.

```suggestion
import TagSelector from 'src/common/TagSelector'
```

frontend/src/common/TagSelector.tsx
GB907762, line 1: What was the problem with reusing the multiselect in MUI?

https://mui.com/material-ui/react-select/#multiple-select

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+}
+
+interface EntryTagSelectorProps {
+  onChange: (newTag: string[]) => void
+  tags: string[]
+  errorText?: string
+  restrictedToAction?: RestrictedActionKeys
+}
+
+const StyledAutocompletePopper = styled('div')(({ theme }) => ({
```
GB907762: There's some problematic styling with the popover that causes it to move around the page or cover the model details if there's more than three entries.

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+      color: '#58a6ff',
+    }),
+  },
+  '& span': {
+    width: '100%',
+  },
+  '& svg': {
+    width: 16,
+    height: 16,
+  },
```
GB907762: Instead of accessing the `svg` element for the whole parent, use the `fontSize` property within the Icon component.

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+              }}
+              disableCloseOnSelect
+              renderValue={() => null}
+              noOptionsText='No tags'
+              renderOption={(props, option, { selected }) => {
+                const { key, ...optionProps } = props
+                return (
+                  <li key={key} {...optionProps}>
+                    <Box
+                      component={Done}
```
GB907762: A Done and an Close icon implies conflicting actions. I'd stick with Close.

frontend/src/entry/EntryOverviewDetails.tsx
```diff
@@ -43,15 +41,19 @@ export default function EntryOverviewDetails({ entry }: OrganisationAndStateDeta
...
-  const handleEntryTagOnChange = async (newTags: string[]) => {
-    setEntryTagUpdateErrorMessage('')
-    const response = await patchEntry(entry.id, { tags: newTags })
-    if (!response.ok) {
-      setEntryTagUpdateErrorMessage(await getErrorMessage(response))
-    } else {
-      mutateEntry()
-    }
-  }
+  const handleEntryTagOnChange = useCallback(
```
GB907762: This seems to be only called when you close the popover causing it to not update when new values are added.

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+                        visibility: selected ? 'visible' : 'hidden',
+                      }}
+                    />
+                  </li>
+                )
+              }}
+              options={[...new Set([...popularTags, ...tags, ...pendingValue])]}
+              renderInput={(params) => (
+                <Box sx={{ display: 'flex', justifyContent: 'center', width: '95%', height: '95%', mx: 'auto', pt: 1 }}>
+                  <TextField {...params} size='small' autoFocus placeholder='Filter existing or create new tag' />
```
GB907762: This text overflows.

I'd suggest something like 


```suggestion
                  <TextField {...params} size='small' autoFocus placeholder='Add tag' />
```

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  const open = Boolean(anchorEl)
+  const id = open ? 'tag-selector' : undefined
+
+  return (
+    <Box sx={{ p: 1 }}>
+      <Box sx={{ fontSize: 13 }}>
+        <Button
+          disabled={restrictedToAction && !userPermissions[restrictedToAction].hasPermission}
```
PE39806: Is it possible for `restrictedToAction` to be defined but `userPermissions[restrictedToAction]` be `undefined`? If so, this could throw.
Possibly change to something like
```typescript
const hasPermission =
  restrictedToAction
    ? userPermissions?.[restrictedToAction]?.hasPermission === true
    : true

disabled={!hasPermission}
```
This result could be reused below as `{restrictedToAction && hasPermission ? (`

GB907762 (changes requested):
I think some of the test are failing since I removed some of the tests. Syncing with main should resolve them.

frontend/src/entry/EntryOverviewDetails.tsx
```diff
@@ -43,15 +41,19 @@ export default function EntryOverviewDetails({ entry }: OrganisationAndStateDeta
...
+    async (newTags: string[]) => {
+      setEntryTagUpdateErrorMessage('')
+      const response = await patchEntry(entry.id, { tags: newTags })
+      if (!response.ok) {
+        setEntryTagUpdateErrorMessage(await getErrorMessage(response))
+        mutateEntry()
+      } else {
+        mutateEntry()
```
PE39806: `mutateEntry()` should not be in both branches

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+        <Grid
+          container
+          spacing={1}
+          sx={{
+            alignItems: 'center',
+          }}
+        >
+          {tags.map((label) => (
+            <Grid key={label}>
+              <Chip variant='filled' label={label} key={label} />
```
PE39806: `key` should not be assigned twice.

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+  },
+  '& svg': {
+    width: 16,
+    height: 16,
+  },
+}))
+
+export default function TagSelector({ onChange, tags, errorText = '', restrictedToAction }: EntryTagSelectorProps) {
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null)
+  const [pendingValue, setPendingValue] = React.useState<string[]>([])
```
GB907762:
```suggestion
  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null)
  const [pendingValue, setPendingValue] = useState<string[]>([])
```

Don't import React from a wildcard alias.

GB907762 (changes requested):

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+  const { tags: popularTags, isTagsError } = useGetPopularEntryTags()
+  const theme = useTheme()
+
+  const handleClick = (event: React.MouseEvent<HTMLElement>) => {
+    setPendingValue(tags)
+    setAnchorEl(event.currentTarget)
+  }
+
+  const handleClose = () => {
+    onChange(pendingValue)
```
PE39806: Do we want to do any client side sanitisation e.g. trimming?

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+                      component={Close}
+                      sx={{ opacity: 0.6, width: 18, height: 18 }}
+                      style={{
+                        visibility: selected ? 'visible' : 'hidden',
+                      }}
+                    />
+                  </li>
+                )
+              }}
+              options={[...new Set([...popularTags, ...tags, ...pendingValue])]}
```
PE39806: Can this be memoised to avoid recreating every render?

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+                  <TextField {...params} size='small' autoFocus placeholder='Filter existing or create new tag' />
+                </Box>
+              )}
+              slots={{
+                popper: PopperComponent,
+              }}
+            />
+            <Typography variant='caption' color={theme.palette.error.main}>
+              {errorText}
+              {isTagsError?.info.message}
```
PE39806: These will be concatenated with no whitespace between them. Could this be something like
```suggestion
              {[errorText, isTagsError?.info?.message].filter(Boolean).join(' ')}
```

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+import Popper from '@mui/material/Popper'
+import { styled, useTheme } from '@mui/material/styles'
+import { useGetPopularEntryTags } from 'actions/entry'
+import { useContext } from 'react'
+import * as React from 'react'
+import UserPermissionsContext from 'src/contexts/userPermissionsContext'
+import { RestrictedActionKeys } from 'types/types'
+
+interface PopperComponentProps {
+  anchorEl?: any
```
PE39806:
```suggestion
  anchorEl?: HTMLElement | null

```

PE39806 (changes requested):

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+  '& span': {
+    width: '100%',
+  },
+  '& svg': {
+    width: 16,
+    height: 16,
+  },
+}))
+
+export default function TagSelector({ onChange, tags, errorText = '', restrictedToAction }: EntryTagSelectorProps) {
```
IR96334: Maybe find different name for this component, as it shares one with muiForms/TagSelector

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+  const handleClose = () => {
+    onChange(pendingValue)
+    if (anchorEl) {
+      anchorEl.focus()
+    }
+    setAnchorEl(null)
+  }
+
+  const open = Boolean(anchorEl)
+  const id = open ? 'tag-selector' : undefined
```
IR96334: What is the function of this variable? Why does it need to be undefined?

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+  fontSize: 13,
+  ...theme.applyStyles('dark', {
+    border: '1px solid #30363d',
+    boxShadow: '0 8px 24px rgb(1, 4, 9)',
+    color: '#c9d1d9',
+    backgroundColor: '#1c2128',
+  }),
+}))
+
+const Button = styled(ButtonBase)(({ theme }) => ({
```
IR96334: Been testing the removing and editing of some of this styling. Some don't seem to do much, e.g. removing or editing or textAlign.

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+  return (
+    <Box sx={{ p: 1 }}>
+      <Box sx={{ fontSize: 13 }}>
+        <Button
+          disabled={restrictedToAction && !userPermissions[restrictedToAction].hasPermission}
+          disableRipple
+          aria-describedby={id}
+          onClick={handleClick}
+          style={{ fontWeight: 'bold', color: theme.palette.primary.main, width: 'auto' }}
```
ARADDCC002: You can add `color='primary'` as a regular button prop which means you don't need to use `theme` directly

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+            </Grid>
+          ))}
+        </Grid>
+      </Box>
+      <StyledPopper id={id} open={open} anchorEl={anchorEl} placement='bottom-start'>
+        <ClickAwayListener onClickAway={handleClose}>
+          <Box sx={{ p: 3 }}>
+            <Box
+              sx={(t) => ({
+                borderBottom: '1px solid #30363d',
```
ARADDCC002: Would prefer to split out the `borderBottom` properties into their three separate properties `borderBottomColour` `borderBottomStyle` `borderBottomWidth`. 

Also, using hex codes directly can affect future theme changes, so if we really need to use new custom colours it's best that we add them to our theme file. Ideally we should use existing theme colours if possible

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+              })}
+            >
+              Add a new tag or select an existing popular tag
+            </Box>
+            <Autocomplete
+              freeSolo
+              disableClearable={true}
+              open
+              multiple
+              onClose={(event: React.ChangeEvent<object>, reason: AutocompleteCloseReason) => {
```
ARADDCC002: Remove `React.ChangeEvent` and import `ChangeEvent` from React to keep it consistent with the rest of the app

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+                if (reason === 'escape') {
+                  handleClose()
+                }
+              }}
+              value={pendingValue}
+              onChange={(event, newValue, reason) => {
+                if (
+                  event.type === 'keydown' &&
+                  ((event as React.KeyboardEvent).key === 'Backspace' ||
```
ARADDCC002: Change these two lines so that they import `KeyboatdEvent` directly from `React`

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+                        visibility: selected ? 'visible' : 'hidden',
+                      }}
+                    />
+                    <Box
+                      component='span'
+                      sx={{
+                        width: 14,
+                        height: 14,
+                        flexShrink: 0,
+                        borderRadius: '3px',
```
ARADDCC002: Is there a reason we're using `3px` here? Would be better if we can use the theme spacing for this to keep it consistent with other border radius' we use for consistency.

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+                        '& span': {
+                          color: '#8b949e',
+                          ...t.applyStyles('light', {
+                            color: '#586069',
+                          }),
+                        },
+                      })}
+                    >
+                      {option}
+                      <br />
```
ARADDCC002: Better to use `margin` and theme spacing instead of `<br />`

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+              renderInput={(params) => (
+                <Box sx={{ display: 'flex', justifyContent: 'center', width: '95%', height: '95%', mx: 'auto', pt: 1 }}>
+                  <TextField {...params} size='small' autoFocus placeholder='Filter existing or create new tag' />
+                </Box>
+              )}
+              slots={{
+                popper: PopperComponent,
+              }}
+            />
+            <Typography variant='caption' color={theme.palette.error.main}>
```
ARADDCC002: We can use `color='error'` as a base `<Typography />` prop instead of needing to import `theme`

ARADDCC002 (changes requested):

frontend/src/common/TagSelector.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+function PopperComponent(props: PopperComponentProps) {
+  const { disablePortal: _disablePortal, anchorEl: _anchorEl, open: _open, ...other } = props
+  return <StyledAutocompletePopper {...other} />
+}
+
+const StyledPopper = styled(Popper)(({ theme }) => ({
+  border: '1px solid #e1e4e8',
+  boxShadow: `0 8px 24px ${'rgba(149, 157, 165, 0.2)'}`,
+  color: '#24292e',
```
IR96334: Use theme colours here, and throughout. One of the problems with this method is its missing some theme standardisations, like borderRadius will be 6px not 6*(whatever the theme defined spacing value is)

frontend/src/common/TagSelector.tsx
ON93754: Agree on trimming - i'd suggest creating a utility function + unit tests outside of this component for re-use elsewhere

frontend/src/common/TagSelector.tsx
JR40159: The main reason was being able to use the example include in the Autocomplete MUI docs 😁 but I also don't know if multiselect allows both free text input as well as selecting options?

frontend/src/entry/EntryOverviewDetails.tsx
JR40159: This is intentional to ensure that when multiple tags are added each tag is not added in a separate request to the backend

frontend/src/common/TagSelector.tsx
JR40159: Improved to use the MUI type

frontend/src/common/TagSelector.tsx
JR40159: How about renaming the other one to `FormTagSelector`? As this one is reused I think it's more difficult to rename

frontend/src/common/TagSelector.tsx
JR40159: We seem to use a mix of both in here?
https://github.com/gchq/Bailo/blob/672422686164dba7a687479f7b0ec325d9b00bc8/frontend/src/entry/EntryOverviewDetails.tsx

JR40159: I've just removed `color` as I don't think it's needed for the MUI Button

backend/src/services/model.ts
```diff
@@ -879,7 +879,13 @@ export async function getModelSystemRoles(user: UserInterface, model: ModelDoc)
     .flat()
 }
 
-export async function popularTagsForEntries() {
-  const tags = await ModelModel.aggregate([{ $unwind: '$tags' }, { $sortByCount: '$tags' }, { $limit: 10 }])
+export async function popularTagsForEntries(limit: number = 10) {
```
PE39806, line 934: Since the zod schema for `limit` is `optional`, this default of `10` will be quietly set when no `limit` is specified. It is better to set the default value on the endpoint where it is visible via the swagger docs, rather than doing this behaviour.

PE39806 (changes requested):

ARADDCC002:
When attempting to add a tag that is already on the model, the selector will say "No options", should this be "Tag already in use" or something similar?

ARADDCC002:
* Tags when uploading files looks messy
* When entering a new tag it would be nice if we can press enter to select it

GB907762:
https://mui.com/material-ui/react-autocomplete/#multiple-values

Could we try the `freeSolo` version of this. This has been lingering for a while. I'm not a fan of the current UX

PR #3509, opened 2026-04-28
Backend scanners early skip file type and size

PE39806:
- Add early skip to too large streams in all backend scanners
- Add new `ArtefactScanState.Skipped` state enum to DB and frontend
- Add `maxFileSizeBytes` and `modelscanSupportedExtensions` to ArtefactScan `GET /info` endpoint
- Add early skip to non-supported file types in backend `ModelScan` scanner
- Add `streamMaxLength` helm var
- Refactor some ArtefactScan type & functions, and improve test isolation & coverage
- Refactor `backend/src/connectors/artefactScanning/*.ts` to simplify `init`
- Better pass through ArtefactScan errors to the backend
- Replace backend `pretty-bytes` package with `bytes` (already a sub-dependency)
- Add backend unit tests
- Remove redundant `vi.clearAllMocks()` calls (covered by `backend/test/testUtils/clearMocks.ts`)


backend/src/connectors/artefactScanning/clamAv.ts
```diff
@@ -9,39 +9,34 @@ import config from '../../utils/config.js'
...
 function safeParseVersion(versionStr: string): string {
-  try {
-    const match = versionStr.match(/ClamAV\s([\d.]+)\//)
-    if (match && match[1]) {
-      return match[1]
-    }
-    return versionStr
-  } catch {
-    return versionStr
+  const match = versionStr.match(/ClamAV\s([\d.]+)\//)
```
ON93754: Comment or function docstring explaining what this regex is looking for?

backend/src/clients/registry.ts
```diff
@@ -346,14 +346,18 @@ export async function getRegistryLayerStream(
   return { stream: result.stream, abort: result.abort }
 }
 
+export async function headLayer(token: string, repoRef: ImageNameRef, digest: string) {
```
ON93754: Could we please have a short docstring for this function? Possibly even a more descriptive name? Seems a little vague

backend/src/connectors/artefactScanning/clamAv.ts
```diff
@@ -6,46 +6,72 @@ import { FileInterfaceDoc } from '../../models/File.js'
...
+  // extract 'x.y.z' from 'ClamAV x.y.z/abc'
+  const match = versionStr.match(/ClamAV\s([\d.]+)/)
+  if (match && match[1]) {
+    return match[1]
+  }
+  return versionStr
+}
+
+function safeParseSize(sizeStr: string): number {
+  const match = sizeStr.trim().match(/^(\d+)\s*([KMG])?B?$/i)
```
ON93754: Comment for what this regex is matching for ?

lib/artefactscan_api/bailo_artefactscan_api/modelscan.py
```diff
@@ -0,0 +1,65 @@
...
+    supported.update(settings.get("supported_zip_extensions", []) or [])
+
+    scanners = settings.get("scanners") or {}
+    if isinstance(scanners, dict):
+        for scanner in scanners.values():
+            if isinstance(scanner, dict):
+                supported.update(scanner.get("supported_extensions", []) or [])
+
+    middlewares = settings.get("middlewares") or {}
+    if isinstance(middlewares, dict):
```
ON93754, line 37: Perhaps a comment like
`# Check for any extensions and if found then add them to the supported set`

backend/src/clients/registry.ts
PE39806: This performs a `HEAD` request against a `layer` reference in the registry. I don't see how else to word this so I would welcome any suggestions.

backend/src/connectors/artefactScanning/Base.ts
```diff
@@ -26,19 +28,24 @@ export type ArtefactScanningConnectorInfo = Pick<ArtefactScanResult, 'toolName'
...
+
   info(): ArtefactScanningConnectorInfo {
     return { toolName: this.toolName, scannerVersion: this.version, artefactKind: this.artefactType }
   }
 
-  abstract init()
+  abstract init(): Promise<void>
 
-  abstract _scan(artefact: ArtefactInterface): Promise<ArtefactScanResult>
+  protected abstract _scan(artefact: ArtefactInterface): Promise<ArtefactScanResult>
```
JR40159: Happy for this to be out of scope for this PR, but I'd like to consider renaming `_scan` to make it obvious what the difference is between `_scan` and `scan`

backend/src/clients/artefactScan.ts
```diff
@@ -6,12 +6,14 @@ import fetch, { Response as FetchResponse } from 'node-fetch'
...
 import config from '../utils/config.js'
-import { BadReq, InternalError } from '../utils/error.js'
+import { GenericError, InternalError } from '../utils/error.js'
 
 const ArtefactScanInfoResponseSchema = z.object({
   apiName: z.string(),
   apiVersion: z.string(),
+  maxFileSizeBytes: z.number().int().positive().optional(),
   modelscanScannerName: z.string(),
+  modelscanSupportedExtensions: z.array(z.string()).optional(),
```
GB907762, line 16: Is it optional. The response schema suggests it isn't?

backend/src/clients/registry.ts
GB907762: You could give the link to the endpoint it implements?

https://docs.docker.com/reference/api/registry/latest/#tag/Blobs/operation/CheckBlobExists

backend/src/connectors/artefactScanning/clamAv.ts
```diff
@@ -5,47 +5,76 @@ import { getObjectStream } from '../../clients/s3.js'
...
   constructor() {
     super()
   }
 
-  async init() {
+  async init(): Promise<void> {
     this.av = await new NodeClam().init({ clamdscan: config.artefactScanning.clamdscan })
     const scannerVersion = await this.av.getVersion()
     this.version = safeParseVersion(scannerVersion)
+    this.maxSize = safeParseSize(config.artefactScanning.clamdscan.streamMaxLength)
```
GB907762: Could you use `bytes` here? We use it as a dependency? 

I just think this might be a more robust choice?
```suggestion
    this.maxSize = bytes(config.artefactScanning.clamdscan.streamMaxLength)
```

We could also use this and drop pretty-bytes, as a backend dependency?

backend/src/connectors/artefactScanning/modelScan.ts
```diff
@@ -70,4 +98,11 @@ export class ModelScanFileScanningConnector extends BaseArtefactScanningConnecto
       }
     }
   }
+
+  protected skipUnsupportedFileType(fileName: string): ArtefactScanResult {
+    return this.scanSkip([
+      UNSUPPORTED_FILE_TYPE_MESSAGE,
+      `${fileName} not covered in [${this.supportedExtensions.join(', ')}].`,
```
GB907762: I'm not super keen on having it explicitly show what are the supportedExtensions? For example, what happens if this array is empty? There's no context provided.

For example:
`this-cool-image.png not covered in [].`

backend/src/connectors/artefactScanning/trivy.ts
```diff
@@ -11,35 +11,44 @@ import {
...
 
+      if (this.maxSize != Infinity) {
+        const layerHeadDetails = await headLayer(
+          repositoryToken,
+          { repository: layer.repository, name: layer.name },
+          layer.layerDigest,
+        )
+        layerSize = parseInt(layerHeadDetails.headers['content-length'] || '') || Infinity
+
+        if (layerSize != Infinity && layerSize > this.maxSize) {
```
GB907762: Does this not cover when the maxSize is unset in config?

```suggestion
        if (layerSize > this.maxSize) {
```

backend/src/connectors/artefactScanning/wrapper.ts
```diff
@@ -12,37 +12,35 @@ export class ArtefactScanningWrapper {
...
-          }
-        }
-        if (attempt > config.connectors.artefactScanners.maxInitRetries) {
+    await Promise.all(Array.from(this.scanners).map((scanner) => this.initialiseScanner(scanner)))
+  }
+
+  private async initialiseScanner(scanner: BaseArtefactScanningConnector) {
+    log.info({ toolName: scanner.toolName }, 'Scanner initialising...')
+
+    for (let attempt = 1; attempt <= config.connectors.artefactScanners.maxInitRetries; attempt++) {
```
GB907762, line 21: Out of scope, but exponential backoff could be handy?

backend/src/connectors/artefactScanning/wrapper.ts
```diff
@@ -12,37 +12,35 @@ export class ArtefactScanningWrapper {
...
+              error,
+            },
           )
+        } else {
+          log.warn({ attempt, toolName: scanner.toolName, error }, 'Could not initialise scanner, retrying.')
         }
-      }),
-    )
+
+        await new Promise((r) => setTimeout(r, config.connectors.artefactScanners.initRetryDelay))
```
GB907762:
```suggestion
        await new Promise((callback) => setTimeout(callback, config.connectors.artefactScanners.initRetryDelay))
```
Clearer naming?

frontend/src/entry/model/files/FileDisplay.tsx
```diff
@@ -91,10 +99,195 @@ type FileDisplayProps = {
...
+        icon: <Warning />,
+      }
+  }
+}
+
+/**
+ * Renders a single scanner's result block inside the popover.
+ * Uses the scanner's own state (not the aggregated state) so each row shows the correct icon/wording for that individual tool.
+ */
+function ScanResultDetail({ scanResult }: { scanResult: ScanResultInterface }) {
```
GB907762: Could you put this component in a separate file so it can be used in other areas. I really don't like the fact that this file is 700 lines.

frontend/src/entry/model/files/FileDisplay.tsx
```diff
@@ -91,10 +99,195 @@ type FileDisplayProps = {
...
+        <Chip
+          size='small'
+          variant='outlined'
+          sx={{ width: 'fit-content' }}
+          label={`Scanner version: ${scannerVersion}`}
+        />
+      )}
+      <Typography>Last ran at: {formatDateTimeString(lastRunAt)}</Typography>
+      {hasFindings && summary && (
+        <ul>
```
GB907762: MUI `List` and `ListItem` might be more appropriate to include styling?

I understand this is reused code. So feel free to suggest this is out of scope?

frontend/src/entry/model/files/FileDisplay.tsx
```diff
@@ -244,118 +396,72 @@ export default function FileDisplay({
...
-          />
-        )}
+        <Chip
+          color={chipDisplay.colour}
+          icon={chipDisplay.icon}
+          onClick={(e) => setAnchorElScan(e.currentTarget)}
+          label={chipDisplay.label}
+        />
         <Popover
-          hidden={file.scanResults && file.scanResults.some((res) => res.state === ScanState.InProgress)}
```
GB907762, line 286: Could you replace this. There's currently an annoying bug where the pop over appears with no elements.

Image

lib/artefactscan_api/bailo_artefactscan_api/modelscan.py
```diff
@@ -0,0 +1,69 @@
...
+    - scanners[*].supported_extensions
+    - middlewares[*].formats[*]
+
+    :param settings: modelscan settings config dict to parse
+    :return: set of file extensions (lowercase, including leading dot) that modelscan will run on
+    """
+    supported: set[str] = set()
+
+    # Top-level zip extensions
+    supported.update(settings.get("supported_zip_extensions", []) or [])
```
GB907762:
```suggestion
    supported.update(settings.get("supported_zip_extensions", []))
```
This is a bit petty but does this matter?

lib/artefactscan_api/bailo_artefactscan_api/modelscan.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # Extensions from each scanner
+    scanners = settings.get("scanners") or {}
```
GB907762:
```suggestion
    scanners = settings.get("scanners", {})
```
Same here?

lib/artefactscan_api/bailo_artefactscan_api/modelscan.py
```diff
@@ -0,0 +1,69 @@
...
+    if isinstance(scanners, dict):
+        for scanner in scanners.values():
+            if isinstance(scanner, dict):
+                supported.update(scanner.get("supported_extensions", []) or [])
+
+    # Extensions from each middleware's format map
+    middlewares = settings.get("middlewares") or {}
```
GB907762: Here as well?
```suggestion
    middlewares = settings.get("middlewares, {}")
```

lib/artefactscan_api/bailo_artefactscan_api/modelscan.py
```diff
@@ -0,0 +1,69 @@
...
+    if isinstance(middlewares, dict):
+        for middleware in middlewares.values():
+            if isinstance(middleware, dict):
+                formats = middleware.get("formats") or {}
+                if isinstance(formats, dict):
+                    for extensions in formats.values():
+                        if isinstance(extensions, (list, set, tuple)):
+                            supported.update(extensions)
```
GB907762: https://peps.python.org/pep-0020/
`Flat is better than nested.`

```suggestion
    if isinstance(scanners, dict):
        for scanner in scanners.values():
            if not isinstance(scanner, dict):
                continue
            supported.update(scanner.get("supported_extensions", []) or [])
    
    # Extensions from each middleware's format map
    middlewares = settings.get("middlewares") or {}
    if isinstance(middlewares, dict):
        for middleware in middlewares.values():
            if not isinstance(middleware, dict):
                continue
    
            formats = middleware.get("formats") or {}
            if not isinstance(formats, dict):
                continue
    
            for extensions in formats.values():
                if not isinstance(extensions, (list, set, tuple)):
                    continue
                supported.update(extensions)
```

backend/src/clients/artefactScan.ts
```diff
@@ -191,6 +193,24 @@ export const TrivyScanResultResponseSchema = z
...
 
+async function handleBadResponse(res: FetchResponse, context?: Record<string, unknown>) {
+  const contentType = res.headers.get('content-type') ?? ''
+  let responseBody: unknown
+
+  try {
+    responseBody = contentType.includes('application/json') ? await res.json() : await res.text()
+  } catch {
+    responseBody = '<unable to parse response body>'
+  }
```
GB907762, line 204: Would it be better rethrowing the error with a useful message?
```suggestion
  } catch (error){
    this.scanError(error)
  }
```

frontend/src/entry/model/files/FileDisplay.tsx
```diff
@@ -91,10 +99,195 @@ type FileDisplayProps = {
...
+      default:
+        return (
+          <>
+            <Done color='success' />
+            <Typography>
+              <span style={{ fontWeight: 'bold' }}>{toolName}</span> found no issues
+            </Typography>
+          </>
+        )
+    }
```
GB907762:
```suggestion
const states = {
  [ArtefactScanState.InProgress]: {
    Icon: HourglassTop,
    color: 'warning',
    text: 'is currently running',
  },
  [ArtefactScanState.NotScanned]: {
    Icon: Warning,
    color: 'warning',
    text: 'could not be run',
  },
  [ArtefactScanState.Error]: {
    Icon: Warning,
    color: 'warning',
    text: 'could not be run',
  },
  [ArtefactScanState.Skipped]: {
    Icon: BlockOutlined,
    color: 'action',
    text: 'was skipped',
  },
  [ArtefactScanState.Complete]: {
    Icon: Done,
    color: 'success',
    text: 'found no issues',
  },
}

const { Icon, color, text } =
  stateMap[state] ?? stateMap[ArtefactScanState.Complete]

return (
  <>
    <Icon color={color} />
    <Typography>
      <strong>{toolName}</strong> {text}
    </Typography>
  </>
)
```
May be cleaner/clearer

- Also note that the `<strong/>` element over `<span style={{ fontWeight: 'bold' }}>` signals semantic importance for accessibility. Disregard if needed?
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/strong

- Could you also remove the colons from the two findings chips to make it consistent?

GB907762 (changes requested):

backend/src/clients/artefactScan.ts
PE39806: It is required to be sent in the latest version of ArtefactScan, but I have made it optional to for the Backend to receive this to maintain backward compatibility.

backend/src/clients/artefactScan.ts
PE39806: The `responseBody` ends up in the later `GenericError` which ends up getting caught and handled as a scan error anyway. I'd rather keep this function simpler by doing less extra handling.

backend/src/connectors/artefactScanning/wrapper.ts
PE39806: Agreed, but won't do in this ticket.

GB907762 (approved):

JR40159 (approved):

PR #3513, opened 2026-04-29
Feature/bai 2359 metrics dashboard

ARADDCC002:


frontend/actions/metrics.ts
```diff
@@ -0,0 +1,93 @@
...
+    bucket,
+    ...(startDate != undefined && { startDate }),
+    ...(endDate != undefined && { endDate }),
+  }
+  const { data, isLoading, error, mutate } = useSWR<
+    {
+      data: ModelVolume
+      ErrorInfo
+    }[]
+  >(`/api/v3/metrics/modelVolume?${qs.stringify(queryParams)}`, fetcher)
```
PE39806: `ErrorInfo` should not be inside this object.

frontend/types/types.ts
```diff
@@ -918,3 +919,76 @@ export type ImageScanResults = {
...
+
+export type ModelVolumeData = {
+  startDate: string
+  endDate: string
+  count: number
+  organisations: Record<string, number>
+}
+
+export type ModelVolume = {
+  data: ModelVolumeData
```
PE39806:
```suggestion
  data: ModelVolumeData[]
```

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,258 @@
...
+            label: schemaItem.schemaName,
+            value: schemaItem.count,
+          }
+        }),
+      )
+    }
+  }, [data.schemaBreakdown])
+
+  const updateModelVolume = useEffectEvent((modelVolume) => {
+    if (modelVolume && modelVolume.data) {
```
PE39806: Earlier return is `modelVolume: data ? data : []` so `modelVolume` is always defined.
```suggestion
    if (modelVolume?.data) {
```

frontend/types/types.ts
```diff
@@ -1,5 +1,6 @@
 import { UiSchema } from '@rjsf/utils'
 import { Dispatch, JSX, SetStateAction } from 'react'
+import OverviewMetrics from 'src/metrics/OverviewMetrics'
```
PE39806: Why import this?

frontend/actions/metrics.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export function useGetVolumeForModel(bucket: string = 'month', startDate?: string, endDate?: string) {
+  const queryParams = {
+    bucket,
+    ...(startDate != undefined && { startDate }),
+    ...(endDate != undefined && { endDate }),
+  }
+  const { data, isLoading, error, mutate } = useSWR<
+    {
+      data: ModelVolume
```
PE39806: Should this be an array?

PE39806 (changes requested):

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,258 @@
...
+  onSelectedOrganisationChange,
+}: OverviewMetricsChartsProps) {
+  const [stateData, setStateData] = useState<PieChartData[]>([])
+  const [schemaData, setSchemaData] = useState<PieChartData[]>([])
+  const [structuredModelVolume, setStructuredModelVolume] = useState<any[]>([])
+  const [startDate, setStartDate] = useState<Dayjs | null>(null)
+  const [endDate, setEndDate] = useState<Dayjs | null>(dayjs(new Date()))
+  const [errorMessage, setErrorMessage] = useState('')
+
+  const setAsFirstDayOfMonth = (date: Dayjs) => {
```
ON93754: Is it worth moving these date utility functions to some shared file, e.g a `utils.ts` or `dateUtils.ts` to make it easier for other files to use down the line?

ON93754: Also return type?

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,258 @@
...
+          }
+        }),
+      )
+    }
+  }, [data.schemaBreakdown])
+
+  const updateModelVolume = useEffectEvent((modelVolume) => {
+    if (modelVolume && modelVolume.data) {
+      const updatedStructure = modelVolume.data.map((volumeData: ModelVolumeData) => {
+        const formattedDate = dayjs(volumeData.startDate).format('MMM YYYY')
```
ON93754: Perhaps another function of the format `asYear` + group it in the same utilities file as the above date functions?

frontend/actions/metrics.ts
ARADDCC002: Nope it's an object

frontend/package.json
PE39806, line 1: Why so many downgrades?

PE39806 (changes requested):

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,253 @@
...
+            openTo='month'
+            views={['year', 'month']}
+            value={startDate}
+            onChange={(newValue) => {
+              setStartDate(newValue)
+              setErrorMessage('')
+            }}
+          />
+          <Typography fontWeight='bold' variant='h6' color='primary'>
+            &
```
IR96334: Ampersand here makes sense, but feels out of place. Maybe try a dash? If it feels too small, could try playing around with fontsize/variant.

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  if (isModelVolumeLoading) {
+    return <Loading />
+  }
+
+  return (
+    <Stack spacing={6} divider={<Divider flexItem />}>
+      <Stack spacing={2}>
+        <Stack direction='row' spacing={2} alignItems='center'>
```
IR96334: `<Stack direction={{sm: 'column', md: 'row'}} ...`
For responsiveness at smaller viewports.

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,253 @@
...
+            <OverviewStatPanel label='Models with releases' value={data.withReleases} minWidth='300px' />
+            <OverviewStatPanel label='Models with access requests' value={data.withAccessRequest} minWidth='320px' />
+          </Stack>
+          <Stack
+            spacing={2}
+            direction={{ lg: 'row', md: 'column' }}
+            justifyContent='space-around'
+            sx={{ width: '100%' }}
+          >
+            <Stack spacing={2}>
```
IR96334: When in columns, the pie charts aren't centered around the pie chart itself, but the pie + key table, making the chart slightly askew.

frontend/src/metrics/PolicyMetricsCharts.tsx
```diff
@@ -0,0 +1,74 @@
...
+      )
+    })
+  }, [data.summary])
+
+  const EntryRow = memoize(({ data: entry }: { data: PolicyModelMetrics }) => {
+    return (
+      <Grid container alignItems='center'>
+        <Grid size={{ sm: 6, xs: 12 }} sx={{ pl: 2 }}>
+          <Box>
+            <Link href={`/model/${entry.entryId}`}>{entry.entryId}</Link>
```
IR96334: Suggestion: As this is relating to model access, could route directly to that bit? With query parameters: `?tab=settings&category=permissions`

frontend/src/metrics/OverviewMetrics.tsx
```diff
@@ -0,0 +1,80 @@
...
+import MessageAlert from 'src/MessageAlert'
+import OverviewMetricsCharts from 'src/metrics/OverviewMetricsCharts'
+import { OverviewBaseMetrics } from 'types/types'
+
+export default function OverviewMetrics() {
+  const router = useRouter()
+
+  const { organisationFromRouter } = router.query
+
+  const { overviewMetrics, isOverviewMetricsLoading, isOverviewMetricsError } = useGetGetOverviewMetrics()
```
IR96334: GetGet

frontend/package.json
ARADDCC002: It's probably not aligned with `main`

frontend/src/metrics/OverviewMetricsCharts.tsx
ARADDCC002: Yeah this has proven quite difficult to align because of how MUI size the charts. Will give it a go though.

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,220 @@
...
+          <Stack spacing={2}>
+            <OverviewStatPanel label='Total entries' value={data.entries} minWidth='300px' />
+            <OverviewStatPanel label='Entries with releases' value={data.withReleases} minWidth='300px' />
+            <OverviewStatPanel label='Entries with access requests' value={data.withAccessRequest} minWidth='320px' />
+          </Stack>
+          <Stack
+            spacing={2}
+            direction={{ lg: 'row', md: 'column' }}
+            justifyContent='space-around'
+            sx={{ width: '100%' }}
```
GB907762, line 202: Entries isn't a user term. Is it worth thinking about the different type of entries?

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,220 @@
...
+          >
```
GB907762, line 203: Data card doesn't have releases?

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,220 @@
...
+              <Typography fontWeight='bold' variant='h6' color='primary'>
+                State
+              </Typography>
+              <PieChart
+                series={[{ innerRadius: 50, outerRadius: 100, data: stateData, arcLabel: 'value' }]}
+                {...pieChartSettings}
+              />
+            </Stack>
+            <Stack spacing={2}>
+              <Typography fontWeight='bold' variant='h6' color='primary'>
```
GB907762, line 233: I know we use State in other parts of the app but would `Life cycle status` be a better term

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,220 @@
...
+                Model card schemas usage
+              </Typography>
+              <PieChart
+                series={[{ innerRadius: 50, outerRadius: 100, data: schemaData, arcLabel: 'value' }]}
+                {...pieChartSettings}
```
GB907762, line 258: There's a weird underline in the styling of the `arcLabels`. I can't find the property to remove it.

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,220 @@
...
+                sx={{
+                  [`.${pieClasses.series}[data-series="outer"] .${pieClasses.arc}`]: {
+                    opacity: 0.6,
+                  },
```
GB907762, line 247: Is this misleading as it should include data cards by schema?

frontend/src/metrics/PolicyMetrics.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const contentRef = useRef<HTMLDivElement>(null)
+
+  const exportModelCard = useReactToPrint({
+    contentRef: contentRef,
+    documentTitle: 'Bailo policy metrics',
+  })
+
+  const handleExportOnClick = () => {
+    if (contentRef) {
+      exportModelCard()
```
GB907762: ??
```suggestion
      exportMetricsOverview()
```
Also we have a preview in the form of dialog on the Overview page. Could we do something similar to keep this a consistent action within the app?

frontend/src/metrics/PolicyMetricsCharts.tsx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export default function PolicyMetricsCharts({ data }: PolicyMetricsChartsProps) {
+  const theme = useTheme()
+  const displayMissingRoleCounts = useMemo(() => {
+    return data.summary.map((roleSummary) => {
+      return (
+        <OverviewStatPanel
+          key={roleSummary.roleId}
+          label={`models missing ${roleSummary.roleId.toUpperCase()}`}
```
GB907762:
```suggestion
          label={`models missing ${roleSummary.roleName}`}
```
may be better?

frontend/src/metrics/PolicyMetricsCharts.tsx
```diff
@@ -0,0 +1,80 @@
...
+          value={roleSummary.count}
+        />
```
GB907762, line 63: Is it worth making these buttons so you could filter results by missing roles?

Let me know if this is out of scope.

frontend/src/metrics/PolicyMetricsCharts.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <Stack spacing={2} sx={{ width: '100%' }}>
+        <Typography fontWeight='bold' variant='h6' color='primary'>
+          Models missing roles
+        </Typography>
+        <Box sx={{ backgroundColor: theme.palette.container.main, p: 2, borderRadius: 1 }}>
+          <Table sx={{ minWidth: 650 }} size='small'>
+            <TableHead>
+              <TableRow>
+                <TableCell>Model ID</TableCell>
+                <TableCell>Model Developer(s)</TableCell>
```
GB907762: Could this be dynamic eg. use the owner name from the roles endpoint?

frontend/src/metrics/PolicyMetricsCharts.tsx
```diff
@@ -0,0 +1,80 @@
...
+                <TableCell>Missing roles</TableCell>
+              </TableRow>
+            </TableHead>
+            <TableBody>{tableRows}</TableBody>
```
GB907762, line 131: Could you please paginate this

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,220 @@
...
+    height: 250,
+    margin: { left: 0 },
+    yAxis: [{ width: 50 }],
+  })
+
+  const pieChartSettings = {
+    margin: { right: 5 },
+    width: 200,
+    height: 200,
+  }
```
GB907762, line 60: The contrast between the black text on the arc and the dark blue isn't very much. Could we make the arc text white and increase the font weight/size?

frontend/pages/metrics.tsx
```diff
@@ -0,0 +1,49 @@
...
+export default function Metrics() {
+  const { currentUser, isCurrentUserLoading, isCurrentUserError } = useGetCurrentUser()
+
+  if (!currentUser || !currentUser.isAdmin) {
+    return (
+      <Forbidden
+        errorMessage='If you think this is in error please contact the Bailo administrators.'
+        noMargin
+        hideNavButton
+      />
```
GB907762, line 24: Worth having a link to the help page here?

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,220 @@
...
+      })
+      setStructuredModelVolume(updatedStructure)
+    }
+  })
+
+  useEffect(() => {
+    if (modelVolume) {
+      updateModelVolume(modelVolume, selectedOrganisation)
+    }
+  }, [modelVolume, selectedOrganisation])
```
GB907762, line 132: It's quite jarring when I update the date. It retriggers a new request. I don't know if you could use something like use optimistic in here? https://react.dev/reference/react/useOptimistic

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,220 @@
...
+          </Typography>
+          <DatePicker
+            openTo='month'
+            views={['year', 'month']}
+            value={startDate}
+            onChange={(newValue) => {
+              setStartDate(newValue)
+              setErrorMessage('')
+            }}
+          />
```
GB907762, line 163: Do we want some sort of limit on these date pickers to a certain year? If I get all models between 1900 and 2100 it's not very forgiving with 200 years worth of models.

GB907762 (changes requested):

GB907762:
https://github.com/gchq/Bailo/blob/9e0c16d8ae495d0b6868719d3e267d17f7554fe5/backend/test/connectors/metrics/base.spec.ts#L338-#L353

Update this to fix unit tests

frontend/src/metrics/OverviewMetricsCharts.tsx
ARADDCC002: This is the preferred term based on user feedback for this page, I agree it's not very consistent, but we can look to change in the future if needed.

frontend/src/metrics/OverviewMetricsCharts.tsx
ARADDCC002: Minor, will look into it if there's time, agree it's visually very off.

frontend/src/metrics/OverviewMetricsCharts.tsx
ARADDCC002: Changed to "Schema usage" for now.

frontend/src/metrics/PolicyMetrics.tsx
ARADDCC002: Can look to add this if there's time

frontend/src/metrics/PolicyMetricsCharts.tsx
ARADDCC002: User feedback prefer a list for now. Time depending I could add a switch with conditional view.

frontend/src/metrics/OverviewMetricsCharts.tsx
ARADDCC002: I can't find an easy way to change font weight in the library, but I've changed the palette which I think is a bit better on the eyes.

backend/package-lock.json
PE39806, line 1: No changes to `backend/package.json`. Remove this change.

.vscode/settings.json
PE39806, line 1: Remove this redundant change.

frontend/src/metrics/OverviewMetricsCharts.tsx
ARADDCC002: I don't think we can add a "sensible limit" because it would be arbitrary. I agree that 100+ years is impossible, but I feel that the onus is on the user to put in realistic parameters.

frontend/src/metrics/OverviewMetricsCharts.tsx
ARADDCC002: reading the docs I'm not 100% convinced this solves this particular use-case. `useOptimistic` is good when you know what sort of feedback you're expecting to get from the API, but in this case we don't _know_ what we're getting back. Happy to be pushed back on this in case I'm mis-understanding the docs.

frontend/pages/metrics.tsx
ARADDCC002: _Technically_ out of scope, but a very small and quick change.

PE39806 (changes requested):

frontend/actions/metrics.ts
```diff
@@ -0,0 +1,45 @@
...
+    isModelVolumeError: error,
+  }
+}
+
+export function useGetOverviewMetrics() {
+  const { data, isLoading, error, mutate } = useSWR<OverviewMetrics, ErrorInfo>('/api/v3/metrics/usage', fetcher)
+
+  return {
+    mutateOverviewMetrics: mutate,
+    overviewMetrics: data ? data : undefined,
```
GB907762: Why's this undefined in a ternary. In what case would this be falsy and not `undefined`

```suggestion
    overviewMetrics: data,
```

frontend/actions/metrics.ts
```diff
@@ -0,0 +1,45 @@
...
+    isOverviewMetricsError: error,
+  }
+}
+
+export function useGetPolicyMetrics() {
+  const { data, isLoading, error, mutate } = useSWR<PolicyMetrics, ErrorInfo>('/api/v3/metrics/compliance', fetcher)
+
+  return {
+    mutatePolicyMetrics: mutate,
+    policyMetrics: data ? data : undefined,
```
GB907762:
```suggestion
    policyMetrics: data,
```
Same here

frontend/src/metrics/OverviewMetricsCharts.tsx
```diff
@@ -0,0 +1,263 @@
+import { Stack, Typography } from '@mui/material'
+import { blueberryTwilightPalette } from '@mui/x-charts'
```
GB907762: This is out of scope but I've suggested colour hashing for certain parts of the app. 

So maybe make the organisation or state chips the same colour as it is in here to avoid conflating different types of information?

I also think unset items should be grey in here?

frontend/src/metrics/OverviewStatPanel.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <Box
+      sx={{
+        border: 'solid',
+        borderWidth: 1,
+        borderRadius: 3,
+        borderColor: theme.palette.divider,
+        p: 2,
+        textAlign: 'center',
+        minWidth: minWidth || 'unset',
```
GB907762:
```suggestion
        minWidth,
```
?

frontend/src/metrics/MetricsHeader.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <Stack spacing={4} sx={{ mt: 2 }}>
+        <Stack direction={{ sm: 'column', md: 'row' }} justifyContent='space-between'>
+          <Stack spacing={1}>
+            <Box>
+              <Stack direction={{ sm: 'column', md: 'row' }} spacing={1} alignItems='center'>
+                <Typography fontStyle='italic'>Showing results for</Typography>
+                <Select
+                  sx={{ maxWidth: '300px' }}
+                  value={selectedOrganisation}
+                  onChange={(e) => handleOrganisationSelectOnChange(e)}
```
GB907762:
```suggestion
                  onChange={handleOrganisationSelectOnChange}
```

frontend/src/metrics/OverviewMetricsCharts.tsx
GB907762: I don't know what case I would need to look to future dates? Could we make it up to the current date and anything after the UNIX epoch?

frontend/src/metrics/PolicyMetrics.tsx
GB907762: Can you commit this suggestion at least

frontend/src/metrics/OverviewMetricsCharts.tsx
ARADDCC002: For now I've used different colour palettes for the pie and bar graph. Tomorrow I can see if I can set custom colours for specific values, but leave other values to the built-in palettes.

GB907762 (dismissed):

backend/package-lock.json
PE39806: This still needs reverting.

PE39806 (dismissed):

GB907762 (approved):

frontend/actions/metrics.ts
```diff
@@ -0,0 +1,45 @@
+import qs from 'querystring'
+import useSWR from 'swr'
+import { ModelVolume, OverviewMetrics, PolicyMetrics } from 'types/types'
+import { ErrorInfo, fetcher } from 'utils/fetcher'
+
+export function useGetVolumeForModel(interval: string = 'month', startDate?: string, endDate?: string) {
```
ON93754, line 6: Worth a return type for these 3 functions? Looks like the same return type could be used for all 3

ARADDCC002: We don't really need to type the returns of these functions as they're stated typing of the `useSWR` hook.

ON93754 (approved):

PR #3515, opened 2026-04-29
Feature/mirror model notifications

GB907762:
- Sanitise emails against XSS.
- Modifies the Model Transfer collection to track model dependencies
- Modifies the Model Transfer collection to use `exportId` as a UID

- Email notifications on imports:
  - On start
  - On success
  - On failure


backend/src/models/ModelTransfer.ts
```diff
@@ -12,39 +12,114 @@ export const TransferStatus = {
 export type TransferStatusKeys = (typeof TransferStatus)[keyof typeof TransferStatus]
 
 export interface ModelTransferInterface {
+  exportId: string
   modelId: string
-  // The Bailo instance ID where the model is being transferred from
-  peerId: string
-  status: TransferStatusKeys
+  peerId: string | undefined
```
PE39806: Prefer
```suggestion
  peerId?: string
```

backend/src/models/ModelTransfer.ts
```diff
@@ -12,39 +12,114 @@ export const TransferStatus = {
...
   {
     timestamps: true,
     collection: 'v2_model_transfers',
+    toJSON: { virtuals: true, getters: true },
+    toObject: { virtuals: true, getters: true },
   },
 )
 
-ModelTransferSchema.plugin(softDeletionPlugin)
+// TODO: Use access request status to infer 'requested' and 'denied' status'
```
PE39806, line 77: Is this TODO within this PR?

backend/src/services/mirroredModel/importers/base.ts
```diff
@@ -70,14 +71,19 @@ export abstract class BaseImporter {
...
+    reject: (reason?: unknown) => void,
+  ): Promise<void> {
     if (isBailoError(error)) {
       reject(error)
     } else {
       reject(
         InternalError('Error processing tarball during import.', { error, metadata: this.metadata, ...this.logData }),
       )
     }
+    await finishTransfer(this.metadata.exportId)
```
PE39806: It is generally best practice to have completed all processing before `reject`/`resolve` is called, otherwise any latter functionality could run in a detached manner where raised errors are not necessarily handled well. If this ordering is intentional then please add a comment explaining why it is needed.

backend/src/services/mirroredModel/importers/documents.ts
```diff
@@ -148,6 +150,18 @@ export class DocumentsImporter extends BaseImporter {
...
+  async handleStreamError(
+    error: unknown,
+    _resolve: (reason?: any) => void,
+    reject: (reason?: unknown) => void,
+  ): Promise<void> {
+    await ModelTransferModel.findOneAndUpdate(
+      { exportId: this.metadata.exportId },
+      { documentStatus: TransferStatus.Failed },
+    )
+    return super.handleStreamError(error, _resolve, reject)
```
PE39806: No need to `return` as this is `void`, but the `super` implementation is now `async`.
```suggestion
    await super.handleStreamError(error, _resolve, reject)
```

backend/src/services/mirroredModel/importers/file.ts
```diff
@@ -75,8 +77,19 @@ export class FileImporter extends BaseImporter {
     }
   }
 
+  async handleStreamError(
+    error: unknown,
+    _resolve: (reason?: any) => void,
+    reject: (reason?: unknown) => void,
+  ): Promise<void> {
+    await updateFile(this.metadata.exportId, this.metadata.filePath, TransferStatus.Failed)
+    return super.handleStreamError(error, _resolve, reject)
```
PE39806:
```suggestion
    await super.handleStreamError(error, _resolve, reject)
```

backend/src/services/mirroredModel/importers/image.ts
```diff
@@ -153,6 +155,15 @@ export class ImageImporter extends BaseImporter {
     }
   }
 
+  async handleStreamError(
+    error: unknown,
+    _resolve: (reason?: any) => void,
+    reject: (reason?: unknown) => void,
+  ): Promise<void> {
+    await updateImage(this.metadata.exportId, this.metadata.distributionPackageName, TransferStatus.Failed)
+    return super.handleStreamError(error, _resolve, reject)
```
PE39806:
```suggestion
    await super.handleStreamError(error, _resolve, reject)
```

backend/src/services/mirroredModel/importers/file.ts
```diff
@@ -75,8 +77,19 @@ export class FileImporter extends BaseImporter {
...
+  }
+
   // Type resolve
-  handleStreamCompletion(resolve: (reason?: FileMirrorInformation) => void, _reject: (reason?: unknown) => void) {
+  async handleStreamCompletion(resolve: (reason?: FileMirrorInformation) => void, _reject: (reason?: unknown) => void) {
+    await updateFile(this.metadata.exportId, this.metadata.filePath, TransferStatus.Completed)
     resolve({ metadata: this.metadata, sourcePath: this.metadata.filePath, newPath: this.updatedPath })
+    await finishTransfer(this.metadata.exportId)
```
PE39806: As above (processing after `resolve`/`reject`).

backend/src/services/mirroredModel/importers/image.ts
```diff
@@ -179,18 +190,20 @@ export class ImageImporter extends BaseImporter {
...
     } else {
+      await updateImage(this.metadata.exportId, this.metadata.distributionPackageName, TransferStatus.Failed)
       reject(
         InternalError('Manifest file (manifest.json) missing or invalid in Tarball file.', {
           metadata: this.metadata,
           ...this.logData,
         }),
       )
     }
+    await finishTransfer(this.metadata.exportId)
```
PE39806: As above (processing after `resolve`/`reject`).

backend/src/services/smtp/emailBuilder.ts
```diff
@@ -17,12 +20,42 @@ export type actions = {
   url: string
 }
 
+export function validateHtmlInput(input: unknown): boolean {
+  if (input === null) {
+    return true
+  }
+
+  if (typeof input === 'string') {
+    return sanitizeHtml(input, { allowedTags: ['br'] }) === input
```
PE39806: Unless there's a specific reason, I think that `allowedTags` should just use the [default options](https://www.npmjs.com/package/sanitize-html#default-options).

backend/src/services/smtp/smtp.ts
```diff
@@ -40,7 +41,10 @@ export async function generateTransporter(transportOption: TransportOption): Pro
   }
 }
 
-async function dispatchEmail(entity: string, emailContent: EmailContent) {
+async function dispatchEmail(entity: string, emailContent: EmailContent | undefined) {
+  if (!emailContent) {
+    return
+  }
```
PE39806: This feels like a strange pattern to me. Why not make `emailContent` required and have the caller handle the missing `emailContent` case?

backend/src/services/smtp/smtp.ts
```diff
@@ -56,6 +60,20 @@ async function dispatchEmail(entity: string, emailContent: EmailContent) {
   await Promise.all(sendEmailResponses)
 }
 
+export async function dispatchEmailToModelRole(modelId: string, role: string, emailContent: EmailContent | undefined) {
+  if (!emailContent) {
+    return
+  }
```
PE39806: As above

backend/src/services/modelTransfer.ts
```diff
@@ -30,57 +28,205 @@ export async function findModelTransferById(user: UserInterface, id: string): Pr
...
+  }
+}
+
+export async function finishTransfer(exportId: string) {
+  const transfer = await ModelTransferModel.findOne({
+    exportId,
+  })
+
+  if (!transfer || !transfer.completed || transfer.completedNotificationSent) {
+    return
```
PE39806, line 213: To me, not finding the `ModelTransferModel` seems more like a failure so I would expect an error to be thrown.

backend/src/services/specification.ts
```diff
@@ -63,8 +63,14 @@ export const systemStatusSchema = z.object({
 
 export const modelTransferSchema = z.object({
```
PE39806, line 64: Is this missing `imageStatus`?

PE39806 (changes requested):

backend/src/models/ModelTransfer.ts
GB907762: This is out of scope. But for future when access requests are used for a model transfer.

backend/src/services/smtp/emailBuilder.ts
GB907762: Yep this function also seems to format elements as well. But I think this is unnecessary.

backend/src/services/smtp/smtp.ts
GB907762: Yep. I think I was struggling with handling a failure case on sanitisation. When an error is thrown in this part of the code it isn't caught.

GB907762: I've decided sanitisation is better than validation.

backend/src/models/ModelTransfer.ts
```diff
@@ -12,39 +12,114 @@ export const TransferStatus = {
...
-  // The Bailo instance ID where the model is being transferred from
-  peerId: string
-  status: TransferStatusKeys
+  peerId?: string
+  documentStatus: TransferStatusKeys
+  fileStatus: Map<string, TransferStatusKeys>
+  imageStatus: Map<string, TransferStatusKeys>
+  startedNotificationSent: boolean
+  completedNotificationSent: boolean
+  completed: boolean
```
PE39806: Move `completed` to `ModelTransferVirtuals`.

backend/src/services/modelTransfer.ts
```diff
@@ -30,57 +28,210 @@ export async function findModelTransferById(user: UserInterface, id: string): Pr
...
-  assertValidObjectId(id)
+export async function updateFile(
+  exportId: string,
+  file: string,
+  status: TransferStatusKeys,
+): Promise<ModelTransferInterface | null> {
+  const updated = await ModelTransferModel.findOneAndUpdate(
+    { exportId },
+    {
+      $set: { [`fileStatus.${file}`]: status },
```
PE39806: I don't like this as a maliciously crafted import where `matadata.filePath` is changed to include mongoose specific strings (e.g. `.`, `$`) could allow for mongoose injection.

An alternative could be to store an array of objects `fileStatus: [{ key: string, status: TransferStatusKeys }]` and then you can safely set the `key` to any string value.

`backend/src/models/ModelTransfer.ts`
```typescript
// ...
export interface TransferArtefactStatus {
  key: string
  status: TransferStatusKeys
}

export interface ModelTransferInterface {
  // ...
  fileStatus: TransferArtefactStatus[]
  imageStatus: TransferArtefactStatus[]
  // ...
}
// ...
```

`backend/src/services/modelTransfer.ts`
```typescript
// ...
export async function updateFile(
  exportId: string,
  file: string,
  status: TransferStatusKeys,
): Promise<ModelTransferInterface | null> {
  const transfer = await ModelTransferModel.findOne({ exportId })

  if (!transfer) {
    log.warn({ exportId, file }, 'The requested model transfer was not found.')
    return null
  }

  const currentStatuses = transfer['fileStatus'] as TransferArtefactStatus[]
  const existingIndex = currentStatuses.findIndex((entry) => entry.key === file)

  if (existingIndex >= 0) {
    currentStatuses[existingIndex].status = status
  } else {
    currentStatuses.push({ key: file, status })
  }

  transfer.markModified('fileStatus')
  await transfer.save()

  return transfer.toObject()
}
// similar for `updateImage`
// ...
```

backend/src/services/modelTransfer.ts
```diff
@@ -30,57 +28,210 @@ export async function findModelTransferById(user: UserInterface, id: string): Pr
...
+export async function updateImage(
+  exportId: string,
+  image: string,
+  status: TransferStatusKeys,
+): Promise<ModelTransferInterface | null> {
+  const updated = await ModelTransferModel.findOneAndUpdate(
+    { exportId },
+    {
+      $set: {
+        [`imageStatus.${image}`]: status,
```
PE39806: As above with `updateFile`.

backend/src/services/modelTransfer.ts
```diff
@@ -30,57 +28,210 @@ export async function findModelTransferById(user: UserInterface, id: string): Pr
...
+  images?: string[],
+  files?: ObjectId[],
+): Promise<ModelTransferInterface | null> {
+  const setUpdates: Record<string, any> = {
+    documentStatus: TransferStatus.Completed,
+  }
+
+  const addKey = (path: string) => {
+    setUpdates[path] = {
+      $ifNull: [`$${path}`, TransferStatus.InProgress],
```
PE39806: This will break for a `path` like `imageStatus.example.com/yolo-v4/alpine:latest` due to mongoose interpreting `.`s etc.

backend/src/services/modelTransfer.ts
```diff
@@ -30,57 +28,210 @@ export async function findModelTransferById(user: UserInterface, id: string): Pr
...
+    }
+  }
+
+  if (images) {
+    for (const image of images) {
+      addKey(`imageStatus.${image}`)
```
PE39806: As above.

backend/src/services/modelTransfer.ts
```diff
@@ -30,57 +28,210 @@ export async function findModelTransferById(user: UserInterface, id: string): Pr
...
+    }
+  }
+
+  if (files) {
+    for (const file of files) {
+      addKey(`fileStatus.${file}`)
```
PE39806: As above.

backend/test/services/modelTransfer.spec.ts
```diff
@@ -102,16 +101,15 @@ describe('services > modelTransfer', () => {
     const input = {
       modelId: 'model-123',
       peerId: 'peer-123',
+      exportId: 'abc-123',
       status: TransferStatus.Requested,
```
PE39806: `status` removed from `createModelTransfer` so remove here too.

PE39806 (changes requested):

backend/src/routes/v2/modelTransfer/getModelTransfer.ts
```diff
@@ -38,7 +38,7 @@ export const getModelTransfer = [
   async (req: Request, res: Response<ModelTransferResponse>): Promise<void> => {
     const { params } = parse(req, getModelTransferSchema)
```
JR40159:
```
    const {
      params: { exportId },
    } = parse(req, getModelTransferSchema
```

backend/src/models/ModelTransfer.ts
```diff
@@ -12,39 +12,114 @@ export const TransferStatus = {
...
+  exportId: string
   modelId: string
-  // The Bailo instance ID where the model is being transferred from
-  peerId: string
-  status: TransferStatusKeys
+  peerId?: string
+  documentStatus: TransferStatusKeys
+  fileStatus: Map<string, TransferStatusKeys>
+  imageStatus: Map<string, TransferStatusKeys>
+  startedNotificationSent: boolean
```
JR40159: What's the value in storing this?

backend/src/models/ModelTransfer.ts
```diff
@@ -12,39 +12,114 @@ export const TransferStatus = {
...
+  completedNotificationSent: boolean
```
JR40159, line 26: What's the value in storing this?

backend/src/services/modelTransfer.ts
```diff
@@ -30,57 +28,210 @@ export async function findModelTransferById(user: UserInterface, id: string): Pr
...
+export async function upsertArtefacts(
+  exportId: string,
+  images?: string[],
+  files?: ObjectId[],
+): Promise<ModelTransferInterface | null> {
+  const setUpdates: Record<string, any> = {
+    documentStatus: TransferStatus.Completed,
+  }
+
+  const addKey = (path: string) => {
```
JR40159: I don't think having this function adds much value

backend/src/services/modelTransfer.ts
```diff
@@ -30,57 +28,210 @@ export async function findModelTransferById(user: UserInterface, id: string): Pr
...
+
+  if (!updated) {
+    log.warn({ exportId, image }, 'The requested model transfer was not found.')
+    return null
+  }
+
+  return updated.toObject()
+}
+
+export async function upsertArtefacts(
```
JR40159: This name is quite vague, `updateArtefactsTranferStatus` or something?

backend/src/services/mirroredModel/importers/file.ts
```diff
@@ -75,8 +77,19 @@ export class FileImporter extends BaseImporter {
     }
   }
 
+  async handleStreamError(
+    error: unknown,
+    _resolve: (reason?: any) => void,
+    reject: (reason?: unknown) => void,
+  ): Promise<void> {
+    await updateFile(this.metadata.exportId, this.metadata.filePath, TransferStatus.Failed)
```
JR40159: `updateFileTransferStatus`?

backend/src/models/ModelTransfer.ts
```diff
@@ -12,39 +12,114 @@ export const TransferStatus = {
...
 
 export interface ModelTransferInterface {
+  exportId: string
   modelId: string
-  // The Bailo instance ID where the model is being transferred from
-  peerId: string
-  status: TransferStatusKeys
+  peerId?: string
+  documentStatus: TransferStatusKeys
+  fileStatus: Map<string, TransferStatusKeys>
```
ON93754: As `documentStatus` `fileStatus` and `imageStatus` are all very similar, is it worth just having a `status` and a separate `transferType` that defines if it is a file/document/image type of transfer ?

backend/src/services/smtp/smtp.ts
```diff
@@ -216,6 +234,90 @@ async function sendEmail(email: Mail.Options) {
...
+  if (!config.smtp.enabled) {
+    log.info('Not sending email due to SMTP disabled')
+    return
+  }
+
+  const mirroredModel = await getModelByIdNoAuth(modelId)
+
+  const information: Info[] = []
+  if (successfulFiles.length) {
+    information.push({ title: 'Successful Files', data: successfulFiles.join('<br />') })
```
ON93754: Worries about having lots of repeated line break strings in the app in case of typos, I wonder if we would benefit by a function to handle the line breaks, something like:

```
function spaceWithLineBreaks(fileNames: string[]): string[] {
    return fileNames.join('<br />')
}
```

Or alternatively just having a constant variable like
`const lineBreak = '<br />'`

backend/src/services/modelTransfer.ts
```diff
@@ -30,57 +28,210 @@ export async function findModelTransferById(user: UserInterface, id: string): Pr
...
+      { startedNotificationSent: true },
+      { new: true },
+    )
+    if (updated) {
+      await startImportNotification(modelId)
+    }
+  }
+}
+
+export async function finishTransfer(exportId: string) {
```
JR40159: I'm not a fan of how when other methods of this service are called, the caller than has to then explicitly call this function even though every time `updateFile` is called, for example, this function then needs to be called. A safer, more maintainable suggestion would be to just call this function at the end of those methods.

JR40159: Additionally, the name `finishTransfer` is a bit confusing as the method isn't finishing the transfer, it's checking if the transfer is complete.

backend/src/services/smtp/emailBuilder.ts
```diff
@@ -17,12 +18,42 @@ export type actions = {
   url: string
 }
 
+export function sanitizeInput<T>(input: T) {
```
JR40159: No need to `export`

backend/src/services/smtp/smtp.ts
```diff
@@ -40,7 +41,10 @@ export async function generateTransporter(transportOption: TransportOption): Pro
   }
 }
 
-async function dispatchEmail(entity: string, emailContent: EmailContent) {
+async function dispatchEmail(entity: string, emailContent: EmailContent | undefined) {
```
JR40159: When could `emailContent` be `undefined`?

backend/src/services/smtp/smtp.ts
```diff
@@ -56,6 +60,20 @@ async function dispatchEmail(entity: string, emailContent: EmailContent) {
   await Promise.all(sendEmailResponses)
 }
 
+export async function dispatchEmailToModelRole(modelId: string, role: string, emailContent: EmailContent | undefined) {
```
JR40159: No need to export

backend/src/services/smtp/smtp.ts
```diff
@@ -56,6 +60,20 @@ async function dispatchEmail(entity: string, emailContent: EmailContent) {
...
+  if (!emailContent) {
+    return
+  }
+  const model = await getModelByIdNoAuth(modelId)
```
JR40159, line 62: From what I can see every time you call this method, you already have the model so why the need to get the model again here?

backend/src/services/modelTransfer.ts
```diff
@@ -30,57 +28,210 @@ export async function findModelTransferById(user: UserInterface, id: string): Pr
...
+    .map(([distributionPackageName]) => distributionPackageName)
+
+  if (transfer?.status === TransferStatus.Completed) {
+    const updated = await ModelTransferModel.findOneAndUpdate(
+      { exportId },
+      { completedNotificationSent: true },
+      { new: true },
+    )
+    if (updated) {
+      await completeImportNotification(transfer.modelId, successfulFiles, successfulImages)
```
JR40159: Do you need to `await` these notifications?

JR40159 (dismissed):

backend/src/models/ModelTransfer.ts
GB907762: To make sure that a second completion email is not sent.

backend/src/services/smtp/smtp.ts
GB907762: In `backend/test/services/smtp/smtp.spec.ts`

backend/src/services/smtp/smtp.ts
GB907762: It can't. I originally validated these so made the response undefined if any content was removed when sanitised. But I just remove the bad content and send the email now.

backend/src/services/smtp/smtp.ts
GB907762: Good spot!

backend/src/models/ModelTransfer.ts
GB907762: I like this idea of a bucket pattern. I think if we decided to add other import types this would make it easier to extend.

backend/src/models/ModelTransfer.ts
GB907762: I suppose this is redundant due to the it being always sent on creation of the document

backend/src/services/modelTransfer.ts
GB907762: I like that idea. I've changed it `finishTransfer -> handleCompletionEmail`

backend/src/services/modelTransfer.ts
GB907762: I've restructured the code to use a bucket pattern. Should hopefully resolve some of these problems. https://www.mongodb.com/docs/manual/data-modeling/design-patterns/group-data/bucket-pattern/

backend/src/models/ModelTransfer.ts
```diff
@@ -38,37 +42,20 @@ const ModelTransferSchema = new Schema<ModelTransferDoc>(
...
+    artefactStatus: {
+      type: [
+        {
+          key: { type: String, required: true },
+          status: {
+            type: String,
+            enum: Object.values(TransferStatus),
+            required: true,
+          },
+          kind: { type: String, required: true },
```
PE39806: This should be an `enum` as above.

backend/src/models/ModelTransfer.ts
```diff
@@ -38,37 +42,20 @@ const ModelTransferSchema = new Schema<ModelTransferDoc>(
     modelId: { type: String, required: true },
     peerId: { type: String },
 
-    documentStatus: {
-      type: String,
-      enum: Object.values(TransferStatus),
+    artefactStatus: {
+      type: [
+        {
+          key: { type: String, required: true },
```
PE39806, line 48: Does this need `unique: true`?

backend/src/models/ModelTransfer.ts
```diff
@@ -38,37 +42,20 @@ const ModelTransferSchema = new Schema<ModelTransferDoc>(
...
-      },
-      required: true,
-      default: {},
-    },
-
-    startedNotificationSent: {
-      type: Boolean,
-      required: true,
-      default: false,
+      default: [{ key: MirrorKind.Documents, status: TransferStatus.InProgress, kind: MirrorKind.Documents }],
```
PE39806, line 58: I don't like this default `key` as I think it should always be unique.

backend/src/services/modelTransfer.ts
```diff
@@ -76,78 +78,125 @@ export async function updateModelTransfer(
   return updated
 }
 
-export async function updateFile(
+export async function updateFileTransferStatus(
```
PE39806: Can `updateFileTransferStatus` and `updateImageTransferStatus` be merged into a single function to reduce duplicate code?

PE39806 (changes requested):

backend/src/models/ModelTransfer.ts
GB907762: I've discovered that in most places are unique property doesn't enforce uniqueness, because the index isn't built before being saved to the database. 
https://mongoosejs.com/docs/validation.html#the-unique-option-is-not-a-validator

That being said I think your correct.

backend/src/models/ModelTransfer.ts
GB907762: The key should be unique to the document, but not necessarily for all transfers since you could have multiple of the same file in different exports, for example?

GB907762: In theory we should only have one document.

backend/src/models/ModelTransfer.ts
```diff
@@ -11,40 +12,91 @@ export const TransferStatus = {
...
+})
 
-// For GET transfers by modelId queries
 ModelTransferSchema.index({ modelId: 1, createdAt: -1 })
+ModelTransferSchema.index({ exportId: 1, createdAt: -1 })
 
-const ModelTransferModel = model<ModelTransferDoc>('v2_Model_Transfer', ModelTransferSchema)
+ModelTransferSchema.plugin(softDeletionPlugin)
 
+const ModelTransferModel = model<ModelTransferDoc>('ModelTransfer', ModelTransferSchema)
```
PE39806: I'm not sure if this should be `v2_Model_Transfer` or `v2_ModelTransfer`
```suggestion
const ModelTransferModel = model<ModelTransferDoc>('v2_ModelTransfer', ModelTransferSchema)
```

backend/src/services/mirroredModel/importers/file.ts
```diff
@@ -75,8 +77,28 @@ export class FileImporter extends BaseImporter {
...
+      this.metadata.filePath,
+      MirrorKind.File,
+      TransferStatus.Failed,
+    )
+    await super.handleStreamError(error, _resolve, reject)
+  }
+
   // Type resolve
-  handleStreamCompletion(resolve: (reason?: FileMirrorInformation) => void, _reject: (reason?: unknown) => void) {
+  async handleStreamCompletion(resolve: (reason?: FileMirrorInformation) => void, _reject: (reason?: unknown) => void) {
```
PE39806: For consistency
```suggestion
  async handleStreamCompletion(resolve: (reason?: FileMirrorInformation) => void, _reject: (reason?: unknown) => void): Promise<void> {
```

backend/src/services/mirroredModel/tarball.ts
```diff
@@ -175,7 +176,7 @@ export async function extractTarGzStream(
     })
 
     // Successful completion of the tar stream
-    untarStream.once('finish', () => {
+    untarStream.once('finish', async () => {
```
PE39806: [`EventEmitter` calls all listeners synchronously](https://nodejs.org/docs/latest-v24.x/api/events.html#asynchronous-vs-synchronous) so use an IIAFE here as in https://github.com/gchq/Bailo/blob/c4c47e8f0ce8d9b04b98f6a3f8277c0ae25b9d3a/backend/src/services/mirroredModel/tarball.ts#L202-L205
`;(async () => { ... })().catch((err) => fail(err))`

backend/src/services/smtp/emailBuilder.ts
```diff
@@ -17,12 +18,42 @@ export type actions = {
...
+  }
+
+  if (Array.isArray(input)) {
+    return input.map(sanitizeInput)
+  }
+
+  if (typeof input === 'object') {
+    const result: Record<string, unknown> = {}
+    for (const [key, val] of Object.entries(input)) {
+      if (key === '__proto__' || key === 'constructor' || key === 'prototype') {
```
PE39806: Maybe add a comment explaining what type of Object is being skipped here.

backend/package.json
```diff
@@ -66,6 +66,7 @@
     "pretty-bytes": "^7.1.0",
     "proxy-agent": "^8.0.1",
     "qs": "^6.15.1",
+    "sanitize-html": "^2.17.3",
```
PE39806, line 69: Please add the types for this `npm i --save-dev @types/sanitize-html`

backend/src/services/modelTransfer.ts
```diff
@@ -1,25 +1,25 @@
...
-
-export async function findModelTransferById(user: UserInterface, id: string): Promise<ModelTransferInterface> {
-  assertValidObjectId(id)
-
+export async function findModelTransferById(user: UserInterface, exportId: string): Promise<ModelTransferInterface> {
   const transfer = await ModelTransferModel.findOne({
-    _id: id,
-  }).lean()
+    exportId,
+  }).lean<ModelTransferInterface>()
```
PE39806, line 19: Worth noting that this is more likely to be of type `<ModelTransferInterface & SoftDeleteInterface>`, possibly with some mongoose extras like `id`. I don't think this needs updating but something to be aware of.

backend/src/services/specification.ts
```diff
@@ -63,8 +63,17 @@ export const systemStatusSchema = z.object({
...
+  peerId: z.string().openapi({ example: 'bailo' }),
+  completedNotificationSent: z.boolean().openapi({ example: true }),
+  startedNotificationSent: z.boolean().openapi({ example: false }),
+  documentStatus: z.nativeEnum(TransferStatus).openapi({ example: TransferStatus.InProgress }),
+  imageStatus: z.object({
+    'example.com/yolo-v4/alpine:latest': z.nativeEnum(TransferStatus).openapi({ example: TransferStatus.InProgress }),
+  }),
+  fileStatus: z.object({
+    '69fa066b3f64c06f027b4921': z.nativeEnum(TransferStatus).openapi({ example: TransferStatus.InProgress }),
+  }),
```
PE39806: I think these are now just `artefactStatus`.

backend/src/services/specification.ts
```diff
@@ -63,8 +63,17 @@ export const systemStatusSchema = z.object({
 
 export const modelTransferSchema = z.object({
   _id: z.string().openapi({ example: '65df1a0e8c2b7c0012f0abcd' }),
-  modelId: z.string().openapi({ example: '65df1a0e8c2b7c0012f0abcd' }),
-  peerId: z.string().openapi({ example: '65df1a0e8c2b7c0012f0abcd' }),
+  modelId: z.string().openapi({ example: 'model-123456' }),
+  peerId: z.string().openapi({ example: 'bailo' }),
+  completedNotificationSent: z.boolean().openapi({ example: true }),
+  startedNotificationSent: z.boolean().openapi({ example: false }),
```
PE39806: I don't think this exists.

backend/test/services/smtp/emailBuilder.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { describe, expect, test, vi } from 'vitest'
+
+import { buildEmail } from '../../../src/services/smtp/emailBuilder.js'
+
+const logMock = vi.hoisted(() => ({
+  error: vi.fn(),
+}))
+vi.mock('../../../src/services/log.js', async () => ({
```
PE39806: This doesn't need to be `async`.

PE39806 (changes requested):

backend/src/models/ModelTransfer.ts
GB907762: `v2_Model_Transfer` to be consistent

backend/src/services/smtp/emailBuilder.ts
GB907762: I'm not really sure that this is useful. In most cases it's likely only useful to sanitise a primitive, an array, or an object. I'm going to get rid of this condition, since it adds unnecessary complexity.

backend/test/services/smtp/emailBuilder.spec.ts
GB907762: It doesn't need to be in that test because I removed validation in favor of sanitisation.

PE39806 (approved):

IR96334 (approved):

PR #3560, opened 2026-05-05
New entry kind for untrusted models

ARADDCC002:


frontend/src/entry/CreateEntry.tsx
IR96334: Can remove this as no access requests allowed anyway. Also, are you allowed to template untrusted models?

frontend/src/entry/overview/TemplatePage.tsx
```diff
@@ -25,7 +26,7 @@ export default function TemplatePage({ entry }: TemplatePageProps) {
     <Box sx={{ maxWidth: 'md', mx: 'auto', my: 4 }}>
       <Stack spacing={4} justifyContent='center' alignItems='center'>
         <Typography component='h2' variant='h6' color='primary' data-test='createEntryCardOverview'>
-          {`Create a ${toTitleCase(EntryCardKindLabel[entry.kind])}`}
```
IR96334, line 28: maybe remove a or an

IR96334:
Can still make Access request to untrusted model via API.

IR96334:
Bug when creating new release, get 404 on GET model request. Issue on line 46. Either remove query parameter (EntryKind.MODEL), or provide model kind dynamically (likely through query parameter of page, e.g. ?kind=UNTRUSTED_MODEL)

IR96334:
> Can still make Access request to untrusted model via API.

Could use a test to cover this possibility as well, like with creating model

frontend/src/entry/EntryCard.tsx
```diff
@@ -1,34 +1,44 @@
...
-import { Button, Card, Stack, Typography } from '@mui/material'
+import { Box, Button, Stack, Typography } from '@mui/material'
 
 interface EntryCardProps {
   title: string
   description: string
   handleClick: () => void
   dataTest?: string
   disabled?: boolean
+  recommended?: boolean
```
IR96334: change name? as most popular is used now

backend/src/utils/config.ts
```diff
@@ -222,6 +222,11 @@ export interface Config {
   inference: {
     authorisationToken: string
   }
+
+  untrustedModels: {
+    defaultState: 'Production'
```
JR40159: How come there are string values here rather than types?

JR40159: Now we have another entry type maybe we should refactor the configuration to have an `entryTypes` config block that contains the config (like if they're enabled) for the different types?

backend/src/services/model.ts
JR40159: Now we've got restrictions for another entry type, it could be worth renaming this function

JR40159: Also, while you're there, please remove the redundant `export` 😄

backend/src/services/model.ts
```diff
@@ -88,6 +100,7 @@ export async function createModel(user: UserInterface, modelParams: CreateModelP
     ...modelParams,
     id: modelId,
     collaborators,
+    state: config.untrustedModels.defaultState || '',
```
JR40159: I think `CreateModelParams` is out of date and the state can be set when creating a model. However, this change will mean that will be overridden?

backend/src/services/model.ts
```diff
@@ -581,6 +594,12 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
       throw BadReq('You cannot have duplicate tags')
     }
   }
+  if (model.kind === EntryKind.UntrustedModel && modelDiff.visibility === EntryVisibility.Public) {
```
JR40159: Building on the current `checkModelRestriction` function it could be worth having a `checkMirroredModelRestrictions` and `checkUntrustedModelRestrictions` where we can group up all the required checks for each entry type into two methods

JR40159: `modelDiff.visibility !== EntryVisibility.Public` is safer and more future proof

JR40159: > Building on the current `checkModelRestriction` function it could be worth having a `checkMirroredModelRestrictions` and `checkUntrustedModelRestrictions` where we can group up all the required checks for each entry type into two methods

Hopefully this will also be more reusable between the different creation and updating

backend/src/utils/__mocks__/config.ts
```diff
@@ -141,6 +141,9 @@ const config: PartialDeep<Config> = {
     contentDirectory: 'content-dir',
     metadataFile: 'meta.json',
   },
+  untrustedModels: {
+    defaultState: 'Production',
```
JR40159, line 152: For the type of `defaultState` could you use the same type as `modelDetails.states` to help ensure they're aligned?

JR40159 (changes requested):

JR40159 (changes requested):
Missing restriction in the backend to prevent registry actions

backend/src/utils/config.ts
ARADDCC002: > Now we have another entry type maybe we should refactor the configuration to have an `entryTypes` config block that contains the config (like if they're enabled) for the different types?

I think this is probably out of scope for this ticket as changing existing config is fiddly. I like the idea and would like to explore it when we refactor the rest of the entry kind permissions code.

frontend/pages/entry/new.tsx
JR40159, line 1: Although I think the categorisation of entry types is an improvement, I think showing all the options in a vertical list now requires the user to do more scrolling to view the options

frontend/pages/model/[modelId].tsx
```diff
@@ -52,19 +52,21 @@ export default function Model() {
...
+              disabledText: 'Select a schema to view this tab.',
             },
             {
               title: 'Access requests',
               path: 'access',
               view: <AccessRequests model={entry} currentUserRoles={currentUserRoles} />,
               datatest: 'accessRequestTab',
-              disabled: !entry.card,
-              disabledText: 'Select a schema to view this tab',
+              disabled: !entry.card || entry.kind === EntryKind.UNTRUSTED_MODEL,
```
Comment (JR40159) [inline frontend/pages/model/[modelId].tsx:62]: Is this something that can also be done in the backend as well?

frontend/src/entry/overview/TemplatePage.tsx
```diff
@@ -47,10 +48,10 @@ export default function TemplatePage({ entry }: TemplatePageProps) {
                 Create from schema
```
JR40159, line 48: Happy for this to be out of scope, but this should only be available to those who have access

JR40159 (changes requested):

frontend/src/entry/overview/TemplatePage.tsx
ARADDCC002: Yeah this is an issue with models as a whole unfortunately.

frontend/pages/model/[modelId].tsx
Comment (ARADDCC002) [reply frontend/pages/model/[modelId].tsx:62]: As in making access requests? I've blocked them in the API already

frontend/pages/entry/new.tsx
ARADDCC002: It's hard to find a nice design that doesn't hide information behind lists/menus

frontend/pages/model/[modelId].tsx
Comment (PE39806) [reply frontend/pages/model/[modelId].tsx:62]: I think it would be good to have this come through as some config value so that there is a single source of truth for each `entry.kind` which is not duplicated for both frontend and backend.

PE39806 (changes requested):

frontend/pages/model/[modelId].tsx
Comment (ARADDCC002) [reply frontend/pages/model/[modelId].tsx:62]: > I think it would be good to have this come through as some config value so that there is a single source of truth for each `entry.kind` which is not duplicated for both frontend and backend.

We could have a system similar to the user permissions whereby for various aspects of the UI we check the list of permissions, and use it hide/disable components using the `<Restricted />` component.

It _might_ be out of scope but I'll see if it can be done in this PR (this is already quite large).

backend/config/default.cjs
```diff
@@ -351,4 +356,8 @@ module.exports = {
   inference: {
     authorisationToken: '',
   },
+
+  untrustedModels: {
```
PE39806, line 364: Missing default `entryDescription` https://github.com/gchq/Bailo/pull/3560/changes#diff-a21de7e92b2f172725594abdd2f17548e1b51e2a45e21bc9044232fdf684c40aR228

infrastructure/helm/bailo/templates/bailo/bailo.configmap.yaml
```diff
@@ -303,6 +308,10 @@ data:
         schemaLocation: '{{ .Values.stroom.schemaLocation }}',
         version: '{{ .Values.stroom.version }}',
       }
+
+      untrustedModels: {
```
PE39806, line 312: Missing `entryDescription` https://github.com/gchq/Bailo/pull/3560/changes#diff-a21de7e92b2f172725594abdd2f17548e1b51e2a45e21bc9044232fdf684c40aR228

infrastructure/helm/bailo/templates/bailo/bailo.configmap.yaml
```diff
@@ -230,6 +230,11 @@ data:
           consumer: '{{ .Values.config.ui.roleDisplayNames.consumer }}',
           contributor: '{{ .Values.config.ui.roleDisplayNames.contributor }}',
         },
+
+        untrustedModel: {
+          enabled: {{ .Values.config.ui.untrustedModel.enabled }}
```
PE39806: Missing a comma
```suggestion
          enabled: {{ .Values.config.ui.untrustedModel.enabled }},
```

frontend/src/entry/settings/EntryDetails.tsx
```diff
@@ -120,12 +130,14 @@ export default function EntryDetails({ entry }: EntryDetailsProps) {
...
               label={publicLabel()}
               data-test='publicButtonSelector'
+              disabled={entry.kind === EntryKind.UNTRUSTED_MODEL}
             />
             <FormControlLabel
               value='private'
               control={<Radio />}
               label={privateLabel()}
               data-test='privateButtonSelector'
+              disabled={entry.kind === EntryKind.UNTRUSTED_MODEL}
```
PE39806: Isn't 'private' the default for untrusted models, meaning that this should not be disabled but, rather, shown as the only option? I'm not keen on this UX for an unchangeable UI form component.

frontend/pages/index.tsx
```diff
@@ -124,6 +124,24 @@ export default function Marketplace() {
...
+    '',
+    selectedTags,
+    selectedOrganisations,
+    selectedStates,
+    selectedPeers,
+    debouncedFilter.length >= 3 ? debouncedFilter : '',
+    false,
+    '',
+    titleOnly,
+  )
```
PE39806, line 143: This will fetch even when `uiConfig.untrustedModel.enabled === false`. Ideally, don't run this unless required. Possibly use a ternary to skip this?

frontend/pages/model/[modelId].tsx
```diff
@@ -52,19 +52,21 @@ export default function Model() {
...
+              disabledText: 'Select a schema to view this tab.',
```
Comment (PE39806) [inline frontend/pages/model/[modelId].tsx]: This text is misleading for untrusted models.
```suggestion
              disabledText: entry.kind === EntryKind.UNTRUSTED_MODEL'
                ? 'Access requests are not available for untrusted models.'
                : 'Select a schema to view this tab.',
```

frontend/src/entry/EntryCard.tsx
```diff
@@ -1,34 +1,44 @@
-import { FileUpload } from '@mui/icons-material'
-import { Button, Card, Stack, Typography } from '@mui/material'
+import { Box, Button, Stack, Typography } from '@mui/material'
 
 interface EntryCardProps {
   title: string
   description: string
   handleClick: () => void
   dataTest?: string
   disabled?: boolean
```
PE39806, line 8: I think this prop is unused.

PE39806 (changes requested):

backend/config/default.cjs
ARADDCC002: This has been moved to UI Config, so I'll delete this one for now.

infrastructure/helm/bailo/templates/bailo/bailo.configmap.yaml
ARADDCC002: Same as above

frontend/src/entry/settings/EntryDetails.tsx
ARADDCC002: This _might_ be one for the UX people, but my understanding is that when a normal form is read-only, it should be disabled with appropriate information as to _why_ it's disabled, but still visible. I'll see how it looks if we remove it and only display the warning - it might just look out of place.

ARADDCC002: I've removed for now, I'm not 100% keen on it but we can see what people think.

frontend/pages/index.tsx
ARADDCC002: We can't call hooks conditionally, it's the same for mirrored models. In the future, we can split these out into their own components which we can then call conditionally.

frontend/src/entry/EntryCard.tsx
ARADDCC002: We use it for mirrored models

backend/src/services/model.ts
```diff
@@ -38,12 +47,30 @@ import { getSchemaById } from './schema.js'
...
 }
 
+type UntrustedModelRestrictionCheck = {
+  state?: string | undefined
+  visibility?: EntryVisibilityKeys
+  [key: string]: any
+}
+function checkUntrustedModelRestrictions(modelKind: EntryKindKeys, partialModel: UntrustedModelRestrictionCheck) {
+  if (modelKind === EntryKind.UntrustedModel) {
+    if (partialModel.visibility) {
```
JR40159: Needed?

backend/src/services/model.ts
JR40159: Bump

JR40159 (changes requested):

frontend/utils/routerUtils.ts
```diff
@@ -0,0 +1,14 @@
...
+  switch (entryKind) {
+    case EntryKind.MODEL:
+      return 'model'
+    case EntryKind.MIRRORED_MODEL:
+      return 'model'
+    case EntryKind.UNTRUSTED_MODEL:
+      return 'model'
+    case EntryKind.DATA_CARD:
+      return entryKind
+  }
```
PE39806, line 15: This should really have some `default` to cover a bad `entryKind`.

PE39806 (changes requested):

PE39806 (dismissed):

backend/src/services/model.ts
```diff
@@ -117,9 +115,14 @@ export async function createModel(user: UserInterface, modelParams: CreateModelP
     ...modelParams,
     id: modelId,
     collaborators,
-    state: config.untrustedModels.defaultState || '',
   })
 
+  if (modelParams.kind === EntryKind.UntrustedModel) {
+    model.state = config.untrustedModels.defaultState || ''
+  } else {
+    model.state = modelParams.kind || ''
```
JR40159: You're setting the state to be the entry kind...?

JR40159 (changes requested):

backend/src/services/model.ts
ARADDCC002: You didn't see anything...

backend/config/default.cjs
```diff
@@ -262,6 +262,11 @@ module.exports = {
       contributor: 'Contributor',
       consumer: 'Consumer',
     },
+
+    untrustedModel: {
+      enabled: true,
```
JRB66955: should this be, by default, disabled akin to mirroredModels?

PE39806 (dismissed):

JRB66955 (dismissed):

backend/config/default.cjs
ARADDCC002: I think you're right

JR40159 (approved):

PE39806 (approved):

JRB66955 (approved):

PR #3567, opened 2026-05-05
vitest backend upgrade 4.1.x

### LINKED ISSUE #3511: Update backend vitest to 4.1.x
Vitest 4.1.x introduces a regression: decorators not accepted in classes. We use class decorators in `backend/src/services/mirroredModel/exporters/base.ts` and its derivatives, so vitest 4.1.x fails to run many tests where these classes end up being imported.

See:
- https://vite.dev/guide/migration#javascript-transforms-by-oxc
- https://github.com/oxc-project/oxc/issues/9170
- https://github.com/vitest-dev/vitest/issues/9876

PE39806:
Closes #3511
Supersedes #3514 which had issues with some tests failing inconsistently.

- Update backend vitest install to 4.1.x
- Pin vite to 7.x
- Remove redundant `vite-tsconfig-paths` frontend package (replaced by `resolve: {tsconfigPaths: true}` in newest vite versions)
- Correct a few dependencies
- Swap deprecated vitest methods with newer ones
- Add "dependencies" to `labeler` GitHub Action


backend/test/clients/stroom.spec.ts
```diff
@@ -45,7 +45,7 @@ describe('clients > stroom', () => {
 
     const resp = sendEvents(events)
 
-    expect(resp).rejects.toThrowError(`Failed to send logs to STROOM - Non-200 response`)
-    expect(fetchMock.default).toBeCalled()
+    expect(resp).rejects.toThrow(`Failed to send logs to STROOM - Non-200 response`)
```
JR40159: Need an `await`?

JR40159 (changes requested):
`toBeCalledTimes` also needs updating

JRB66955 (dismissed):

JR40159 (approved):

GB907762 (approved):

PR #3577, opened 2026-05-08
Separate route file into v1/2/3

JR40159:


backend/src/routes/v3/routes.ts
```diff
@@ -0,0 +1,12 @@
+import { Router } from 'express'
+
+import { generateV3SwaggerSpec } from '../../services/specification.js'
+import { getImageByDigest } from '../v2/model/images/getImage.js'
+
+const router = Router()
+
+router.get('/api-docs/swagger.json', (req, res) => res.json(generateV3SwaggerSpec()))
+
+router.get('/model/:modelId/image/:name/:tag/:digest', ...getImageByDigest)
```
GB907762, line 13: Could you move the handler to `backend/src/routes/v3`?

GB907762 (changes requested):

backend/src/routes/v3/routes.ts
JR40159: It looked to me like v2 and v3 are using the same handler so I just left it in v2?

GB907762: No they use different handlers but within the same file so the  so probably worth splitting?

JR40159: Ah, good spot- I'll do that

GB907762 (dismissed):

GB907762 (dismissed):

GB907762 (approved):

ARADDCC002 (dismissed):

GB907762 (changes requested):
Needs merge updates

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -141,7 +141,7 @@ export default function SideNavigation({
             />
             <Divider aria-hidden='true' />
             <NavMenuItem
-              href='/api/v2/docs'
+              href='/api/docs'
```
GB907762, line 144: https://github.com/gchq/Bailo/blob/e6260120822a9eea4a9ae6e5af37e7a15757a858/frontend/pages/docs/users/programmatically-using-bailo/open-api.mdx?plain=1#L9

Update inline

backend/src/routes/v3/model/images/getImage.ts
```diff
@@ -0,0 +1,74 @@
...
+  tag: z.string({
+    required_error: 'Must specify image tag as param',
+  }),
+})
+
+const getImageByDigestParams = getImageParams.extend({
+  digest: z.string({
+    required_error: 'Must specify image digest as param',
+  }),
+})
```
GB907762:
```suggestion
const getImageByDigestParams = z.object({
  modelId: z.string({
    required_error: 'Must specify model id as param',
  }),
  name: z.string({
    required_error: 'Must specify image name as param',
  }),
  tag: z.string({
    required_error: 'Must specify image tag as param',
  }),
  digest: z.string({
    required_error: 'Must specify image digest as param',
  }),
})
```
Small change

GB907762 (changes requested):

GB907762 (approved):

ARADDCC002 (approved):

PR #3600, opened 2026-05-14
Combine and send mirrored release images as a single archive

JRB66955:


backend/src/services/registry.ts
```diff
@@ -82,6 +82,14 @@ export function joinDistributionPackageName(distributionPackageName: Distributio
   return `${distributionPackageName.path}@${distributionPackageName['digest']}`
 }
 
+export function parseManifestFilePath(manifestFilePath: string) {
+  const match = manifestFilePath.match(/images\/manifests\/(.+)\/(.+)\.json$/)
```
PE39806, line 86: Replace dot-all with anything-but-slash (unless extra slashes should be included in the match).
```suggestion
  const match = manifestFilePath.match(/images\/manifests\/([^\/]+)\/([^\/]+)\.json$/)
```

backend/src/services/mirroredModel/exporters/image.ts
```diff
@@ -12,77 +12,73 @@ import { BaseExporter } from './base.js'
...
   protected async _init() {
-    const imageCheck = this.release.images.find((image) => image.id === this.image.id)
-    if (!imageCheck) {
-      throw InternalError('Could not find image associated with release.', {
-        modelId: this.model.id,
-        semver: this.release.semver,
-        imageId: this.image.id,
-        ...this.logData,
-      })
-    }
```
PE39806, line 51: It's still worth checking if the specified images are actually in the release.

backend/src/services/mirroredModel/exporters/image.ts
```diff
@@ -12,77 +12,73 @@ import { BaseExporter } from './base.js'
...
+    for (const image of this.images) {
+      const releaseAuth = await authorisation.release(this.user, this.model!, ReleaseAction.View, this.release)
+      if (!releaseAuth.success) {
+        throw Forbidden(releaseAuth.info, {
+          userDn: this.user.dn,
+          modelId: this.model.id,
+          semver: this.release.semver,
+          ...this.logData,
+        })
+      }
```
PE39806, line 66: This check does not include `image` so should be moved out of this loop.

backend/src/services/mirroredModel/exporters/image.ts
```diff
@@ -12,77 +12,73 @@ import { BaseExporter } from './base.js'
...
+
+      const imageAuth = await authorisation.image(this.user, this.model!, {
```
PE39806, line 68: Remove the loop and make a single call to `authorisation.images`.

backend/src/services/mirroredModel/importers/image.ts
```diff
@@ -13,34 +13,55 @@ import config from '../../../utils/config.js'
...
+export type ImageMirrorMetadataV1 = BaseMirrorMetadata & {
   importKind: MirrorKindKeys<'Image'>
   distributionPackageName: string
+  schemaVersion: 1
 }
+
+export type ImageMirrorMetadataV2 = BaseMirrorMetadata & {
+  importKind: MirrorKindKeys<'Image'>
+  // TODO FIX - temporary to resolve build issues
+  distributionPackageName: string
```
PE39806, line 28: I'd image the easiest fix is:
```suggestion
  distributionPackageNames: string[]
```

backend/src/services/mirroredModel/importers/image.ts
```diff
@@ -13,34 +13,55 @@ import config from '../../../utils/config.js'
...
 export type ImageMirrorInformation = {
-  metadata: ImageMirrorMetadata
-  image: { modelId: string; imageName: string; imageTag: string }
+  metadata: ImageMirrorMetadataV1 | ImageMirrorMetadataV2
+  images: Array<{ modelId: string; imageName: string; imageTag: string }>
 }
 
 export class ImageImporter extends BaseImporter {
-  declare protected readonly metadata: ImageMirrorMetadata
+  declare protected readonly metadata: ImageMirrorMetadataV1 | ImageMirrorMetadataV2
```
PE39806, line 38: It's probably easier to have two classes `ImageImporterV1` for the legacy handler and `ImageImporterV2` for the new `metadata: ImageMirrorMetadataV2`, and then update https://github.com/gchq/Bailo/blob/edd8b2ada4ba1b56446107c3a6f104504b89fd9b/backend/src/services/mirroredModel/mirroredModel.ts#L201 with the new class depending on the `metadata.schemaVersion`. This will preserve the legacy functionality, improve readability, reduce switches, and make the diff much easier to follow.
Common functionality could be put into some `BaseImageImporter` class if helpful.

backend/src/services/mirroredModel/importers/image.ts
```diff
@@ -13,34 +13,55 @@ import config from '../../../utils/config.js'
...
 import log from '../../log.js'
-import { splitDistributionPackageName } from '../../registry.js'
+import { parseManifestFilePath, splitDistributionPackageName } from '../../registry.js'
 import { BaseImporter, BaseMirrorMetadata } from './base.js'
 
-export type ImageMirrorMetadata = BaseMirrorMetadata & {
+export type ImageMirrorMetadataV1 = BaseMirrorMetadata & {
   importKind: MirrorKindKeys<'Image'>
   distributionPackageName: string
+  schemaVersion: 1
```
PE39806, line 22: This should be optional. If a legacy (pre-this-PR) version of Bailo exports an image then this field would be missing.

backend/src/services/mirroredModel/importers/image.ts
```diff
@@ -13,34 +13,55 @@ import config from '../../../utils/config.js'
...
+  protected manifestBody?: ImageManifestV2 | null = null
+  protected manifests = new Map<string, { imageName: string; imageTag: string; manifest: ImageManifestV2 }>()
 
-  static readonly manifestRegex = new RegExp(
+  static readonly manifestRegexV1 = new RegExp(
     String.raw`^${escapeRegExp(config.modelMirror.contentDirectory)}/manifest\.json$`,
   )
-  static readonly blobRegex = new RegExp(
+  static readonly manifestRegexV2 = new RegExp(
+    String.raw`^${escapeRegExp(config.modelMirror.contentDirectory)}/images/manifests.+json$`,
```
PE39806, line 50:
```suggestion
    String.raw`^${escapeRegExp(config.modelMirror.contentDirectory)}/images/manifests.+\.json$`,
```

PE39806 (changes requested):

backend/src/services/mirroredModel/importers/image.ts
GB907762: Yep. I started to implement a strategy pattern in the way you've described.

https://refactoring.guru/design-patterns/strategy/python/example

JRB66955:
Closing, not likely to be worked on further anytime soon due to changing priorities.

PR #3604, opened 2026-05-15
Dev/UI ux changes

ARADDCC002:


backend/package-lock.json
PE39806, line 1: No changes to `backend/package.json` - revert this.

frontend/package-lock.json
PE39806, line 1: No changes to `frontend/package.json` - revert this.

frontend/src/wrapper/TopNavigation.tsx
```diff
@@ -98,7 +98,7 @@ export default function TopNavigation({ drawerOpen = false, pageTopStyling = {},
         // TODO - use "theme.applyStyles" when implementing dark mode
         background: '#242424',
         ...theme.applyStyles('light', {
-          background: 'linear-gradient(276deg, rgba(214,37,96,1) 0%, rgba(84,39,142,1) 100%)',
+          background: 'linear-gradient(276deg, rgba(181, 73, 125,1) 0%,#4d3075 100%)',
```
PE39806: Is it possible to reuse `palette.primary.main` rather than `#4d3075` here?

lib/landing/src/docs/DocsWrapper.tsx
```diff
@@ -62,7 +72,7 @@ export default function DocsWrapper({ children }: DocsWrapperProps): ReactElemen
       return (
         <Fragment key={doc.slug}>
           {doc.header && doc.slug ? (
-            <ListItem dense sx={{ pl: paddingLeft }}>
+            <ListItem dense sx={{ pl: 1 + paddingLeft }} id={path.replaceAll('/', '-')}>
```
PE39806: Could refactor `path.replaceAll('/', '-')` to a new function as it is used in a few places.

PE39806 (dismissed):

lib/landing/src/docs/DocsWrapper.tsx
ARADDCC002: This change is just because the original `<DocsWrapper />` is copied over from `frontend` to `lib/landing`. Because we're technically only using it in two places I don't think it's worth moving to a new function.

GB907762 (dismissed):
Could you add some spacing to the paginate component?

ARADDCC002:
TODO: Fix the naming error of the updated marketplace image.

ARADDCC002:
> Could you add some spacing to the paginate component?

Yes this a good idea. I'll add an optional spacing prop to `<Paginate />`, that way we can decide as and when to use the spacing as I believe it looks better without in some places.

backend/docs/_static/custom.css
```diff
@@ -1,6 +1,6 @@
 :root {
-  --primary: #54278e;
-  --secondary: #d62560;
+  --primary: #4d3075;
```
PE39806, line 2: This change also needs to be copied to https://github.com/gchq/Bailo/blob/0f6520e930450b3d38722b0bfcac343827e1cacc/backend/docs/conf.py#L56

PE39806 (changes requested):

PE39806 (approved):

frontend/src/accessibility/report.mdx
PE39806, line 1: This change makes the `lib/landing` versions of these pages full width and hard to read. You _can_ go to a lot of effort to copy in all of the extra files but it still has strange spacing for the header and tab area (see below). As is, I don't think this looks good enough to be okay.

Image

PE39806 (changes requested):

frontend/src/accessibility/report.mdx
ARADDCC002: This is a good point. 

I wonder if we leave out the accessibility page changes in this PR so that we can get the others in, and update these at a later point?

lib/landing/pages/index.tsx
```diff
@@ -22,7 +22,7 @@ export default function Home() {
       <Box
         sx={{
           backgroundColor: '#f8e6dc',
-          background: 'linear-gradient(#54278e, #d62560)',
+          background: 'linear-gradient(#4d3075, #d62560)',
```
PE39806:
```suggestion
          background: 'linear-gradient(#4d3075, #b5497d)',
```
Or, ideally, replace both values with `${theme.palette.secondary.main}` etc.

frontend/src/theme.ts
PE39806:
```suggestion
            borderColor: '#b5497d',
```

PE39806:
```suggestion
            borderColor: '#b5497d',
```

PE39806 (changes requested):
Also update `frontend/src/common/Loading.tsx` `#d62560` -> `#b5497d`

PE39806 (approved):

JRB66955 (approved):

PR #3605, opened 2026-05-18
Correct agenda start order and rethrow on error

PE39806:


ARADDCC002 (approved):

backend/src/services/schedule/scheduler.ts
```diff
@@ -26,18 +26,18 @@ export async function startScheduler(jobRegistrars: JobRegistrar[] = []) {
...
-    await registerJob(agenda)
-  }
-
   try {
     await agenda.start()
     started = true
     log.info('Scheduler started')
   } catch (err) {
     log.error({ err }, 'Failed to start the scheduler')
+    throw err
```
GB907762, line 35: Would this crash the container on start up if agenda can't be resolved with the database for example?

PE39806: Yes, it will stop the backend from starting. This is similar to how the backend will not start if the scanners are unavailable but specified in config.

GB907762: Yeah I just wanted to check this was intentional 😄.
LGTM

GB907762 (approved):

PR #3606, opened 2026-05-19
Add new review kind for model lifecycle

ARADDCC002:


backend/src/models/Review.ts
```diff
@@ -15,6 +15,7 @@ export interface ReviewInterface {
 
   kind: ReviewKindKeys
   role: string
+  dueDate?: Date
```
JR40159: This type could be improved so that the due date is required for the kind of 'lifecycle'. A similar improvement could be made for the other review kinds

frontend/types/types.ts
```diff
@@ -568,20 +568,34 @@ type PartialReviewRequestInterface =
       semver: string
     }
 
+type PartialDueDateReviewRequestInterface =
+  | {
+      kind: 'release' | 'access'
+      dueDate?: Date
```
JR40159: Should be `never`?

JR40159 (commented):
Does this ticket include adding a response edit for this new review type or adding a single V3 endpoint for submitting responses for all the review types?

frontend/types/types.ts
ARADDCC002: Personally, I think it should be optional for these other types in case we want to use them in the future, but I can set to `never` if we feel it's not something we want to do in the foreseeable future.

ARADDCC002:
> Does this ticket include adding a response edit for this new review type or adding a single V3 endpoint for submitting responses for all the review types?

I can certainly add them - will pad the PR out a bit!

To add onto this, I think a `v3` end point to allow for any review response type is a good option. Will need to look into how `zod` handles conditional typing based on what `ReviewKind` it is.

backend/src/models/Review.ts
ARADDCC002: I agree - the frontend typing for this object is actually much more strict and accurate, should be able to copy over the updated fontend type to the backend. Another reason for shared types!

frontend/types/types.ts
JR40159: I think we should write the implementation for the use-case we currently have and updating these types will be very simple if we decide to update the other review types.

backend/src/routes/v3/review/postReviewResponse.ts
```diff
@@ -0,0 +1,84 @@
...
+      content: {
+        'application/json': {
+          schema: z.object({
+            response: responseInterfaceSchema,
+          }),
+        },
+      },
+    },
+  },
+})
```
JR40159: Need to add `v3` to update the V3 API docs

JR40159 (changes requested):

backend/src/models/Review.ts
JR40159: Still outstanding?

JR40159 (changes requested):

JR40159 (dismissed):

JR40159 (dismissed):

backend/src/services/v3/response.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  await reviewResponse.save()
+  await sendReviewResponseNotification(review, reviewResponse, user)
+
+  sendWebhooks(
+    review.modelId,
+    WebhookEvent.CreateReviewResponse,
+    `A new response has been added to a review requested for Model ${review.modelId}`,
+    { review: review },
+  )
```
PE39806, line 50: COULD: `sendWebhooks` is async so we should really have some `.catch` to prevent throwing unhandled/uncaught exceptions e.g.
```suggestion
  sendWebhooks(
    review.modelId,
    WebhookEvent.CreateReviewResponse,
    `A new response has been added to a review requested for Model ${review.modelId}`,
    { review: review },
  ).catch((err) =>
    log.error({
      err,
      modelId: review.modelId,
      eventKind: WebhookEvent.CreateReviewResponse,
      reviewId: review.id,
    }, 'Failed to send webhook'),
  )
```
This is not the pattern we currently follow in the code so I am happy for this to not be done here.

backend/test/services/v3/review.spec.ts
```diff
@@ -0,0 +1,35 @@
+import { describe, expect, test, vi } from 'vitest'
+
+import { findReviewById } from '../../../src/services/v3/review.js'
+import { getTypedModelMock } from '../../testUtils/setupMongooseModelMocks.js'
+
+const ReviewRoleModelMock = getTypedModelMock('ReviewRoleModel')
+
+vi.mock('../../../src/connectors/authentication/index.js', async () => ({
```
PE39806: Remove redundant `async`
```suggestion
vi.mock('../../../src/connectors/authentication/index.js', () => ({
```

backend/test/services/v3/review.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+  default: { getEntities: vi.fn(() => ['user:test']) },
+}))
+
+const modelMock = vi.hoisted(() => ({
+  getModelById: vi.fn(),
+}))
+vi.mock('../../../src/services/model.js', async () => modelMock)
```
PE39806: Remove redundant `async`
```suggestion
vi.mock('../../../src/services/model.js', () => modelMock)
```

backend/test/services/v3/response.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { describe, expect, test, vi } from 'vitest'
+
+import { Decision } from '../../../src/models/Response.js'
+import { respondToReview } from '../../../src/services/v3/response.js'
+import { getTypedModelMock } from '../../testUtils/setupMongooseModelMocks.js'
+import { testReleaseReview } from '../../testUtils/testModels.js'
+
+vi.mock('../../../src/connectors/authorisation/index.js')
+vi.mock('../../../src/connectors/authentication/index.js', async () => ({
```
PE39806: Remove redundant `async`
```suggestion
vi.mock('../../../src/connectors/authentication/index.js', () => ({
```

backend/test/services/v3/response.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+}))
+
+const ResponseModelMock = getTypedModelMock('ResponseModel')
+
+const reviewV3Mock = vi.hoisted(() => ({
+  findReviewById: vi.fn(function () {
+    return [testReleaseReview]
+  }),
+}))
+vi.mock('../../../src/services/v3/review.js', async () => reviewV3Mock)
```
PE39806: Remove redundant `async`
```suggestion
vi.mock('../../../src/services/v3/review.js', () => reviewV3Mock)
```

backend/test/services/v3/response.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const responseV2Mock = vi.hoisted(() => ({
+  sendReviewResponseNotification: vi.fn(),
+}))
+vi.mock('../../../src/services/response.js', async () => responseV2Mock)
```
PE39806: Remove redundant `async`
```suggestion
vi.mock('../../../src/services/response.js', () => responseV2Mock)
```

backend/test/services/v3/response.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const mockWebhookService = vi.hoisted(() => {
+  return {
+    sendWebhooks: vi.fn(),
+  }
+})
```
PE39806: Simplify to
```suggestion
const mockWebhookService = vi.hoisted(() => ({
  sendWebhooks: vi.fn(),
}))
```

backend/test/routes/v3/review/postReviewResponse.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+import { createFixture, testPost } from '../../../testUtils/routes.js'
+import { testReviewResponse } from '../../../testUtils/testModels.js'
+
+vi.mock('../../../../src/connectors/audit/index.js')
+
+const mockResponseService = vi.hoisted(() => {
+  return {
+    respondToReview: vi.fn(() => testReviewResponse),
+  }
+})
```
PE39806: Simplify
```suggestion
const mockResponseService = vi.hoisted(() => ({
  respondToReview: vi.fn(() => testReviewResponse),
}))
```

PE39806 (changes requested):

JRB66955 (dismissed):

backend/src/models/Review.ts
```diff
@@ -1,23 +1,34 @@
...
+      dueDate?: never
+      semver: string
+      accessRequestId: never
+    } & PartialReviewInterface)
+  | ({
+      kind: 'lifecycle'
+      dueDate: Date
+      semver?: never
+      accessRequestId?: never
+    } & PartialReviewInterface)
```
GB907762, line 24: So `never` is type for describing the return type when a function throws an error or doesn't resolve. See bellow for usage.

```typescript
// Function returning never must not have a reachable end point
function error(message: string): never {
  throw new Error(message);
}
 
// Inferred return type is never
function fail() {
  return error("Something failed");
}
 
// Function returning never must not have a reachable end point
function infiniteLoop(): never {
  while (true) {}
}
```
https://www.typescriptlang.org/docs/handbook/basic-types.html#never

Instead you want undefined. have a look through this PR if you can find any other examples

```suggestion
export type ReviewInterface = PartialReviewInterface &
  (
    | {
        kind: 'access'
        accessRequestId: string
        semver?: undefined
        dueDate?: undefined
      }
    | {
        kind: 'release'
        semver: string
        accessRequestId?: undefined
        dueDate?: undefined
      }
    | {
        kind: 'lifecycle'
        dueDate: Date
        semver?: undefined
        accessRequestId?: undefined
      }
  )
```

backend/src/services/v3/review.ts
```diff
@@ -0,0 +1,74 @@
...
+      .unwind({ path: '$model' })
+      .match(await findUserInCollaborators(user))
+      .limit(1)
+  ).at(0)
+  if (!review) {
+    throw NotFound(`Unable to find Review to respond to.`, { reviewId })
+  }
+
+  // Authorisation check to make sure the user can access a model
+  await getModelById(user, review.modelId)
```
GB907762, line 37: Could you do the auth check in first instance? Normal pattern.

backend/src/services/v3/response.ts
GB907762, line 1: Replace `null` with `undefined` in here

backend/src/services/v3/review.ts
GB907762, line 1: I'm not sure about adding a `v3` services folder. At the route layer this is separated because we use separate routers but I think in the service layer we should keep things inplace?

GB907762 (changes requested):

backend/src/services/v3/review.ts
ARADDCC002: Unfortunately, we don't have access to the `modelId` until after we've done the `aggregate` query. We _could_ pass in `modelId` as part of the request, but I'd prefer if possible to keep the API as small as possible.

backend/src/services/v3/review.ts
ARADDCC002: I discussed this with other members of the team and they preferred this approach. We can discuss again and if consensus is to move it back that's fine.

backend/src/services/v3/response.ts
ARADDCC002: I'd prefer to stay consistent in this PR but I agree this is something we should change in the future.

GB907762 (approved):

PE39806 (approved):

PR #3617, opened 2026-05-21
UI changes for review lifecyle

ARADDCC002:


PE39806 (changes requested):
Observations & queries:

1. It is possible to set the "next review date" _before_ a schema has been chosen. Is this intended behaviour? I would expect the review to only be possible once a schema has been chosen as, otherwise, there is nothing to review.
2. The "Review" button in the sidebar does not add to the counter for upcoming nor overdue open "Model card" reviews. I would certainly expect overdue open reviews to count, and also  likely expect upcoming open reviews to be included. This may be a regression as a pending Access Request review also does not show the count. Does the head request require updating?
3. The "Model card" reviews does not display the `dueDate` field, nor have an order by for this property. I would expect "Model card" reviews to be sorted by `dueDate` (ascending/soonest first) by default.
4. If a user has appropriate role(s), it would be good to have some to "jump to the model card review" button on the model, especially as the review has a "Back to model" button.
5. When reviewing, leaving "Next review date" blank returns the error "Bad Request: Due date of next review cannot be in the past." which is a little misleading. Instead, their should be some indication that this is a required field.
6. If the review outcome is "Request Changes" then the archived review states "You have requested changes as a undefined". This is not a problem if the review outcome was "Approve".
7. If the review outcome is "Request Changes" then the model card's "Next review due:" is shown as blank rather than the date chosen in the review. This is not a problem if the review outcome was "Approve".
8. Clicking "View full model card" while reviewing opens an empty popup - not the actual model card.
9. I cannot view the comments on any archived "Model card" reviews (neither outcome).
10. Change the Title Case "Request Changes" text to Sentence case "Request changes".

ARADDCC002:
> Observations & queries:
> 
>     1. It is possible to set the "next review date" _before_ a schema has been chosen. Is this intended behaviour? I would expect the review to only be possible once a schema has been chosen as, otherwise, there is nothing to review.
> 
>     2. The "Review" button in the sidebar does not add to the counter for upcoming nor overdue open "Model card" reviews. I would certainly expect overdue open reviews to count, and also  likely expect upcoming open reviews to be included. This may be a regression as a pending Access Request review also does not show the count. Does the head request require updating?
> 
>     3. The "Model card" reviews does not display the `dueDate` field, nor have an order by for this property. I would expect "Model card" reviews to be sorted by `dueDate` (ascending/soonest first) by default.
> 
>     4. If a user has appropriate role(s), it would be good to have some to "jump to the model card review" button on the model, especially as the review has a "Back to model" button.
> 
>     5. When reviewing, leaving "Next review date" blank returns the error "Bad Request: Due date of next review cannot be in the past." which is a little misleading. Instead, their should be some indication that this is a required field.
> 
>     6. If the review outcome is "Request Changes" then the archived review states "You have requested changes as a undefined". This is not a problem if the review outcome was "Approve".
> 
>     7. If the review outcome is "Request Changes" then the model card's "Next review due:" is shown as blank rather than the date chosen in the review. This is not a problem if the review outcome was "Approve".
> 
>     8. Clicking "View full model card" while reviewing opens an empty popup - not the actual model card.
> 
>     9. I cannot view the comments on any archived "Model card" reviews (neither outcome).
> 
>     10. Change the Title Case "Request Changes" text to Sentence case "Request changes".

1. In my head I thought it might be worth adding in case people wanted reminding to set a schema, but yes looking back I can disable that, at least in the UI (I don't think it needs locking down in the API as it won't break anything).
2. This intended as currently if there's a due date set there will always be an open review for that model. In the future we can certainly change this until the review is due X weeks before the date set.
3. This is a good suggestion
4. As reviews are always active for a model if the date is set I don't want to make it as distracting as the release/access request ones but yes I like this suggestion.
5. Good suggestion
6. Will look into this!
7. ^
8. This should be fixed in the latest commit
9. Yep this is something I might address in a future PR as to not make this one bigger than it needs to be
10. Good suggestion

backend/src/services/v3/response.ts
```diff
@@ -49,6 +55,41 @@ export async function respondToReview(
...
+    user,
+    existingReviews.map((review) => review.model),
+    ModelAction.Update,
+  )
+  existingReviews = existingReviews.filter((_, i) => auths[i].success)
+  if (existingReviews) {
+    for (const existingReview of existingReviews) {
+      const reviewToDelete = await ReviewModel.findOne({ _id: existingReview._id })
+      if (reviewToDelete) {
+        reviewToDelete.delete()
```
JRB66955: This should be `await`'ed.

backend/src/services/v3/review.ts
```diff
@@ -72,3 +75,80 @@ export async function findUserInCollaborators(user: UserInterface) {
...
+    user,
+    existingReviews.map((review) => review.model),
+    ModelAction.Update,
+  )
+  existingReviews = existingReviews.filter((_, i) => auths[i].success)
+  if (existingReviews) {
+    for (const existingReview of existingReviews) {
+      const reviewToDelete = await ReviewModel.findOne({ _id: existingReview._id })
+      if (reviewToDelete) {
+        reviewToDelete.delete()
```
JRB66955: This should be `await`'ed.

backend/src/services/v3/response.ts
```diff
@@ -49,6 +55,41 @@ export async function respondToReview(
...
+
+  let existingReviews = await ReviewModel.aggregate(stages)
+
+  const auths = await authorisation.models(
+    user,
+    existingReviews.map((review) => review.model),
+    ModelAction.Update,
+  )
+  existingReviews = existingReviews.filter((_, i) => auths[i].success)
+  if (existingReviews) {
```
JRB66955: TypeScript interprets this as an array, therefore this will always be truthy.
```suggestion
  if (existingReviews.length > 0) {
```

backend/src/services/v3/response.ts
```diff
@@ -49,6 +55,41 @@ export async function respondToReview(
     { review: review },
   )
 
+  const stages: PipelineStage[] = [
```
JRB66955: This aggregate pipeline is very similar to the `v3/review.ts` with most of its stages, perhaps there's a way to simplify/abstract this in some way? Not too fussed either way though.

JRB66955 (changes requested):

backend/src/services/v3/response.ts
ARADDCC002: I think if that's the case we can get rid of the check altogether, if the array is empty it won't trigger the for loop.

backend/src/services/v3/response.ts
ARADDCC002: I'm tempted to leave for now, just because I don't know where else we'd use this as it's very specific to these two use-cases. I'll have a think, though.

JRB66955 (dismissed):

backend/src/services/v3/review.ts
```diff
@@ -72,3 +75,80 @@ export async function findUserInCollaborators(user: UserInterface) {
...
+      throw BadReq(`Creating ${reviewContent.kind} reviews is not supported yet in V3.`)
+
+    default:
+      throw BadReq('Unknown review kind')
+  }
+}
+
+async function createLifecycleReview(user: UserInterface, modelId: string, dueDate: Date): Promise<ReviewInterface> {
+  // Authorisation check to make sure the user can access a model
+  await getModelById(user, modelId)
```
GB907762: This doesn't check authorisation at all. It just checks whether you can see a model which most people can. 
Do something like:

```suggestion
  const auth = await authorisation.model(user, modelId,  Model.Update)
  if (!auth.success) {
    throw Forbidden(auth.info, { userDn: user.dn, accessRequestId })
  }
```

backend/src/services/v3/review.ts
```diff
@@ -72,3 +75,80 @@ export async function findUserInCollaborators(user: UserInterface) {
...
+      $sort: { createdAt: -1 },
+    },
+    { $lookup: { from: 'v2_models', localField: 'modelId', foreignField: 'id', as: 'model' } },
+    { $unwind: { path: '$model' } },
+  ]
+
+  stages.push({ $lookup: { from: 'v2_responses', localField: '_id', foreignField: 'parentId', as: 'responses' } })
+  stages.push({ $addFields: { responseCount: { $size: '$responses' } } })
+  stages.push({ $match: { responseCount: 0 } })
+  stages.push({ $unset: ['responses', 'responseCount'] })
```
GB907762: I don't like this. Could we preserve all documents when you update one.

backend/src/services/v3/review.ts
```diff
@@ -72,3 +75,80 @@ export async function findUserInCollaborators(user: UserInterface) {
...
+    ModelAction.Update,
+  )
+  existingReviews = existingReviews.filter((_, i) => auths[i].success)
+  if (existingReviews) {
+    for (const existingReview of existingReviews) {
+      const reviewToDelete = await ReviewModel.findOne({ _id: existingReview._id })
+      if (reviewToDelete) {
+        await reviewToDelete.delete()
+      }
+    }
```
GB907762: Look here: https://mongoosejs.com/docs/api/model.html#Model.deleteMany() 

But refer to my previous comment about removing this entirely.

GB907762 (changes requested):

JR40159 (dismissed):
Need to update the reviews section of the documentation

ARADDCC002:
@GB907762 Logic has now changed based off feedback. If there's an existing review for a model it shouldn't create a new one.

PE39806 (changes requested):
The "Undo review" button when accessing an archived review throws the error `Bad Request: Path: body.decision - Message: Invalid literal value, expected "approve"`

frontend/src/reviews/ReviewItem.tsx
```diff
@@ -32,7 +32,10 @@ export default function ReviewItem({ review }: ReviewItemProps) {
   }
 
   function handleListItemClick() {
-    router.push(`/model/${review.model.id}/${determineReviewPath()}/review?role=${review.role}`)
+    // Lifecycle reviews cannot be re-reviewed
+    if (review.kind !== ReviewKind.LIFECYCLE) {
```
PE39806: I like that this link is disabled, but I think that it being clickable is confusing and makes it look like something should be happening. Is it possible to, instead, make the list item so that it cannot be clicked and doesn't have the mouse pointer etc?

PE39806 (changes requested):

backend/src/services/v3/response.ts
```diff
@@ -49,16 +51,8 @@ export async function respondToReview(
...
-  if (review.kind === ReviewKind.Lifecycle && response.decision === Decision.Approve) {
-    const newReview = new ReviewModel({
-      modelId: review.modelId,
-      kind: review.kind,
-      role: review.role,
-      dueDate,
-    })
-    await newReview.save()
+  if (review.kind === ReviewKind.Lifecycle && dueDate) {
+    createLifecycleReview(user, review.modelId, dueDate)
```
PE39806: This should be `await`ed
```suggestion
    await createLifecycleReview(user, review.modelId, dueDate)
```

backend/src/services/review.ts
```diff
@@ -42,6 +43,10 @@ export async function findReviews(
         ...(modelId && { modelId }),
         ...(semver && { semver }),
         ...(accessRequestId && { accessRequestId }),
+        ...(reviewId &&
+          Types.ObjectId.isValid(reviewId) && {
```
PE39806: I'd prefer we throw if `reviewId` is specified but invalid - otherwise this is silently ignored which might have confusing results.

backend/src/services/v3/review.ts
```diff
@@ -72,3 +75,80 @@ export async function findUserInCollaborators(user: UserInterface) {
...
+      $sort: { createdAt: -1 },
+    },
+    { $lookup: { from: 'v2_models', localField: 'modelId', foreignField: 'id', as: 'model' } },
+    { $unwind: { path: '$model' } },
+  ]
+
+  stages.push({ $lookup: { from: 'v2_responses', localField: '_id', foreignField: 'parentId', as: 'responses' } })
+  stages.push({ $addFields: { responseCount: { $size: '$responses' } } })
+  stages.push({ $match: { responseCount: 0 } })
+  stages.push({ $unset: ['responses', 'responseCount'] })
```
PE39806: This is a lot of stages and I worry about the efficiency of this. I _think_ the following has equivalent functionality but should be quicker due to fewer `$lookup`s.
```suggestion
  const stages: PipelineStage[] = [
    {
      $match: {
        modelId,
        kind: ReviewKind.Lifecycle,
      },
    },
    {
      $lookup: {
        from: 'v2_responses',
        localField: '_id',
        foreignField: 'parentId',
        as: 'responses',
      },
    },
    {
      $match: {
        responses: { $size: 0 },
      },
    },
    {
      $limit: 1,
    },
  ]
```

frontend/actions/review.ts
```diff
@@ -47,33 +54,52 @@ export function useGetReviewRequestsForUser(open?: boolean) {
...
-    (modelId && semver) || (modelId && accessRequestId)
+    modelId
       ? `/api/v2/reviews?${qs.stringify({
           mine: false,
           modelId,
           ...(semver && { semver }),
           ...(accessRequestId && { accessRequestId }),
+          ...(reviewId && { reviewId }),
+          ...(kind && { kind }),
+          ...(open && { open }),
```
PE39806: Be careful with this `falsey` value
```suggestion
          ...(open !== undefined && { open }),
```

frontend/actions/review.ts
```diff
@@ -47,33 +54,52 @@ export function useGetReviewRequestsForUser(open?: boolean) {
...
+  open,
+}: GetReviewRequestsForModelQuery) {
   const { data, isLoading, error, mutate } = useSWR<
     {
       reviews: ReviewRequestInterface[]
     },
     ErrorInfo
   >(
-    (modelId && semver) || (modelId && accessRequestId)
+    modelId
```
PE39806: Is it possible for `modelId` to be defined but still be getting other values like `semver`, `accessRequestId` and/or `reviewId` (e.g. other requests are slow)? If so, this could fail.

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -82,21 +82,34 @@ export default function SideNavigation({
...
   const { responses, isResponsesLoading, isResponsesError } = useGetUserResponses()
 
   useEffect(() => {
     async function fetchReviewCount() {
       onResetErrorMessage()
-      if (reviewCountHeader) {
-        setReviewCount(reviewCountHeader)
+      if (releaseReviewCountHeader) {
+        setReviewCount(releaseReviewCountHeader + accessRequestReviewCountHeader)
       }
```
PE39806: This won't change when the count goes to 0, and doesn't work when only `accessRequestReviewCountHeader` >0. This will also go wrong if we ever get a `NaN` which has happened in the past. Prefer
```suggestion
      const releaseCount = releaseReviewCountHeader ?? 0
      const accessCount = accessRequestReviewCountHeader ?? 0
      setReviewCount(releaseCount + accessCount)
```

frontend/utils/dateUtils.ts
```diff
@@ -27,6 +27,10 @@ export const formatDateStringAsMonthAndYear = (value: string) => {
   return dayjs(value).format('MMM YYYY')
 }
 
+export const formatSDateStringAsDayMonthAndYear = (value: string) => {
```
PE39806: Typo?
```suggestion
export const formatDateStringAsDayMonthAndYear = (value: string) => {
```

PE39806 (changes requested):

frontend/actions/review.ts
ARADDCC002: I've changed this as non-optional. The API _can_ call be called without `modelId`, but from a fontend perspective this should take in a `modelId` due to context of the hook name. 

Alternatively, we change the name of the hook to `useGetReviewRequests` instead? That way we can keep `modelId` optional.

PE39806 (dismissed):

frontend/pages/model/[modelId]/lifecycle/[reviewId]/review.tsx
```diff
@@ -0,0 +1,153 @@
...
+    if (!reviewId) {
+      return setErrorMessage('Could not find review ID')
+    }
+    if (decision === Decision.Approve && !dueDate) {
+      return setErrorMessage('Please provide a valid due date.')
+    }
+
+    setIsReviewButtonLoading(true)
+    const res = await postGenericReviewResponse({
+      kind: 'lifecycle',
```
Comment (GB907762) [inline frontend/pages/model/[modelId]/lifecycle/[reviewId]/review.tsx]: Consider an enum?
```suggestion
      kind: ReviewKind.Lifecycle,
```

frontend/pages/model/[modelId]/lifecycle/[reviewId]/review.tsx
```diff
@@ -0,0 +1,153 @@
...
+            <ReviewWithComment
+              onSubmit={handleSubmit}
+              reviews={reviews}
+              loading={isReviewButtonLoading}
+              modelId={modelId as string}
+              includeDueDate
+              hideRequestChangesButton
+            />
+            <MessageAlert message={errorMessage} severity='error' />
+            <Divider />
```
Comment (GB907762) [inline frontend/pages/model/[modelId]/lifecycle/[reviewId]/review.tsx]: Is this divider a typo copied from the other component?
```suggestion
```

GB907762 (approved):

PE39806 (approved):

PR #3619, opened 2026-05-22
Add compliance role

JR40159:


frontend/actions/user.ts
```diff
@@ -45,6 +64,25 @@ export function useGetCurrentUser() {
   }
 }
 
+export function useGetCurrentUserV3(): UseGetCurrentUserV3Result {
```
2000090079, line 67: There seems to be duplicated loading and error handling logic between `useGetCurrentUser` and `useGetCurrentUserV3`. Could some of this logic be abstracted into a reusable helper or shared hook?

frontend/pages/metrics.tsx
```diff
@@ -1,17 +1,16 @@
...
 import Link from 'src/Link'
 import OverviewMetrics from 'src/metrics/OverviewMetrics'
 import PolicyMetrics from 'src/metrics/PolicyMetrics'
 
 export default function Metrics() {
-  const { currentUser, isCurrentUserLoading, isCurrentUserError } = useGetCurrentUser()
+  const currentUser = useContext(CurrentUserContext)
 
-  if (!currentUser || !currentUser.isAdmin) {
+  if (!currentUser.systemRoles.includes('compliance')) {
```
2000090079: Should we consider using optional chaining here (`currentUser?.systemRoles?.includes(...)`) to make the check safer during initial renders or unexpected context states?

frontend/src/Wrapper.tsx
```diff
@@ -93,57 +99,67 @@ export default function Wrapper({ children }: WrapperProps): ReactElement {
...
-                  <Announcement message={uiConfig.announcement.text} onClose={handleAnnouncementOnClose} />
-                )}
-                <Box paddingTop={4}>
-                  <MessageAlert message={errorMessage} severity='error' />
-                  {children}
-                  <Copyright sx={{ p: 2 }} />
-                </Box>
-              </>
-            )}
+    <CurrentUserContext.Provider value={currentUserV3}>
```
2000090079: There appears to be a large duplicated layout block after introducing `CurrentUserContext.Provider`. Could some of this shared layout/UI structure be extracted into a reusable wrapper component to improve maintainability?

2000090079 (commented):
Reviewed the frontend-related changes and left a few maintainability and readability suggestions.

frontend/actions/user.ts
```diff
@@ -45,6 +64,25 @@ export function useGetCurrentUser() {
...
+  const { data, isLoading, error, mutate } = useSWR<UserResponseV3, ErrorInfo>('/api/v3/entities/me', fetcher)
+
+  if (isLoading) {
+    return { isCurrentUserLoading: true }
+  }
+  if (error) {
+    return { isCurrentUserError: error }
+  }
+  if (!data?.user) {
```
ON93754: Optional: This might be more readable if it was instead:
`if (!data || !data.user) {`

frontend/pages/metrics.tsx
ON93754: Optional: Consider using the enum value rather than the string value 'compliance' itself - easy to make typos

frontend/src/Wrapper.tsx
ARADDCC002: We should add this context wrapper inside `_app.tsx` where we have all of our other contexts. We're now also getting to the point where we've got _lots_ of context providers, and should look for a better way of managing them.

ARADDCC002 (changes requested):

frontend/actions/user.ts
JR40159: For V3 of the current user functionality introduced in this PR we're using a context which will massively reduce duplication when the uses of the `useGetCurrentUser` are moved to using this new context. When everything uses the context then the loading and error handling will be done once in the context, however refactoring the rest of the frontend is out of scope for this PR and will be done in a separate PR.

frontend/pages/metrics.tsx
JR40159: We type the responses the frontend gets back from the backend and are responsible for ensuring those types are accurate. Introducing optional chaining here risks masking a broken contract between the frontend and the backend — if `systemRoles` were missing, the guard would silently pass and potentially allow unauthorised access rather than failing visibly.

frontend/src/Wrapper.tsx
JR40159: I'm not 100% sure I understand your comment here, however, in the latest changes I've add an app context component which hopefully simplifies the use of contexts.

JR40159:
> Reviewed the frontend-related changes and left a few maintainability and readability suggestions.

Thank you very much for the feedback, really appreciated! We'd always welcome more in the future!

PE39806 (dismissed):

backend/src/routes/v3/routes.ts
```diff
@@ -16,4 +17,6 @@ router.get('/metrics/usage', ...getUsageMetrics)
 router.get('/metrics/compliance', ...getComplianceMetrics)
 router.get('/metrics/entryVolume', ...getEntryVolume)
```
GB907762, line 19: You've added a compliance role although these are not accessible via that endpoint. They're only useful to admins. Could you amend the backend the same way you've locked off the frontend.

backend/test/routes/v3/entities/getCurrentUser.spec.ts
GB907762, line 1: Could you just have a look at `backend/test/connectors/metrics/base.spec.ts` I don't think this tests role access very well.

frontend/pages/metrics.tsx
```diff
@@ -1,17 +1,17 @@
...
 import OverviewMetrics from 'src/metrics/OverviewMetrics'
 import PolicyMetrics from 'src/metrics/PolicyMetrics'
+import { Roles } from 'types/types'
 
 export default function Metrics() {
-  const { currentUser, isCurrentUserLoading, isCurrentUserError } = useGetCurrentUser()
+  const currentUser = useContext(CurrentUserContext)
 
-  if (!currentUser || !currentUser.isAdmin) {
+  if (!currentUser.systemRoles.includes(Roles.Compliance)) {
```
GB907762: Should this also include Admins?

frontend/pages/metrics.tsx
```diff
@@ -1,17 +1,17 @@
...
     return (
       <Forbidden
```
GB907762: Instead of making this whole page forbidden would it make more sense to hide this on the sidebar entirely? We already do this with non admins?

If you have a compliance role but not an admin role the 403 gets rendered in the frontend as an alert from the backend. If you so happened to go this page without the link in the sidebar?

backend/src/routes/v3/entities/getCurrentUser.ts
```diff
@@ -0,0 +1,46 @@
...
+  'v3',
+)
+
+export const getCurrentUser = [
+  async (req: Request, res: Response<GetCurrentUserResponse>): Promise<void> => {
+    req.audit = AuditInfo.ViewCurrentUserInformation
+    const systemRoles = [
+      ...((await authentication.hasRole(req.user, Roles.Admin)) ? [Roles.Admin] : []),
+      ...((await authentication.hasRole(req.user, Roles.Compliance)) ? [Roles.Compliance] : []),
+    ]
```
GB907762, line 40: As a consideration perhaps we could use this since this looks a bit challenging to maintain long term? 

[ListGroupsCommand](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/cognito-identity-provider/command/ListGroupsCommand/)

GB907762 (changes requested):

backend/src/routes/v3/routes.ts
JR40159: Really good spot!

frontend/pages/metrics.tsx
JR40159: No, I think the frontend should be kept simple and but then in the backend the compliance role could be given to admins if needed.

GB907762 (dismissed):

ARADDCC002 (dismissed):

ARADDCC002 (approved):

PE39806 (approved):

PR #3620, opened 2026-05-22
Add middleware to protect against DoS

JR40159:
Following guidance https://cheatsheetseries.owasp.org/cheatsheets/Nodejs_Security_Cheat_Sheet.html#monitor-the-event-loop


ARADDCC002 (dismissed):

PE39806 (dismissed):

backend/src/routes/middleware/tooBusy.ts
```diff
@@ -0,0 +1,12 @@
+import toobusy from 'toobusy-js'
+
+import log from '../../services/log.js'
+
+export function tooBusy(req, res, next) {
+  if (toobusy()) {
+    log.warn('Server response time too long, preventing more requests from being handled.')
+    res.status(503).send('Server Too Busy')
```
GB907762: throw a `ConfigurationError`

GB907762 (changes requested):
I think rate limiting might be good? You could abuse this by just hogging Bailo resources? Meaning no one else can use the tool?

backend/src/routes/middleware/tooBusy.ts
ARADDCC002: I naively assumed this library was only limiting requests from the same origin. Also, is `ConfigurationError` what we want? To me that would imply that there was a problem with the configuration, whereas this is a purposely thrown error.

GB907762: I think our naming for this error isn't correct. Really it should be `ServiceUnavailable` but we should let our error handler handle it so that error is consistent for the user, inline with all other errors within the app.

Have a look [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/503) for usage of a 503

`Common causes are that a server is down for maintenance or overloaded`

GCHQDeveloper36:
That library hasn't been updated in 10 years (although how much maintenance does it need) - maybe a slightly different approach by using rate limiting like suggested above - perhaps: https://www.npmjs.com/package/express-rate-limit instead?

GB907762:
> That library hasn't been updated in 10 years (although how much maintenance does it need) - maybe a slightly different approach by using rate limiting like suggested above - perhaps: https://www.npmjs.com/package/express-rate-limit instead?

https://github.com/STRML/node-toobusy/blob/master/toobusy.js#L35-L41

`too-busy.js` looks equally old. Although it looks fairly minimal. You could make a small function

JR40159:
Thanks for all the feedback! I've updated the PR to use [express-rate-limit](https://github.com/express-rate-limit/express-rate-limit) which is IP based. Happy for further discussion on a sensible default, for now I've gone with what they have in their docs as the actual default is 5 requests in a minute. Also interested in opinions if this should be across the whole app, route specific or should exclude the frontend.

JR40159:
Another option could be [express-slow-down](https://www.npmjs.com/package/express-slow-down) which slows down responses rather than blocking them.

PE39806:
This might need disabling for our integration tests...

sylvesterkaczmarek (commented):
The IP rate limit is not proxy-safe in Bailo's current deployment topology and can turn into a shared/global limiter.

Bailo's NGINX configurations proxy requests to the backend and explicitly set `X-Forwarded-For`, but the Express app never configures `trust proxy`. With Express' default `trust proxy = false`, `express-rate-limit` keys on `req.ip`, which is the direct reverse-proxy address rather than the originating client. Users behind the same NGINX/ingress therefore share the same 100-request/15-minute bucket, so one client can exhaust it and cause 429s for everyone else behind that proxy.

Recent `express-rate-limit` versions also detect exactly this combination (`X-Forwarded-For` present while `trust proxy` is false) as `ERR_ERL_UNEXPECTED_X_FORWARDED_FOR` because it prevents accurate client identification.

Could proxy trust be configured explicitly and safely for the supported deployment topology (specific hop count/CIDR rather than `true`), or could the limiter use a key generator based on a validated proxy chain? I would add an integration test with two different X-Forwarded-For client addresses behind the same proxy and verify they get independent buckets.

PR #3626, opened 2026-05-26
Statically host swagger docs on GitHub pages

### LINKED ISSUE #1575: Swagger API not resolving
line from here https://gchq.github.io/Bailo/docs/users/programmatically-using-bailo/open-api

to bailo api end points - not resolving 

https://gchq.github.io/docs/api

PE39806:
- Add static Swagger docs for API (closes #1575) by generating OpenAPI spec `backend/src/scripts/generateOpenApiSpecs.ts`
- Add new navigation links (API, Python Docs, Python SDK)
- Split navigation into primary (AppBar + Hero) and all (+ Footer)
- Embed Python docs in `iframe` within React app to reuse `AppBar`
- Add Copyright to static pages
- Fix broken `isDocsPage` logic
- Add missing eslint ignores etc.
- Remove some files that are copied by `lib/landing/scripts/generate.js` from `frontend` dir


lib/landing/next-env.d.ts
ARADDCC002, line 1: Was this added by mistake?

ARADDCC002 (changes requested):

lib/landing/public-static/api/docs/index.html
```diff
@@ -0,0 +1,75 @@
...
+            { url: './swagger-v3.json', name: 'v3.0.0(beta)' },
+          ],
+          dom_id: '#swagger-ui',
+          deepLinking: true,
+          presets: [SwaggerUIBundle.presets.apis, SwaggerUIStandalonePreset],
+          plugins: [SwaggerUIBundle.plugins.DownloadUrl],
+          layout: 'StandaloneLayout',
+          supportedSubmitMethods: [],
+        })
+      }
```
GB907762: The documentation suggests you can use this as a react component. Might be better and make the theming more consistent?

lib/landing/scripts/generate.js
```diff
@@ -96,6 +96,19 @@ async function run() {
...
+  mkdirSync('./public/api/docs', { recursive: true })
+  cp('../../backend/openapi-specs/swagger-v2.json', './public/api/docs/')
+  cp('../../backend/openapi-specs/swagger-v3.json', './public/api/docs/')
+  // see https://swagger.io/docs/open-source-tools/swagger-ui/usage/installation/#static-files-without-http-or-html
+  cp('./node_modules/swagger-ui-dist/swagger-ui-bundle.js', './public/api/docs/')
+  cp('./node_modules/swagger-ui-dist/swagger-ui-standalone-preset.js', './public/api/docs/')
+  cp('./node_modules/swagger-ui-dist/swagger-ui.css', './public/api/docs/')
+  cp('./node_modules/swagger-ui-dist/index.css', './public/api/docs/')
+  cp('./node_modules/swagger-ui-dist/favicon-16x16.png', './public/api/docs/')
+  cp('./node_modules/swagger-ui-dist/favicon-32x32.png', './public/api/docs/')
```
GB907762: Check my other comment. I think this might be able to be removed?

lib/landing/scripts/generate.js
```diff
@@ -96,6 +96,19 @@ async function run() {
...
+  cp('./public-static/api/docs/index.html', './public/api/docs/')
```
GB907762: I don't understand moving this? Should it not be in public regardless?

lib/landing/src/Wrapper.tsx
```diff
@@ -43,7 +45,9 @@ type WrapperProps = {
 }
 
 export default function Wrapper({ title, page, children }: WrapperProps): ReactElement {
-  const isDocsPage = useMemo(() => page.startsWith('docs'), [page])
+  const router = useRouter()
+
+  const isDocsPage = router.pathname === '/docs' || router.pathname.startsWith('/docs/')
```
GB907762: Potentially problematic. I know this is the case in Bailo frontend. Checkout `router.isReady`

GB907762 (changes requested):

lib/landing/next-env.d.ts
PE39806: Yes - from https://nextjs.org/docs/app/api-reference/config/typescript#next-envdts
> We recommend adding `next-env.d.ts` to your `.gitignore` file.

lib/landing/scripts/generate.js
PE39806: `public` is set up to only hold copied files and is deleted as a step in https://github.com/gchq/Bailo/blob/373dc07408ae06b1062e21c8c01d0cb627dac54b/lib/landing/scripts/generate.js#L74

lib/landing/public-static/api/docs/index.html
PE39806: We originally rendered this as a React component for the built app's docs and then switched away from it due to `swagger-ui-react` not having support for React v19 (in #2770). Seeing as https://github.com/swagger-api/swagger-ui/issues/10243 is still open, I assume we cannot just render this as a react component without downgrading react which we do not want to do.

PE39806: Apologies, I had misunderstood what exactly each package did. This has now been replaced with a React component.

GB907762 (approved):

lib/landing/public-static/api/docs/index.html
GB907762: Looks awesome 🎉 Thanks.

ARADDCC002 (approved):

Issue #2422, opened 2025-07-01
Support 'recommended' fields on a model card schema


GCHQDeveloper36:
## Is your feature request related to a problem? Please describe.

Model card schemas define the fields an organisation would like the owners/developers to consider, but there isn't any way to enforce/require/suggest which fields are 'required' vs 'optional'. Therefore when a release is created, and a reviewer considers the card, there might be some key information missing that's needed to make an informed decision.

## Describe the solution you would like

Some sort of modification to the schema definitions to allow for annotating fields if they're "required"/"recommended" (wording TBD). I'd imagine that from an end-user perspective, we could highlight "recommended" fields during an edit session, and allow the list of (missing) fields to be seen in a simple list when a collaborator views a model card.

## Describe alternatives you have considered

I think jsonschema supports 'required' fields, however when a card is strictly validated, these would get flagged and defeat the point of the living, editable card. Extending the support for jsonschema, or ignoring 'required' fields in a validation response, might be needed.

GCHQDeveloper36:
That could definitely work too, although does that start to break the separation of model card schemas + core system code? Since the states are defined in system configuration, but schemas are considered 'separate' at the moment?

JR40159:
#3740 Adds the capability to recommend fields based on those fields being required for other model states. Would this adequately address this issue?

PR #3634, opened 2026-05-27
Backend changes for required fields for states

JR40159:


backend/src/routes/v2/schema/getSchema.ts
```diff
@@ -41,9 +44,12 @@ interface GetSchemaResponse {
 export const getSchema = [
   async (req: Request, res: Response<GetSchemaResponse>): Promise<void> => {
     req.audit = AuditInfo.ViewSchema
-    const { params } = parse(req, getSchemaSchema)
+    const {
+      query: { modelState },
```
GB907762, line 49: I'm wondering if this should be at route level. It may be worth doing some logic at the database level. Like a virtual or otherwise?

Because there's a transitive dependency on the metadata you're having to update different pieces of logic for each route it might be worth just managing this from the database client?

backend/src/services/schema.ts
```diff
@@ -38,9 +39,42 @@ export async function getSchemaById(schemaId: string) {
...
 }
 
+function enforceModelStateFields(schema: object, targetState: string) {
+  const validStates = config.ui.modelDetails.states
+  if (!validStates.includes(targetState)) {
+    throw BadReq('The value for modelState is not a valid', { validStates, modelState: targetState })
+  }
+  const jsonSchema = schema
+  traverse(jsonSchema, { allKeys: true }, (subschema, pointer, root, parentPointer, parentKeyword, parentSchema) => {
+    if (!subschema || typeof subschema !== 'object') {
```
GB907762, line 73: Be incredibly careful here when comparing `typeof` with an object it's truthy for things like `null`, instances of classes ect.. It shouldn't be too bad here but might want some further checks?

backend/src/services/schema.ts
```diff
@@ -38,9 +39,42 @@ export async function getSchemaById(schemaId: string) {
...
+
   return schema
 }
 
+function enforceModelStateFields(schema: object, targetState: string) {
+  const validStates = config.ui.modelDetails.states
+  if (!validStates.includes(targetState)) {
+    throw BadReq('The value for modelState is not a valid', { validStates, modelState: targetState })
+  }
+  const jsonSchema = schema
```
PE39806: I _think_ this will be a pointer rather than a copy of the object, so risks any mutations to `jsonSchema` (which is returned from `enforceModelStateFields` so any mutations to that too) affecting the original `schema` too. Instead, it would be safer to do
```suggestion
  const jsonSchema = structuredClone(schema)
```
which will be a new object, so doesn't risk any unintended mutations.

backend/src/services/schema.ts
```diff
@@ -38,9 +39,42 @@ export async function getSchemaById(schemaId: string) {
...
+  traverse(jsonSchema, { allKeys: true }, (subschema, pointer, root, parentPointer, parentKeyword, parentSchema) => {
+    if (!subschema || typeof subschema !== 'object') {
+      return
+    }
+
+    if (Array.isArray(subschema.state) && subschema.state.includes(targetState)) {
+      if (parentKeyword === 'properties' && parentSchema) {
+        const propertyName = pointer.split('/').pop()
```
PE39806: From https://datatracker.ietf.org/doc/html/rfc6901#section-3
> Because the characters '\~' (%x7E) and '/' (%x2F) have special meanings in JSON Pointer, '\~' needs to be encoded as '\~0' and '/' needs to be encoded as '\~1' when these characters appear in a reference token.
```suggestion
        const propertyName = pointer.replace(/~1/g, '/').replace(/~0/g, '~').split('/').pop()
```

backend/src/routes/v2/schema/getSchema.ts
```diff
@@ -12,6 +12,9 @@ export const getSchemaSchema = z.object({
   params: z.object({
     schemaId: z.string(),
   }),
+  query: z.object({
+    modelState: z.string().optional(),
```
PE39806: Prefer
```suggestion
    modelState: z.enum(config.ui.modelDetails.states).optional(),
```

backend/src/services/schema.ts
```diff
@@ -38,9 +39,42 @@ export async function getSchemaById(schemaId: string) {
     throw NotFound(`The requested schema was not found.`, { schemaId })
   }
 
+  if (modelState) {
+    const jsonSchema = enforceModelStateFields(schema.jsonSchema, modelState)
+    schema.jsonSchema = jsonSchema
```
PE39806: I don't particularly like mutating the Mongoose object in place like this. If a POJO is acceptable then maybe do
```suggestion
    return {
      ...schema.toObject(),
      jsonSchema
    }
```

backend/src/services/schema.ts
```diff
@@ -38,9 +39,42 @@ export async function getSchemaById(schemaId: string) {
...
   return schema
 }
 
+function enforceModelStateFields(schema: object, targetState: string) {
+  const validStates = config.ui.modelDetails.states
+  if (!validStates.includes(targetState)) {
+    throw BadReq('The value for modelState is not a valid', { validStates, modelState: targetState })
+  }
+  const jsonSchema = schema
+  traverse(jsonSchema, { allKeys: true }, (subschema, pointer, root, parentPointer, parentKeyword, parentSchema) => {
```
PE39806: I'd imagine that traversing large schemas is quite slow and, since they are immutable, a cache could be used here using the `schemaId` and `state` as the key.

PE39806 (changes requested):

backend/src/services/schema.ts
GB907762: This is a good shout. I had the same thought, hence I said maybe something at the database layer but either or.

JR40159:
@GCHQDeveloper36 would this resolve #2422 ?

backend/src/services/schema.ts
```diff
@@ -38,7 +52,42 @@ export async function getSchemaById(schemaId: string) {
...
+  if (!validStates.includes(targetState)) {
+    throw BadReq('The value for modelState is not a valid', { validStates, modelState: targetState })
+  }
+  const jsonSchema = structuredClone(schema)
+  traverse(jsonSchema, { allKeys: true }, (subschema, pointer, root, parentPointer, parentKeyword, parentSchema) => {
+    if (!subschema || typeof subschema !== 'object') {
+      return
+    }
+
+    if (Array.isArray(subschema.requiredByModelSates) && subschema.requiredByModelSates.includes(targetState)) {
```
IR96334: typo:
requiredByModelSates
-->
requiredByModelStates

IR96334:
As status adherence is optional, could we add a config value, e.g. modelCardStateAdherenceStrict: true ? And subsequently add a check for this value when getting schema, meaning if it isn't true and the status query is provided then we skip enforceModelStateFields ?

IR96334:
Also, we will need a check in: patchModel.ts > model.ts > func updateModel. Where if the adherence is strict and there is a model state change, we only allow it if all required model questions are answered.

backend/src/services/schema.ts
```diff
@@ -1,18 +1,21 @@
...
 import { UserInterface } from '../models/User.js'
 import { SchemaKind, SchemaKindKeys } from '../types/enums.js'
 import config from '../utils/config.js'
-import { Forbidden, NotFound } from '../utils/error.js'
+import { BadReq, Forbidden, NotFound } from '../utils/error.js'
 import { handleDuplicateKeys } from '../utils/mongo.js'
 import log from './log.js'
 import { addReviewsForNewRole } from './review.js'
 
+const schemaCache = new NodeCache()
```
PE39806, line 19: Please add a `maxKeys` to this.

backend/src/services/schema.ts
```diff
@@ -38,7 +52,42 @@ export async function getSchemaById(schemaId: string) {
     throw NotFound(`The requested schema was not found.`, { schemaId })
   }
 
-  return schema
+  const schemaObject = schema.toObject()
```
PE39806, line 56: If only using the object, prefer `lean()` for efficiency https://mongoosejs.com/docs/tutorials/lean.html
```suggestion
  const schema = await SchemaModel.findOne({
    id: schemaId,
  }).lean()

  if (!schema) {
    throw NotFound(`The requested schema was not found.`, { schemaId })
  }
```

PE39806 (changes requested):

backend/src/services/schema.ts
JR40159: `If the number of keys exceeds this limit, it will throw an error when trying to set more keys than the maximum` From the code-cache readme.
Adding setting that configuration would mean also having to handle the error, which I assume we have been doing elsewhere in the code-base where we are using that configuration....

backend/src/services/schema.ts
JR40159: I don't think this will work as we need to use the getter and setter on `jsonSchema`.

backend/src/services/schema.ts
JR40159: I'm also not sure the size of the cache is an issue for the schema cache

PE39806: Oh, I assumed that this capped the number of keys and just aged off the oldest one(s)... I'll be making a PR to address this shortly in other areas of the codebase.

PE39806 (dismissed):

backend/src/services/schema.ts
```diff
@@ -38,7 +53,42 @@ export async function getSchemaById(schemaId: string) {
...
+  return schemaObject
+}
+
+function enforceModelStateFields(schema: object, targetState: string) {
+  const validStates = config.ui.modelDetails.states
+  if (!validStates.includes(targetState)) {
+    throw BadReq('The value for modelState is not a valid', { validStates, modelState: targetState })
+  }
+  const jsonSchema = structuredClone(schema)
+  traverse(jsonSchema, { allKeys: true }, (subschema, pointer, root, parentPointer, parentKeyword, parentSchema) => {
```
IR96334: root and parentPointer are unused

backend/src/services/schema.ts
```diff
@@ -38,7 +53,42 @@ export async function getSchemaById(schemaId: string) {
...
+    schemaObject.jsonSchema = enforceModelStateFields(schemaObject.jsonSchema, modelState)
+  }
+  schemaCache.set(JSON.stringify({ schemaId, modelState }), schemaObject)
+  return schemaObject
+}
+
+function enforceModelStateFields(schema: object, targetState: string) {
+  const validStates = config.ui.modelDetails.states
+  if (!validStates.includes(targetState)) {
+    throw BadReq('The value for modelState is not a valid', { validStates, modelState: targetState })
```
JRB66955:
```suggestion
    throw BadReq('The value for modelState is not a valid model state', { validStates, modelState: targetState })
```

JRB66955 (dismissed):

IR96334 (approved):

ARADDCC002 (approved):

PR #3638, opened 2026-05-28
Bump nginx image version

PE39806:
Not sure why dependabot didn't pick up on this one. It's worth keeping an eye on this and potentially finding a solution in the future if this continues to not be handled by dependabot.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -47,7 +47,7 @@ services:
       - minioVolume:/bitnami/minio
 
   nginx:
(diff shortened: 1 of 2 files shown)


GB907762 (approved):

GB907762:
I think our infrastructure default upgrades have failed before, for stuff like the registry. Worth looking into?

ARADDCC013 (approved):

JRB66955 (approved):

PR #3640, opened 2026-05-29
fixed an issue where the metrics item widget was not parsing decimal …

ARADDCC002:
…points

### DIFF
--- a/frontend/utils/stringUtils.ts
+++ b/frontend/utils/stringUtils.ts
@@ -39,7 +39,7 @@ export const isValidPortNumber = (portNumber: string) => {
 }
 
 export const isValidNumber = (numberOrString: string) => {
-  return numberOrString.length > 0 && !isNaN(Number(numberOrString))
+  return !(isNaN(Number(numberOrString)) || isNaN(parseFloat(numberOrString)))
 }
 
 export const toKebabCase = (value: string): string => {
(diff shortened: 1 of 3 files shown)


PE39806 (dismissed):

JRB66955 (dismissed):

frontend/utils/stringUtils.ts
```diff
@@ -39,8 +39,7 @@ export const isValidPortNumber = (portNumber: string) => {
 }
 
 export const isValidNumber = (numberOrString: string) => {
-  const numericString = Number(numberOrString)
-  return numberOrString.length > 0 && !isNaN(numericString) && Number.isInteger(numericString)
+  return numberOrString.length > 0 && !isNaN(Number(numberOrString))
```
GB907762:
```suggestion
  return !(isNaN(numberOrString) || isNaN(parseFloat(numberOrString)))
```
This is probably quite petty but this clause is truthy for things like the following. Feel free to discard. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number
```js
'0xff'; // true (hexadecimal notation)
'0b11111111'; // true (binary notation)
'0.255e3'; // true (decimal exponential notation)
```

GB907762 (dismissed):

GB907762 (approved):

PR #3644, opened 2026-06-01
Fix static docs relative path resolution

PE39806:


GB907762 (approved):

GB907762 (changes requested):
What are your thoughts on adding accessibility to the navigation?

PE39806:
> What are your thoughts on adding accessibility to the navigation?

Accessibility isn't a primary "feature" of Bailo and is more of an administrative page, hence being moved to the footer in #3626.

GB907762 (approved):

JRB66955 (approved):

ARADDCC002 (approved):

PR #3687, opened 2026-06-03
Bump mongoose to 9.4.x and fix typing/build issues

PE39806:
From https://mongoosejs.com/docs/typescript.html
> To define a property of type `ObjectId`, you should use `Types.ObjectId` in the TypeScript document interface. You should use `'ObjectId'` or `Schema.Types.ObjectId` in your schema definition.

I've had to replace the `ReleaseSchema` `semver` string -> object and vice-versa with a new `semverString` virtual due to newly introduced typing issues.


ARADDCC002 (dismissed):

PE39806:
It looks like the `ReleaseModel` getter when the `semver` is a string doesn't behave how I would like. Presumably it is searching by `String` and not by `SemverObject`. I'll try and get a neater fix sorted soon.

JRB66955 (dismissed):

backend/src/models/Release.ts
```diff
@@ -114,6 +110,24 @@ ReleaseSchema.plugin(softDeletionPlugin)
...
+    filter.semver = semverStringToObject(filter.semver)
+  } else if (filter.semver !== null && typeof filter.semver === 'object' && '$in' in filter.semver) {
+    filter.semver.$in = (filter.semver.$in as Array<string | SemverObject>).map((s) =>
+      typeof s === 'string' ? semverStringToObject(s) : s,
+    )
+  }
+}
+
+// Querying by semver can be performed either with a string or a semver object this resolves the type to an object for database lookup
+ReleaseSchema.pre(['find', 'findOne', 'findOneAndUpdate'], function () {
```
PE39806: From https://mongoosejs.com/docs/queries.html we _might_ want to cover a few more queries e.g.
```suggestion
ReleaseSchema.pre(['find', 'findOne', 'findOneAndUpdate', 'findOneAndDelete', 'findOneAndReplace', 'updateOne', 'updateMany', 'replaceOne', 'deleteOne', 'deleteMany', 'countDocuments'], function () {
```
Equally, I'm happy to leave this as is if we're confident we cover the existing use cases.

GB907762: Good suggestion

PE39806:
I am happy with these changes - will need others to approve as this is a PR that I made!

GB907762 (approved):

JRB66955 (approved):

PR #3688, opened 2026-06-03
Add lifecycle emails

GB907762:
Emails are sent per entry to each model owner:
- Sends an email for specific time periods before a lifecycle review.
- Sends reminder emails every period of time after a lifecycle review has expired.

### DIFF
--- a/backend/config/default.cjs
+++ b/backend/config/default.cjs
@@ -69,7 +69,7 @@ module.exports = {
 
   smtp: {
     // Enable / disable all email sending
-    enabled: true,
+    enabled: false,
     transporter: 'smtp',
 
     // Connection information for an SMTP server.  Settings are passed directly to 'node-mailer', see reference for options:
--- a/backend/config/default.cjs
+++ b/backend/config/default.cjs
@@ -69,7 +69,7 @@ module.exports = {
 
   smtp: {
     // Enable / disable all email sending
-    enabled: false,
+    enabled: true,
     transporter: 'smtp',
 
     // Connection information for an SMTP server.  Settings are passed directly to 'node-mailer', see reference for options:
--- a/backend/config/default.cjs
+++ b/backend/config/default.cjs
@@ -86,7 +86,7 @@ module.exports = {
 
     // Refer to: https://github.com/agenda/human-interval#uses
     lifecycle: {
-      preReminderIntervals: ['1 hour', '2 weeks', '10 weeks'],
+      preReminderIntervals: ['1 day', '2 weeks', '10 weeks'],
       postReminderInterval: '1 day',
     },
 
--- a/backend/config/default.cjs
+++ b/backend/config/default.cjs
@@ -84,6 +84,12 @@ module.exports = {
       },
     },
 
+    // Refer to: https://github.com/agenda/human-interval#uses
+    lifecycle: {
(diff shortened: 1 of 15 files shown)


backend/config/default.cjs
```diff
@@ -84,6 +84,12 @@ module.exports = {
       },
     },
 
+    // Refer to: https://github.com/agenda/human-interval#uses
+    lifecycle: {
+      preReminderIntervals: ['1 hour', '2 weeks', '10 weeks'],
```
ARADDCC002: Up for debate, but I think changing the first value to `1 day` might be better. Giving people one hour's final notice might seem a bit too urgent.

backend/src/services/smtp/smtp.ts
```diff
@@ -174,6 +175,30 @@ export async function notifyReviewResponseForRelease(reviewResponse: ResponseInt
   await dispatchEmail(toEntity('user', release.createdBy), await emailContent)
 }
 
+export async function notifyLifeCycleReview(modelId: string, reviewId: string, dueIn?: string) {
+  if (!config.smtp.enabled) {
+    log.info('Not sending email due to SMTP disabled')
```
ARADDCC002, line 180: Personally, I think `Notification email not sent as SMTP is disabled` reads better.

backend/src/services/smtp/smtp.ts
```diff
@@ -174,6 +175,30 @@ export async function notifyReviewResponseForRelease(reviewResponse: ResponseInt
...
+  }
+
+  const model = await getModelByIdNoAuth(modelId)
+  const emailContent = buildEmail(
+    dueIn
+      ? `A lifecycle review for ${model.name} is due in ${dueIn}`
+      : `A lifecycle review for ${model.name} has past it's due date`,
+    [],
+    [
+      { name: `See ${toTitleCase(model.kind, '-')}`, url: `${appBaseUrl}/${model.kind}/${modelId}` },
```
ARADDCC002: Unfortunately, regular models, untrusted models and mirrored models all use the same path for their model page of "<baseurl>/model/<model_id>`

ARADDCC002: This is probably going to be a problem with reviewing in the app too, so probably not exclusive to this PR

backend/src/utils/config.ts
```diff
@@ -87,6 +87,11 @@ export interface Config {
       }
     }
 
+    lifecycle: {
+      preReminderIntervals: string[]
```
ARADDCC002, line 91: We will probably need to add this to the rest of the config in helm.

ARADDCC002 (changes requested):

backend/src/services/smtp/smtp.ts
GB907762: Yeah. I think this whole part of the code needs refactoring. This condition should be really checked once. I'll have a look. But I might move this out of scope considering it's in about 10 places

backend/src/utils/config.ts
GB907762: Good catch forgot about that.

ARADDCC002 (approved):

JRB66955 (approved):

JR40159 (approved):

PR #3701, opened 2026-06-04
Added previous model card review details to overview panel

ARADDCC002:


frontend/src/entry/overview/ReviewDateDialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <Dialog open={open} onClose={onClose} fullWidth slotProps={{ transition: Transition }}>
+      <DialogTitle color='primary'>Set a review date for this model card</DialogTitle>
+      <DialogContent sx={{ m: 'auto' }}>
+        <DatePicker
+          value={reviewDate}
+          sx={{ backgroundColor: 'unset', borderRadius: 1 }}
+          onChange={(newValue) => {
+            setReviewDate(newValue)
+          }}
+          minDate={dayjs(new Date())}
```
GB907762: I think the next review should be a day after today. This seems to round down to midnight. Which causes a `400`

backend/src/services/v3/review.ts
```diff
@@ -111,6 +111,13 @@ export async function createLifecycleReview(
   modelId: string,
   dueDate: Date,
 ): Promise<ReviewInterface> {
+  if (!dueDate || dueDate.getTime() === 0) {
+    throw BadReq('Lifecycle review responses should have a valid due date.')
+  }
+  if (dueDate.getTime() <= Date.now()) {
+    throw BadReq('Due date of next review cannot be in the past.')
+  }
```
GB907762: Could you reuse the code in `validateLifecycleReview`

`dueDate.getTime() === 0` Identifies any time before the UNIX epoch which are valid times.

Additionally you shouldn't need a condition here because you've already validated this with zod?

```suggestion
  if (dueDate.getTime() <= Date.now()) {
    throw BadReq('Due date of next review cannot be in the past.')
  }
```

If you were really fancy you could validate the date on the route. Something like this:

```typescript
dueDate: z.coerce.date({
          required_error: 'Please provide a next review due date when submitting this review.',
}).refine((date) => date > new Date(), {
  message: "Due date of next review cannot be in the past.",
});
```

GB907762 (commented):
I've noticed a small bug while testing.

- Approve a review for a model card
- Press the back button in your browser to go back to the closed review
- `Undo review` and `Approve` review appears but are invalid options because it's closed

If there are two model owners and one approves before the other, because of an email they both get they can access this screen.

What I would expect is as the second owner could see what comment the previous one has left but not the options to change it?

ARADDCC002:
> I've noticed a small bug while testing.
> 
>     * Approve a review for a model card
> 
>     * Press the back button in your browser to go back to the closed review
> 
>     * `Undo review` and `Approve` review appears but are invalid options because it's closed
> 
> 
> If there are two model owners and one approves before the other, because of an email they both get they can access this screen.
> 
> What I would expect is as the second owner could see what comment the previous one has left but not the options to change it?

So this review will move to `Archived` for all model owners. I'll see if I can change how the page displays, though, so it's clearer. There certainly shouldn't be an `undo review` button.

backend/src/services/v3/review.ts
ARADDCC002: Good idea to put it in the schema. Personally, as much validation should be done via `Zod` just so that it frees up the service layer. In `v2` we have a lot of large functions because the first half of them are just doing validation.

ARADDCC002:
Changed the review of the lifecycle page so that it displays if it has already been reviewed, and who by.

JRB66955 (dismissed):

frontend/src/entry/EntryOverviewDetails.tsx
```diff
@@ -194,6 +200,20 @@ export default function EntryOverviewDetails({ entry }: OrganisationAndStateDeta
                 </Button>
               )}
             </Stack>
+            {archivedReviews.length > 0 && (
+              <>
+                <Typography fontWeight='bold' color='primary'>
+                  Last reviewed:
+                </Typography>
+                <Typography>
+                  {responses[0] ? formatDateStringAsDayMonthAndYear(responses[0].createdAt.toString()) : 'Invalid date'}
```
JR40159: Does this rely on the responses being in date order? If so, I'm not sure how safe that is. Perhaps you could add a query param to get responses to only get the latest response for something?

JR40159 (dismissed):

frontend/src/entry/EntryOverviewDetails.tsx
ARADDCC002: For lifecycle reviews there should only ever be one response, you can't "undo" them like you can for releases/access requests, and if a second model owner tries to review it, they will get an error. 

I could do this check anyway, though.

frontend/src/entry/EntryOverviewDetails.tsx
```diff
@@ -194,6 +200,20 @@ export default function EntryOverviewDetails({ entry }: OrganisationAndStateDeta
...
+                </Typography>
+                <Typography fontWeight='bold' color='primary'>
+                  Last reviewed by:
+                </Typography>
+                {responses[0] ? <UserDisplay dn={responses[0].entity} /> : <Typography>Invalid user</Typography>}
```
IR96334: The user dn is hoverable, but doesn't _look_ hoverable. To do this, could add showIcon prop to make more consistent with what is known to be hoverable, e.g. entry access. Any better solutions are welcome.

backend/src/routes/v3/review/postReviewResponse.ts
```diff
@@ -28,9 +28,13 @@ export const postReviewResponseSchema = z.object({
...
-        dueDate: z.coerce.date({
-          required_error: 'Please provide a next review due date when submitting this review.',
-        }),
+        dueDate: z.coerce
+          .date({
+            required_error: 'Please provide a next review due date when submitting this review.',
+          })
+          .refine((date) => date > new Date(), {
+            message: 'Due date of next review cannot be in the past.',
+          }),
```
GB907762, line 37: Could you update '/api/v3/review/{modelId}' with the same validation? Make it more consistent and simplify the code a bit.

frontend/pages/model/[modelId]/lifecycle/[reviewId]/review.tsx
```diff
@@ -130,15 +137,30 @@ export default function LifecycleReview() {
...
+                  hideRequestChangesButton
+                />
+                <MessageAlert message={errorMessage} severity='error' />
+              </>
+            )}
+            {responses.length > 0 && (
+              <>
+                <Stack direction='row' spacing={1}>
+                  <Typography>This lifecycle review has already been approved by</Typography>
+                  <UserDisplay dn={responses[0].entity} />
```
Comment (GB907762) [inline frontend/pages/model/[modelId]/lifecycle/[reviewId]/review.tsx]: It might be worth displaying the owner's comment as well, in in a similar way we do access requests with review responses. 

You may write some notes that you would like other owners to see?

GB907762 (changes requested):

frontend/pages/model/[modelId]/lifecycle/[reviewId]/review.tsx
```diff
@@ -130,15 +137,30 @@ export default function LifecycleReview() {
...
-            <ReviewWithComment
-              onSubmit={handleSubmit}
-              reviews={reviews}
-              loading={isReviewButtonLoading}
-              modelId={modelId as string}
-              includeDueDate
-              hideRequestChangesButton
-            />
-            <MessageAlert message={errorMessage} severity='error' />
+            {responses.length === 0 && (
```
Comment (IR96334) [inline frontend/pages/model/[modelId]/lifecycle/[reviewId]/review.tsx]: If this first clause (responses.length === 0) is untrue, the next clause (responses.length > 0) will always be true. To make this logical flow clear, we can instead use the `responses.length === 0 ? _blah_ : _blah_` format

frontend/src/entry/EntryOverviewDetails.tsx
```diff
@@ -161,14 +165,9 @@ export default function EntryOverviewDetails({ entry }: OrganisationAndStateDeta
               Next review due:
             </Typography>
             {updateReviewDatePermission && reviews.length === 0 && (
```
GB907762:
```suggestion
            {updateReviewDatePermission.hasPermission && reviews.length === 0 && (
```

frontend/src/entry/EntryOverviewDetails.tsx
GB907762:
```suggestion
              {(!updateReviewDatePermission.hasPermission || reviews.length > 0) && (
```

GB907762:
```suggestion
              {updateReviewDatePermission.hasPermission && reviews[0] && (
```

GB907762 (changes requested):
Might need to revisit how this context works at some point. And make a custom hook instead of repeating this memo?

GB907762 (dismissed):

GB907762 (approved):

PE39806 (approved):

PR #3710, opened 2026-06-09
Various async await issues

PE39806:
- Handle webhooks more safely
- Handle bucket creation safely
- Clean up some incorrectly `async` functions
- Add a few missing `await`s

### DIFF
--- a/backend/src/index.ts
+++ b/backend/src/index.ts
@@ -17,8 +17,9 @@ shelljs.exec('update-ca-certificates', { fatal: false, async: false })
 // that nobody should notice unless they want to upload an image
 // within the first few milliseconds of the _first_ time it's run
 if (config.s3.automaticallyCreateBuckets) {
-  ensureBucketExists(config.s3.buckets.uploads)
-  ensureBucketExists(config.s3.buckets.registry)
+  for (const bucket of [config.s3.buckets.uploads, config.s3.buckets.registry]) {
+    ensureBucketExists(bucket).catch((err) => log.error({ err, bucket }, 'Unable to automatically create bucket.'))
+  }
 }
 
 // connect to Mongo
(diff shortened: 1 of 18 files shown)


backend/src/index.ts
```diff
@@ -17,8 +17,9 @@ shelljs.exec('update-ca-certificates', { fatal: false, async: false })
 // that nobody should notice unless they want to upload an image
 // within the first few milliseconds of the _first_ time it's run
 if (config.s3.automaticallyCreateBuckets) {
-  ensureBucketExists(config.s3.buckets.uploads)
-  ensureBucketExists(config.s3.buckets.registry)
+  for (const bucket of [config.s3.buckets.uploads, config.s3.buckets.registry]) {
+    ensureBucketExists(bucket).catch((err) => log.error({ err, bucket }, 'Unable to automatically create bucket.'))
+  }
```
GB907762: This happens on start up I believe? Do we really want this to fail silently if someone is configuring Bailo and the backend can't communicate with the object store?

GB907762 (changes requested):

backend/src/index.ts
PE39806: I think that this is what the existing behaviour is, although the unhandled error _could_ potentially just crash the pod. The only alternative is to `await` the bucket creation which would also be a valid change. Which approach do you think is best?

GB907762: I think the current behaviour is better. We attach the logger to the request for the reason of forwarding this to the user. But for set up I think it's best if the backend crashes is there's something unexpected with start up?

I wouldn't mind awaiting this? There's a comment above suggesting there's some unexpected behaviour because this runs asynchronously?

JRB66955: It does fail silently already in the existing code/implementation - although I think it would be more useful to await the bucket creation in this instance.

JRB66955 (approved):

GB907762 (approved):

PR #3711, opened 2026-06-09
Scheduled review notifications for lifecycle reviews deleted when mod…

ARADDCC002:
…el deleted


backend/src/services/model.ts
```diff
@@ -269,6 +270,8 @@ export async function removeModel(user: UserInterface, modelId: string, kind?: E
           ),
         ),
       ]),
+    // The Mongo implementations for this data is handled using a third-party library so we don't need to pass in the session.
+    () => cancelLifecycleJobsForModel(model.id),
```
GB907762, line 274: I think in future we can generalise this for all jobs on a model. But looks good for now.

GB907762 (approved):

JRB66955 (approved):

PR #3713, opened 2026-06-10
Added review responses to model cards

ARADDCC002:


backend/src/routes/v3/response/postComment.ts
```diff
@@ -0,0 +1,64 @@
...
+import { ReviewKind } from '../../../types/enums.js'
+import { getEnumValues } from '../../../utils/enum.js'
+import { parse } from '../../../utils/validate.js'
+
+export const postCommentSchema = z.object({
+  query: z.object({
+    modelId: z.string().openapi({ example: 'test-model-123' }),
+    identifier: z.string().optional().openapi({ example: '1.0.0' }),
+    kind: z.enum(getEnumValues(ReviewKind)).openapi({ example: ReviewKind.Release }),
+  }),
```
PE39806: `identifier` meaning changes too much depending on `kind` for me to be happy with this. I much prefer splitting the zod schema to be more explicit e.g.
```suggestion
  query: z.discriminatedUnion('kind', [
    z.object({
      kind: z.literal(ReviewKind.Release).openapi({ example: ReviewKind.Release }),
      modelId: z.string().openapi({ example: 'test-model-123' }),
      identifier: z
        .string()
        .openapi({
          example: '1.0.0',
          description: 'Semantic version identifying the release.',
        }),
    }),
    z.object({
      kind: z.literal(ReviewKind.Access).openapi({ example: ReviewKind.Access }),
      modelId: z.string().openapi({ example: 'test-model-123' }),
      identifier: z
        .string()
        .openapi({
          example: 'access-req-456',
          description: 'Access request identifier.',
        }),
    }),
    z.object({
      kind: z.literal(ReviewKind.Lifecycle).openapi({ example: ReviewKind.Lifecycle }),
      modelId: z.string().openapi({ example: 'test-model-123' }),
      identifier: z
        .undefined()
        .openapi({
          description: 'Not used for lifecycle comments.',
        }),
    }),
  ]),
```

PE39806 (dismissed):
The "Reply" option under a review in the model's review history currently doesn't do anything.

ARADDCC002:
> The "Reply" option under a review in the model's review history currently doesn't do anything.

Thanks for spotting this. We decided against allowing regular comments on models for now, so this option should have been hidden.

backend/src/routes/v3/routes.ts
```diff
@@ -24,6 +25,8 @@ router.post('/review/:reviewId/response', ...postReviewResponse)
 router.post('/review/:modelId', ...postReview)
 router.get('/review/:reviewId/responses/latest', ...getLatestResponse)
 
+router.post('/response/comment', ...postComment)
```
GB907762: I can't decide whether a more restful option might be to split these. As in have.
```typescript
router.post('/model/:modelId/access-request/:accessRequestId/comment', ...postAccessRequestComment)
router.post('/model/:modelId/release/:semver/comment', ...postReleaseComment)
router.post('/model/:modelId/comment', ...postModelComment)
```

I think this is the more restful approach and will avoid the wrangling to try and resolve the parent document. Which could cause complexity issues and speed issues down the line?

backend/src/services/v3/response.ts
```diff
@@ -67,3 +73,83 @@ export async function getLatestResponseForReview(reviewId: string) {
...
+
+async function resolveCommentParent(user: UserInterface, modelId: string, kind: ReviewKindKeys, identifier?: string) {
+  const model = await getModelById(user, modelId)
+  if (!model) {
+    throw NotFound('The requested model was not found', { modelId })
+  }
+  const auth = await authorisation.model(user, model, ModelAction.View)
+  if (!auth.success) {
+    throw Forbidden(auth.info, { userDn: user.dn, modelId: model.id })
+  }
```
GB907762: Already has these checks

```suggestion
  const model = await getModelById(user, modelId)
```

backend/src/services/v3/response.ts
```diff
@@ -67,3 +73,83 @@ export async function getLatestResponseForReview(reviewId: string) {
...
+
+  switch (kind) {
+    case ReviewKind.Release: {
+      if (!identifier) {
+        throw BadReq('A valid semver must be provided for release comments.')
+      }
+      const semverObj = semverStringToObject(identifier)
+      const release = await ReleaseModel.findOne({ modelId: model.id, semver: semverObj })
```
GB907762: Handled by database already.

```suggestion
      const release = await ReleaseModel.findOne({ modelId: model.id, semver: identifier })
```

backend/src/services/v3/response.ts
```diff
@@ -67,3 +73,83 @@ export async function getLatestResponseForReview(reviewId: string) {
...
+      return model._id
+    }
+    default: {
+      const exhaustiveCheck: never = kind
+      throw InternalError(`Unsupported review kind: ${exhaustiveCheck}`)
+    }
+  }
+}
+
+export async function newComment(
```
GB907762:
```suggestion
export async function createResponse(
```
Might be better named? Inline with all other services?

backend/src/services/v3/response.ts
```diff
@@ -67,3 +73,83 @@ export async function getLatestResponseForReview(reviewId: string) {
...
+  identifier?: string,
+) {
+  const parentId = await resolveCommentParent(user, modelId, kind, identifier)
+
+  // Store the response
+  const commentResponse = new ResponseModel({
+    entity: toEntity('user', user.dn),
+    kind: ResponseKind.Comment,
+    comment,
+    createdAt: new Date().toISOString(),
```
GB907762: I'm a little confused why we have these dotted around the codebase.
We use mongoose timestamps for every schema, that are synthesised in the database layer. There's no need to add this.

https://mongoosejs.com/docs/timestamps.html#timestamps

```suggestion
```

frontend/src/entry/overview/ReviewHistoryDialog.tsx
```diff
@@ -0,0 +1,39 @@
...
+  entry: EntryInterface
+  mutateEntry: () => void
+}
+
+export default function ReviewHistoryDialog({ entry, mutateEntry, open, onClose }: ReviewHistoryDialogProps) {
+  const handleOnClose = useCallback(() => {
+    onClose()
+  }, [onClose])
+
+  return (
```
GB907762, line 19: I'm not so sure about this `Dialog`. Would this be better on a separate page?

I thought this might be more akin to the AccessRequest component in `frontend/pages/model/[modelId]/access-request/[accessRequestId].tsx` Whether this can be reused?

frontend/src/entry/LastReviewOverviewDetails.tsx
```diff
@@ -32,6 +39,13 @@ export default function LastReviewOverviewDetails({ reviewId }: LastReviewOvervi
         Last reviewed by:
       </Typography>
       {response ? <UserDisplay dn={response.entity} showIcon /> : <Typography>Invalid user</Typography>}
+      <Button onClick={() => setReviewHistoryOpen(true)}>Review history</Button>
```
GB907762: Slight change of wording?
```suggestion
      <Button onClick={() => setReviewHistoryOpen(true)}>View Life Cycle History</Button>
```
We have option to see revisions for each model card, through an actions button.

I don't know whether adding those comments into there might give greater visibility on what people have reviewed and what changes have been made?

GB907762 (changes requested):

backend/src/routes/v3/routes.ts
ARADDCC002: I think now that we're not having model card comments for the time being I might revert the v3 response end point and go back to how we were doing it before.

backend/src/services/v3/response.ts
ARADDCC002: N/A

backend/src/services/v3/response.ts
ARADDCC002: N/A

backend/src/services/v3/response.ts
ARADDCC002: N/A

backend/src/services/v3/response.ts
ARADDCC002: N/A

frontend/src/entry/overview/ReviewHistoryDialog.tsx
ARADDCC002: I'm not convinced this needs to be a new page as it's just a list of review responses. I think it can be moved in the future if we decide to add more functionality, though.

GB907762 (approved):

ON93754 (approved):

PR #3714, opened 2026-06-11
Add announcement printout to python package

PE39806:
Re-use the existing `announcements` config from `GET /api/v2/config/ui` in the Python client to allow displaying important messages to non-UI users.

Announcements follow the frontend logic and are only shown when all of the following is met:
- `enabled` is `true`
- `text` is set
- `startTimestamp` is either unset or is a past date

Announcements are only shown on the first connection to the specific Bailo `Client`, and may then be redisplayed once every 24 hours. Announcements are printed to stdout via its own logger `bailo.announcements`, which allows for individual logger configuration/suppression. Announcements may be completely disabled using the `Client(announcements=False)` kwarg for headless Python clients.

### DIFF
--- a/lib/python/src/bailo/core/client.py
+++ b/lib/python/src/bailo/core/client.py
@@ -7,9 +7,18 @@ import time
 from io import BytesIO
 from typing import Any
 
+try:
+    from backports.datetime_fromisoformat import MonkeyPatch  # pyright: ignore[reportMissingImports]
+except ImportError:
+    MonkeyPatch = None
+
 from bailo.core.agent import Agent, TokenAgent
 from bailo.core.enums import CollaboratorEntry, EntryKind, ModelVisibility, SchemaKind
 from bailo.core.utils import filter_none, normalise_query_params
+
+if MonkeyPatch is not None:
+    # backport `datetime.fromisoformat` 3.11 changes https://pypi.org/project/backports-datetime-fromisoformat/
+    MonkeyPatch.patch_fromisoformat()
 
 logger = logging.getLogger(__name__)
 # Give users a clean way to suppress/reformat announcements without affecting debug/error logs
@@ -74,14 +83,12 @@ class Client:
                     return
 
                 start_timestamp = announcement.get("startTimestamp", "")
-                # Python 3.10 support - see https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat for origin
-                start_timestamp = start_timestamp.replace("Z", "+00:00")
                 now = datetime.datetime.now(datetime.timezone.utc)
                 if not start_timestamp == "" and now < datetime.datetime.fromisoformat(start_timestamp):
                     return
 
                 for line in text.splitlines():
-                    announcement_logger.info("remote: %s", line)
+                    announcement_logger.warning("remote: %s", line)
             except Exception:  # pylint: disable=broad-exception-caught
                 logger.debug("Failed to check for server announcements", exc_info=True)
 
--- a/lib/python/src/bailo/core/client.py
+++ b/lib/python/src/bailo/core/client.py
@@ -72,6 +72,8 @@ class Client:
                 return
 
             start_timestamp = announcement.get("startTimestamp", "")
+            # Python 3.10 support - see https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat for origin
+            start_timestamp = start_timestamp.replace("Z", "+00:00")
             now = datetime.datetime.now(datetime.timezone.utc)
             if not start_timestamp == "" and now < datetime.datetime.fromisoformat(start_timestamp):
                 return
--- a/lib/python/src/bailo/core/client.py
+++ b/lib/python/src/bailo/core/client.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import datetime
 import logging
-import sys
+import threading
 import time
 from io import BytesIO
 from typing import Any
@@ -12,6 +12,8 @@ from bailo.core.enums import CollaboratorEntry, EntryKind, ModelVisibility, Sche
 from bailo.core.utils import filter_none, normalise_query_params
 
 logger = logging.getLogger(__name__)
+# Give users a clean way to suppress/reformat announcements without affecting debug/error logs
+announcement_logger = logging.getLogger("bailo.announcements")
 
 
 class Client:
@@ -34,12 +36,12 @@ class Client:
         self._agent = agent or Agent()
         self._announcements_enabled = announcements
         self._last_announcement_check: float | None = None
-        self._checking_announcement = False
+        self._announcement_lock = threading.Lock()
 
     @property
     def agent(self) -> Agent:
         """Return the HTTP agent, checking for announcements periodically."""
-        if self._announcements_enabled and not self._checking_announcement:
+        if self._announcements_enabled:
             self._maybe_check_announcement()
         return self._agent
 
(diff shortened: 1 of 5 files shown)


lib/python/src/bailo/core/client.py
```diff
@@ -15,14 +21,78 @@ class Client:
...
+            self._maybe_check_announcement()
+        return self._agent
+
+    @agent.setter
+    def agent(self, value: Agent) -> None:
+        self._agent = value
+
+    def _maybe_check_announcement(self) -> None:
+        """Check for announcements if enough time has elapsed since the last check."""
+        now = time.monotonic()
```
GB907762: https://docs.python.org/3/library/time.html#time.monotonic

Not sure how I feel about using the monotonic clock especially on an elapsed duration of 1 day?

lib/python/src/bailo/core/client.py
```diff
@@ -15,14 +21,78 @@ class Client:
...
+        if (
+            self._last_announcement_check is None
+            or (now - self._last_announcement_check) >= self._ANNOUNCEMENT_CHECK_INTERVAL
+        ):
+            self._check_announcement()
+            self._last_announcement_check = now
```
GB907762, line 69: I don't think this is going to persist between processes. May be funky.

lib/python/src/bailo/core/client.py
```diff
@@ -15,14 +21,78 @@ class Client:
...
+
+            start_timestamp = announcement.get("startTimestamp", "")
+            # Python 3.10 support - see https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat for origin
+            start_timestamp = start_timestamp.replace("Z", "+00:00")
+            now = datetime.datetime.now(datetime.timezone.utc)
+            if not start_timestamp == "" and now < datetime.datetime.fromisoformat(start_timestamp):
+                return
+
+            for line in text.splitlines():
+                print(f"remote: {line}", file=sys.stderr)
```
GB907762: Hmm. Do we think we should log this to `stderr`. Often announcements could be non fatal.

Would it be better using the existing logging functionality. Such as a warning or something. 

For automated jobs this might be something you don't want to see?

GB907762 (changes requested):

lib/python/src/bailo/core/client.py
PE39806: I've followed the pattern that `git` uses in writing to `stderr`. I think this is the right approach as it avoids writing to `stdout` so the remote message does not get caught by things like pipes.

PE39806: After further discussion, I have switched to using a unique `logger` which can be more easily suppressed.

lib/python/src/bailo/core/client.py
PE39806: Yes, this will be per client.

lib/python/src/bailo/core/client.py
```diff
@@ -15,14 +23,76 @@ class Client:
...
+            or (now - self._last_announcement_check) >= self._ANNOUNCEMENT_CHECK_INTERVAL
+        ):
+            self._check_announcement()
+            self._last_announcement_check = now
+
+    def _check_announcement(self) -> None:
+        """Fetch the server announcement and print it to stderr if enabled and current."""
+        with self._announcement_lock:
+            try:
+                response = self._agent.get(f"{self.url}/v2/config/ui").json()
```
ON93754, line 75: Why not call get_ui_config() here instead of repeating the request logic?

PE39806: `get_ui_config` will trigger the `agent` getter which would be infinitely recursive.

lib/python/src/bailo/core/client.py
```diff
@@ -15,14 +23,76 @@ class Client:
     :param agent: An agent object to handle requests
     """
 
-    def __init__(self, url: str, agent: Agent | None = None):
+    _ANNOUNCEMENT_CHECK_INTERVAL = 60 * 60 * 24  # 24 hours
+
+    def __init__(self, url: str, agent: Agent | None = None, *, announcements: bool = True):
```
GB907762, line 37:
```suggestion
    def __init__(self, url: str, agent: Agent | None = None, announcements: bool = True):
```
Not sure about this wildcard?

GB907762 (changes requested):

lib/python/src/bailo/core/client.py
PE39806: The wildcard makes it so that any subsequent kwargs (in this case `announcements`) must be passed as a [keyword-only argument](https://peps.python.org/pep-3102/). I have chosen to do this so that `announcements` must be explicitly specified which reduces ambiguity for an optional config-only flag.

lib/python/src/bailo/core/client.py
```diff
@@ -1,12 +1,20 @@
...
 from io import BytesIO
 from typing import Any
 
 from bailo.core.agent import Agent, TokenAgent
 from bailo.core.enums import CollaboratorEntry, EntryKind, ModelVisibility, SchemaKind
 from bailo.core.utils import filter_none, normalise_query_params
 
+logger = logging.getLogger(__name__)
+# Give users a clean way to suppress/reformat announcements without affecting debug/error logs
+announcement_logger = logging.getLogger("bailo.announcements")
```
GB907762, line 25: The default level is a `WARNING` I assume this is intentional - won't appear in normal use only with `logging.getLogger("bailo.announcements").setLevel(logging.INFO)`

lib/python/src/bailo/core/client.py
```diff
@@ -15,14 +23,76 @@ class Client:
...
+                if not announcement.get("enabled", False):
+                    return
+
+                text = announcement.get("text", "")
+                if not text:
+                    return
+
+                start_timestamp = announcement.get("startTimestamp", "")
+                # Python 3.10 support - see https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat for origin
+                start_timestamp = start_timestamp.replace("Z", "+00:00")
```
GB907762: I'm not so sure..

```
 uiConfig.announcement.enabled &&
          (!dismissedTimestamp || new Date(dismissedTimestamp) < new Date(uiConfig.announcement.startTimestamp))
``` 
This is the snippet that decides whether an announcement is shown.

The key being a js `Date`. It has a lot of edge cases. For example this could be a UNIX timestamp, a random string ect..

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#several_ways_to_create_a_date_object

GB907762 (changes requested):

lib/python/src/bailo/core/client.py
PE39806: Ah, no, I meant to make the new logger use `warning`...

lib/python/src/bailo/core/client.py
PE39806: I've found https://pypi.org/project/backports-datetime-fromisoformat/ which backports the Python >=3.11 `datetime.fromisoformat` change to have better handling for different date string formats.

GB907762: Hmm. Good find but I'm not sure this covers values for things such as `startTimestamp: '0'`.

PE39806:
Closing for now - we will re-investigate how to do this in a way in the future that doesn't risk impacting automated pipelines like this approach does.

PR #3716, opened 2026-06-12
Enforce model states and clean up user experience of states and organisations

GB907762:
This may be a breaking change for clients.

Model state have been updated so that it can only be changed with a model card.


frontend/src/common/EntrySelect.tsx
```diff
@@ -0,0 +1,100 @@
...
+      if (field === 'state') {
+        router.replace({
+          query: { ...router.query, requiredState: event.target.value },
+        })
+      }
+      setErrorMessage(await getErrorMessage(response))
+    } else {
+      if (field === 'state') {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const { requiredState, ...queryWithoutFiltered } = router.query
```
ARADDCC002: `_requiredState`, should remove the need for this linting disable comment

frontend/src/common/EntrySelect.tsx
```diff
@@ -0,0 +1,100 @@
...
+            {options.map((option: string) => (
+              <MenuItem key={option} value={option}>
+                {option}
+              </MenuItem>
+            ))}
+            <MenuItem value=''>Unset</MenuItem>
+          </Select>
+        </Stack>
+      ) : (
+        <Stack direction='row' alignItems='center'>
```
ARADDCC002: Ahead of the MUI v9 upgrade, can we move `alignItems` inside `sx`

frontend/src/common/EntrySelect.tsx
```diff
@@ -0,0 +1,100 @@
...
+        })
+      }
+      mutate()
+    }
+    setIsEdit(false)
+  }
+
+  return (
+    <>
+      <Typography id={`${label.toLowerCase()}-label`} sx={{ fontWeight: 'bold', color: theme.palette.primary.main }}>
```
ARADDCC002: You can add the prop `color='primary'` directly to `<Typography />` and then you don't need to import `theme`.

frontend/src/common/EntrySelect.tsx
```diff
@@ -0,0 +1,100 @@
...
+        {`${label}:`}
+      </Typography>
+      {isEdit ? (
+        <Stack direction='row' alignItems='center'>
```
ARADDCC002: Same as before, `alignItems` can be put inside `sx`.

frontend/src/common/EntrySelect.tsx
```diff
@@ -0,0 +1,100 @@
...
+            id={label.toLowerCase()}
+            value={value ?? ''}
+            onChange={handleSelectOption}
+          >
+            {options.map((option: string) => (
+              <MenuItem key={option} value={option}>
+                {option}
+              </MenuItem>
+            ))}
+            <MenuItem value=''>Unset</MenuItem>
```
ARADDCC002: Wrap the text inside `<em />` to italicise it.

frontend/src/common/EntrySelect.tsx
```diff
@@ -0,0 +1,100 @@
...
+          </Select>
+        </Stack>
+      ) : (
+        <Stack direction='row' alignItems='center'>
+          <Typography>{value ? value : 'Unset'}</Typography>
```
ARADDCC002: `Unset` should be in italics.

ARADDCC002 (dismissed):

frontend/src/common/EntrySelect.tsx
GB907762: So this pattern removes `requiredState` from the query.

Idea being we can use this on an error to filter questions that are required to the user.  

So the underline prefix wont remove this property from the query correctly. But this is a good spot!

backend/src/services/model.ts
```diff
@@ -612,6 +613,16 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
...
 
+  if (modelDiff.state && model.card) {
+    const schema = await getSchemaById(model.card.schemaId, modelDiff.state)
+    const { valid } = new Validator().validate(model.card.metadata, schema.jsonSchema, {
+      required: true,
+    })
+    if (!valid) {
+      throw BadReq(`Please fill in all required fields in the model card, to update the state to ${modelDiff.state}`)
+    }
+  }
```
PE39806, line 624: This check should be moved so that it is run _after_ the auth check.

backend/src/services/model.ts
```diff
@@ -612,6 +613,16 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
 
   checkUntrustedModelRestrictions(model.kind, modelDiff)
 
+  if (modelDiff.state && model.card) {
+    const schema = await getSchemaById(model.card.schemaId, modelDiff.state)
+    const { valid } = new Validator().validate(model.card.metadata, schema.jsonSchema, {
```
PE39806: I think declaring some `const jsonSchemaValidator = new Validator()` outside of this function and then just reusing `jsonSchemaValidator` would be more efficient.

PE39806 (changes requested):

PE39806 (dismissed):

backend/src/services/model.ts
```diff
@@ -617,6 +620,16 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
     throw Forbidden(auth.info, { userDn: user.dn })
   }
 
+  if (modelDiff.state && model.card) {
+    const schema = await getSchemaById(model.card.schemaId, modelDiff.state)
+    const { valid } = jsonSchemaValidator.validate(model.card.metadata, schema.jsonSchema, {
```
JR40159: Duplication of `validateContentAgainstSchema`

JR40159 (changes requested):

frontend/src/common/EntrySelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+        const { requiredState, ...queryWithoutFiltered } = router.query
+        router.replace({
+          query: queryWithoutFiltered,
+        })
+      }
+      mutate()
+    }
+    setIsEdit(false)
+  }
+
```
ON93754, line 63: `const labelLowerCase = label.toLowerCase()`

Then just use `labelLowerCase` below instead

backend/src/services/model.ts
JR40159: Also no unit tests to cover this.

JR40159 (changes requested):

JR40159 (changes requested):
Some bugs from manual testing:
- Model state validation not working for empty model card
- The dropdown for editing the state and org is fixed width so breaks on window resizing
- Once in edit model for state and org the user can't get out of edit mode without changing the value

backend/src/services/accessRequest.ts
```diff
@@ -41,11 +41,10 @@ export async function createAccessRequest(
   if (schema.hidden) {
     throw BadReq('Cannot create new Access Request using a hidden schema.', { schemaId: accessRequestInfo.schemaId })
   }
-  try {
-    await validateContentAgainstSchema(accessRequestInfo.schemaId, accessRequestInfo.metadata)
-  } catch (error) {
+  const { valid, errors } = await validateContentAgainstSchema(accessRequestInfo.schemaId, accessRequestInfo.metadata)
+  if (!valid) {
```
JR40159, line 45: Could we handle this in the `validateContentAgainstSchema` once to reduce duplication of handling the invalid case?

JR40159 (changes requested):

backend/src/services/accessRequest.ts
GB907762: I think this is legacy behaviour, in the fact that we seem to rethrow the error that is presented as a result of `validateContentAgainstSchema`.

I think it's an improvement to abstract what error's thrown away from this function, since we may want to handle validation errors differently depending on the usage.

JR40159 (dismissed):

frontend/src/entry/EntryOverviewDetails.tsx
PE39806, line 1: I don't like how the selected item in the dropdown is blank when the value is `Unset`

Image

backend/test/services/model.spec.ts
```diff
@@ -633,9 +653,7 @@ describe('services > model', () => {
 
   test('saveImportedModelCard > unknown error when trying to validate model card', async () => {
```
PE39806: Should probably rename this since the funciton has changed
```suggestion
  test('saveImportedModelCard > throws bad request when model card validation fails', async () => {
```

PE39806 (changes requested):

backend/test/services/model.spec.ts
GB907762: Good spot. Actually I think I misunderstood this test. It's to handle the previous case when `validate` throws a non validation error. Which is handled implicitly via the error handler.

I'll remove this since it's a duplicate.

frontend/src/entry/EntryOverviewDetails.tsx
GB907762: I thought about this. If you think it's clearer by having `Unset` I'll do that.

JR40159 (approved):

PE39806 (approved):

PR #3720, opened 2026-06-15
Freeze Registry

GB907762:
Freeze the registry to include s3 driver prefix bug.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -60,7 +60,8 @@ services:
       - backend
 
   registry:
-    image: registry:3.1.1
+    # https://github.com/distribution/distribution/pull/4898
+    image: distribution/distribution@sha256:1eda63c5b5ba2c53b8b7e7071d7e67ca42e93c4ea5d9015c2b8160eb09a95d61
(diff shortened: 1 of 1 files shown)


PE39806 (approved):

PE39806 (changes requested):
This won't apply to k8s deployments - `infrastructure/helm/bailo/values.yaml` needs updating, but as it's a digest rather than a tag so `infrastructure/helm/bailo/templates/registry/registry.deployment.yaml` will also need a change to use `@` rather than `:` for the `image`.

JR40159:
> This won't apply to k8s deployments - `infrastructure/helm/bailo/values.yaml` needs updating, but as it's a digest rather than a tag so `infrastructure/helm/bailo/templates/registry/registry.deployment.yaml` will also need a change to use `@` rather than `:` for the `image`.

The plan is to first try this with the local dev deployments before rolling it out to k8s deployments.

JR40159 (approved):

PE39806 (approved):

PR #3728, opened 2026-06-16
Added button for additional reviews

ARADDCC002:
Gives users who have made a release/access request the means to notify reviewers who have requested changes that the work is ready for a new review. 

Next to the `[role] has requested changes` text, will be a button you can press to send an email notification.


backend/src/services/v3/review.ts
```diff
@@ -167,3 +169,37 @@ export async function createLifecycleReview(
...
+    log.warn(
+      {
+        err: e instanceof Error ? e.message : 'Unknown error',
+        reviewId,
+      },
+      'Failed to notify reviewers',
+    )
+    throw InternalError(
+      'Notification to reviewer could not be sent, please contact Bailo support if the problem persists.',
+    )
```
PE39806: This will be thrown if SMTP is disabled which is opaque behaviour for a user. Other bits of code use
```
  if (!config.smtp.enabled) {
    log.info('Not sending email due to SMTP disabled')
    return
  }
```
which might be more appropriate here?

I wonder if this new button should really only be enabled when SMTP is also enabled, but this is not currently exposed by UIConfig and I wouldn't want the existing config value to be duplicated.

PE39806 (changes requested):

backend/src/services/v3/review.ts
```diff
@@ -167,3 +170,34 @@ export async function createLifecycleReview(
...
+  try {
+    const model = await getModelById(user, review.modelId)
+    // The above only determines view access to a model, we should make sure the user has write access too
+    const auth = await authorisation.model(user, model, ModelAction.Write)
+    if (!auth.success) {
+      throw Forbidden(auth.info, { userDn: user.dn, modelId: model.id })
+    }
+    await notifyReviewerOfAdditionalReview(user, review)
+  } catch (_e) {
+    throw InternalError(
```
PE39806: Prefer:
```suggestion
  } catch (err) {
    log.error(err, 'Failed to notify reviewer')
    throw InternalError(
```
This way if we start getting other errors (e.g. `BadReq`) we can at least see what is causing them.

backend/src/services/v3/review.ts
```diff
@@ -167,3 +170,34 @@ export async function createLifecycleReview(
...
+
+export async function notifyReviewer(user: UserInterface, reviewId: string) {
+  if (!config.smtp.enabled) {
+    log.info('Not sending email due to SMTP disabled')
+    return
+  }
+  const review = await findReviewById(user, reviewId)
+  if (review.lastNotificationAt && Date.now() < review.lastNotificationAt.getTime() + 300000) {
+    throw BadReq('A notification was already sent recently.')
+  }
```
PE39806, line 181: This is susceptible to a race condition for multiple concurrent requests (as the `find` and the `findOneAndUpdate` are not a single event. Prefer something like
```suggestion
  const updated = await ReviewModel.findOneAndUpdate(
    {
      _id: reviewId,
      $or: [
        { lastNotificationAt: { $exists: false } },
        { lastNotificationAt: { $lt: new Date(Date.now() - 5 * 60 * 1000) } },
      ],
    },
    { $set: { lastNotificationAt: new Date() } },
  )

  if (!updated) {
    throw BadReq('A notification was already sent recently.')
  }
```
then remove the latter `await ReviewModel.findOneAndUpdate`. This might need some further reordering with the auth check, but this will at least be atomic.

backend/src/models/Review.ts
```diff
@@ -29,6 +29,7 @@ type PartialReviewInterface = {
   role: string
   createdAt: string
   updatedAt: string
+  lastNotificationAt?: Date
```
PE39806: `createdAt` and `updatedAt` are `string`s. I'm happy for this to be a `Date` but I think we need to standardise this better (possibly in a separate PR).

backend/src/services/v3/review.ts
```diff
@@ -167,3 +170,34 @@ export async function createLifecycleReview(
...
   return newReview
 }
+
+export async function notifyReviewer(user: UserInterface, reviewId: string) {
+  if (!config.smtp.enabled) {
+    log.info('Not sending email due to SMTP disabled')
+    return
+  }
+  const review = await findReviewById(user, reviewId)
+  if (review.lastNotificationAt && Date.now() < review.lastNotificationAt.getTime() + 300000) {
```
GB907762: `300000` Could we make this configurable?

backend/src/services/v3/review.ts
```diff
@@ -167,3 +170,34 @@ export async function createLifecycleReview(
...
+    throw BadReq('A notification was already sent recently.')
+  }
+  try {
+    const model = await getModelById(user, review.modelId)
```
GB907762:
```suggestion
    const model = await getModelByIdNoAuth(user, review.modelId)
```
So you don't have to authorise twice 😄

backend/src/services/v3/review.ts
GB907762: I think this rethrows the `403`.

Could you only wrap the try around `notifyReviewerOfAdditionalReview`

backend/src/services/v3/review.ts
PE39806: Also the `5 * 60 * 1000` should probably be some config value (maybe `config.review.notifyCooldownMs`).

backend/src/services/smtp/smtp.ts
```diff
@@ -241,6 +245,65 @@ export async function notifyReviewResponseForAccess(
   await dispatchEmailToModelRole(accessRequest.modelId, reviewResponse.role, await emailContent)
 }
 
+async function notifyRole(
+  user: UserInterface,
```
PE39806: Unused `user`

PE39806 (changes requested):

backend/src/services/v3/review.ts
```diff
@@ -167,3 +170,34 @@ export async function createLifecycleReview(
...
+    // The above only determines view access to a model, we should make sure the user has write access too
+    const auth = await authorisation.model(user, model, ModelAction.Write)
+    if (!auth.success) {
+      throw Forbidden(auth.info, { userDn: user.dn, modelId: model.id })
+    }
+    await notifyReviewerOfAdditionalReview(user, review)
+  } catch (_e) {
+    throw InternalError(
+      'Notification to reviewer(s) could not be sent, please contact Bailo support if the problem persists.',
+    )
```
GB907762:
```suggestion
  } catch (error) {
    throw InternalError(
      'Notification to reviewer(s) could not be sent', { modelId: model.id, error}
    )
```
Better context. Also we have a support link in the UI component.

frontend/pages/model/[modelId]/access-request/[accessRequestId].tsx
```diff
@@ -32,6 +35,11 @@ export default function AccessRequest() {
     modelId: modelId as string,
     accessRequestId: accessRequestId || '',
   })
+  const {
+    responses: reviewResponses,
+    isResponsesLoading: isReviewResponsesLoading,
+    isResponsesError: isReviewResponsesError,
+  } = useGetResponses([...reviews.map((review) => review._id)])
```
Comment (GB907762) [inline frontend/pages/model/[modelId]/access-request/[accessRequestId].tsx]:
```suggestion
  } = useGetResponses(reviews.map((review) => review._id))
```

frontend/src/entry/model/reviews/ReviewDisplay.tsx
```diff
@@ -84,25 +109,41 @@ export default function ReviewDisplay({
...
-                    </Stack>
-                  )
-                })}
-            </Stack>
-          </Tooltip>
+          <>
+            <Tooltip title={`${plural(orderedReviewResponses.length, 'review')}`}>
+              <Stack>
+                {orderedReviewResponses
+                  .filter((reviewResponse) => reviewResponse.decision === Decision.RequestChanges)
```
GB907762, line 93: Do we also want this when there is no decision?

I realise we have a notification on access requests when they're created so might be too much.

frontend/src/entry/model/reviews/ReviewDisplay.tsx
```diff
@@ -84,25 +109,41 @@ export default function ReviewDisplay({
...
+                        <HourglassEmpty color='warning' fontSize='small' />
+                        <Typography variant='caption'>
+                          {showCurrentUserResponses
+                            ? `You have requested changes as a ${roleNameDisplay(response)}`
+                            : `Changes requested by  ${roleNameDisplay(response)}`}
+                        </Typography>
+                        <Restricted action='editRelease' fallback={<></>}>
+                          <>
+                            {
+                              <Button
```
GB907762: It would be cool if we could grey this out when there's a cooldown?

frontend/src/entry/model/reviews/ReviewDisplay.tsx
```diff
@@ -84,25 +109,41 @@ export default function ReviewDisplay({
...
+                                size='small'
+                                onClick={() => handleNotifyReviewerOnClick(response.parentId)}
```
GB907762: I'd be in favour of splitting this map. Maybe we could make a new component. This handler triggers on both if you click one, which I don't think is expected.

frontend/src/entry/model/reviews/ReviewDisplay.tsx
```diff
@@ -84,25 +109,41 @@ export default function ReviewDisplay({
...
+                                startIcon={<Refresh />}
+                                loading={isNotifyButtonLoading}
+                              >
+                                Request re-review
```
GB907762: What happens when I don't have emails enabled on a Bailo instance?

Seems to just hang..

GB907762 (dismissed):

backend/src/services/v3/review.ts
ARADDCC002: I'll try again, but I'm sure GH was complaining about this, will try and see!

frontend/src/entry/model/reviews/ReviewDisplay.tsx
ARADDCC002: The cooldown is stored server-side so not sure we'd be able to do that as it would be renabled when the page re-renders

frontend/src/entry/model/reviews/ReviewDisplay.tsx
ARADDCC002: It will check the config to make sure it's enabled. Do you have it enabled in config but not running?

frontend/src/entry/model/reviews/ReviewDisplay.tsx
ARADDCC002: I think it's fine just for when requesting changes personally, but happy to change if others prefer.

backend/src/services/v3/review.ts
ARADDCC002: I've added the interval to config.

As for the race condition, I'm not sure the solution presented would work. If the notification fails for some reason, the date will have been updated, and the user will need to wait for the cool down to have finished.

frontend/src/entry/model/reviews/ReviewDisplay.tsx
GB907762: This component already gets the UI config (with a cooldown period). And it also gets `lastNotificationAt`. Couldn't we have some frontend validation?

ARADDCC002: The config is set inside `smtp` which isn't sent to the frontend. We _could_ add it to `uiConfig` instrad, but I'm wary of having two sets of `smtp` config stored in separate parts of our config and would prefer that the backend handles the validation in this case.

frontend/src/entry/model/reviews/ReviewDisplay.tsx
GB907762: So when it's disabled in config, It seems the notification executes fine which may be misleading for the user. In the fact they think a reviewer has received an email?

backend/src/services/v3/review.ts
```diff
@@ -177,21 +177,24 @@ export async function notifyReviewer(user: UserInterface, reviewId: string) {
...
-  if (review.lastNotificationAt && Date.now() < review.lastNotificationAt.getTime() + 300000) {
+  if (
+    review.lastNotificationAt &&
+    Date.now() < review.lastNotificationAt.getTime() + config.smtp.review.lastNotifiedCoolDownMs
+  ) {
     throw BadReq('A notification was already sent recently.')
   }
+
+  const model = await getModelByIdNoAuth(review.modelId)
+  // The above only determines view access to a model, we should make sure the user has write access too
```
PE39806: No longer true
```suggestion
  // `getModelById` would only determine view access to a model, we must make sure the user has write access too
```

infrastructure/helm/bailo/templates/bailo/bailo.configmap.yaml
```diff
@@ -126,6 +126,10 @@ data:
           postReminderInterval: '{{ .Values.config.smtp.postReminderInterval }}',
         },
 
+        review: {
+          lastNotifiedCoolDownMs: {{ .Values.config.smtp.lastNotifiedCoolDownMs }},
```
PE39806: Please also update `infrastructure/helm/README.md` with this new config value.

PE39806 (changes requested):

backend/src/models/Review.ts
```diff
@@ -29,7 +29,7 @@ type PartialReviewInterface = {
   role: string
   createdAt: string
   updatedAt: string
-  lastNotificationAt?: Date
+  lastNotificationAt?: Date | null
```
PE39806: Is `| null` required when a type has a `?` already?

frontend/types/types.ts
```diff
@@ -588,6 +588,7 @@ type PartialReviewInterface = {
   role: string
   createdAt: string
   updatedAt: string
+  lastNotificationAt?: Date | null
```
PE39806: As above.

backend/src/services/v3/review.ts
```diff
@@ -167,3 +170,48 @@ export async function createLifecycleReview(
...
+    },
+    // We want to store the original document for rollback purposes
+    { new: false },
+  )
+
+  if (!originalReview) {
+    throw BadReq('Could not find existing review.', { reviewId })
+  }
+
+  if (
```
JR40159: Rather than having to store and then maintain this information in our database I think it would be much simpler to just rate limit this endpoint

JR40159 (changes requested):

backend/src/services/v3/review.ts
```diff
@@ -167,3 +170,48 @@ export async function createLifecycleReview(
 
   return newReview
 }
+
+export async function notifyReviewer(user: UserInterface, reviewId: string) {
```
JR40159, line 174: Unit test

backend/src/routes/v3/review/postNotifyReviewer.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export const postNotifyReviewer = [
+  async (req: Request, res: Response<PostNotifyReviewerResponse>): Promise<void> => {
+    const {
+      params: { reviewId },
+    } = parse(req, postNotifyReviewerSchema)
+    await notifyReviewer(req.user, reviewId)
+
+    res.json({ status: 'Notification sent to reviewer.' })
```
JR40159: This response body is pretty pointless, a 200 response would probably be enough?

backend/src/routes/v3/review/postNotifyReviewer.ts
JR40159, line 1: Missing auditing

JR40159 (changes requested):

backend/src/routes/v3/routes.ts
```diff
@@ -11,6 +13,18 @@ import { postNotifyReviewer } from './review/postNotifyReviewer.js'
 import { postReview } from './review/postReview.js'
 import { postReviewResponse } from './review/postReviewResponse.js'
 
+const reviewNotifierLimiter = rateLimit({
```
JR40159: As this is specific to one route I would maybe define this and add it to the route in `postNotifyReviewer.ts` to keep this `routes.ts` file nice and clean.

backend/src/routes/v3/routes.ts
```diff
@@ -11,6 +13,18 @@ import { postNotifyReviewer } from './review/postNotifyReviewer.js'
...
+  windowMs: config.smtp.review.lastNotifiedCoolDownMs,
```
JR40159: This might be controversial, but do we need to make this configurable? I'm not sure what is considered to be a sensible cool-down period will frequently differ in different deployments of Bailo. Our application configuration is getting very large so I think we need to be sure before making it even larger.

JR40159 (changes requested):

PE39806 (dismissed):

backend/src/routes/v3/routes.ts
ARADDCC002: I'm happy to remove it for now - if people feel it should be configurable, we can re-add it. I agree our config is getting massive!

backend/src/connectors/audit/Base.ts
JR40159, line 1: The Stroom audit class has an error as it's missing the audit funciton

JR40159 (dismissed):

PE39806 (approved):

JRB66955 (approved):

PR #3740, opened 2026-06-16
Feature/bai 2511 view all question states

IR96334:


frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -112,8 +128,18 @@ export default function JsonSchemaForm({
...
+
+  const handleFilterStateClick = (state: string) => {
+    if (requiredByModelState === state) {
+      state = ''
+    }
+    router.replace({
+      query: { ...router.query, requiredByModelState: state } as RouterQueryParams,
     })
+    setRequiredByModelState(state)
   }
```
GB907762:
```suggestion
  const handleFilterStateClick = (state: string) => {
    if (state === requiredByModelState) {
      // eslint-disable-next-line @typescript-eslint/no-unused-vars
      const { requiredByModelState, ...queryWithoutFiltered } = router.query
      router.replace({
        query: queryWithoutFiltered,
      })
      setRequiredByModelState('')
    } else {
      router.replace({
        query: { ...router.query, requiredByModelState: state } as RouterQueryParams,
      })
      setRequiredByModelState(state)
    }
  }
```
`state = ''` Doesn't remove the query parameter. Use this instead.

GB907762: Could you also add this functionality to when the form becomes read only? 

For example when I am not in edit mode I want this query parameter to disappear.

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -184,16 +218,31 @@ export default function JsonSchemaForm({
...
             </Box>
           )}
           {canEdit && (
-            <Typography sx={{ pt: 1 }}>
-              Required fields for this state are marked with an asterisk
-              <span style={{ color: theme.palette.error.main }}>*</span>
-            </Typography>
+            <Stack direction={'column'} spacing={1}>
+              <Stack spacing={2} direction={'row'}>
+                <Typography>Filter fields by: </Typography>
```
GB907762: I'm a bit concerned about the compatibility of schemas with no required fields. This option is very prominent and has no use if you're not interested in this feature. #3716 Uses the query parameter to highlight different fields already. 

I wonder if we remove this chip selector entirely. And just redirect through that?

frontend/utils/formUtils.ts
```diff
@@ -59,12 +59,9 @@ export function createStep({
     mirroredState,
     type,
     index,
-
     section,
     schemaRef,
-
     shouldValidate: false,
-
     isComplete,
```
GB907762, line 65:
```typescript
    index,
    section,
    schemaRef,
    shouldValidate: false,
    isComplete,
```
Remove empty lines

frontend/utils/formUtils.ts
```diff
@@ -253,21 +250,28 @@ function isPrimitiveSchema(schema: any): boolean {
   return ['string', 'number', 'boolean'].includes(schema.type)
 }
 
-function countQuestionsFromSchema(schema: any): number {
+export function requiredByStateFilter(requiredByModelState?: string, schema?: any) {
```
GB907762: In my mind a box highlighting the required fields may be a better option.

I think hiding information may cause some UX problems.

For example I may want fill in non required fields but in order to view them, I have to turn the filter off and scroll down to them.

Happy to be swayed, but I just feel it's a bad user experience.

frontend/utils/formUtils.ts
```diff
@@ -253,21 +250,28 @@ function isPrimitiveSchema(schema: any): boolean {
...
+  return requiredByModelState &&
+    (!schema.requiredByModelStates || !schema.requiredByModelStates.includes(requiredByModelState))
+    ? false
+    : true
```
GB907762, line 281:
```suggestion
  return Boolean(
    requiredByModelState && !(schema.requiredByModelStates && schema.requiredByModelStates.includes(requiredByModelState)),
  )
```

GB907762 (changes requested):

frontend/utils/formUtils.ts
IR96334: This might be bug on github as I have none

GB907762: Oh no. Eslint has removed them. Interesting why they are there in the first place. ignore this comment.

backend/src/services/schema.ts
```diff
@@ -56,38 +58,74 @@ export async function getSchemaById(schemaId: string, modelState?: string): Prom
...
 
-function enforceModelStateFields(schema: object, targetState: string) {
+function addToParentRequired(
+  pointer: string,
+  modifiedSchemas: WeakSet<object>,
+  parentKeyword?: string,
+  parentSchema?: traverse.SchemaObject,
+) {
+  if (parentKeyword === 'properties' && parentSchema) {
+    const propertyName = pointer.replace(/~1/g, '/').replace(/~0/g, '~').split('/').pop()
```
ON93754, line 74: Comment to explain what/why would be good for this line

backend/src/services/schema.ts
```diff
@@ -56,38 +58,74 @@ export async function getSchemaById(schemaId: string, modelState?: string): Prom
...
-    schemaObject.jsonSchema = enforceModelStateFields(schemaObject.jsonSchema, modelState)
-  }
+  schemaObject.jsonSchema = enforceModelStateFields(schemaObject.jsonSchema, modelState)
+
   schemaCache.set(JSON.stringify({ schemaId, modelState }), schemaObject)
   return schemaObject
 }
 
-function enforceModelStateFields(schema: object, targetState: string) {
+function addToParentRequired(
```
ON93754, line 67: Worth adding a couple of unit tests for this function? I'd be worried about a bug making some of this logic fiddly to debug, so some unit tests would give us more confidence

frontend/src/common/EntrySelect.tsx
```diff
@@ -0,0 +1,105 @@
...
+        {isEdit ? (
+          <FormControl sx={{ maxWidth: 240 }} fullWidth size='small'>
+            <Select
+              onClose={handleEditChange}
+              error={Boolean(errorMessage)}
+              id={labelLowerCase}
+              value={value ?? ''}
+              onChange={handleSelectOption}
+              displayEmpty
+              renderValue={(value: any) => (value ? value : <em>Unset</em>)}
```
ON93754: Should this type be `string` rather than any?

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -184,16 +203,33 @@ export default function JsonSchemaForm({
...
+                        onClick={() => handleFilterStateClick(state)}
+                        color={requiredByModelState === state ? 'primary' : 'default'}
+                      />
+                    </Tooltip>
+                  ))}
+                </Stack>
+              )}
+              <Typography sx={{ pt: 1 }}>
+                Required fields for this state are marked with an asterisk
+                <span style={{ color: theme.palette.error.main }}> *</span>``
```
GB907762:
```suggestion
                <span style={{ color: theme.palette.error.main }}> *</span>
```

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -184,16 +203,33 @@ export default function JsonSchemaForm({
...
           )}
           {canEdit && (
-            <Typography sx={{ pt: 1 }}>
-              Required fields for this state are marked with an asterisk
-              <span style={{ color: theme.palette.error.main }}>*</span>
-            </Typography>
+            <Stack direction={'column'} spacing={1}>
+              {stateList && (
+                <Stack spacing={2} direction={'row'}>
+                  <Typography>Filter fields by: </Typography>
```
GB907762: if the states are an empty list this will still appear?

GB907762 (changes requested):

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -184,16 +203,33 @@ export default function JsonSchemaForm({
...
             </Box>
           )}
           {canEdit && (
-            <Typography sx={{ pt: 1 }}>
-              Required fields for this state are marked with an asterisk
-              <span style={{ color: theme.palette.error.main }}>*</span>
-            </Typography>
+            <Stack direction={'column'} spacing={1}>
+              {stateList && (
+                <Stack spacing={2} direction={'row'}>
```
PE39806: I would like for the chip(s) to be vertically aligned with the preceding `Typography`.

frontend/src/Form/FormTemplates.tsx
```diff
@@ -1,18 +1,22 @@
 import { Share } from '@mui/icons-material'
 import AddIcon from '@mui/icons-material/Add'
 import CloseIcon from '@mui/icons-material/Close'
+import Done from '@mui/icons-material/Done'
+import Error from '@mui/icons-material/Error'
```
PE39806: We tend to use the outlined variant of these icons for other errors in Bailo
```suggestion
import Error from '@mui/icons-material/ErrorOutline'
```

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -43,11 +52,14 @@ export default function JsonSchemaForm({
...
   displayStats?: boolean
+  stateList?: string[]
 }) {
-  const [activeStep, setActiveStep] = useState(0)
-  const [sharedSection, setSharedSection] = useState('')
   const theme = useTheme()
   const router = useRouter()
+  const [activeStep, setActiveStep] = useState(0)
+  const [sharedSection, setSharedSection] = useState('')
+  const requiredByModelState = useSearchParams().get('requiredByModelState')
```
PE39806: Should this be compared to some const/enum to validate known states? Users could mess with their own URL to manipulate otherwise and have invalid states set, which could bypass the frontend checks. The backend will still perform validation as expected and reject missing required answers.

frontend/utils/formUtils.ts
IR96334: The original version is required for completion bar, but as we are no longer filtering and highlighting it can be used for questions in schema

frontend/src/Form/FormTemplates.tsx
```diff
@@ -53,6 +57,45 @@ export function DescriptionFieldTemplate() {
...
+        </Stack>
+        {children}
+      </Box>
+    )
+  }
+
+  return <>{children}</>
+}
+
+export function ErrorListTemplate({ registry }) {
```
PE39806: Add typing e.g.
```
import { ErrorListProps } from '@rjsf/utils'

export function ErrorListTemplate({ registry }: ErrorListProps) {
```

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -184,16 +203,33 @@ export default function JsonSchemaForm({
...
-            <Typography sx={{ pt: 1 }}>
-              Required fields for this state are marked with an asterisk
-              <span style={{ color: theme.palette.error.main }}>*</span>
-            </Typography>
+            <Stack direction={'column'} spacing={1}>
+              {stateList && (
+                <Stack spacing={2} direction={'row'}>
+                  <Typography>Filter fields by: </Typography>
+                  {stateList.map((state) => (
+                    <Tooltip key={state} title={`Filter out questions necessary for ${state}`}>
```
PE39806: This is the wrong way around
```suggestion
                    <Tooltip key={state} title={`Highlight questions required for ${state}`}>
```

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -184,16 +203,33 @@ export default function JsonSchemaForm({
           {displayStats && (
             <Box>
               <Box>
-                Entries Completed: {formStats.totalAnswers}/{formStats.totalQuestions}
+                {requiredByModelState} Entries Completed: {formStats.totalAnswers}/{formStats.totalQuestions}
```
PE39806: Prefer sentence case, and remove the leading space when no `requiredByModelState` is set e.g.
```suggestion
                {toSentenceCase(
                  `${requiredByModelState ? `${requiredByModelState} ` : ''}entries completed: ${formStats.totalAnswers}/${formStats.totalQuestions}`,
                )}
```

frontend/src/Form/FormTemplates.tsx
```diff
@@ -53,6 +57,45 @@ export function DescriptionFieldTemplate() {
...
+    registry.formContext.requiredByModelState &&
+    schema.requiredByModelStates &&
+    schema.requiredByModelStates.includes(registry.formContext.requiredByModelState)
+  if (requiredByState) {
+    return (
+      <Box
+        p={0.5}
+        sx={{
+          backgroundColor: alpha(
+            answered ? registry.formContext.theme.palette.primary.main : registry.formContext.theme.palette.error.main,
```
PE39806: Why do this rather than using `const theme = useTheme()` as elsewhere?

backend/src/services/schema.ts
```diff
@@ -57,9 +58,8 @@ export async function getSchemaById(schemaId: string, modelState?: string): Prom
   const schemaObject = schema.toObject()
   schemaObject.jsonSchema = structuredClone(schema.jsonSchema)
 
-  if (modelState) {
-    schemaObject.jsonSchema = enforceModelStateFields(schemaObject.jsonSchema, modelState)
-  }
+  schemaObject.jsonSchema = enforceModelStateFields(schemaObject.jsonSchema, modelState)
```
PE39806: Do we need `structuredClone` both here and in `enforceModelStateFields`?

backend/src/services/schema.ts
```diff
@@ -84,9 +84,20 @@ function addToParentRequired(
   }
 }
 
-function enforceModelStateFields(schema: object, targetState: string) {
+function addUniqueStates(_root: traverse.SchemaObject, states: string[]) {
```
PE39806: Why does `_root` have the `_` prefix? This is typically used to indicate a parameter is unused.

backend/src/services/schema.ts
```diff
@@ -84,9 +84,20 @@ function addToParentRequired(
...
+  if (!_root.stateList) {
+    _root.stateList = []
+  }
+  for (const state of states) {
+    if (!_root.stateList.includes(state)) {
+      _root.stateList.push(state)
+    }
+  }
```
PE39806:
```suggestion
  const validStates = new Set(config.ui.modelDetails.states)

  _root.stateList = Array.from(
    new Set([...(_root.stateList ?? []), ...states.filter((state) => validStates.has(state))]),
  )
```

frontend/src/theme.ts
```diff
@@ -22,6 +22,7 @@ declare module '@mui/material/styles' {
     vulnerabilityHigh: PaletteColorOptions
     vulnerabilityMedium: PaletteColorOptions
     vulnerabilityLow: PaletteColorOptions
+    correct: PaletteColorOptions
```
PE39806: I think this is unused.

PE39806 (commented):
I think it would also be a good idea to add `"requiredByModelStates": ["Review", "Production"],` to `backend/src/scripts/example_schemas/minimal_model_schema.json` (maybe under `modelSummary`) to demonstrate this feature to new admins from the get-go.

frontend/utils/formUtils.ts
```diff
@@ -253,21 +292,28 @@ function isPrimitiveSchema(schema: any): boolean {
   return ['string', 'number', 'boolean'].includes(schema.type)
 }
 
-function countQuestionsFromSchema(schema: any): number {
+export function requiredByStateFilter(requiredByModelState?: string, schema?: any) {
```
GB907762: I don't think schema can be ever `undefined`

GB907762 (changes requested):

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -184,16 +214,35 @@ export default function JsonSchemaForm({
           {displayStats && (
             <Box>
               <Box>
-                Entries Completed: {formStats.totalAnswers}/{formStats.totalQuestions}
+                {toSentenceCase(
+                  `${requiredByModelState ? `${requiredByModelState} ` : ''}entries completed: ${formStats.totalAnswers}/${formStats.totalQuestions}`,
```
GB907762:
```suggestion
                  `${requiredByModelState ? `${requiredByModelState} ` : ''}fields completed: ${formStats.totalAnswers}/${formStats.totalQuestions}`,
```
I know this was like this before but I find entries a `confusing` term

frontend/src/Form/JsonSchemaForm.tsx
GB907762, line 1: It may be worth some unit tests in here. You could reuse `frontend/utils/test/formUtils.test.ts`

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -158,21 +241,30 @@ export default function JsonSchemaForm({
                 <ListItem
                   key={step.schema.title}
                   disablePadding
-                  sx={{ display: 'flex', justifyContent: 'flex-end', width: '100%' }}
+                  secondaryAction={
+                    canEdit &&
+                    sectionCompletion[step.schema.title] > 0 && (
+                      <ListItemIcon>
+                        <Circle color='error' sx={{ fontSize: 12 }} />
```
GB907762: I don't like the circle. I don't think it has very good UX.

You've already used the following styling to express that something needs to be filled in because of it's State. Could you do similar here? Maybe a tooltip would be more helpful?

```typescript
alpha(answered ? theme.palette.primary.main : theme.palette.error.main, 0.1)
```

GB907762 (changes requested):

PE39806 (changes requested):
Update the Python integration tests as these tests update the `state` without setting the now-required fields in the model card. It would be good to keep one of the existing tests and check that is _does_ fail, and then add another test that should be expected to pass.

frontend/src/entry/overview/FormEditPage.tsx
```diff
@@ -48,9 +43,19 @@ type FormEditPageProps = {
...
+export type RouterQueryParams = {
+  page?: number
+  requiredByModelState?: string
+  isEdit?: boolean
+}
+
 export default function FormEditPage({ entry, mutateEntry }: FormEditPageProps) {
-  const [isEdit, setIsEdit] = useState(false)
+  const router = useRouter()
+  const isEdit = router.query.isEdit === 'true'
```
PE39806: I'm not so keen on this as it allows any user to enter edit mode via URL manipulation. Our backend stops any saves via auth but it's not good that users can do this.

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -43,16 +64,52 @@ export default function JsonSchemaForm({
...
 }) {
+  const theme = useTheme()
   const router = useRouter()
-  const [activeStep, setActiveStep] = useState<number>(
-    router.query && router.query.page !== undefined && typeof router.query.page === 'string'
-      ? Number(router.query.page)
-      : 0,
+  const [activeStep, setActiveStep] = useState(0)
+  const [sharedSection, setSharedSection] = useState('')
+  const requiredByModelState = useSearchParams().get('requiredByModelState')
```
PE39806: Why not use just `router.query.requiredByModelState`?

frontend/src/Form/FormTemplates.tsx
```diff
@@ -59,6 +57,45 @@ export function DescriptionFieldTemplate() {
...
+  if (requiredByState) {
+    return (
+      <Box
+        p={0.5}
+        sx={{
+          backgroundColor: alpha(answered ? theme.palette.primary.main : theme.palette.error.main, 0.1),
+        }}
+      >
+        <Stack direction='row' alignItems='center'>
+          {answered ? <Done color='success' fontSize='small' /> : <Error color='error' fontSize='small' />}
```
PE39806: It would be good to add `aria-label`s e.g.
```suggestion
          {answered ? <Done color='success' fontSize='small' aria-label={`Answered field required for ${registry.formContext.requiredByModelState}`} /> : <Error color='error' fontSize='small' aria-label={`Unanswered field required for ${registry.formContext.requiredByModelState}`} />}
```

PE39806 (changes requested):

frontend/src/entry/overview/FormEditPage.tsx
IR96334: added auth. alternative would be moving state to Overview and passing state  setter prop through to the Select, think this is less clunky

frontend/src/Form/FormTemplates.tsx
```diff
@@ -59,6 +63,57 @@ export function DescriptionFieldTemplate() {
...
+    registry.formContext.requiredByModelState &&
+    schema.requiredByModelStates &&
+    schema.requiredByModelStates.includes(registry.formContext.requiredByModelState)
+
+  if (requiredByState) {
+    return (
+      <Box
+        sx={{
+          backgroundColor: alpha(answered ? theme.palette.primary.main : theme.palette.error.main, 0.1),
+          p: 0.5,
```
ARADDCC002: Personally, I think this should be set to `1`

frontend/src/Form/FormTemplates.tsx
```diff
@@ -59,6 +63,57 @@ export function DescriptionFieldTemplate() {
...
+        }}
+      >
+        <Stack direction='row' sx={{ alignItems: 'center' }}>
```
ARADDCC002: Add `spacing={0.5}`

frontend/src/Form/FormTemplates.tsx
```diff
@@ -59,6 +63,57 @@ export function DescriptionFieldTemplate() {
...
+  const theme = useTheme()
+  const answered = isQuestionAnswered(id, schema, registry.formContext)
+  const requiredByState =
+    registry.formContext.requiredByModelState &&
+    schema.requiredByModelStates &&
+    schema.requiredByModelStates.includes(registry.formContext.requiredByModelState)
+
+  if (requiredByState) {
+    return (
+      <Box
```
ARADDCC002: Change this to `<Stack />` and add `spacing={0.5}`

frontend/src/Form/FormTemplates.tsx
```diff
@@ -59,6 +63,57 @@ export function DescriptionFieldTemplate() {
...
+              aria-label={`Answered field required for ${registry.formContext.requiredByModelState}`}
+            />
+          ) : (
+            <Error
+              color='error'
+              fontSize='small'
+              aria-label={`Unanswered field required for ${registry.formContext.requiredByModelState}`}
+            />
+          )}
+          <Typography sx={{ fontSize: 12 }} color={answered ? 'success' : 'error'}>
```
ARADDCC002: Rather than using a new `fontSize`, we can probably use the `variant='caption'` prop instead

frontend/src/Form/FormTemplates.tsx
```diff
@@ -59,6 +63,57 @@ export function DescriptionFieldTemplate() {
...
+    )
+  }
+
+  return <>{children}</>
+}
+
+export function ErrorListTemplate() {
+  const theme = useTheme()
+  return (
+    <Typography color={theme.palette.error.main} sx={{ mb: 2 }}>
```
ARADDCC002: We can simplify the `color` prop by doing `color='error'`

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -120,6 +185,29 @@ export default function JsonSchemaForm({
     })
   }
 
+  const handleHighlightStateClick = (state: string) => {
+    if (state === requiredByModelState) {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
ARADDCC002: To avoid needing to disable linting for this line, we can replace this section with:

```
const { ...queries } = router.query
delete queries.requiredByModelState
router.replace({ query: queries })
```

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -145,22 +233,32 @@ export default function JsonSchemaForm({
...
-                    >
-                      {step.schema.title}
-                    </Typography>
+                    <ListItemText sx={{ pr: 1 }}>
+                      <Typography
+                        sx={{
+                          wordBreak: 'break-word',
+                          color:
+                            !step.isComplete(step) && displayLabelValidation
+                              ? theme.palette.error.main
```
ARADDCC002: We should be able to just put `error` here to make it consistent.

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -172,16 +270,35 @@ export default function JsonSchemaForm({
...
               </Box>
               <LinearProgressWithLabel value={formStats.percentageQuestionsComplete} />
             </Box>
           )}
           {canEdit && (
-            <Typography sx={{ pt: 1 }}>
-              Required fields for this state are marked with an asterisk
-              <span style={{ color: theme.palette.error.main }}>*</span>
-            </Typography>
+            <Stack direction={'column'} spacing={1}>
```
ARADDCC002: You can remove the `direction` prop as `column` is the default direction.

frontend/src/Form/JsonSchemaForm.tsx
```diff
@@ -172,16 +270,35 @@ export default function JsonSchemaForm({
...
+              {stateList && stateList.length > 0 && (
+                <Stack spacing={2} direction={'row'} sx={{ alignItems: 'center' }}>
```
ARADDCC002: `direction='row'`

ARADDCC002 (changes requested):

frontend/src/entry/overview/FormEditPage.tsx
GB907762, line 126: I know it doesn't specifically applies to this ticket but when change tab a confirm dialog appears, because the route has changed. 

Could we remove this since it's more prominent with this filtering.

GB907762 (changes requested):

frontend/src/entry/overview/FormEditPage.tsx
IR96334: Kept as its important so users don't lose progress when moving pages. But now ignores the page query param, as that is the problem.

GB907762: Yep also remove the `requiredByStates` param.

frontend/src/common/EntrySelect.tsx
GB907762:
```suggestion
```
Really small but I think removing this whole if condition is better. If this is removed it wont trigger the chip selection to disappear when you select a new state?

GB907762 (dismissed):

frontend/src/entry/EntryOverviewDetails.tsx
PE39806: This change is not needed as we have changed Untrusted Model behaviour to allow setting any state.

PE39806 (changes requested):

frontend/src/hooks/useUnsavedChanges.ts
```diff
@@ -20,10 +20,25 @@ export default function useUnsavedChanges(): UnsavedChangesHook {
...
+      return remaining ? `${pathname}?${remaining}` : pathname
+    }
+    const handleBrowseAway = (newUrl: string) => {
       if (!unsavedChanges) {
         return
       }
+      // Skip warning if the only difference between the URLs is the 'page' query param
+      if (getComparableUrl(router.asPath) === getComparableUrl(newUrl)) {
+        return
+      }
```
PE39806, line 42: Please take the tests from #3835 and then I will close that PR in favour of this one implementing the fix.

PE39806 (changes requested):

frontend/src/hooks/useUnsavedChanges.ts
PE39806, line 1: Editing a model card then clicking on "Save" now erroneously triggers the "Any unsaved changes will be lost - are you sure you wish to leave this page?" alert.

PE39806 (changes requested):

PE39806 (approved):

ARADDCC002 (approved):

PR #3742, opened 2026-06-17
Configure dayjs to enforce UTC usage throughout

JRB66955:


PE39806 (changes requested):
I really like this change, but please could you add a new rule to `frontend/eslint.config.mjs` to enforce using `import ... from @dayjs` rather than `import ... from dayjs`, similar to what we do with `zod` in `backend/eslint.config.mjs`? This will ensure we maintain this practice going forwards.

JRB66955:
> I really like this change, but please could you add a new rule to `frontend/eslint.config.mjs` to enforce using `import ... from @dayjs` rather than `import ... from dayjs`, similar to what we do with `zod` in `backend/eslint.config.mjs`? This will ensure we maintain this practice going forwards.

The `eslint.config.mjs` config file has been updated to achieve this.

frontend/eslint.config.mjs
```diff
@@ -50,6 +50,27 @@ export default defineConfig([
       'react-hooks/set-state-in-effect': 'off',
     },
   },
+  {
+    files: ['src/**/*.ts', 'src/**/*.tsx'],
```
PE39806:
```suggestion
    files: [
      'actions/**/*.ts',
      'actions/**/*.tsx',
      'pages/**/*.ts',
      'pages/**/*.tsx',
      'src/**/*.ts',
      'src/**/*.tsx',
      'utils/**/*.ts',
      'utils/**/*.tsx',
    ],
```

PE39806 (changes requested):

PE39806 (dismissed):

GB907762 (dismissed):

GB907762 (approved):

PE39806 (approved):

Issue #3748, opened 2026-06-19
Import model card from text using LLM extraction


M422110:
### Problem

When onboarding models into Bailo, users often have existing model documentation — HuggingFace READMEs, system cards, internal model docs — but must manually re-enter this information field by field into Bailo's schema-driven form. This is time-consuming and error-prone.

### Summary
Allow users to import model card data from open-source model card text files (e.g., HuggingFace model card) by pasting text into a dialog and using an LLM endpoint to extract structured data into the Bailo schema format.

This reduces the manual effort required to curate compliant model cards when existing documentation is available.

### Proposed Solution
**User Flow**
User creates a model and selects a schema (existing workflow, unchanged)
User enters edit mode on the model card form
User clicks "Import from Text" button (new, appears next to existing "Add JSON to form")
A dialog opens where the user pastes their model card text
The backend sends the text + schema to a configured LLM endpoint
The LLM extracts structured data matching the schema format
The extracted data populates the form fields for user review
User reviews, edits, and saves as normal

### Technical Approach
**Backend:**
New configurable LLM integration using OpenAI-compatible chat completions API format
New service (modelCardImport.ts) that builds a schema-guided prompt and validates the LLM response
New route: POST /api/v2/model/:modelId/import-model-card-text
Config section (llm) with endpoint URL, API key, model name, timeout — disabled by default

**Frontend:**
New ImportModelCardTextDialog component for pasting text and viewing extraction progress
Integration into FormEditPage alongside existing "Add JSON to form" feature
Feature-flagged via uiConfig.llmImport.enabled — completely hidden when LLM is not configured

**Anti-hallucination strategy:**
System prompt strictly forbids inference/guessing — only explicitly stated information is extracted
Schema field titles AND description/help text are included in the prompt to guide accurate extraction
Server-side validation strips unknown keys and wrong types from LLM response
Internal-reference fields (entity selectors, data card selectors) excluded from the prompt entirely
Fields that cannot be populated from the source document remain blank
User always reviews extracted data in the form before saving
Deployment Considerations
Feature is opt-in and off by default — deployments without an LLM endpoint configured will not see the "Import from Text" button at all
No new dependencies required for the LLM client (uses native fetch with OpenAI-compatible API format)

### Acceptance Criteria
 User can paste model card text and have it populate schema form fields
 Only information explicitly present in the source text is extracted
 Fields that cannot be populated remain blank (no hallucinated data)
 Extracted data respects the schema structure (correct types, valid keys only)
[...]

PR #3750, opened 2026-06-19
Feature/import model card text

M422110:
Feature implementation for issue: https://github.com/gchq/Bailo/issues/3748


backend/src/clients/llm.ts
GB907762, line 1: Could we have some tests for this file please.

backend/src/clients/llm.ts
```diff
@@ -0,0 +1,64 @@
...
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        ...(config.llm.apiKey && { Authorization: `Bearer ${config.llm.apiKey}` }),
+      },
+      body: JSON.stringify({
+        model: config.llm.model,
+        messages,
+        max_tokens: config.llm.maxTokens,
+        temperature: 0,
```
GB907762: I'm not sure about using greedy decoding on a model card. I understand a deterministic approach would be good, but could cause lack of nuance or over simplification. 

Additionally not every model will allow for temperate to be 0.

Would it be better to make this configurable?

backend/src/routes/v2/model/modelcard/postImportModelCardText.ts
```diff
@@ -0,0 +1,60 @@
...
+    modelId: z.string({
+      required_error: 'Must specify model id as param',
+    }),
+  }),
+  body: z.object({
+    text: z
+      .string({
+        required_error: 'Must provide model card text to import',
+      })
+      .max(500000),
```
GB907762: See my other comment about changing the request text size. Personally I wouldn't validate with both zod and a middleware, and pick one or the other.

Also should this be `512_000` or `bytes.parse('500kb')`?

backend/src/services/modelCardImport.ts
GB907762, line 1: Could we have some tests for this file.

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,278 @@
...
+interface FieldDescriptor {
+  path: string
+  title: string
+  type: string
+  description?: string
+  format?: string
+}
+
+function describeObjectFields(schema: JsonSchema): string {
+  if (!schema.properties) return '{}'
```
GB907762: Eslint curly rule
```suggestion
  if (!schema.properties) {
    return '{}'
  }
```

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,278 @@
...
+  const parts: string[] = []
+  for (const [k, v] of Object.entries(schema.properties) as [string, JsonSchema][]) {
+    if (v.type === 'array' && v.items) {
+      const items = v.items as JsonSchema
+      if (items.type === 'object' && items.properties) {
+        parts.push(`${k}: array of objects with fields: ${describeObjectFields(items)}`)
+      } else if (items.enum) {
```
GB907762: I've tried updating some enums explicitly. I can't seem to get any to update. I can't work out why.

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,278 @@
...
+        )
+      } else {
+        parts.push(`${k}: array of ${items.type || 'string'}`)
+      }
+    } else if (v.type === 'object' && v.properties) {
+      parts.push(`${k}: object with fields: ${describeObjectFields(v)}`)
+    } else {
+      parts.push(`${k}: ${v.type || 'string'}`)
+    }
+  }
```
GB907762: I'd really love some better variable names especially when they're only one letter, even if `key` and `value` are used. `k` `v` and `e`

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,278 @@
...
+  return `{ ${parts.join(', ')} }`
+}
+
+export function buildSchemaDescription(schema: JsonSchema, basePath = ''): FieldDescriptor[] {
```
GB907762: This could be expensive. At the very least you could use `NodeCache` as the schema will rarely change. 

I was thinking of using prompt caching but I think it's infeasible in this general approach?

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,278 @@
...
+  if (schema.type !== 'object' || !schema.properties) {
+    return data
+  }
+
+  const result: Record<string, unknown> = {}
+  const props = schema.properties as Record<string, JsonSchema & { widget?: string; $ref?: string }>
+
+  for (const [key, value] of Object.entries(data)) {
+    let propSchema = props[key]
+    if (!propSchema) continue
```
GB907762:
```suggestion
    if (!propSchema) {
      continue
    }
```
Same here

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,278 @@
...
+      if (resolved) {
+        propSchema = { ...resolved, ...propSchema, $ref: undefined } as typeof propSchema
+      }
+    }
+
+    if (propSchema.widget && EXCLUDED_WIDGETS.has(propSchema.widget)) continue
+
+    if (value === null || value === undefined) continue
+
+    if (isPlaceholderValue(value)) continue
```
GB907762: https://eslint.org/docs/latest/rules/curly#all We use this eslint rule for readability.
```suggestion
    if (propSchema.widget && EXCLUDED_WIDGETS.has(propSchema.widget)) {
      continue
    }
  
    if (value === null || value === undefined) {
      continue
    }
  
    if (isPlaceholderValue(value)) {
      continue
    }
```

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,278 @@
...
+        .map((item) => {
+          if (itemSchema?.type === 'object' && typeof item === 'object' && item !== null) {
+            return stripUnknownKeys(item as Record<string, unknown>, itemSchema, root)
+          }
+          if (itemSchema?.enum && !itemSchema.enum.includes(item)) {
+            return null
+          }
+          return item
+        })
+        .filter((item) => item !== null && item !== undefined && !isPlaceholderValue(item))
```
GB907762:
```suggestion
        .filter((item) => !isPlaceholderValue(item))
```
isPlaceholderValue already deals with these types.

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,278 @@
...
+    } else if (propSchema.type === 'boolean' && typeof value === 'boolean') {
+      result[key] = value
+    }
+  }
+
+  return result
+}
+
+function isPlaceholderValue(value: unknown): boolean {
+  if (typeof value !== 'string') return false
```
GB907762:
```suggestion
  if (typeof value !== 'string') {
    return false
  }
```

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,278 @@
...
+    const date = new Date(value.trim())
+    if (isNaN(date.getTime())) {
+      return true
+    }
+  }
+  return false
+}
+
+function buildPromptMessages(fields: FieldDescriptor[], text: string): ChatMessage[] {
+  const systemMessage = `You are a data extraction assistant. You will be given a model card text document and a target JSON schema. Extract information from the document into the schema format.
```
GB907762: Have you considered how different models may interpret this system prompt. I think this would be best configurable.

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,278 @@
...
+
+CRITICAL RULES:
+- Only extract information that is EXPLICITLY stated in the source document.
+- If a field cannot be populated from the document, omit it entirely from the output.
+- Do NOT infer, guess, or hallucinate any values.
+- Do NOT generate placeholder text like "Not specified", "N/A", or "Unknown" — leave the field out.
+- Do NOT generate placeholder or example URLs. Only include a URL if it appears verbatim in the source document.
+- Do NOT infer or generate dates. Only populate date fields if an explicit date is clearly stated in the source document for that specific purpose.
```
GB907762: There's a weird bug, that when you explicitly change the date the it is not populated in the edit field. 
But when saved it is populated.

I think there's some problem with the widget in the frontend.

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  const cleaned = stripUnknownKeys(parsed, schema.jsonSchema)
+
+  const validationResult = new Validator().validate(cleaned, schema.jsonSchema)
+  if (validationResult.errors.length > 0) {
+    log.warn(
+      { modelId, errors: validationResult.errors.map((e) => `${e.property}: ${e.message}`) },
+      'Extracted data has validation issues, stripping invalid fields.',
+    )
+  }
```
GB907762, line 429:
```suggestion
    const { valid, errors } = validateContentAgainstSchema(cleaned, schema.jsonSchema)
    if (!valid) {
      throw BadReq('Extracted data has validation issues, stripping invalid fields.', {
        errors,
      })
  }
```
This snippet should feed this to the user via a `400` with the validation error's of the schema. Instead of just logging it to the backend.

backend/src/utils/config.ts
```diff
@@ -239,6 +239,15 @@ export interface Config {
...
   }
+
+  llm: {
+    enabled: boolean
+    endpoint: string
+    apiKey: string
+    model: string
+    maxTokens: number
+    timeoutMs: number
+  }
```
GB907762, line 247: There's some stuff that is separately configured in HELM that are configured using a configmap. If you add the following to these files, the custom variables should be mounted correctly to the pod on deployment.

`infrastructure/helm/bailo/values.yaml`
```yaml
...
    untrustedModel:
      enabled: false
      untrustedModelDescription: 'These are private only models.'

    llmImport:
      enabled: false
...
      mountPath: /certs/artefactscan-ca.pem
  env: # Override settings here
    # TRIVY_DB_INSECURE: false
    # TRIVY_DB_HOSTNAME: ghcr.io

llm:
  enabled: false
  endpoint: ''
  model: ''
  maxTokens: 16384
  timeoutMs: 120000
```

*Note: Sensitive credentials such as `apiKey` should be mounted to the pod as an environment variable and configured using `custom-environment-variables.cjs`*

infrastructure/helm/bailo/templates/bailo/bailo.configmap.yaml
```json
...
      untrustedModels: {
        defaultState: '{{ .Values.untrustedModels.defaultState }}'
      },

      llm: {
        enabled: {{ .Values.llm.enabled }},
        endpoint: '{{ .Values.llm.endpoint }}',
        apiKey: '{{ .Values.llm.apiKey }}',
        model: '{{ .Values.llm.model }}',
        maxTokens: {{ .Values.llm.maxTokens }},
        timeoutMs: {{ .Values.llm.timeoutMs }},
      }
    }
...
                secretAccessKey: 'KMS_ACCESS_KEY',
              },
            },
          },
        },
      },
      llm: {
        apiKey: LLM_API_KEY,
      }
    }
```

infrastructure/helm/bailo/templates/bailo/bailo.deployment.yaml
```yaml
...
            {{- if .Values.inference.enabled }}
            - name: INFERENCING_WRITE_TOKEN
              valueFrom:
                secretKeyRef:
                  name: {{ include "bailo.fullname" . }}-inference
                  key: auth-token
            {{- end }}
            {{- if .Values.llm.enabled }}
            - name: LLM_API_KEY
              valueFrom:
                secretKeyRef:
                  name: {{ include "bailo.fullname" . }}-llm-api
                  key: auth-token
            {{- end }}
...
```

backend/src/routes.ts
```diff
@@ -15,7 +15,13 @@ import { httpLog } from './services/log.js'
 
 export const server = express()
 
-server.use([bodyParser.json(), httpLog, helmet()])
+const LARGE_BODY_ROUTES = ['/api/v2/model/', '/import-model-card-text']
+const jsonParser = (req, res, next) => {
+  const needsLargeBody = LARGE_BODY_ROUTES.every((segment) => req.path.includes(segment))
+  const limit = needsLargeBody ? '500kb' : '100kb'
```
GB907762, line 21: Could you clarify why this is neccessary. I think a `100kb` prompt is reasonable. Large requests, especially on this could trigger denial of service or consumption of excessive tokens.

frontend/pages/docs/administration/getting-started/app-configuration.mdx
```diff
@@ -60,4 +49,32 @@ When deploying using `helm`, configuration is primarily handled through `values.
 `helm/bailo/templates/bailo/bailo.configmap.yaml`. Configuration is loaded when the application starts. The application
 must be restarted when configuration changes.
 
+## LLM Import Configuration
+
+Bailo can optionally use an LLM to extract model card data from free text (e.g. HuggingFace model cards). This feature
```
GB907762: We have the ability to search through hugging face models on the market place. 

Maybe something for the future, but being able to click on a hugging face model from the market place and import it would be quite nice.

frontend/src/entry/overview/ImportModelCardTextDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  function handleClose() {
+    if (!loading) {
+      setErrorMessage('')
+      onClose()
+    }
+  }
+
+  return (
+    <Dialog maxWidth='md' fullWidth open={open} onClose={handleClose} TransitionComponent={Transition}>
```
GB907762:
```suggestion
    <Dialog maxWidth='md' fullWidth open={open} onClose={handleClose} slots={{ transition: Transition }}>
```
There's a deprecation warning on this.

frontend/src/entry/overview/ImportModelCardTextDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <DialogTitle>Import Model Card from Text</DialogTitle>
+      <DialogContent sx={{ p: 2 }}>
```
GB907762:
```suggestion
      <DialogContent sx={{ p: 2, overflow: 'hidden' }}>
```
There's a double scroll bar on small windows. This will remove the one on the dialog but not on the text field.

backend/src/clients/llm.ts
```diff
@@ -0,0 +1,64 @@
+import fetch, { Response } from 'node-fetch'
+
+import log from '../services/log.js'
+import config from '../utils/config.js'
+import { ConfigurationError, InternalError } from '../utils/error.js'
+
+export interface ChatMessage {
+  role: 'system' | 'user' | 'assistant'
```
GB907762: Make this an enum for future use.

```suggestion
  export enum ChatMessageRole {
    USER = 'system',
    GROUP = 'user',
    ASSISTANT = 'assistant'
    
  } as const

  export type ChatMessageRoleKeys = (typeof ChatMessageRole)[keyof typeof ChatMessageRole]

export interface ChatMessage {
  role: ChatMessageRoleKeys
```

GB907762 (changes requested):
Thanks for your first contribution to Bailo!

This looks like a great feature, and something that should be valuable to people filling in model cards.

If there's anything in our issue tracker, or anything you think would be of interest to Bailo in future. Feel free to raise an issue or open another PR.

backend/src/clients/llm.ts
M422110: Yes making it configurable sounds sensible - I think best to make it deterministic if we can but could do some experimentation to see if we get better results.

M422110: Commited a change to make this configurable **https://github.com/gchq/Bailo/pull/3750/changes/8bbe9a9e4ce28e4e72bceb21fcc4bfb3bd5f4cac**

backend/src/services/modelCardImport.ts
M422110: Ive added some backend tests for this file - https://github.com/gchq/Bailo/pull/3750/commits/bfc78dcd43d6c879fbf1152055d9154cb94d34e2

backend/src/services/modelCardImport.ts
M422110: run Eslint across the board and fixed in this commit: https://github.com/gchq/Bailo/pull/3750/commits/f8a08f214968c5dfdaf626efed856c6fbc622e38

backend/src/routes.ts
M422110: When I tested with a very large source file - such as the claude opus system card document - 

https://www-cdn.anthropic.com/0b4915911bb0d19eca5b5ee635c80fef830a37ea/Claude%20Opus%204.8%20System%20Card.pdf 

the 100kb limit wasn't enough, I decided to apply only to a small set of routes to reduce the risk of denial of service.

backend/src/clients/llm.ts
M422110: These roles in the openai spec are pretty common, but happy to apply this

backend/src/routes.ts
GB907762: Sounds good to me!

backend/src/clients/llm.ts
M422110: https://github.com/gchq/Bailo/pull/3750/commits/bfc78dcd43d6c879fbf1152055d9154cb94d34e2

backend/src/routes/v2/model/modelcard/postImportModelCardText.ts
M422110: https://github.com/gchq/Bailo/pull/3750/commits/3189d16559c8a1fc60b69faccf7a6e1a2f6819c9

backend/src/services/modelCardImport.ts
M422110: https://github.com/gchq/Bailo/pull/3750/commits/cfc2c179d06f5dc97dddef80d549ff2246d4541f

backend/src/services/modelCardImport.ts
GB907762: I think it's not very forgiving since an enum needs to be an exact string. The LLM doesn't seem to pick up on the possible strings, correctly.

backend/src/services/modelCardImport.ts
IR96334, line 1: The functions: describeObjectFields, buildSchemaDescription, resolveRef, stripUnknownKeys, and isPlaceholderValue are all quite complex and dense. I would like more context through comments, more descriptive/apt variable names. This should make it more readable and maintainable for the future

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,290 @@
...
+      }
+    } else if (propSchema.type === 'array' && Array.isArray(value)) {
+      let itemSchema = propSchema.items as (JsonSchema & { $ref?: string }) | undefined
+      if (itemSchema?.$ref) {
+        const resolved = resolveRef(itemSchema.$ref, root)
+        if (resolved) {
+          itemSchema = { ...resolved, ...itemSchema, $ref: undefined } as typeof itemSchema
+        }
+      }
+      const cleaned = value
```
ON93754: This function is getting very long, high cognitive load - Could move this logic out to a separate function, something like:

```
/**
* A docstring explaining what this function does
*/
function cleanValue(rawValue: sometype): somereturntype {
    return rawValue.map((item) => {
          if (itemSchema?.type === 'object' && typeof item === 'object' && item !== null) {
            return stripUnknownKeys(item as Record<string, unknown>, itemSchema, root)
          }
          if (itemSchema?.enum && !itemSchema.enum.includes(item)) {
            return null
          }
          return item
        })
        .filter((item) => !isPlaceholderValue(item))
}
```

This gives an added benefit of making it easy to write some unit tests for this if we wanted additional confidence that it does what is expected

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,290 @@
...
+  }
+  const normalised = value.trim().toLowerCase()
+  if (
+    ['n/a', 'not specified', 'not available', 'unknown', 'none', 'tbd', 'to be determined', 'not applicable'].includes(
+      normalised,
+    )
+  ) {
+    return true
+  }
+  if (/^https?:\/\/(example\.com|www\.example\.|placeholder|your-|insert-|link-here)/i.test(value.trim())) {
```
ON93754: A comment here would be useful, i'd even consider moving this out to a utility function with a helpful name, such as:

```
/**
* A docstring explaining what this does
*/
function isPlaceholderUrl(value: string): boolean {
    return /^https?:\/\/(example\.com|www\.example\.|placeholder|your-|insert-|link-here)/i.test(value.trim())
}
```

Then update this line to:
`if isPlaceholderUrl(value) {`

backend/src/services/modelCardImport.ts
```diff
@@ -0,0 +1,290 @@
...
+    return true
+  }
+  if (/^\d{4}-\d{2}-\d{2}$/.test(value.trim())) {
```
ON93754: Same for this, that regex needs a comment at least, or perhaps a function such as:

```
function isIsoDateString(value: string): boolean {
  return /^\d{4}-\d{2}-\d{2}$/.test(value.trim());
}
```

Then update this line to:
`if isIsoDateString(value) {`

backend/src/services/modelCardImport.ts
ON93754: Optionally you could then write some unit tests for `isPlaceholderUrl` to be confident it works as expected

backend/src/services/modelCardImport.ts
M422110: Sounds like we may need to improve the system prompt to handle the enum matching correctly.

backend/src/services/modelCardImport.ts
M422110: Yes making the system prompt configurable makes sense rather than needing a code change

backend/src/services/modelCardImport.ts
M422110: https://github.com/gchq/Bailo/pull/3750/commits/cd12fc24fb96e7e17040b1c384272f619877f89d 

Added a node cache with a TTL of 1 hour.

backend/src/services/modelCardImport.ts
M422110: The root cause was buildSchemaDescription didn't resolve $ref references, and so the options were never passed to the LLM. Have also added some additional structure to the prompt and made the matching case insensitive to ensure we dont drop matches. Tested and seems to work better now

https://github.com/gchq/Bailo/pull/3750/commits/0b1a9f92d178b85fd9b4a64846cca95c568eacfb

GB907762: Yep that sounds great. During testing I did see it wasn't matching lower case enums. This looks like an improvement ❤️

GB907762:
This looks like it's coming along nicely!

I've just ran some Github actions. There's some failing styles within the docs. 

You can fix this with `npm run style` in the root directory.

Alternatively we have a `pre-commit` when you install husky via `npm i`. That will do this automatically whenever you commit a change.

backend/src/services/modelCardImport.ts
M422110: Added in commit here - https://github.com/gchq/Bailo/pull/3750/commits/7012e8a8caa7f774740ee8348eea8f03bba3c009

backend/src/services/modelCardImport.ts
M422110: This seems to be a quirk in the picker added this fix which should mean that it refreshes on the value update:

https://github.com/gchq/Bailo/pull/3750/commits/5fc2e018df4f21e2a81ba466e088cacd19fe19df

GB907762: Awesome

backend/src/services/modelCardImport.ts
M422110: Im not sure I agree that we should fail the entire process, its likely in this scenario many of the fields in the schema have been mapped correctly  (such as the long text fields)- doing a lot of the heavy lifting for the importer. And some field likely to be where we need to map to a type like a date, or perhaps an enum has not been set correctly. Failing the whole thing would mean the user would have to fall back to filling in the entire card. Perhaps a better solution would be rather than failing with a 400, we return the cleaned data and include the validation warning in the response?

M422110: This now returns validation errors in the response message back to the user but will not fail the entire request:
https://github.com/gchq/Bailo/pull/3750/commits/915f2bdba3e45fef99f54e74adc8b152a14e6b9e

GB907762: Yep I'm happy with this approach. This is a good suggestion.

backend/src/utils/config.ts
M422110: Added the helm config:

https://github.com/gchq/Bailo/pull/3750/commits/d84ca920375ba45136203f5b384ceeb95a3f2ec7

backend/src/services/modelCardImport.ts
M422110: Ive added lots more comments to this file and also refacored the variable names to be more descriptive

https://github.com/gchq/Bailo/pull/3750/commits/97bcb1a4edaa4b10a42eca199d1fc0b51b88ff0b

backend/src/services/modelCardImport.ts
M422110: https://github.com/gchq/Bailo/pull/3750/commits/6a0a9b78eb44d9faf777b4efd28702c454a58138

backend/src/services/modelCardImport.ts
M422110: https://github.com/gchq/Bailo/pull/3750/commits/6a0a9b78eb44d9faf777b4efd28702c454a58138

backend/src/services/modelCardImport.ts
M422110: Ive tried to refactor to pull out some helper methods, and since the review comment have also refactored to include more comments/more descriptive variable names:

https://github.com/gchq/Bailo/pull/3750/commits/5e94640801387f2a188da1dd8747565026b30e58

frontend/src/entry/overview/ImportModelCardTextDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <TextField
+          size='small'
+          placeholder='Paste model card text here...'
+          multiline
+          minRows={12}
+          maxRows={20}
+          value={text}
+          onChange={(e) => setText(e.target.value)}
+          disabled={loading}
+          fullWidth
```
GB907762:
```suggestion
    <Dialog
      maxWidth='md'
      fullWidth
      open={open}
      onClose={handleClose}
      slots={{ transition: Transition }}
      sx={{ '& .MuiDialog-paper': { height: '100%' } }}
    >
      <DialogTitle>Import Model Card from Text</DialogTitle>
      <DialogContent sx={{ p: 2, overflow: 'hidden', display: 'flex', flexDirection: 'column' }}>
        <Typography variant='body2' sx={{ mb: 2 }}>
          Paste the contents of an existing model card (e.g. from HuggingFace, a system card, or other documentation).
          An LLM will extract relevant information into your schema fields.
        </Typography>
        <Alert severity='warning' sx={{ mb: 2 }}>
          You must review any information provided for correctness. AI-extracted data may be incomplete or inaccurate.
        </Alert>
        <TextField
          size='small'
          placeholder='Paste model card text here...'
          multiline
          value={text}
          onChange={(e) => setText(e.target.value)}
          disabled={loading}
          fullWidth
          sx={{
            flex: '1 1 auto',
            overflow: 'hidden',
            '& .MuiInputBase-root': { height: '100%' },
            '& .MuiInputBase-input': { height: '100% !important', overflow: 'auto !important' },
          }}
```
There's still some problems on small windows. Error's seem to be covered by the text block.

frontend/pages/docs/administration/getting-started/app-configuration.mdx
```diff
@@ -60,4 +49,32 @@ When deploying using `helm`, configuration is primarily handled through `values.
...
+```javascript
+// In your configuration override (e.g. production.cjs or local.js)
+{
+  llm: {
+    enabled: true,
+    endpoint: 'https://api.openai.com/v1/chat/completions', // Any OpenAI-compatible endpoint
+    apiKey: 'your-api-key',
+    model: 'gpt-4o',          // Model identifier for the LLM provider
+    maxTokens: 16384,         // Maximum tokens in the LLM response
+    timeoutMs: 120000,        // Timeout for LLM requests in milliseconds
```
GB907762:
```suggestion
    timeoutMs: 120000,        // Timeout for LLM requests in milliseconds
    temperature: 0            // Temperature for LLM provider
```

GB907762 (changes requested):
This is looking a lot more robust!

Just a couple of tweaks with regards accessibility and documentation.

frontend/pages/docs/administration/getting-started/app-configuration.mdx
GB907762: Maybe a small description about why a low temperature is preferred for model cards?

M422110: https://github.com/gchq/Bailo/pull/3750/commits/2a9f0b339ed05b4e80f93e7ddb868b3d61a390a0

GB907762 (dismissed):

backend/config/default.cjs
```diff
@@ -369,4 +373,32 @@ module.exports = {
   inference: {
     authorisationToken: '',
   },
+
+  llm: {
+    enabled: false,
```
PE39806: This value seems to be a duplicate of the above `ui.llmImport.enabled` which risks one of these being enabled while the other is disabled. I would prefer if this key (`llm.enabled`) is removed and the code only has `ui.llmImport.enabled` so that there is a single source of truth.

backend/src/routes/v2/model/modelcard/postImportModelCardText.ts
```diff
@@ -0,0 +1,59 @@
...
+  tags: ['modelcard'],
+  description: 'Extract model card metadata from free text using an LLM.',
+  schema: postImportModelCardTextSchema,
+  responses: {
+    200: {
+      description: 'Extracted model card metadata matching the schema structure.',
+      content: {
+        'application/json': {
+          schema: z.object({
+            metadata: z.unknown(),
```
PE39806: This doesn't match `PostImportModelCardTextResponse` below
```suggestion
            metadata: z.record(z.string(), z.unknown()),
            warnings: z.array(z.string()),
```
Ideally, also add some `.openapi({ example: 'foo' })` as well so that the API docs can show what the expected data could be.

frontend/src/entry/overview/SaveAndCancelFormButtons.tsx
```diff
@@ -29,6 +34,17 @@ export default function SaveAndCancelButtons({
         mb: { xs: 2 },
       }}
     >
+      {showImportFromText && openImportTextDialog && (
+        <Button
+          variant='contained'
+          startIcon={<AutoAwesomeIcon />}
+          color='secondary'
+          onClick={openImportTextDialog}
+          data-test='importModelCardTextButton'
```
PE39806, line 43: This is the same `data-test` value as in `frontend/src/entry/overview/ImportModelCardTextDialog.tsx` but these should really be unique.

frontend/pages/docs/administration/getting-started/app-configuration.mdx
```diff
@@ -60,4 +60,46 @@ When deploying using `helm`, configuration is primarily handled through `values.
...
+// for the llm and also schema that you are using for model cards.
+{
+  llm: {
+    enabled: true,
+    endpoint: 'https://api.openai.com/v1/chat/completions', // Any OpenAI-compatible endpoint
+    apiKey: 'your-api-key',
+    model: 'gpt-4o',          // Model identifier for the LLM provider
+    maxTokens: 16384,         // Maximum tokens in the LLM response
+    timeoutMs: 120000,        // Timeout for LLM requests in milliseconds
+    temperature: 0,            // Temperature for LLM provider - a low temprature will make the responses more deterministic
```
PE39806:
```suggestion
    temperature: 0,            // Temperature for LLM provider - a low temperature will make the responses more deterministic
```

PE39806 (changes requested):

backend/config/default.cjs
M422110: Yep agree its ambiguous at the moment whilst there is only 1 llm powered feature. https://github.com/gchq/Bailo/pull/3750/commits/5402789c665ee584dbb240f2b4bae4681a20556b

backend/src/routes/v2/model/modelcard/postImportModelCardText.ts
M422110: Updated the shema and added a .openapi example for the docs: https://github.com/gchq/Bailo/pull/3750/commits/02a96863e356db63ce4ebefa1555958f322bfcf4

frontend/src/entry/overview/SaveAndCancelFormButtons.tsx
M422110: Yes agree - updated to make it unique to avoid potential clashes. https://github.com/gchq/Bailo/pull/3750/commits/f6ed3f17e571971e7230abf089615e3d2a4d3c42

GB907762 (dismissed):

GB907762 (dismissed):

frontend/pages/docs/users/models/model-card/importing-model-card-text.mdx
PE39806, line 1: Please update this page to follow our [docs style-guide](https://github.com/gchq/Bailo/blob/151c23a4616a36e8fbf35128502d0eea3894b6e2/frontend/pages/docs/style-guide.mdx).

The main changes will be adding a "Common questions callout", a "How to use" section, a "What to expect" section, and (optionally) "Next steps" and/or "Related pages".
Please also replace any em-dashes `—` with regular hyphens `-` (something I need to add to the style guide).

PE39806 (changes requested):

frontend/pages/docs/users/models/model-card/importing-model-card-text.mdx
M422110: @PE39806 Docs updated to match style guide: https://github.com/gchq/Bailo/pull/3750/commits/9d1596e456addd01179944a1658704da9f6f28d1

PE39806: I'm afraid that the latest commit made changes to `frontend/pages/docs/administration/getting-started/app-configuration.mdx` rather than `frontend/pages/docs/users/managing-models-and-releases/upload-to-bailo/importing-model-card-text.mdx`. Please could you revert the latest commit and update `importing-model-card-text.mdx` instead?
Lastly, a new entry needs adding to `flatDirectory` in `frontend/pages/docs/directory.tsx` to add the new page `importing-model-card-text.mdx` to the docs tree to make it discoverable.

M422110: The users documentation is intended for users of the application isnt it? They wouldnt be the ones administrating the LLM configuration.

M422110: If we dont want it in the app-configuration page, we could have a seperate llm section within the admin pages?

frontend/pages/docs/administration/llm/llm-import.mdx
```diff
@@ -0,0 +1,74 @@
...
+  ui: {
+    llmImport: {
+      enabled: true,
+    },
+  },
+}
+```
+
+For Helm deployments, these values are set in `values.yaml` under `llm` and `config.ui.llmImport`. See
+[App Configuration](/docs/administration/getting-started/app-configuration) for details on how configuration inheritance
```
PE39806: We have to use relative imports for our docs as these are also hosted on the static GitHub pages. I will update our style guide with this.
```suggestion
[App Configuration](../app-configuration) for details on how configuration inheritance
```

frontend/pages/docs/administration/llm/llm-import.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+The extracted data is validated against the schema before being applied. Any fields that do not match the schema or
+contain placeholder values are automatically filtered out. Validation warnings are shown to the user so they can review
+the result before saving.
+
+## Related pages
+
+- [Importing Model Card Text](/docs/users/models/model-card/importing-model-card-text) - User guide for the import
+  feature
+- [App Configuration](/docs/administration/getting-started/app-configuration) - General application configuration
```
PE39806:
```suggestion
- [Importing Model Card Text](../../users/models/model-card/importing-model-card-text) - User guide for the import
  feature
- [App Configuration](../getting-started/app-configuration) - General application configuration
```

frontend/pages/docs/users/models/model-card/importing-model-card-text.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+- HuggingFace model card README files
+- AI system cards (for example Claude or GPT model cards)
+- Internal model documentation
+- Any structured or unstructured text describing a model
+
+## Related pages
+
+- [Model Card](/docs/users/models/model-card/model-card) - Filling in the model card and configuring access
+- [LLM Import](/docs/administration/llm/llm-import) - Enabling and configuring the LLM import feature
```
PE39806:
```suggestion
- [Model Card](model-card) - Filling in the model card and configuring access
- [LLM Import](../../../administration/llm/llm-import) - Enabling and configuring the LLM import feature
```

frontend/pages/docs/users/models/model-card/importing-model-card-text.mdx
```diff
@@ -0,0 +1,53 @@
...
+> - What text formats does the import feature support?
+> - What happens when I run the import?
+> - Why can I not see the "Import from Text" button?
+
+If your model already has documentation - such as a HuggingFace model card, a system card, or other written
+documentation - you can use Bailo's text import feature to automatically extract relevant information into your model
+card form.
+
+This feature requires an administrator to configure an LLM endpoint. If you do not see the **Import from Text** button,
+contact your Bailo administrator. See [LLM Import](/docs/administration/llm/llm-import) for setup instructions.
```
PE39806:
```suggestion
contact your Bailo administrator. See [LLM Import](../../../administration/llm/llm-import) for setup instructions.
```

PE39806 (changes requested):

frontend/pages/docs/directory.tsx
```diff
@@ -28,7 +28,12 @@ export const flatDirectory: Array<DirectoryEntry> = [
 
   { title: 'Models', slug: 'users/models', header: true },
   { title: 'Creating a Model', slug: 'users/models/creating-a-model' },
-  { title: 'Model Card', slug: 'users/models/model-card' },
+  { title: 'Model Card', slug: 'users/models/model-card', header: true },
+  { title: 'Model Card', slug: 'users/models/model-card/model-card' },
```
PE39806, line 32: Because of this change, `frontend/pages/docs/users/models/creating-a-model.mdx` needs the URLs `](model-card)` updated to be `](model-card/model-card)` (two instances).

PE39806 (changes requested):

frontend/pages/docs/directory.tsx
M422110: Also included in the commit to fix links

https://github.com/gchq/Bailo/pull/3750/commits/03a342702d248a35451d2ae1561be3ea990b0695

ON93754 (approved):

PE39806 (approved):

PR #3767, opened 2026-06-23
Improve python error handling for backend failures

PE39806:
- Update python core to correctly handle non-success responses from the backend
- Update unit and integration tests
- Update docs (including Jupyter notebooks) to show how to use the new error handling
- Bump package version for a new release

Note that this error handling is fully backwards compatible so is not a breaking change.

### DIFF
--- a/lib/python/src/bailo/core/client.py
+++ b/lib/python/src/bailo/core/client.py
@@ -5,6 +5,9 @@ from json import JSONDecodeError
 from typing import Any
 
 import requests
+
+# isort: split
+
 from bailo.core.agent import Agent, TokenAgent
 from bailo.core.enums import CollaboratorEntry, EntryKind, ModelVisibility, SchemaKind
 from bailo.core.exceptions import BailoException, ResponseException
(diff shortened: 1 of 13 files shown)


lib/python/src/bailo/core/client.py
```diff
@@ -24,6 +27,29 @@ def __init__(self, url: str, agent: Agent | None = None):
         self.url = url.rstrip("/") + "/api"
         self.agent = agent or Agent()
 
+    def _parse_json(self, res: requests.Response) -> dict[str, Any]:
```
GB907762: `self` looks like a redundant parameter. It isn't accessed. Would it be best to add something to `Agent.__request`

GB907762 (changes requested):

lib/python/src/bailo/core/client.py
PE39806: Good spot. I've opted to make this a `@staticmethod` rather than moving it into `Agent.__request`. This avoids changing behaviour for non-JSON responses (e.g. `get_download_file`) without needing to pass in any extra args or use extra helper functions which might risk this becoming a breaking change.

GB907762 (approved):

ON93754 (approved):

PR #3768, opened 2026-06-23
Add email on release approval

GB907762:
Release email is sent to everyone who has a system role on a model (not review roles). 
Release email is sent to everyone on a model with an approved access request.


backend/src/routes/v3/response/getLatestResponseForReview.ts
```diff
@@ -14,25 +14,28 @@ export const getLatestResponseSchema = z.object({
...
+              response: responseInterfaceSchema,
+            }),
+          },
         },
       },
     },
   },
-})
+  'v3',
+)
```
GB907762, line 38: While debugging. I noticed this route was on the `v2` swagger docs.

backend/src/services/smtp/smtp.ts
```diff
@@ -272,6 +284,29 @@ export async function startImportNotification(modelId: string) {
...
 }
 
+export async function notifyReleaseOnApproval(modelId: string, release: ReleaseDoc) {
+  if (!config.smtp.enabled) {
+    log.info('Not sending email due to SMTP disabled')
+    return
+  }
+
+  const emailContent = buildEmail(
+    `A new release is available`,
```
PE39806: I wonder if this should be reworded to "A new release has been approved" as the release was already available once created and before being reviewed?

backend/src/services/smtp/smtp.ts
```diff
@@ -272,6 +284,29 @@ export async function startImportNotification(modelId: string) {
...
+    [],
```
PE39806: From the email, it's not clear which model has the new release. Please could we add the model name and new release semver here? This way, users can immediately know if the new release is something they are interested in or not.

I appreciate that we send other emails without these details, but I think this is something we should be doing going forwards.

backend/test/services/smtp/smtp.spec.ts
```diff
@@ -106,9 +107,19 @@ const responseService = vi.hoisted(() => ({
       kind: 'review',
     }
   }),
+  checkAccessRequestsApproved: vi.fn(function () {
+    return true
+  }),
```
PE39806: This could be condensed to be shorter
```suggestion
  checkAccessRequestsApproved: vi.fn(() => true),
```
I understand that this line currently follows the same styling used in this file, which I think came from me making a misunderstanding in the last vitest major bump 😅

backend/test/services/smtp/smtp.spec.ts
```diff
@@ -106,9 +107,19 @@ const responseService = vi.hoisted(() => ({
...
 }))
 vi.mock('../../../src/services/response.js', async () => responseService)
 
+const AccessRequestModelMock = vi.hoisted(() => ({
+  find: vi.fn(function () {
+    return [] as any[]
+  }),
```
PE39806: As above
```suggestion
  find: vi.fn(() => [] as any[]),
```

backend/test/services/response.spec.ts
```diff
@@ -32,6 +35,9 @@ const smtpMock = vi.hoisted(() => ({
   notifyReviewResponseForRelease: vi.fn(function () {
     return Promise.resolve()
   }),
+  notifyReleaseOnApproval: vi.fn(function () {
+    return Promise.resolve()
+  }),
```
PE39806: As above
```suggestion
  notifyReleaseOnApproval: vi.fn(() => Promise.resolve())
```

backend/src/services/response.ts
```diff
@@ -168,11 +172,15 @@ export async function sendReviewResponseNotification(
         log.error({ review }, 'Unable to send notification for review response. Cannot find semver.')
         return
       }
-
       const release = await getReleaseBySemver(user, review.modelId, review.semver)
       notifyReviewResponseForRelease(reviewResponse, release).catch((error) =>
         log.warn({ error }, 'Error when notifying collaborators about review response.'),
       )
+      if ((await checkReleaseApproved(release.modelId, release.semver)) && !isApproved) {
```
PE39806: It may be that TS/JS optimises this already under the hood, but generally it's best to have the pre-computed values on the left for short-circuit evaluation
```suggestion
      if (!isApproved && (await checkReleaseApproved(release.modelId, release.semver))) {
```

PE39806 (changes requested):

backend/src/services/smtp/smtp.ts
GB907762: Required roles already get emailed when a release is created. This is a good clarification

backend/src/services/response.ts
GB907762: Yep good spot. Additionally makes it more readable I think.

PE39806 (approved):

JRB66955 (approved):

PR #3770, opened 2026-06-24
Add required by model state docs

GB907762:
Documentation for #3740
There's already some documentation in the admin section for setup.


frontend/pages/docs/users/models/model-card.mdx
PE39806, line 18: Add the new section to these common questions
```suggestion
> - What is the model card version history?
> - What are required model status fields?
```

frontend/pages/docs/users/models/model-card.mdx
```diff
@@ -35,6 +36,20 @@ schema.
...
+
+Some model cards will require you to fill in specific fields for different points in that model's lifecycle. This helps
+ensure the structured document is up to date and accurate, for consumers.
+
+To update to a new lifecycle status, select an option in status chips. This will highlight fields required by that
+status. Fill in all these fields on all pages.
+
+<Box sx={{ maxWidth: '75%' }}>
+  <Image alt='selecting user model roles' src={bailoUpdateModelCardForProduction} />
+</Box>
```
PE39806, line 50: Add a plain-text caption sentence for the image as per https://github.com/gchq/Bailo/blob/main/frontend/pages/docs/style-guide.mdx#images

PE39806 (changes requested):

frontend/pages/docs/users/models/model-card.mdx
GB907762: Good spot 😄

PE39806: This needs to be separate from the `alt` text so that a RAG model can capture detail about the image. I think that RAG models don't contextualise from `alt` text very well.
For example,
https://github.com/gchq/Bailo/blob/b467c60a0422f40cb2d230012b770ef7050e4db1/frontend/pages/docs/users/models/model-card.mdx#L48-L53

PE39806 (changes requested):

PE39806 (approved):

JRB66955 (approved):

PR #3775, opened 2026-06-24
changes to model details

ARADDCC002:


frontend/src/entry/LastReviewOverviewDetails.tsx
```diff
@@ -26,12 +26,15 @@ export default function LastReviewOverviewDetails({ reviewId }: LastReviewOvervi
...
       <Typography>
-        {response ? formatDateStringAsDayMonthAndYear(response.createdAt.toString()) : 'Invalid date'}
+        {response ? (
+          <Stack direction='row' spacing={1}>
+            <div>{formatDateStringAsDayMonthAndYear(response.createdAt.toString())} by</div>
+            <UserDisplay dn={response.entity} showIcon />
+          </Stack>
+        ) : (
+          <em>Error fetching review</em>
+        )}
```
GB907762, line 36: I think because this information pertains the model card, it might be better placed at the top of the model card. Instead of on the details. As in `Last reviewed by Joe Blogs on 01/01/1970`?

Additionally I think `Next review due:` is irrelevant to a consumer. So maybe a warning at the top of the screen (for expiry) if you are an owner, in the same way we have ones to intact a schema migration might be better.

frontend/src/entry/EntryOverviewDetails.tsx
```diff
@@ -192,18 +184,35 @@ export default function EntryOverviewDetails({ entry }: OrganisationAndStateDeta
...
+                  setReviewHistoryOpen(true)
+                }}
+              >
+                History
+              </Button>
             </Stack>
             {archivedReviews.length > 0 && <LastReviewOverviewDetails reviewId={archivedReviews[0]._id} />}
           </Stack>
         )}
+        <Stack spacing={1} sx={{ width: { sm: '100%', md: 'max-content' } }}>
```
GB907762:
```suggestion
        <Stack spacing={1}>
```
I know the styling's copy pasted, but I think it's redundant?

GB907762 (changes requested):

frontend/src/entry/LastReviewOverviewDetails.tsx
ARADDCC002: I think it takes up too much real estate at the top. The top of the page should really be reserved for the most important information in my opinion, and because the review information is only of interest to either a few people depending on circumstance I think it's better inside `Model details`

JRB66955 (dismissed):

frontend/src/entry/EntryOverviewDetails.tsx
```diff
@@ -178,32 +144,49 @@ export default function EntryOverviewDetails({ entry }: OrganisationAndStateDeta
...
-                <Typography>
+                <Typography sx={{ pr: 1 }}>
                   {reviews[0].dueDate
                     ? formatDateStringAsDayMonthAndYear(reviews[0].dueDate.toString())
                     : 'Invalid date'}
                 </Typography>
               )}
               {updateEntryPermission.hasPermission && reviews[0] && (
                 <Button
-                  variant='outlined'
```
PE39806, line 195: I think I prefer this remaining `outlined` as it then aligns with the style used for both releases and access requests.

PE39806 (changes requested):

PE39806 (dismissed):

JRB66955 (dismissed):

JR40159 (changes requested):
- Could we change the title of the section in Model details to be "Model card review"? Seems better suited for when no review date has been set
- The "Set review date" and "History" buttons aren't consistent in terms of size
- I don't think there's a case where you should be able to be able to set the date of a review and view the history of the reviews at the same time in the model details
- Could we avoid having to have a dialog box for setting the date by having the date picker in the model details box with a confirmation button next to it?

JR40159:
Could you also fix the behaviour of the tags section when no tags exist and the user is not listed on the model?

ARADDCC002:
> * Could we change the title of the section in Model details to be "Model card review"? Seems better suited for when no review date has been set
> 
>     * The "Set review date" and "History" buttons aren't consistent in terms of size
> 
>     * I don't think there's a case where you should be able to be able to set the date of a review and view the history of the reviews at the same time in the model details
> 
>     * Could we avoid having to have a dialog box for setting the date by having the date picker in the model details box with a confirmation button next to it?

1. Resolved
2. I've removed the `History` button when there's no archived reviews
3. ^
4. Resolved

JR40159 (approved):

GB907762 (approved):

PR #3776, opened 2026-06-24
Feature/bai 2554 add argon2 support

ON93754:
- Moved hashing logic out to its own service file
- Created tests for hashing logic
- Updated the creation of tokens to make secret keys far more complex


backend/src/services/hash.ts
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Deserialises a stored "saltHex:hashHex" string into salt and hash buffers.
+ */
+export function fromSecretKey(stored: string): { salt: Buffer; hash: Buffer } {
+  const [saltHex, hashHex] = stored.split(':')
+  return {
+    salt: Buffer.from(saltHex, 'hex'),
+    hash: Buffer.from(hashHex, 'hex'),
+  }
+}
```
GB907762: Unused

backend/src/services/token.ts
```diff
@@ -15,15 +15,15 @@ interface CreateTokenProps {
   actions: Array<string>
 }
 
-const nanoid = customAlphabet('0123456789ABCDEFGHIJKLMNOPQSRTUVWXYZ')
+const nanoid = customAlphabet('0123456789ABCDEFGHIJKLMNOPQSRTUVWXYZabcdefghijklmnopqrstuvwxyz')
```
GB907762:
```suggestion
const nanoid = customAlphabet('0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz')
```

backend/src/services/token.ts
```diff
@@ -15,15 +15,15 @@ interface CreateTokenProps {
...
 export async function createToken(user: UserInterface, { description, scope, modelIds, actions }: CreateTokenProps) {
   if (user.token) {
     // Prevent escalating token privileges
     throw Forbidden('A token cannot be used to create another token', { accessKey: user.token.accessKey })
   }
 
   const accessKey = `BAC_${nanoid(8)}`
```
GB907762: I think we should use the current alphabet for consistency.

backend/src/services/hash.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Generates an Argon2id hash for the given key using a provided or random salt.
+ * Returns both the salt and computed hash.
+ */
+export function createArgon2Hash(key: string, salt?: Buffer) {
+  const nonce = salt ?? randomBytes(16)
+
+  const hash = argon2Sync('argon2id', {
```
GB907762: I think `argon2Sync` would block the event loop on the server leading to denial of service?

It's an expensive task.

backend/src/services/hash.ts
```diff
@@ -0,0 +1,120 @@
+import bcrypt from 'bcryptjs'
+import { argon2Sync, createHash, randomBytes, timingSafeEqual } from 'crypto'
+
+import { BadReq } from '../utils/error.js'
+
+export const HashType = {
+  argon2: 'argon2',
```
GB907762:
```suggestion
  ARGON2: 'argon2',
```

backend/src/models/Token.ts
```diff
@@ -117,53 +118,34 @@ TokenSchema.pre('save', function userPreSave() {
...
+  return new Promise((resolve) => {
     if (!this.secretKey) {
       resolve(false)
       return
     }
 
-    if (this.hashMethod === HashType.Bcrypt) {
-      bcrypt.compare(candidateToken, this.secretKey, (err: Error | null, result: boolean | undefined) => {
-        if (err) {
-          reject(err)
```
GB907762, line 151: This seems somewhat critical. Could we keep this?

backend/src/services/hash.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  return timingSafeEqual(hash, expectedHash)
+}
+
+/**
+ * Creates a bcrypt hash for the provided key using the configured cost factor.
+ * Throws if hashing fails.
+ */
+export function createBcryptHash(key: string): string {
+  const hash = bcrypt.hashSync(key, BCRYPT_CONFIG.rounds)
```
GB907762: I think just use the async version here.

backend/src/services/hash.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  if (!hash) {
+    throw BadReq('Unable to create token')
+  }
```
GB907762: Should never be triggered and I think this error is vague.
```suggestion
```

GB907762 (changes requested):

backend/src/services/hash.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  return `${nonce.toString('hex')}:${hash.toString('hex')}`
+}
+
+/**
+ * Verifies a key against a stored Argon2 hash formatted as "salt:hash".
+ * Recomputes the hash using the extracted salt and compares using a timing-safe check.
+ */
+export async function verifyArgon2Hash(key: string, stored: string) {
+  const [saltHex, hashHex] = stored.split(':')
```
GB907762, line 47: I'm a little worried about inconsistent data between the salt and the hash. For example what happens on an off chance the data doesn't have a `:` because of some sort of failure or edge case. Could we have some catches?

GB907762 (changes requested):

JRB66955 (dismissed):

backend/src/utils/id.ts
```diff
@@ -1,11 +1,15 @@
 import { customAlphabet } from 'nanoid'
 import { v4 as generateUuidv4 } from 'uuid'
 
+const shortAlphabetCharacters = '0123456789abcdefghijklmnopqrstuvwxyz'
```
GB907762:
```suggestion
const shortAlphabetCharacters = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
```
Sorry missed this. Otherwise looks good.

GB907762 (changes requested):

GB907762 (approved):

JRB66955 (approved):

PR #3780, opened 2026-06-25
Fix docs broken relative links

PE39806:
- Fix GitHub Pages docs routing
- Also fixes some links to Python Notebook files

Supersedes #3779


frontend/pages/docs/administration/federation/peer-integration.mdx
GB907762: I think this doesn't resolve to the correct path on Github Pages

GB907762 (changes requested):

JRB66955 (approved):

GB907762 (approved):

PR #3786, opened 2026-06-26
Updated frontend to eslint 10

ARADDCC002:


ARADDCC002:
It seems we don't need to use the `eslint-plugin-react` dependency, and as it hasn't been updated to use `eslint` v10, I think it's safe to remove it so we can upgrade.

JRB66955 (approved):

GB907762:
> It seems we don't need to use the `eslint-plugin-react` dependency, and as it hasn't been updated to use `eslint` v10, I think it's safe to remove it so we can upgrade.

Nice spot that we don't use this.

I think this is a useful plugin. I often see deprecation warnings or no keys on maps on some components. Are these things worth adding?

frontend/package.json
```diff
@@ -60,7 +60,7 @@
     "uuid": "^14.0.0"
   },
   "devDependencies": {
-    "@eslint/js": "^9.39.2",
+    "@eslint/js": "10.0.1",
```
PE39806, line 63: Do we want this (and other dependencies) to be explicit or should they have a carat `^` as some of the other dependencies do?

ARADDCC002: We've been burned in the past where minor/patch updates have caught us out with linting when new rules are introduced. 

I can always add it back in for now, though.

PE39806 (approved):

GB907762 (approved):

PR #3787, opened 2026-06-29
Added a new admin panel

ARADDCC002:
Reduce the side-bar buttons by one by creating a new page for admins. Instead of having both review roles and schema buttons on the side-bar, there's now just one for admins.

In the future, we can put more content on this page without needing to further increase the number of links on the side-bar.


JRB66955 (dismissed):

frontend/cypress/e2e/bailo/accessibility.cy.ts
```diff
@@ -37,17 +37,6 @@ describe('Check A11y violations', () => {
     cy.checkA11y(undefined, undefined, printA11yViolations)
   })
 
-  it('Check A11y violations, Schemas Page', () => {
```
IR96334, line 40: Should there be a subsequent Admins Page check?

frontend/pages/admin.tsx
```diff
@@ -0,0 +1,114 @@
...
+  )
+}
+
+export type AdminSectionKeys = (typeof AdminSection)[keyof typeof AdminSection]
+
+export default function Admin() {
+  const router = useRouter()
+  const { section } = router.query
+
+  const [selectedSection, setSelectedSection] = useState<AdminSectionKeys>(AdminSection.SCHEMA_LIST)
```
IR96334: Does adding:
`isAdminSection(section) ? section : AdminSection.SCHEMA_LIST`
eliminate the need for the useEffect? It seems the keep the same functionality

frontend/cypress/e2e/bailo/accessibility.cy.ts
ARADDCC002: The e2e tests don't currently have admin permissions so can't see them - so I'm not removing functionality here, I assume we were previously just testing the 403 page inadvertently

frontend/pages/admin.tsx
ARADDCC002: Defaulting to schema list for now and not worrying about the routing. Can be redone in a future ticket as it wasn't working as expected before.

frontend/pages/admin.tsx
```diff
@@ -0,0 +1,90 @@
...
+                selected={selectedSection === AdminSection.SCHEMA_LIST}
+                onClick={() => handleListItemClick(AdminSection.SCHEMA_LIST)}
+                icon={<Schema color={selectedSection === AdminSection.SCHEMA_LIST ? 'secondary' : 'inherit'} />}
+              >
+                Schema list
+              </SimpleListItemButton>
+              <SimpleListItemButton
+                selected={selectedSection === AdminSection.COMPARE}
+                onClick={() => handleListItemClick(AdminSection.COMPARE)}
+                icon={<Compare color={selectedSection === AdminSection.COMPARE ? 'secondary' : 'inherit'} />}
```
IR96334: This icon is slightly crowded for my tastes. Maybe stick to CompareArrows
`import CompareArrowsIcon from '@mui/icons-material/CompareArrows';`

frontend/src/schemas/SchemaTab.tsx
IR96334: Doesn't seem to work. Doesn't change with the selectedCategory state and continues to exist in unrelated tabs.

frontend/src/schemas/SchemaTab.tsx
```diff
@@ -22,26 +22,20 @@ export default function SchemaTab() {
...
+        <MenuItem value={kind} key={kind}>
           {camelCaseToTitleCase(kind)}
-        </SimpleListItemButton>
+        </MenuItem>
       )),
-    [handleListItemClick, selectedCategory],
+    [],
   )
 
+  const handleOrganisationSelectOnChange = useCallback((event: SelectChangeEvent) => {
```
IR96334: Change name, maybe to: `handleSchemaChange`

frontend/src/schemas/SchemaTab.tsx
IR96334: With these state alongside the query param there are 2 sources of truth that must be kept in-sync. Instead, we can remove the state entirely and replace with something like: `const category: string = (router.query?.category as string | undefined) ?? SchemaKind.MODEL`.
This allows us to remove the useEffect and have the select handler instead: `router.replace({
      query: { ...router.query, category: event.target.value },
    })`

frontend/src/schemas/SchemaTab.tsx
IR96334: Does this need to be a useCallback?

frontend/src/schemas/SchemaTab.tsx
ARADDCC002: I've removed for now. I don't think we really need it as it's just the admin panel.

ARADDCC002: As posted above I've removed for now.

frontend/src/schemas/SchemaTab.tsx
ARADDCC002: I don't think so as I don't think it was will cause re-render issues

IR96334 (dismissed):

PE39806 (changes requested):
This fails our 300% zoom a11y statement by adding a horizontal scrollbar.

ON93754 (dismissed):

IR96334 (approved):

PE39806 (approved):

PR #3817, opened 2026-07-01
Add diff to history

GB907762:


frontend/src/MuiForms/CustomTextInput.tsx
```diff
@@ -51,19 +52,48 @@ export default function CustomTextInput({
   }
 
   const mirroredState = getMirroredState(id, registry.formContext)
+  const compareFromState = getCompareFromState(id, registry.formContext) as string | undefined
+  const compareFromMirroredState = getCompareFromMirroredState(id, registry.formContext) as string | undefined
+  const inCompareMode = !!registry.formContext.compareMode && !registry.formContext.editMode
+
+  if (inCompareMode && !registry.formContext.mirroredModel) {
+    const from = compareFromState !== undefined ? compareFromState : (mirroredState as string | undefined)
```
IR96334: think can replace x ? x : y with x ?? y

IR96334:
The diff is (seemingly) only available for text-based widgets, and not other ones. Users will likely expect such a page to show changes for all widget types, not just the textual changes. Does this not make the page misrepresentative/confusing to any user who would assume it relates to all questions?

frontend/src/entry/overview/EntryCardHistoryDialog.tsx
```diff
@@ -12,33 +12,133 @@ import {
...
+        const checked = isSelected(snapshot)
+        const hideCheckbox = selected.length >= MAX_SELECTED && !checked
+        return (
+          <EntryCardRevision
+            key={snapshot.key}
+            snapshot={snapshot}
+            entryKind={entry.kind}
+            isChecked={checked}
+            hideCheckbox={hideCheckbox}
+            onRowClick={() => {
```
IR96334, line 134: Bug: When clicking a row or checkbox after already selecting 2 revisions, it seems to reroute you.

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,453 @@
...
+
+  return (
+    <Container>
+      <Paper sx={{ p: 4, my: 4 }}>
+        <Stack direction='row' spacing={1} sx={{ alignItems: 'center', mb: 2 }}>
+          <Button startIcon={<ArrowBackIosIcon />} onClick={() => router.push(backHref)}>
+            {`Back to ${kindLabel}`}
+          </Button>
+        </Stack>
+        <Accordion defaultExpanded={!isSameRevision} sx={{ mb: 3 }}>
```
IR96334: Don't feel this is large enough or "in the way" as to require an Accordion. Plus the arrow to expand is small and very miss able.

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,453 @@
...
+                    value={fromMirroredOption}
+                    getOptionLabel={renderVersionLabel}
+                    isOptionEqualToValue={(option, value) => option.version === value.version}
+                    onChange={(_event, value) => setVersion(QueryDiffSide.From, QueryCardType.Mirror, value?.version)}
+                    renderInput={(params) => <TextField {...params} label='Mirrored Version' />}
+                  />
+                )}
+              </Stack>
+              <Stack sx={{ justifyContent: 'center', alignItems: 'center' }}>
+                <CompareArrowsIcon color='primary' fontSize='large' />
```
IR96334: This needs to be centered.

frontend/src/entry/overview/EntryCardRevision.tsx
```diff
@@ -1,46 +1,68 @@
...
-            spacing={1}
-            sx={{
-              alignItems: 'center',
-            }}
-          >
-            <Typography sx={{ color: theme.palette.secondary.main }}>{entryCard.version}</Typography>
-            <Typography variant='caption'>{entryCard.mirrored && `(Mirrored)`}</Typography>
+          <Stack direction='column' spacing={0}>
+            {snapshot.local !== undefined && (
+              <Typography sx={{ color: changedIsLocal ? primaryColor : mutedColor }}>{`v${snapshot.local}`}</Typography>
```
IR96334: The version number should be the most important thing in the row and what attracts your eye first. Currently the user dn is more noticeable. Could try underlining, increasing font size of, or emboldening the version number to make it more eye-catching, thus more "important".

frontend/src/entry/overview/EntryCardHistoryDialog.tsx
```diff
@@ -60,12 +160,16 @@ export default function EntryCardHistoryDialog({ entry, setOpen }: EntryCardHist
...
+                <TableCell>Compare</TableCell>
               </TableRow>
             </TableHead>
-            {sortedEntryCardRevisions}
+            {renderedSnapshots}
           </Table>
         </TableContainer>
         <DialogActions>
+          <Button color='primary' variant='contained' disabled={!compareEnabled} onClick={onCompareSelected}>
+            Compare selected
```
IR96334: Currently, you can select one to compare and then select one to view. This mixing of a "comparison mode" and  "single-view mode" is a bit confusing. 

For clarity, it could be wise to adjust this button to, "Compare revisions". And if selected, add the checkboxes and disable the ability to click to view just ONE revision. This effectively separates the dialog into a "comparison mode" and "single-view mode"

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,453 @@
...
+          <AccordionSummary expandIcon={<ExpandMore />}>
+            <Typography variant='h6' component='h1' color='primary'>
+              {`Compare ${toTitleCase(kindLabel)} Cards`}
+            </Typography>
+          </AccordionSummary>
+          <AccordionDetails>
+            <Stack direction={{ xs: 'column', md: 'row' }} spacing={2} sx={{ alignItems: 'stretch' }}>
+              <Stack spacing={2} sx={{ flex: 1 }}>
+                <Typography sx={{ fontWeight: 'bold' }}>From</Typography>
+                <Autocomplete
```
IR96334: Add the version numbers as seen in dialog. Especially helpful if a user makes multiple changes in one day.

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,453 @@
...
+                  isOptionEqualToValue={(option, value) => option.id === value.id}
+                  onInputChange={(_event, value, reason) => {
+                    if (reason === 'input') {
+                      setFromSearch(value)
+                    }
+                  }}
+                  onChange={(_event, value) => setModel(QueryDiffSide.From, value)}
+                  renderInput={(params) => <TextField {...params} label={`${kindLabel}`} />}
+                />
+                <Autocomplete
```
IR96334: Similar to my dialog comment, I think it should be more obvious whether you are viewing 1 or 2 revisions (alongside removal of accordion). If viewing one (single-view mode, if you will), could keep the comparison tool in view, but alter the "to" section with something that makes it more obviously empty. I like the idea of a button, e.g.
| + Compare |
| -------- | 

Pressing this would add back the "to" section, unpopulated.

frontend/src/entry/overview/EntryCardHistoryDialog.tsx
GB907762: This is not a bug, but a feature. There's two streams to this feature.

1. Compare to another revision
2. View the revision

I don't want to mix up this functionality with each other. I'll try to improve the UX on this. So have the compare to fields need to be populated if you are using the second stream.

frontend/src/entry/overview/EntryCardRevision.tsx
GB907762: I've made this bold which should improve visibility of it. Additionally I've made the mirrored display italic.

I do disagree that the version numbers are important. I feel this an internal attribute, and users would rather see the name and the date

ARADDCC002:
image

When viewing the diff between two versions that have the same day timestamp you can't distinguish between the two. I would either add the revision number itself somewhere and / or display the time as well.

GB907762:
image
> When viewing the diff between two versions that have the same day timestamp you can't distinguish between the two. I would either add the revision number itself somewhere and / or display the time as well.

Yeah I think this is fair athough I do think the version doesn't carry a lot of meaning to end users. I'd be tempted just to provide a timestamp.

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,453 @@
...
+      <Paper sx={{ p: 4, my: 4 }}>
+        <Stack direction='row' spacing={1} sx={{ alignItems: 'center', mb: 2 }}>
+          <Button startIcon={<ArrowBackIosIcon />} onClick={() => router.push(backHref)}>
+            {`Back to ${kindLabel}`}
+          </Button>
+        </Stack>
+        <Accordion defaultExpanded={!isSameRevision} sx={{ mb: 3 }}>
+          <AccordionSummary expandIcon={<ExpandMore />}>
+            <Typography variant='h6' component='h1' color='primary'>
+              {`Compare ${toTitleCase(kindLabel)} Cards`}
```
ARADDCC002: I would lower case this, so that it would read `Compare model card` or `Compare mirrored model card`

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,453 @@
...
+    <Container>
+      <Paper sx={{ p: 4, my: 4 }}>
+        <Stack direction='row' spacing={1} sx={{ alignItems: 'center', mb: 2 }}>
+          <Button startIcon={<ArrowBackIosIcon />} onClick={() => router.push(backHref)}>
+            {`Back to ${kindLabel}`}
+          </Button>
+        </Stack>
+        <Accordion defaultExpanded={!isSameRevision} sx={{ mb: 3 }}>
+          <AccordionSummary expandIcon={<ExpandMore />}>
+            <Typography variant='h6' component='h1' color='primary'>
```
ARADDCC002: I imagine `h1` is too high. That implies it's the page heading, whereas it's being used as a filter sub-heading.

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,453 @@
...
+              {`Compare ${toTitleCase(kindLabel)} Cards`}
+            </Typography>
+          </AccordionSummary>
+          <AccordionDetails>
+            <Stack direction={{ xs: 'column', md: 'row' }} spacing={2} sx={{ alignItems: 'stretch' }}>
```
ARADDCC002: I think it should be `sm: 'column'`

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,453 @@
...
+
+  const canRenderForm = !loading && !!chosenSchema && splitSchema.steps.length > 0 && !schemasDiverge
+
+  const isSameRevision =
+    !!fromEntryId && fromEntryId === toEntryId && fromVersion === toVersion && fromMirroredVersion === toMirroredVersion
+
+  return (
+    <Container>
+      <Paper sx={{ p: 4, my: 4 }}>
+        <Stack direction='row' spacing={1} sx={{ alignItems: 'center', mb: 2 }}>
```
ARADDCC002: How come you're using a `<Stack />` for one element?

frontend/src/entry/overview/EntryCardRevision.tsx
GB907762: Additionally I've added some styling to mirrored models to make the edits and versions very clear.

frontend/src/entry/EntryCardCompare.tsx
GB907762: https://github.com/gchq/Bailo/pull/3817#issuecomment-4903803315

ARADDCC002 (dismissed):

frontend/src/entry/EntryCardCompare.tsx
GB907762: Good spot. I'll use the same component as we use else ware - A button within a link.

frontend/src/entry/EntryCardCompare.tsx
GB907762: This is the same pattern used everywhere else in the app.

I'll remove the accordion to make this more clear.

frontend/src/entry/EntryCardCompare.tsx
GB907762: I'll remove the accordion and make this clearer.

frontend/src/entry/EntryCardCompare.tsx
GB907762: https://mui.com/material-ui/customization/breakpoints/#default-breakpoints

The `xs` breakpoint will apply to the smallest screens. Otherwise if you shrink the page down enough the page will overflow.

frontend/src/entry/EntryCardCompare.tsx
GB907762: Not keen on this. I've made them unpopulated but visible, for better UX.

JRB66955 (dismissed):

frontend/src/MuiForms/TagSelector.tsx
IR96334: formContext is provided through registry object. change this and add replace formContext with registry.formContext throughout and this widget should start working

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,497 @@
...
+                  onChange={(_event, value) => setVersion(QueryDiffSide.From, QueryCardType.Mirror, value?.version)}
+                  renderInput={(params) => <TextField {...params} label='Mirrored Version' />}
+                />
+              )}
+            </Stack>
+            <Stack sx={{ justifyContent: 'center', alignItems: 'center' }}>
+              <Typography>&nbsp;</Typography>
+              <Tooltip title='Swap comparison sides'>
+                <span>
+                  <IconButton
```
IR96334, line 431: Needs to look more like a button. Either buttonify it, maybe adding a border? Or add a swap button below and add the swap handle there instead.

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,497 @@
...
+          <Stack direction={{ xs: 'column', md: 'row' }} spacing={2} sx={{ alignItems: 'stretch' }}>
+            <Stack spacing={2} sx={{ flex: 1 }}>
+              <Typography sx={{ fontWeight: 'bold' }}>From</Typography>
+              <Autocomplete
+                disablePortal
+                options={fromEntries}
+                loading={isFromEntriesLoading}
+                fullWidth
+                size='small'
+                value={fromEntry ?? null}
```
IR96334, line 381: can remove nullish coalesce as fromEntry resolves to undefined when empty and value takes null/undefined.

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,497 @@
...
+            </Stack>
+            <Stack spacing={2} sx={{ flex: 1 }}>
+              <Typography sx={{ fontWeight: 'bold' }}>To</Typography>
+              <Autocomplete
+                disablePortal
+                options={toEntries}
+                loading={isToEntriesLoading}
+                fullWidth
+                size='small'
+                value={toEntry ?? null}
```
IR96334, line 471: can remove nullish coalesce as fromEntry resolves to undefined when empty and value takes null/undefined.

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,497 @@
...
+        fromEntryId,
+        fromLocalCard?.version,
+        fromMirroredCard?.version,
+        toEntryId,
+        toLocalCard?.version,
+        toMirroredCard?.version,
+        fromOnly ? 'single' : 'diff',
+      ].join(':')
+    : undefined
+  if (currentKey && currentKey !== lastKey) {
```
IR96334, line 258: Add newline between const and if

backend/test/services/model.spec.ts
PE39806, line 1: Add a new search test with multiple `kind`s e.g. `kind: ['model', 'data-card']`.

backend/src/connectors/peer/huggingface.ts
```diff
@@ -88,11 +88,11 @@ export class HuggingFaceHubConnector extends BasePeerConnector {
     }
 
     // Only supports searching for models, not data cards etc.
-    if (opts.kind && opts.kind != 'model') {
+    if (opts.kind && opts.kind.length > 0 && !opts.kind.includes('model')) {
```
PE39806: This change allows searches across multiple kinds but HF only supports `'model'`, so we should restrict this to only allow `'model'`
```suggestion
    if (opts.kind && opts.kind.length === 1 && !opts.kind.includes('model')) {
```

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,497 @@
...
+    setLastKey(currentKey)
+    setSplitSchema(initialSplit)
```
PE39806: Does this update state during the render, and if so should these be moved to some `useEffect`?

frontend/src/MuiForms/AdditionalInformation.tsx
```diff
@@ -218,7 +212,9 @@ export default function AdditionalInformation({
               <Box sx={{ wordBreak: 'break-word' }}>{mirroredStateDisplay()}</Box>
             )}
           </Stack>
-          {children && (
+          {/* Some components are conditionally rendered this filters between children that have a ReactNode*/}
+          {((children && !Array.isArray(children)) ||
+            (Array.isArray(children) && children.some((child) => !!child))) && (
```
PE39806, line 206: Maybe use [`React.Children.count(children)`](https://react.dev/reference/react/Children#children-count) here?

frontend/utils/stringUtils.ts
```diff
@@ -38,6 +38,17 @@ export const isValidPortNumber = (portNumber: string) => {
   )
 }
 
+export const parseNat = (numberOrString: string) => {
+  if (isNaN(Number(numberOrString)) || Number(numberOrString) < 0) {
+    return NaN
+  }
+  return parseInt(numberOrString)
```
PE39806, line 45: Prefer `Number` due to speed, in which case convert once and store as `const` for `if` and `return` https://medium.com/@emreavcilar/javascript-parseint-vs-number-8186ef8e29c1

frontend/src/entry/EntryCardCompare.tsx
IR96334, line 258: Upon testing, the contents of this if statement will run alongside the useEffect on line 255, effectively executing the same setSplitSchema call twice. This is because the useEffect's dep is initialSplit, which itself has the deps of (basically) key. So when the key is changed the useEffect is pretty much always run, and if the "new" key is different from lastKey, the useEffect will still run (as initialSplit has changed) ALONGSIDE this code block. I would _probably_ remove the setSplitSchema from this block.

frontend/src/MuiForms/Metrics.tsx
```diff
@@ -36,28 +37,41 @@ interface MetricsProps {
   onChange: (newValue: MetricValue[]) => void
   value: MetricValue[]
   label: string
+  formContext?: Registry['formContext']
```
IR96334: change to registry as above

frontend/src/MuiForms/CheckboxInput.tsx
```diff
@@ -65,27 +100,33 @@ export default function CheckboxInput({
...
-      {!registry.formContext.editMode && (
-        <>
-          <Typography>{value ? 'Yes' : 'No'}</Typography>
-        </>
+      ) : inCompareMode && registry.formContext.mirroredModel && value ? (
+        <InlineDiff from={formatBoolean(compareFromState)} to={formatBoolean(value)} />
+      ) : (
+        value && <Typography>{formatBoolean(value)}</Typography>
       )}
     </AdditionalInformation>
```
PE39806: Given `compareFromState`, `compareFromMirroredState` and `inCompareMode` are repeatedly redefined, I think these should be abstracted to a new utility e.g.
`frontend/src/hooks/useCompareField.ts`
```ts
import { Registry } from '@rjsf/utils'
import { getCompareFromMirroredState, getCompareFromState, getMirroredState } from 'utils/formUtils'

export interface CompareFieldState<T> {
  mirroredState: T | undefined
  compareFromState: T | undefined
  compareFromMirroredState: T | undefined
  inCompareMode: boolean
  isMirroredModel: boolean
  inMirroredCompare: boolean
  editMode: boolean
}

export default function getCompareFieldState<T = unknown>(
  id: string,
  formContext: Registry['formContext'],
): CompareFieldState<T> {
  const mirroredState = getMirroredState(id, formContext) as T | undefined
  const compareFromState = getCompareFromState(id, formContext) as T | undefined
  const compareFromMirroredState = getCompareFromMirroredState(id, formContext) as T | undefined
  const inCompareMode = !!formContext.compareMode && !formContext.editMode
  const isMirroredModel = !!formContext.mirroredModel
  const editMode = !!formContext.editMode

  return {
    mirroredState,
    compareFromState,
    compareFromMirroredState,
    inCompareMode,
    isMirroredModel,
    inMirroredCompare: inCompareMode && isMirroredModel,
    editMode,
  }
}
```
with a new wrapper component e.g.
`frontend/src/common/CompareField.tsx`
```tsx
import { SxProps } from '@mui/material/styles'
import { ReactNode } from 'react'
import InlineDiff from 'src/common/InlineDiff'
import { CompareFieldState } from 'src/hooks/useCompareField'
import AdditionalInformation from 'src/MuiForms/AdditionalInformation'

interface CompareFieldProps {
  id: string
  label?: string
  required?: boolean
  description?: string
  compare: CompareFieldState<unknown>
  value: unknown
  formatter?: (val: unknown) => string | undefined
  markdown?: boolean
  hasValue?: boolean
  fallbackMirroredContent?: ReactNode
  sx?: SxProps
  children: ReactNode
}

export default function CompareField({
  id,
  label,
  required,
  description,
  compare,
  value,
  formatter,
  markdown,
  hasValue,
  fallbackMirroredContent,
  sx,
  children,
}: CompareFieldProps) {
  const format = (val: unknown): string | undefined => {
    if (formatter) {
      return formatter(val)
    }
    return val as string | undefined
  }

  if (compare.inCompareMode && !compare.isMirroredModel) {
    const from = compare.compareFromState ?? compare.mirroredState
    return (
      <AdditionalInformation
        editMode={false}
        display={false}
        label={label}
        id={id}
        required={required}
        mirroredModel={false}
        description={description}
      >
        <InlineDiff from={format(from)} to={format(value)} markdown={markdown} />
      </AdditionalInformation>
    )
  }

  const mirroredContent = compare.inMirroredCompare ? (
    <InlineDiff
      from={format(compare.compareFromMirroredState)}
      to={format(compare.mirroredState)}
      markdown={markdown}
    />
  ) : (
    (fallbackMirroredContent ?? compare.mirroredState)
  )

  const effectiveHasValue = hasValue ?? !!value
  const display = compare.inMirroredCompare ? true : effectiveHasValue

  return (
    <AdditionalInformation
      editMode={compare.editMode}
      mirroredState={mirroredContent}
      display={display && compare.isMirroredModel}
      label={label}
      id={id}
      required={required}
      mirroredModel={compare.isMirroredModel}
      description={description}
      sx={sx}
    >
      {children}
    </AdditionalInformation>
  )
}
```
This will then require each widget in `frontend/src/MuiForms/` to be updated e.g.

```suggestion
  const compare = getCompareFieldState<boolean>(id, registry.formContext)

  const formatBoolean = (val?: unknown): string | undefined => {
    if (val === undefined) {
      return undefined
    }
    return val ? 'Yes' : 'No'
  }

  if (!compare.editMode && !compare.inCompareMode && value === undefined) {
    return (
      <AdditionalInformation
        editMode={false}
        mirroredState={compare.mirroredState}
        display={compare.isMirroredModel && value !== undefined}
        label={label}
        id={id}
        required={required}
        mirroredModel={compare.isMirroredModel}
        description={schema.description}
      >
        <Typography
          sx={{
            fontStyle: 'italic',
            color: theme.palette.customTextInput.main,
          }}
          aria-label={`Label for ${label}`}
        >
          Unanswered
        </Typography>
      </AdditionalInformation>
    )
  }

  const fallbackMirrored =
    compare.mirroredState !== undefined ? <Typography>{formatBoolean(compare.mirroredState)}</Typography> : undefined

  return (
    <CompareField
      id={id}
      label={label}
      required={required}
      description={schema.description}
      formContext={registry.formContext}
      compare={compare}
      value={value}
      formatter={formatBoolean}
      hasValue={value !== undefined}
      fallbackMirroredContent={fallbackMirrored}
    >
      {compare.editMode ? (
        <RadioGroup onChange={handleChange} value={value} aria-label={`radio input field for ${label}`} id={id}>
          <FormControlLabel value={true} control={<Radio data-test={`${id}-yes-option`} />} label='Yes' />
          <FormControlLabel value={false} control={<Radio data-test={`${id}-no-option`} />} label='No' />
        </RadioGroup>
      ) : compare.inMirroredCompare && value ? (
        <InlineDiff from={formatBoolean(compare.compareFromState)} to={formatBoolean(value)} />
      ) : (
        value && <Typography>{formatBoolean(value)}</Typography>
      )}
    </CompareField>
```

PE39806 (changes requested):

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,497 @@
...
+    })
+  }
+
+  return (
+    <Container>
+      <Paper sx={{ p: 4, my: 4 }}>
+        <Stack spacing={4}>
+          <Link href={link}>
+            <Button sx={{ width: 'fit-content' }} startIcon={<ArrowBack />}>
+              {`Back to ${kindLabel}`}
```
PE39806: I'm not sure about this when comparing different models as it will always point to the From model and not To, but users may have navigated to compare from the To model.

frontend/pages/model-card/compare.tsx
PE39806, line 1: The only route to this page is to have a model with a model card with multiple versions. It is unlikely, but a user may wish to compare a model with a model card with only 1 version (e.g. from a template) to another model, in which case they cannot do this from the model with a model card with only 1 version.

PE39806 (changes requested):

frontend/src/MuiForms/Metrics.tsx
```diff
@@ -36,28 +37,41 @@ interface MetricsProps {
...
 }
 
 export default function Metrics({ onChange, value, label, id, registry, required, schema }: MetricsProps) {
-  const [metricsWithIds, setMetricsWithIds] = useState<MetricValueWithId[]>(
-    value.map((metric) => ({
-      ...metric,
-      id: uuidv4(),
-    })),
-  )
+  const [metricsWithIds, setMetricsWithIds] = useState<MetricValueWithId[]>([])
```
IR96334, line 47: What is the reason for this change?

frontend/src/MuiForms/AdditionalInformation.tsx
GB907762: Hmm I don't think this solves the issue, due to the following and the fact that conditional components short circuit to `false`

https://react.dev/reference/react/Children#children-count-caveats

I think I can role this back since the refactor ensures that each component has one child at most.

frontend/utils/stringUtils.ts
GB907762: I'll keep this in mind for future. The reason why I've done it this way is to prevent erroneous edge cases such as `Infinity`, `0xff`, `1e10` ect.. when parsing through the query.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number

It's small and probably not so important but there's a variety of inputs on the page that don't sanitise it this way and the speed cost is imperceptible.

frontend/src/entry/EntryCardCompare.tsx
GB907762: I'll put some conditional logic to render another model link.

frontend/src/MuiForms/Metrics.tsx
GB907762: I don't know where the breaking change has come from, but this component seems to have been unable to add metrics via the `Add item` button. I've added a use effect to update `metricsWithIds` when the value changes on the form, so these happen automatically.

What was previously happening was the value was changing but no rerender happens.

frontend/src/MuiForms/Metrics.tsx
GB907762: Removed. Unused prop

frontend/src/entry/EntryCardCompare.tsx
GB907762: I'm going to remove this. I already have a use effect for on render

frontend/pages/model-card/compare.tsx
GB907762: Could you clarify what you mean. You can get to this page by selecting an individual version and just clicking on the listitem, instead of the checkbox.

If you think this is unclear. Let me know.

frontend/src/entry/EntryCardCompare.tsx
GB907762: Wont do. A MUI Autocomplete treat's `undefined` as a value, within a controlled component and thus has a remove adornment besides it.

https://mui.com/material-ui/api/autocomplete/#autocomplete-prop-defaultValue

frontend/src/entry/EntryCardCompare.tsx
GB907762: https://github.com/gchq/Bailo/pull/3817#discussion_r3597266877

frontend/src/entry/EntryCardCompare.tsx
GB907762: I've made this a FAB
https://mui.com/material-ui/react-floating-action-button/

frontend/pages/model-card/compare.tsx
PE39806: Ah, my mistake.

frontend/cypress/e2e/bailo/upload-model.cy.ts
PE39806, line 1: Ideally, replace this test rather than just removing it.

frontend/pages/model-card/compare.tsx
```diff
@@ -0,0 +1,39 @@
...
+    toModel?: string
+    toVersion?: string
+    toMirroredVersion?: string
+  } = router.query
+
+  return (
+    <>
+      <Title text='Compare Model Cards' />
+      <EntryCardCompare
+        entryKind={EntryKind.MODEL}
```
PE39806, line 29: I think this is incorrect as this could be an MM or UM, and `EntryCardCompare` uses this value in the `Go to ${EntryKindLabel[entryKind]}` button.

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,507 @@
...
+        (fromVersion !== undefined && isFromLocalLoading) ||
+        (fromMirroredVersion !== undefined && isFromMirroredLoading))) ||
+    (toEntryId !== undefined &&
+      (isToRevisionsLoading ||
+        (toVersion !== undefined && isToLocalLoading) ||
+        (toMirroredVersion !== undefined && isToMirroredLoading))) ||
+    (!!chosenSchemaId && isSchemaLoading)
+
+  const renderVersionLabel = (revision: EntryCardRevisionInterface) =>
+    `Updated by ${revision.createdBy} on ${formatDateTimeString(revision.createdAt)}`
```
PE39806: I like the information this surfaces but I also think that a) this should include the version (same as in `EntryCardRevision`) and b) this should use `UserDisplay` to resolve the user instead of `${revision.createdBy}` as we do elsewhere in the application.

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,507 @@
...
+    <Container>
+      <Paper sx={{ p: 4, my: 4 }}>
+        <Stack spacing={4}>
+          <Typography variant='h6' component='h1' color='primary'>
+            {`Compare ${toTitleCase(kindLabel)} Cards`}
+          </Typography>
+          <Stack direction={{ xs: 'column', md: 'row' }} spacing={2} sx={{ alignItems: 'stretch' }}>
+            <Stack spacing={2} sx={{ flex: 1 }}>
+              <Typography sx={{ fontWeight: 'bold' }}>From</Typography>
+              {renderGotoEntryButton(fromEntryId)}
```
PE39806: I wonder if these should be on the same line with a slightly larger gap between this and the following dropdowns, as this will make the distinction between the user selection and the navigation button clearer. The current approach is fully functional but I can imagine a user accidentally clicking on the navigation button rather than the dropdown filter with the current UX.
The same would apply to "To" below.

PE39806 (changes requested):
The "Share" button in `frontend/src/Form/FormTemplates.tsx` correctly updates the compare page's URL but the link that is copied to the clipboard is incorrect (e.g. `http://localhost:8080/data-card/compare?page=0#root` instead of `http://localhost:8080/data-card/compare?fromModel=datacard1-abc123&toModel=datacard1-abc123&fromVersion=1&toVersion=2&page=0#root`)

frontend/src/entry/overview/EntryCardHistoryDialog.tsx
```diff
@@ -12,33 +12,134 @@ import {
...
+
+  // Newest-first for display in the table.
+  const snapshotsNewestFirst = useMemo(() => [...snapshotsOldestFirst].reverse(), [snapshotsOldestFirst])
+
+  const buildHref = (from: EntryCardSnapshot, to?: EntryCardSnapshot): string => {
+    const query = new URLSearchParams()
+    query.set('fromModel', entry.id)
+    if (to) {
+      query.set('toModel', entry.id)
+    }
```
PE39806: Comparing data cards is valid so prefer
```suggestion
    query.set('fromEntry', entry.id)
    if (to) {
      query.set('toEntry', entry.id)
    }
```

PE39806 (changes requested):

ARADDCC002:
I'm personally not a fan of the box shadow on the icon button to swap the comparison of two schemas.

frontend/pages/model-card/compare.tsx
GB907762: So I've set this up in a way you can compare Models, MMs and UMs together `Go to ${EntryKindLabel[entryKind]}` Means that it willl say `Go to model` on UMs for example. This is the same approach we use for the back buttons in the rest of the app.

GB907762:
> The "Share" button in `frontend/src/Form/FormTemplates.tsx` correctly updates the compare page's URL but the link that is copied to the clipboard is incorrect (e.g. `http://localhost:8080/data-card/compare?page=0#root` instead of `http://localhost:8080/data-card/compare?fromModel=datacard1-abc123&toModel=datacard1-abc123&fromVersion=1&toVersion=2&page=0#root`)

I'm not a great fan of copying hash url's automatically since they are invariably static links, but I'll do this anyway.

frontend/src/entry/EntryCardCompare.tsx
GB907762: I'll make this into a link as well.

frontend/src/entry/EntryCardCompare.tsx
```diff
@@ -0,0 +1,529 @@
...
+                  />
+                  <RevisionDetails revision={fromMirroredOption} />
+                </>
+              )}
+            </Stack>
+            <Stack sx={{ justifyContent: 'center', alignItems: 'center' }}>
+              <Typography>&nbsp;</Typography>
+              <Tooltip title='Swap comparison sides'>
+                <span>
+                  <Fab
```
IR96334: Fab feels out of place here. Replace with IconButton and add styling: 
```java
sx={{
                      border: '1px solid',
                      borderColor: 'primary.main',
                      '&.Mui-disabled': {
                        borderColor: 'action.disabled',
                      },
                    }}
```

PE39806 (approved):

IR96334 (approved):

PR #3820, opened 2026-07-02
Untrusted model file upload guidance 

JR40159:


frontend/src/entry/model/files/FileUploadDialog.tsx
```diff
@@ -146,15 +151,25 @@ export default function FileUploadDialog({ open, onDialogClose, model, mutateMod
...
+    <Dialog disableAutoFocus open={open} onClose={onDialogClose} maxWidth='md' fullWidth>
       <DialogContent>
         <Stack spacing={2}>
           <label htmlFor='add-files-button'>
-            <Button loading={isFilesUploading} component='span' variant='outlined' sx={{ width: '100%' }}>
+            <Button loading={isFilesUploading} endIcon={<FileUpload />} component='span' variant='outlined'>
               Select files
             </Button>
           </label>
+          {model.kind === EntryKind.MODEL && (
```
PE39806:
```suggestion
          {model.kind === EntryKind.UNTRUSTED_MODEL && (
```

PE39806 (changes requested):

frontend/src/entry/model/files/FileUploadDialog.tsx
JR40159: Good spot, left over from testing!

PE39806 (changes requested):
This PR needs to also display the string when uploading a file in the:
- [x] Create release page
- [x] Edit release page

frontend/actions/uiConfig.ts
```diff
@@ -11,10 +11,30 @@ export function useGetUiConfig() {
...
+  if (error) {
+    return {
+      isUiConfigError: error,
+    }
+  }
+
+  if (!data || !data.uiConfig) {
+    return {
+      isUiConfigError: {
+        ...new Error('Unable to get data for the current user'),
```
PE39806: I don't understand why this creates and then immediately spreads an error. Please could you either add a comment here, or just provide the required data directly (e.g. `name` and `message`)?

frontend/src/contexts/AppContextProvider.tsx
```diff
@@ -21,6 +25,8 @@ export default function AppContextProvider({ children }: AppContextProviderProps
   const themeModeValue = useThemeMode()
   const unsavedChangesValue = useUnsavedChanges()
   const userPermissionsValue = useUserPermissions()
+  const { scanners, isScannersError, isScannersLoading } = useGetArtefactScannerInfo()
+  const { uiConfig, isUiConfigLoading, isUiConfigError } = useGetUiConfig()
```
PE39806, line 29: Should this not be something like `const uiConfig = useContext(UiConfigContext)` or `const uiConfig = useUiConfigContext()`?

frontend/src/entry/model/files/FileUploadDialog.tsx
```diff
@@ -146,15 +151,25 @@ export default function FileUploadDialog({ open, onDialogClose, model, mutateMod
   )
 
   return (
-    <Dialog open={open} onClose={onDialogClose} maxWidth='md' fullWidth>
+    <Dialog disableAutoFocus open={open} onClose={onDialogClose} maxWidth='md' fullWidth>
```
PE39806: Why add `disableAutoFocus`? this might be an a11y regression.

frontend/src/entry/model/files/FileUploadDialog.tsx
```diff
@@ -146,15 +151,25 @@ export default function FileUploadDialog({ open, onDialogClose, model, mutateMod
...
       <DialogContent>
         <Stack spacing={2}>
           <label htmlFor='add-files-button'>
-            <Button loading={isFilesUploading} component='span' variant='outlined' sx={{ width: '100%' }}>
+            <Button loading={isFilesUploading} endIcon={<FileUpload />} component='span' variant='outlined'>
               Select files
             </Button>
           </label>
+          {model.kind === EntryKind.UNTRUSTED_MODEL && (
```
PE39806, line 168: It's probably safer to check that the box won't be empty
```suggestion
          {model.kind === EntryKind.UNTRUSTED_MODEL && fileUploadGuidance && (
```

frontend/src/contexts/AppContextProvider.tsx
```diff
@@ -39,18 +45,30 @@ export default function AppContextProvider({ children }: AppContextProviderProps
...
     }
 
-    if (isCurrentUserV3Loading || !currentUserV3) {
+    if (isUiConfigError) {
+      return <MessageAlert message={isUiConfigError.info.message} severity='error' />
+    }
+
+    if (isScannersError) {
+      return <MessageAlert message={isScannersError.info.message} severity='error' />
+    }
```
PE39806, line 54: Most pages don't use the scanners, but if the scanner endpoint(s) become unavailable I think this will break _all_ pages rather than just scanning ones. I'm not sure if this matters or not.

PE39806 (changes requested):

frontend/src/entry/model/files/FileUploadDialog.tsx
JR40159: Is this necessary as there's a default string set in the default configuration?

frontend/src/entry/model/files/FileUploadDialog.tsx
JR40159: No idea where that came from....

frontend/src/contexts/AppContextProvider.tsx
JR40159: But we need to handle the error and loading states?

frontend/src/contexts/AppContextProvider.tsx
JR40159: Imo making the error more prevalent like this will make it easier to notice as that endpoint shouldn't be erroring

frontend/src/entry/model/files/FileUploadDialog.tsx
PE39806: There's nothing stopping an admin from setting the string to `''` which would then look odd.

frontend/src/contexts/AppContextProvider.tsx
PE39806: In that case, I think I don't quite understand the new `UiConfigContext` yet 😅

frontend/src/entry/model/releases/ReleaseForm.tsx
```diff
@@ -359,6 +361,15 @@ export default function ReleaseForm({
             >{`Files (${formData.files.length})`}</Typography>
           </AccordionSummary>
           <AccordionDetails>
+            {!isReadOnly && model.kind === EntryKind.MODEL && (
```
PE39806: Again...?
```suggestion
            {!isReadOnly && model.kind === EntryKind.UNTRUSTED_MODEL && (
```

PE39806 (changes requested):

PE39806 (approved):

ON93754 (approved):

PR #3832, opened 2026-07-07
UI for no release metrics

ARADDCC002:


JRB66955 (dismissed):

frontend/src/metrics/PolicyMetrics.tsx
```diff
@@ -1,45 +1,102 @@
...
+      return metricData.byOrganisation.find((subset) => subset.organisation === selectedOrganisation)
+    },
+    [selectedOrganisation],
+  )
+
+  useEffect(() => {
+    if (!selectedMetric) {
+      setSelectedData(rolePolicyMetrics)
+    } else {
+      if (!filteredDataset) {
```
IR96334: I think filteredDataset will always be truthy? If so, this is dead code.

frontend/src/metrics/PolicyMetrics.tsx
```diff
@@ -1,45 +1,99 @@
...
+                <EmptyBlob text='No items to display.' />
+              ),
+            )
+          }
+          break
+      }
     }
-    return rolePolicyMetrics.byOrganisation.find((subset) => subset.organisation === selectedOrganisation)
-  }, [rolePolicyMetrics, selectedOrganisation])
+  }, [selectedMetric, rolePolicyMetrics, noReleasesPolicyMetrics, filteredDataset])
```
IR96334:
```suggestion
 const filteredDataset = useCallback(
    (metricData) => {
      if (selectedOrganisation === 'All') {
        return metricData.global
      }
      return metricData.byOrganisation.find((subset) => subset.organisation === selectedOrganisation)
    },
    [selectedOrganisation],
  )

  const [selectedMetric, setSelectedMetric] = useState<SelectedMetricKindKeys>(SelectedMetricKind.MISSING_ROLES)
  const selectedData: undefined | PolicyRoleMetrics | BaseNoReleaseMetrics = (() => {
    switch (selectedMetric) {
      case SelectedMetricKind.MISSING_ROLES:
        return rolePolicyMetrics
      case SelectedMetricKind.NO_RELEASES:
        return noReleasesPolicyMetrics
      default:
        return rolePolicyMetrics
    }
  })()

  const selectedChart: ReactElement = useMemo(() => {
    if (!selectedData) {
      return <></>
    }
    const filtered = filteredDataset(selectedData)
    if (!filtered || filtered.entries.length === 0) {
      return <EmptyBlob text='No items to display.' />
    }
    switch (selectedMetric) {
      case SelectedMetricKind.MISSING_ROLES:
        return <PolicyRoleMetricsCharts data={filteredDataset(rolePolicyMetrics)} />

      case SelectedMetricKind.NO_RELEASES:
        return <PolicyNoReleasesMetricsCharts data={filteredDataset(noReleasesPolicyMetrics)} />

      default:
        return <></>
    }
  }, [filteredDataset, noReleasesPolicyMetrics, rolePolicyMetrics, selectedData, selectedMetric])
```

IR96334: Don't think useState + useEffect combo is necessary here and I don't think changing selectedData and selectedChart should cause rerender, so can make const.

frontend/src/metrics/PolicyNoReleasesMetricsCharts.tsx
```diff
@@ -0,0 +1,79 @@
...
+interface PolicyMetricsChartsProps {
+  data: GlobalNoReleasesMetrics
+}
+
+export default function PolicyNoReleasesMetricsCharts({ data }: PolicyMetricsChartsProps) {
+  const theme = useTheme()
+
+  const { entryRoles, isEntryRolesLoading, isEntryRolesError } = useGetEntryRoles()
+
+  const ownerRoleDisplayName = useMemo(() => {
```
IR96334, line 21: Is this expensive enough to require memoisation?

frontend/src/metrics/PolicyRoleMetricsCharts.tsx
IR96334, line 49: Is this expensive enough to required memoisation?

ARADDCC002: It relies on `data.summary` which in turn is retrieved from the backend, so it's best to memoise it.

frontend/src/metrics/PolicyNoReleasesMetricsCharts.tsx
ARADDCC002: It relies on data retrieved via the API so it should be memoised

frontend/src/metrics/PolicyRoleMetricsCharts.tsx
```diff
@@ -51,18 +65,50 @@ export default function PolicyMetricsCharts({ data }: PolicyMetricsChartsProps)
...
+        >
+          {data.summary.map((roleSummary) => (
+            <MenuItem key={roleSummary.roleId} onClick={() => handleChipFilterOnClick(roleSummary.roleId)}>
+              {missingRoleFilters.includes(roleSummary.roleId) ? (
+                <ListItemIcon>
+                  <Check />
+                </ListItemIcon>
+              ) : (
+                <></>
+              )}
```
PE39806:
```suggestion
              {missingRoleFilters.includes(roleSummary.roleId) && (
                <ListItemIcon>
                  <Check />
                </ListItemIcon>
              )}
```

PE39806 (changes requested):

frontend/src/common/FilterMenuButton.tsx
IR96334: Resetting filters isn't intuitive enough. To reset I would expect a reset filters button, or the option that effectively resets, i.e. All, should be differentiated somehow. This is done elsewhere through italics. It could also be done through other means.

IR96334: Edit: its done with italics for _no organisation_, in the org filter

frontend/actions/metrics.ts
```diff
@@ -47,6 +47,20 @@ export function useGetRolePolicyMetrics() {
   }
 }
 
+export function useGetNoReleasesPolicyMetrics() {
+  const { data, isLoading, error, mutate } = useSWR<BaseNoReleaseMetrics, ErrorInfo>(
+    '/api/v3/metrics/compliance/no-releases',
+    fetcher,
+  )
```
PE39806, line 54: It looks like there's a bug in this endpoint. The response includes data cards in the entry types, and data cards can never have releases so should be excluded.

frontend/src/metrics/PolicyNoReleasesMetricsCharts.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  if (isEntryRolesLoading) {
+    return <Loading />
+  }
+
+  return (
+    <Stack spacing={4}>
+      <Stack spacing={2} sx={{ width: '100%' }}>
+        <Typography sx={{ fontWeight: 'bold' }} variant='h6' color='primary'>
+          Entries with no releases
```
PE39806:
```suggestion
          Models with no releases
```

frontend/src/metrics/PolicyNoReleasesMetricsCharts.tsx
```diff
@@ -0,0 +1,79 @@
...
+        </Typography>
+        <Box sx={{ backgroundColor: theme.palette.container.main, p: 2, borderRadius: 1 }}>
+          <Table size='small'>
+            <TableHead>
+              <TableRow>
```
PE39806, line 68: I wonder if it's worth adding another cell for the model type (Model, Mirrored model or Untrusted model)? I'm not sure if different types may be treated a bit differently so surfacing this information early may be useful. I am happy for you to push back on this if you don't think that's appropraite/necessary.

PE39806 (changes requested):

PE39806 (approved):

JRB66955 (approved):

PR #3833, opened 2026-07-07
Fixed an issue with datacard selector not working

ARADDCC002:


GB907762 (changes requested):
Could you have a look into the `Metrics` widget as part of this, since the `Add item` button fails

GB907762 (changes requested):
This doesn't seem to cover mirrored models with an unpopulated field.

IR96334:
TagSelector is also broken.
Fix:
Update TagSelectorProps to expect `registry?: Registry`, instead of `formContext?: Registry['formContext']`.

ARADDCC002:
> Could you have a look into the `Metrics` widget as part of this, since the `Add item` button fails

As we aren't currently using that widget I feel it's best to fix it in another PR.

GB907762:
> > Could you have a look into the `Metrics` widget as part of this, since the `Add item` button fails
> 
> As we aren't currently using that widget I feel it's best to fix it in another PR.

I'll tackle it as part of #3817 Since I need to rewrite it anyway as part of a new diff

GB907762 (approved):

PE39806 (approved):

IR96334 (approved):

Issue #2557, opened 2025-08-08
Support richer file management with support for directories


GCHQDeveloper36:
### Problem

Uploading files to an entry is done in a flat style, with no support for hierarchy or mirroring files distributed in a richer format with subdirectories for organisation.

### Proposed Solution

Allow users to name files in a path-like structure (e.g. filename is `subdir/file.json`). When showing a list of files, show these prefixes/paths as folder-like objects that can be expanded or navigated into.

### Alternatives

Prefix filenames like `subdir-file.json` which isn't clear, and breaks naming of files.

### What is the priority of your issue?

Low

PR #3856, opened 2026-07-16
Feature/hierarchical file management

M422110:
Implemented folder management to address issue: https://github.com/gchq/Bailo/issues/2557

## Context

Bailo currently stores and displays files in a flat list with no directory structure. Users uploading models with many files (weights, configs, documentation) have no way to organise them. The GitHub issue requests path-like filenames (e.g. `subdir/file.json`) displayed as navigable folder structures. This impacts both the file management tab and the releases view.

## Design Decision: Encode Paths in the `name` Field

Store the full relative path in the existing `FileInterface.name` field (e.g. `"weights/resnet50.bin"` instead of `"resnet50.bin"`).

**Why this approach:**
- Zero schema changes, zero database migrations
- Fully backward-compatible: existing flat filenames appear at root level
- Matches browser `webkitRelativePath` for folder uploads
- S3 keys remain UUID-based and unchanged (`beta/model/{modelId}/files/{uuid}`)
- Single source of truth for display, search, and uniqueness

**Convention:** `/` is the path separator. Names must not start/end with `/`. Empty segments (`//`) and traversal segments (`.`, `..`) are forbidden.

**Why `/` is safe:** 

Investigation confirmed that S3 keys, model mirroring/export, ClamAV scanning, and frontend download links all use `fileId`, never `file.n
ame`, for path construction. Two areas need targeted fixes (Express route and Python client)

### New folder upload option which will stage all files and subfolders. Ability to navigate directory (including breadcrumbs)
bailo_file_upload

### Ability to add folders to new releases directly
bailo_release_file_upload

### Toggle between flat and directory view when there are nested folders

Screenshot 2026-07-16 at 11 33 25
Screenshot 2026-07-16 at 11 33 43

### Search operates across all files/folders showing counts in directory view 
Screenshot 2026-07-16 at 11 36 05


backend/src/routes/v2/model/file/getDownloadFile.ts
```diff
@@ -66,7 +67,10 @@ const binaryContent = {
 const modelIdParam = z.object({ modelId: z.string() })
 const fileIdParam = z.object({ fileId: z.string() })
 const semverParam = z.object({ semver: z.string() })
-const fileNameParam = z.object({ fileName: z.string() })
+// Wildcard route (*fileName) returns path segments as an array, join them back into a slash-delimited filename
+const fileNameParam = z.object({
+  fileName: z.union([z.string(), z.array(z.string())]).transform((v) => (Array.isArray(v) ? v.join('/') : v)),
```
PE39806, line 72: Prevent arbitrary wildcard paths by validating early
```suggestion
  fileName: z.union([z.string(), z.array(z.string())]).transform((v) => (Array.isArray(v) ? v.join('/') : v)).transform((v) => validateFileName(v)),
```

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+    try {
+      setIsDeleting(true)
+      setDeleteError('')
+      for (const file of allFilesInFolder) {
+        const res = await deleteEntryFile(modelId, file._id)
+        if (!res.ok) {
+          setDeleteError(await getErrorMessage(res))
+          return
+        }
+      }
```
PE39806, line 340: I'm not so keen on this approach as a failure midway through leaves the folder partially deleted. Instead, a new backend bulk-delete endpoint which uses transactions is the safer approach for atomicity. This bulk delete option could take a folder/directory and then recursively delete all child items via a single API call.

frontend/src/common/MultiFileInput.tsx
```diff
@@ -52,6 +61,7 @@ export default function MultiFileInput({
           ...newFiles.map((newFile) => ({ fileName: newFile.name, metadata: { text: '', tags: [] } })),
         ])
       }
```
PE39806, line 63: I'm not keen on directly mutating the native browser objects and would prefer something more like
```suggestion
      const newFiles = event.target.files ? Array.from(event.target.files) : []

      if (newFiles.length) {
        if (files) {
          const updatedFiles = [
            ...files.filter((file) => !newFiles.some((newFile) => newFile.name === file.name)),
            ...newFiles,
          ]
          onFilesChange(updatedFiles)
        } else {
          onFilesChange(newFiles)
        }
        onFilesMetadataChange([
          ...filesMetadata.filter(
            (fileMetadata) => !newFiles.some((newFile) => newFile.name === fileMetadata.fileName),
          ),
          ...newFiles.map((newFile) => ({
            fileName: newFile.webkitRelativePath || newFile.name,
            metadata: { text: '', tags: [] },
          })),
        ])
      }
```

PE39806 (changes requested):
Thank you for putting in the effort to add this useful functionality. From a quick test, the implementation seems good and I'm very happy with the way you've carefully handled `validateFileName` and `create(getFileBaseName(file.name))`. I'll have a more in depth look through the code soon, but here are a few initial thoughts for now.

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+    const items: BrowseListItem[] = []
+    for (const child of currentNode.children) {
+      if (child.isDirectory) {
+        const nestedNames = collectAllFileNames(child)
+        items.push({
+          key: `folder-${child.fullPath}`,
+          kind: 'folder',
+          node: child,
+          name: child.name,
+          searchableText: [child.name, ...nestedNames].join(' '),
```
GB907762, line 95: This does a partial match on all the paths themselves. It can be confusing when searching for a specific folder and getting results because of a specific directory.

I think a better option is to do a partial match on the current directory. Whether you are searching for files or not.

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+          size: child.totalFileCount,
```
GB907762, line 96: Many file explorers do not allow you to sort directories by size, since they may be ambiguous either meaning quantity of files or the total size of the directory.

Name ordering by default might be best?

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+            mutator={mutator}
+            releases={releases}
+            displayName={data.file.name.includes('/') ? data.file.name.split('/').pop() : undefined}
+          />
+        </Stack>
+      </Box>
+    )
+  }
+
+  const viewToggle = hasNested ? (
```
GB907762, line 181: What do you think the value is in having both of these views. I think possibly the legacy function should be removed and we should default to a folder structure?

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+          )}
+          <Paginate
+            list={files.map((f) => ({ key: f._id, ...f }))}
+            emptyListText='No files found'
+            searchFilterProperty='name'
+            sortingProperties={[
+              { value: 'name', title: 'Name', iconKind: 'text' },
+              { value: 'size', title: 'Size', iconKind: 'size' },
+              { value: 'createdAt', title: 'Date uploaded', iconKind: 'date' },
+              { value: 'updatedAt', title: 'Date updated', iconKind: 'date' },
```
GB907762, line 213: I'm unsure of what's causing it but these filters seem to be broken with a mixture of files and folders. As in the dates do not appear in order.

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+            emptyListText='No files in this folder'
+            searchFilterProperty='searchableText'
+            onSearchChange={setFolderSearchQuery}
+            sortingProperties={[
+              { value: 'name', title: 'Name', iconKind: 'text' },
+              { value: 'size', title: 'Size', iconKind: 'size' },
+              { value: 'createdAt', title: 'Date uploaded', iconKind: 'date' },
+              { value: 'updatedAt', title: 'Date updated', iconKind: 'date' },
+            ]}
+            searchPlaceholderText='Search files and folders'
```
GB907762, line 269: I'm unsure of what is causing this.

The search seems to match folders containing files suffixed with a whitespace.

For example `file.txt ` matches `folder/file.txt` and shows a folder. However this is not shown in the file view.

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+          spacing={2}
+          sx={{ alignItems: 'center', cursor: 'pointer', flex: 1 }}
+          onClick={() => onNavigate(node.fullPath)}
+        >
+          <Folder color='action' />
+          <Typography variant='h6'>{node.name}</Typography>
+          <Typography variant='caption' sx={{ width: 'max-content' }}>
+            {searchQuery && matchingCount !== totalCount
+              ? `${matchingCount} of ${totalCount} file${totalCount !== 1 ? 's' : ''} match`
+              : `${totalCount} file${totalCount !== 1 ? 's' : ''}`}
```
GB907762, line 378:
```suggestion
              ? `${matchingCount} of ${plural(totalCount, 'file')} match`
              : `plural(totalCount, 'file')}
```

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+        </Stack>
+        {canDelete && (
+          <>
+            <IconButton
+              size='small'
+              onClick={(e) => setAnchorEl(e.currentTarget)}
+              data-test={`folder-menu-${node.name}`}
+            >
+              <MoreVert />
+            </IconButton>
```
GB907762, line 389: Image

The styling between folders and files are inconsistent. For example the menu is misaligned and different colours. 

Personally something like we use for the `ExistingFileSelector` seems to be a more consistent form.

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+            <Menu anchorEl={anchorEl} open={Boolean(anchorEl)} onClose={() => setAnchorEl(null)}>
+              <MenuItem
+                onClick={() => {
+                  setAnchorEl(null)
+                  setDeleteOpen(true)
+                }}
+              >
+                <ListItemIcon>
+                  <Delete color='error' fontSize='small' />
```
GB907762, line 398: I agree that this should be red. In the file component could you update this with the same styling.

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+        title={`Delete folder "${node.name}"?`}
+        onConfirm={handleDeleteConfirm}
+        onCancel={() => setDeleteOpen(false)}
+        confirmLoading={isDeleting}
+        errorMessage={deleteError}
+      >
+        <Stack spacing={1}>
+          <Typography>
+            This will delete <strong>{allFilesInFolder.length}</strong> file{allFilesInFolder.length !== 1 ? 's' : ''}{' '}
+            in this folder and all sub-folders.
```
GB907762, line 417:
```suggestion
            This will delete <strong>{plural(allFilesInFolder.length, 'file')}</strong> in this folder and all
            sub-folders.
```

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+          </Typography>
+          {associatedReleases.length > 0 && (
+            <Typography color='error'>
+              Warning: {associatedReleases.length} release{associatedReleases.length !== 1 ? 's' : ''} reference files
```
GB907762, line 421:
```suggestion
              Warning: {plural(associatedReleases.length, 'release')} reference files
```

frontend/src/entry/model/files/FileBrowser.tsx
```diff
@@ -0,0 +1,437 @@
...
+              in this folder: {associatedReleases.map((r) => r.semver).join(', ')}
+            </Typography>
+          )}
+          <Typography variant='caption'>Files to be deleted:</Typography>
+          <Box sx={{ maxHeight: 200, overflow: 'auto', pl: 2 }}>
+            {allFilesInFolder.map((file) => (
+              <Typography key={file._id} variant='body2'>
+                {file.name}
```
GB907762, line 429: I wonder if there's some nicer styling with this?
If you have a long file name such as `thisisareallylongfoldernamethatihavemadetotestscrollbars/thisisareallylongfilenamethatihavemadetotestscrollbars.txt`, the text overflow's. Similar case with many files in a folder. I wonder if this could be truncated at least.

It may also be good if this could show the directories structure.

frontend/src/entry/model/files/FileUploadDialog.tsx
```diff
@@ -30,20 +31,28 @@ export default function FileUploadDialog({ open, onDialogClose, model, mutateMod
...
-          filesToBeUploaded.find((existingFile) => existingFile.file.name === newFile.file.name) === undefined,
+          filesToBeUploaded.find(
+            (existingFile) => (existingFile.uploadPath || existingFile.file.name) === newFile._dedupeKey,
+          ) === undefined,
       )
 
-      setFilesToBeUpload([...filteredNewFiles, ...filesToBeUploaded])
+      setFilesToBeUpload([...filteredNewFiles.map(({ _dedupeKey: _, ...rest }) => rest), ...filesToBeUploaded])
+      // Reset input value so re-selecting the same folder works
+      event.target.value = ''
```
GB907762, line 55: Is it worth having some sort of warning, if a file or folder, to be uploaded already exists?

I can see it probably being unobvious to users, if they accidently upload the same folder structure, if a file or folder already exists.

frontend/src/entry/model/files/FileUploadDialog.tsx
```diff
@@ -154,17 +164,21 @@ export default function FileUploadDialog({ open, onDialogClose, model, mutateMod
     <Dialog open={open} onClose={onDialogClose} maxWidth='md' fullWidth>
       <DialogContent>
         <Stack spacing={2}>
-          <label htmlFor='add-files-button' style={{ width: '100%', display: 'flex', justifyContent: 'center' }}>
+          <Stack direction='row' spacing={2} sx={{ justifyContent: 'center' }}>
+            <label htmlFor='add-files-button'>
+              <Button loading={isFilesUploading} endIcon={<FileUpload />} component='span' variant='outlined'>
```
GB907762, line 169: `isFilesUploading` is a shared state. Either use the drop-zone in another comment. Or disable both of these during upload since we already have a loading bar.

frontend/src/entry/model/files/FileUploadDialog.tsx
```diff
@@ -154,17 +164,21 @@ export default function FileUploadDialog({ open, onDialogClose, model, mutateMod
...
               loading={isFilesUploading}
-              endIcon={<FileUpload />}
-              component='span'
+              endIcon={<FolderOpen />}
               variant='outlined'
-              sx={{ width: '40%' }}
+              onClick={() => folderInputRef.current?.click()}
             >
-              Select files
+              Select folder
```
GB907762, line 179: This is a bit of tech debt but I have suggested a drop zone in the past. It may be better than a button?

https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/File_drag_and_drop

Do you have any thoughts?

frontend/src/entry/model/releases/ExistingFileSelector.tsx
```diff
@@ -121,6 +181,60 @@ export default function ExistingFileSelector({ model, existingReleaseFiles, onCh
     </ListItem>
   ))
 
+  const BrowseRow = ({ data }: { data: SelectorListItem }) => {
```
GB907762, line 184: Is it worth having a checkbox here for selecting folders for a release. You may wish to have supporting documents such as white papers and so forth that you wish to easily attach to a release.

frontend/src/entry/model/releases/ExistingFileSelector.tsx
```diff
@@ -137,21 +251,68 @@ export default function ExistingFileSelector({ model, existingReleaseFiles, onCh
...
-              { value: 'updatedAt', title: 'Date updated', iconKind: 'date' },
-            ]}
-            defaultSortProperty='createdAt'
-            searchFilterProperty='name'
-            searchPlaceholderText='Search by file name'
-          >
-            {FileRow}
-          </Paginate>
+          {hasNested ? (
+            <>
```
GB907762, line 255: I think some `Create Path` button here could be handy, especially when you don't want to use the folder structure on your local directory.

This could create a popup and change the current path on form completion, giving users naming flexibility.

frontend/src/entry/model/releases/ExistingFileSelector.tsx
```diff
@@ -137,21 +251,68 @@ export default function ExistingFileSelector({ model, existingReleaseFiles, onCh
...
+              </Breadcrumbs>
+              <Paginate
+                list={listItems}
+                emptyListText='No files found'
+                sortingProperties={[
+                  { value: 'name', title: 'Name', iconKind: 'text' },
+                  { value: 'size', title: 'Size', iconKind: 'size' },
+                  { value: 'createdAt', title: 'Date uploaded', iconKind: 'date' },
+                ]}
+                defaultSortProperty='name'
```
GB907762, line 292: This sorts folders and files alphabetically, meaning folders and files are mixed. I think a better option would be to sort folders first and then files?

I'd imagine folders should usually take precedence, since they might contain more generic data about the model.

frontend/src/common/MultiFileInput.tsx
GB907762, line 1: I think potentially changing the chips to something else, could be useful. If you have a really long names, chips truncate which can be difficult to determine what you are uploading. For example if you had different versions of weights. 

Image

GB907762 (changes requested):

JR40159 (changes requested):
Definitely agree with https://github.com/gchq/Bailo/pull/3856#discussion_r3620510301 and I think the directory structure of the files on a model in Bailo should be agnostic of the directory structure on the host machine when a file is uploaded. I think this is a key enabler to having multiple contributors on a model as you don't have to enforce all model contributors having the same directory structure on their host machine.

frontend/src/entry/model/releases/ExistingFileSelector.tsx
M422110: Ive added a "new folder" button in the file browser so that you can create empty directories like a normal file browser, and then the upload dialog is path aware so that it knows the relative directory to upload files to

https://github.com/gchq/Bailo/pull/3856/commits/0d4625c69706dd1bf8c29d43499d428f6b168445

frontend/src/entry/model/releases/ExistingFileSelector.tsx
M422110: yes i agree - have added this functionality here: [b778e46](https://github.com/gchq/Bailo/pull/3856/commits/b778e46515c4201010c93fed5de229813753273c)

frontend/src/entry/model/files/FileBrowser.tsx
M422110: have added a bulk delete endpoint for files: [e75bbb9](https://github.com/gchq/Bailo/pull/3856/commits/e75bbb9a440ff11b90da03e7536041e57ecf5cb3)

backend/src/routes/v2/model/file/getDownloadFile.ts
M422110: fixed in [5b92416](https://github.com/gchq/Bailo/pull/3856/commits/5b9241639bdeaaf03be4f18e8b46389d0fdfd33f)

frontend/src/entry/model/files/FileBrowser.tsx
M422110: I thought the flat structure may still be useful if you were only managing a few files - but agree it would make the ui less busy, and more consistent with the release file picker

frontend/src/common/MultiFileInput.tsx
M422110: thought about this, and I think the best approach would be to follow the same workflow as the file management page - upload folders/files in a dialogue and then display the files/folders as part of the release in the standard file management viewer rather than this view

frontend/src/entry/model/files/FileBrowser.tsx
M422110: Deprecated the old view in this commit - [371830c](https://github.com/gchq/Bailo/pull/3856/commits/371830c1e89c2718df51f67aec59f7819d402bd6)

frontend/src/common/MultiFileInput.tsx
M422110: Consolidate the mechanism for file upload in releases and file management so its more consistent. [da5671f](https://github.com/gchq/Bailo/pull/3856/commits/da5671fc7d61a4c4106a15be1389ec92c77b4589)

frontend/src/entry/model/files/FileBrowser.tsx
M422110: I think the default ordering is name - agree size is a little ambigious

frontend/src/entry/model/files/FileBrowser.tsx
M422110: Looks like this bug was to do with how the searchable text was created from the files path - updated the seperator and added test [68680fb](https://github.com/gchq/Bailo/pull/3856/commits/68680fb7671c702fb9dc29cd241fbb8c5d671cc4)

frontend/src/entry/model/files/FileBrowser.tsx
M422110: Ive updated the way folders dates are recorded, and also the sorting so it matches the UX of windows: [68680fb](https://github.com/gchq/Bailo/pull/3856/commits/68680fb7671c702fb9dc29cd241fbb8c5d671cc4) and [bee5c24](https://github.com/gchq/Bailo/pull/3856/commits/bee5c2479ce5a32029a564dea6f40c5ecd55833a)

frontend/src/entry/model/files/FileBrowser.tsx
M422110: [9fc9e98](https://github.com/gchq/Bailo/pull/3856/commits/9fc9e9818089d47834d23b9ce515e11b107b1b45)

frontend/src/entry/model/files/FileBrowser.tsx
M422110: Ive updated the row padding: [0044729](https://github.com/gchq/Bailo/pull/3856/commits/00447296b3015943649609f12bf0e9934640a447)

frontend/src/entry/model/releases/ExistingFileSelector.tsx
M422110: Folder/file sorting should now match what happens in windows i.e grouping the files and folders together and then sorting asc/desc within each group. The folder date is infered by the children - because folders are inferred by file name and not actually stored.

frontend/src/entry/model/files/FileBrowser.tsx
M422110: I think i prefer the UX of filtering on folder and file name here, would provide a better experience if you have a highly nested strucutre

frontend/src/entry/model/files/FileUploadDialog.tsx
M422110: added support for override detection and resolution: https://github.com/gchq/Bailo/pull/3856/commits/8f7c91daec01203ae8aa66470b213544270e1062

sylvesterkaczmarek (commented):
The new batch-delete path appears to allow cross-model file deletion because the supplied `modelId` is never bound to each `fileId`.

`removeFiles()` first loads the caller-supplied model A. For each ID it then calls `getFileById(user, fileId)` without passing model A, so a file from model B is accepted as long as the caller can *view* model B. The subsequent Delete check is `authorisation.file(user, modelA, fileB, FileAction.Delete)`, and `authorisation.files()` bases Delete permission on the supplied model roles; it does not verify `file.modelId === model.id`. A user who can manage model A and merely view a file in model B can therefore pass A + B's file ID, satisfy the Delete check via A, and reach `FileModel.findByIdAndDelete(fileB._id, session)`. The scan records for B are also deleted, while `removeFileFromReleases()` queries model A, potentially leaving B's release references inconsistent.

Could each file lookup be scoped to both `_id` and `modelId` (or explicitly reject `file.modelId !== modelId`) before any authorization/deletion work? I would add a regression test with a user who owns model A, can view model B, and submits B's file ID to A's batch-delete endpoint.

PR #3861, opened 2026-07-17
Edit mirrored model source id before import

PE39806:
Allow changing a mirrored model's (MM's) source ID before it has been mirrored, but not once it has been mirrored. This may be useful is a user has set up an MM with various collaborators etc. but made a typo in the source ID, by not needing to delete and create an entirely new MM.

Also updates the python client to support `settings` in the API requests, and adds a mutate to the frontend model & data card pages to immediately reflect updated settings (source model ID, name, description).

TODO: Update Python request handling of:
- [x] get model
- [x] post model
- [x] patch model


frontend/src/entry/settings/EntryDetails.tsx
```diff
@@ -125,6 +136,23 @@ export default function EntryDetails({ entry }: EntryDetailsProps) {
           <EntryNameInput autoFocus value={name} kind={entry.kind} onChange={(value) => setName(value)} />
           <EntryDescriptionInput value={description} onChange={(value) => setDescription(value)} />
         </>
+        {isMirroredModel && (
+          <>
+            <Divider />
+            <Typography variant='h6' component='h2'>
+              Source model
```
GB907762: Does this extra heading need to be here? 

My thought is `Source model` and `Source Model ID` are very similar

frontend/src/entry/settings/EntryDetails.tsx
```diff
@@ -125,6 +136,23 @@ export default function EntryDetails({ entry }: EntryDetailsProps) {
...
+            </Typography>
+            {isImported ? (
+              <MessageAlert
+                message={`Source Model ID: ${entry.settings.mirror?.sourceModelId}`}
+                subHeading='The source model ID cannot be changed after import.'
+                severity='info'
+              />
+            ) : (
+              <SourceModelInput value={sourceModelId} onChange={(value) => setSourceModelId(value)} />
```
GB907762: Do you have any thoughts about making the ID at the top of the screen editable like with states and organisations?

I think the location on the settings page is potentially quite hidden.

GB907762 (changes requested):

frontend/src/entry/settings/EntryDetails.tsx
PE39806: This is a good idea - I will update the mirrored model `MessageAlert` in `frontend/pages/model/[modelId]/index.tsx` to allow editing the source model ID from there, and remove the new Settings section.

frontend/src/entry/settings/EntryDetails.tsx
PE39806: No longer needed as this section has been removed.

GB907762 (approved):

ON93754 (approved):

PR #3862, opened 2026-07-17
Dark mode

ARADDCC002:


PE39806 (changes requested):
This is a really good start. I've had a play and have noticed the following:
- [x] clickable chips (e.g. in the marketplace Filters) do not change appearance on selection
- [x] Text field "Search by full text" in the marketplace Filters has no border/background differentiation until mouse hover
- [x] Text colour in the UI config banner is affected by dark mode (but the bg colour is not)
- [x] Contrast between the top app bar and main content is not enough for clear distinction
- [x] Text quotes in doc pages and other markdown `> ` have a light bg
- [x] Text character for User circles is affected by theme but colour is not, making them hard to read
- [x] `frontend/public/emptyBlob.svg` contains black elements so looks odd in dark mode (possibly just need a dark-mode variant)
- [x] `entitySelector` (e.g. in access request from `backend/src/scripts/example_schemas/minimal_access_request_schema.json`) has very light bg with light text so lacks contract
- [x] `outline` buttons lack contrast against lighter backgrounds (e.g. model > settings > mirrored models > releases to export ? button) (`HelpDialog`)
- [x] URLs in a11y pages have poor contrast
- [x] "Ready for review" banner has light text on light bg
- [x] Python client docs icon `<>` is same colour as the document image behind it
- [x] Schema comparison diff still uses light theme

Does this PR also aim to update the `lib/landing` pages with a dark mode? We could use the same approach as the Swagger docs and add a clickable button for light/dark mode in the app bar, as there is no settings page. There may even be a way to pass this to the Swagger page in `lib/landing`.

ARADDCC002:
> This is a really good start. I've had a play and have noticed the following:
> 
>     * [ ]  clickable chips (e.g. in the marketplace Filters) do not change appearance on selection
> 
>     * [ ]  Text field "Search by full text" in the marketplace Filters has no border/background differentiation until mouse hover
> 
>     * [ ]  Text colour in the UI config banner is affected by dark mode (but the bg colour is not)
> 
>     * [ ]  Contrast between the top app bar and main content is not enough for clear distinction
> 
>     * [ ]  Text quotes in doc pages and other markdown `> ` have a light bg
> 
>     * [ ]  Text character for User circles is affected by theme but colour is not, making them hard to read
> 
>     * [ ]  `frontend/public/emptyBlob.svg` contains black elements so looks odd in dark mode (possibly just need a dark-mode variant)
> 
>     * [ ]  `entitySelector` (e.g. in access request from `backend/src/scripts/example_schemas/minimal_access_request_schema.json`) has very light bg with light text so lacks contract
> 
>     * [ ]  `outline` buttons lack contrast against lighter backgrounds (e.g. model > settings > mirrored models > releases to export ? button) (`HelpDialog`)
> 
>     * [ ]  URLs in a11y pages have poor contrast
> 
>     * [ ]  "Ready for review" banner has light text on light bg
> 
>     * [ ]  Python client docs icon `<>` is same colour as the document image behind it
> 
>     * [ ]  Schema comparison diff still uses light theme
> 
> 
> Does this PR also aim to update the `lib/landing` pages with a dark mode? We could use the same approach as the Swagger docs and add a clickable button for light/dark mode in the app bar, as there is no settings page. There may even be a way to pass this to the Swagger page in `lib/landing`.

Very good feedback, thank you! I think overall I don't mind a _few_ things not to be 100% up to par, as even if released this would be a beta feature to improve over time, but next time I work on this I'll work for a few on this list.

I'm happy for `lib/landing` to be handled on a different PR.

PE39806 (dismissed):

frontend/pages/_document.tsx
```diff
@@ -17,6 +17,17 @@ export default function MyDocument(props: DocumentProps & DocumentHeadTagsProps)
...
+          {/* Set the markdown editor colour mode before hydration to avoid a flash of the wrong theme. */}
+          <script
+            dangerouslySetInnerHTML={{
+              __html: `
+                try {
+                  var d = localStorage.getItem('dark_mode_enabled') === 'true';
+                  document.documentElement.setAttribute('data-color-mode', d ? 'dark' : 'light');
+                } catch (e) {}
+              `,
+            }}
```
GB907762, line 29: I'd prefer the use of cookies, so this persists between browsers.

frontend/src/settings/ProfileTab.tsx
```diff
@@ -17,6 +24,26 @@ export default function ProfileTab({ user }: ProfileTabProps) {
...
+          variant='body1'
+          sx={{
+            fontWeight: 'bold',
+          }}
+        >
+          Appearance
+        </Typography>
+        <FormControlLabel
+          control={
+            <Switch
```
GB907762: In future having a system toggle as well as dark and light would be good.

So whatever your browser defaults to.

frontend/src/settings/ProfileTab.tsx
```diff
@@ -17,6 +24,26 @@ export default function ProfileTab({ user }: ProfileTabProps) {
...
+              checked={isDark}
+              onChange={toggleDarkMode}
+              slotProps={{ input: { 'aria-label': 'Toggle dark mode' } }}
+            />
+          }
+          label='Dark mode (beta)'
```
GB907762: I'd put some sort of disclaimer about a11y here. So it's clear we don't support it at the moment.

frontend/public/emptyBlobLight.svg
GB907762, line 1: This doesn't have great contrast with light mode. Look into the `EmptyBlob` component. Maybe some switch is needed?

GB907762 (changes requested):

frontend/public/emptyBlobLight.svg
PE39806: The new `<style>` & `@media (prefers-color-scheme: dark)` allows changing the colours depending on the theme. Do you have any suggestions for what these could be for dark mode?

frontend/public/emptyBlobDark.svg
PE39806, line 1: Remove this and, instead, just update the new `<style>` `@media (prefers-color-scheme: dark)` in `frontend/public/emptyBlob.svg` to change the class colours to the values chosen here.

PE39806 (changes requested):

frontend/public/emptyBlobLight.svg
GB907762: [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-color-scheme) uses the system theme for the operating system. Where as we are evaluating the theme using localStorage/cookies. 

These could be misaligned.

frontend/pages/_document.tsx
ON93754: We use local storage in a couple of places already so that would be good for persistence

frontend/public/emptyBlobLight.svg
PE39806, line 1: After some further reading, it looks like I have misunderstood how `@media (prefers-color-scheme: dark)` works as it follows the OS preference, ignoring MUI's theme. Sorry!
I think the light & dark approach you did in https://github.com/gchq/Bailo/pull/3862/commits/32d9d5c4623ba01eb93994b6477045ed299ccb05 will have to return, although you could probably simplify it to
```ts
  const theme = useTheme()
  const src = theme.palette.mode === 'dark' ? '/emptyBlobDark.svg' : '/emptyBlobLight.svg'

  ...
    <Image src={src} alt='Empty blob' width={120} height={120} data-test='emptyBlobImage' />
```

PE39806 (changes requested):

frontend/pages/_document.tsx
GB907762: `localStorage` are not interchangeable with `cookies` since cookies can be modified and updated by the server allowing cookies from the same domain or different domains (with CORS).

Although I'm happy that the approach of using `localStorage` in this case makes sense.

PE39806 (dismissed):

PE39806 (approved):

GB907762 (approved):

PR #3864, opened 2026-07-21
Registry action auditing

JRB66955:


backend/src/connectors/audit/Base.ts
```diff
@@ -44,6 +45,7 @@ export const ResourceKind = {
   ArtefactScanning: 'artefact scanning',
   Metric: 'metric',
   User: 'user',
+  Registry: 'reigstry',
```
PE39806:
```suggestion
  Registry: 'registry',
```

backend/src/routes/v1/registryAuth.ts
```diff
@@ -347,7 +359,7 @@ export const getDockerRegistryAuth = [
     const authResults = await Promise.all(requestedAccesses.map((a) => checkAccess(a, user, admin)))
     const grantedAccesses: Access[] = []
 
-    authResults.forEach((result, idx) => {
+    authResults.forEach(async (result, idx) => {
```
PE39806: This for loop is not `await`ed. Maybe use some `for (const [idx, result] of authResults.entries()) {` for sequential, `await`ed processing?

backend/src/routes/v1/registryAuth.ts
```diff
@@ -360,6 +372,19 @@ export const getDockerRegistryAuth = [
...
+        await audit.onRegistryImagePush(req, user)
+      }
+      if (access.actions.includes('pull')) {
+        req.audit = AuditInfo.RegistryImagePull
+        await audit.onRegistryImagePull(req, user)
+      }
+      if (access.actions.includes('delete')) {
+        req.audit = AuditInfo.RegistryImageDelete
+        await audit.onRegistryImageDelete(req, user)
+      }
```
PE39806: These aren't strictly the push, pull or delete actions themselves, but are granting access to then perform these operations. More accurate audit names:
```ts
RegistryAuthorisePull
RegistryAuthorisePush
RegistryAuthoriseDelete
```

The actual push, pull, delete etc. could be audited by extending the existing webhook receiver `backend/src/routes/internal/registry/events.ts`.

PE39806: Also, this is in a loop so each `req.audit = ...` overwrites the previously set value.

backend/src/connectors/audit/stdout.ts
```diff
@@ -445,6 +446,42 @@ export class StdoutAuditConnector extends BaseAuditConnector {
     req.log.info(event, req.audit.description)
   }
 
+  async onRegistryLogin(req: Request, user: UserInterface): Promise<void> {
+    this.checkEventType(AuditInfo.RegistryLogin, req)
+    const event = this.generateEvent(req, { userDn: user.dn })
```
PE39806: If we can get extra detail like the `repository` and `actions` from `req` for these then definitely add them to this event.

PE39806 (changes requested):

backend/src/routes/v1/registryAuth.ts
```diff
@@ -323,6 +327,14 @@ export const getDockerRegistryAuth = [
       // User requesting no scope, they're just trying to login
       // Because this token has no permissions, it is safe to
       // provide.
+
+      req.audit = AuditInfo.RegistryLogin
+      await audit.onRegistryLogin(req, user)
+
+      // necessary?
```
ON93754: Does this comment need actioning? Removing?

backend/src/routes/v1/registryAuth.ts
```diff
@@ -376,6 +400,9 @@ export const getDockerRegistryAuth = [
       throw Forbidden({ requestedAccesses }, 'Requested image is not accessible - no authorised scopes.', rlog)
     }
 
+    req.audit = AuditInfo.RegistryIssueAccessToken
+    await audit.onRegistryIssueAccessToken(req, user)
+
     const accessToken = await issueAccessToken(user, grantedAccesses)
```
PE39806: Usually we audit after the event
```suggestion
    const accessToken = await issueAccessToken(user, grantedAccesses)

    req.audit = AuditInfo.RegistryIssueAccessToken
    await audit.onRegistryIssueAccessToken(req, user)
```

backend/src/connectors/audit/Base.ts
```diff
@@ -443,6 +445,42 @@ export const AuditInfo = {
...
+  },
+  RegistryIssueAccessToken: {
+    typeId: 'RegistryIssueAccessToken',
+    description: 'Registry has issued an access token for a user',
+    auditKind: AuditKind.Create,
+    resourceKind: ResourceKind.Registry,
+  },
+  RegistryIssueRefreshToken: {
+    typeId: 'RegistryIssueRefreshToken',
+    description: 'Registry has refreshed an already issues access token for a user',
```
PE39806:
```suggestion
    description: 'Registry has refreshed an already issued access token for a user',
```

PE39806 (changes requested):

PE39806 (dismissed):

backend/src/connectors/audit/Base.ts
```diff
@@ -443,6 +444,42 @@ export const AuditInfo = {
...
+    auditKind: AuditKind.View,
+    resourceKind: ResourceKind.Registry,
+  },
+  RegistryAuthorisePush: {
+    typeId: 'RegistryAuthorisePush',
+    description: 'Registry has authorised a user to push an image',
+    auditKind: AuditKind.Create,
+    resourceKind: ResourceKind.Registry,
+  },
+  RegistryAuthoriseDelete: {
```
PE39806: I think these names are a bit misleading as the webhook events happens when these actions have occurred rather than when they are authorised. Prefer `RegistryImagePulled` etc.

backend/src/routes/internal/registry/events.ts
```diff
@@ -65,6 +67,20 @@ export const handleRegistryEvents = [
      */
 
     for (const event of events) {
+      const user = event.actor?.name ?? ''
+
+      if (event?.action === 'pull') {
+        const auditReq = Object.assign(Object.create(req), { audit: AuditInfo.RegistryAuthorisePull }) as Request
```
PE39806: Repeated use of `Object.assign(Object.create(req) ,...)` - maybe extract so some helper?
```ts
function withAudit(req: Request, auditInfo: AuditInfoKeys): Request {
  return Object.assign(Object.create(req), { audit: auditInfo })
}
```
then call with
```
await audit.onRegistryImagePull(
  withAudit(req, AuditInfo.RegistryAuthorisePull),
  user,
)
```

backend/src/connectors/audit/Base.ts
```diff
@@ -443,6 +444,42 @@ export const AuditInfo = {
...
+    auditKind: AuditKind.View,
+    resourceKind: ResourceKind.Registry,
+  },
+  RegistryIssueAccessToken: {
+    typeId: 'RegistryIssueAccessToken',
+    description: 'Registry has issued an access token for a user',
+    auditKind: AuditKind.Create,
+    resourceKind: ResourceKind.Registry,
+  },
+  RegistryIssueRefreshToken: {
```
PE39806: Are these three actually used?

PE39806 (changes requested):

backend/src/connectors/audit/Base.ts
```diff
@@ -443,6 +444,24 @@ export const AuditInfo = {
...
+    description: 'Registry has authorised a user to push an image',
+    auditKind: AuditKind.Create,
+    resourceKind: ResourceKind.Registry,
+  },
+  RegistryAuthoriseDeleted: {
+    typeId: 'RegistryAuthoriseDeleted',
+    description: 'Registry has authorised a user to delete an image',
+    auditKind: AuditKind.Delete,
+    resourceKind: ResourceKind.Registry,
+  },
```
PE39806, line 464: These names & descriptions don't really match how the types are used.

PE39806 (changes requested):

backend/src/routes/v1/registryAuth.ts
PE39806, line 1: Are these changes needed?

PE39806 (changes requested):

backend/src/connectors/audit/Base.ts
JRB66955: I've updated these to be consistent with the audit function names now (e.g. `RegistryImagePushed` etc.)

PE39806 (approved):

ARADDCC002 (approved):

PR #3872, opened 2026-07-21
Implemented CSV download

ON93754:
- Added a button to the entries tab in the metrics page to export as CSV


frontend/utils/csvUtils.ts
PE39806, line 1: Typo in file name, correct to `csvUtils.ts`

frontend/src/metrics/EntryMetrics.tsx
```diff
@@ -117,6 +130,8 @@ export default function EntryMetrics() {
             selectedOrganisation={selectedValue('organisation')}
             exportDocumentTitle='Bailo entry metrics'
```
PE39806:
```suggestion
            exportDocumentTitle={exportDocumentTitle}
```

frontend/src/metrics/components/MetricsHeader.tsx
```diff
@@ -93,10 +97,15 @@ export default function MetricsHeader({
               <em>Last updated {formatDateStringWithMinutes(lastUpdated)}</em>
             </Typography>
           </Stack>
-          <Stack>
+          <Stack direction='row' spacing={1} sx={{ alignItems: 'center' }}>
             <Button variant='contained' onClick={() => setDialogOpen(true)}>
               Export as PDF
             </Button>
+            {csvExportEnabled && (
+              <Button variant='contained' onClick={onCsvExport}>
```
PE39806: Could disable this when there's no data to export.

frontend/utils/csvUtils.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * Escapes a single CSV field. Wraps the value in quotes (and doubles any
+ * embedded quotes) if it contains a comma, quote, or newline.
+ */
+function escapeCsvField(value: string): string {
```
PE39806, line 11: For some maliciously crafted value, this will not protect against spreadsheet injection e.g. `'=HYPERLINK("https://malicious.site")'`. I think this can be fixed by prefixing "dangerous" characters with a `'` e.g.
```ts
const dangerousPrefixes = /^[=+\-@]/

const safeValue = dangerousPrefixes.test(value)
  ? `'${value}`
  : value
```
but please double check this.

frontend/utils/test/csvUtils.test.ts
PE39806, line 1: Add tests for:
- escaping empty values
- escaping values containing carriage return(s)
- preventing spreadsheet formula injection
- handling empty datasets
- handling embedded quotes and commas together

PE39806 (changes requested):

frontend/src/metrics/components/MetricsHeader.tsx
```diff
@@ -15,6 +15,8 @@ interface MetricsHeaderProps {
   onMetricChange?: (newMetric: SelectedMetricKindKeys) => void
   exportDocumentTitle: string
   titleObjectType?: string
+  csvExportEnabled?: boolean
+  onCsvExport?: () => void
```
GB907762: Why is this optional. Should cause the button to be unclickable if it's `undefined`

I'd prefer proper error handling. Or even removing `csvExportEnabled` in favour for the function.

frontend/src/metrics/components/MetricsHeader.tsx
```diff
@@ -93,10 +97,15 @@ export default function MetricsHeader({
...
           </Stack>
-          <Stack>
+          <Stack direction='row' spacing={1} sx={{ alignItems: 'center' }}>
             <Button variant='contained' onClick={() => setDialogOpen(true)}>
               Export as PDF
             </Button>
+            {csvExportEnabled && (
+              <Button variant='contained' onClick={onCsvExport}>
+                Export as CSV
+              </Button>
```
GB907762: I think it's worth having some confirmation so it's clear it's being downloaded. To me it's not obvious that this button downloads the CSV.

frontend/src/metrics/EntryMetrics.tsx
```diff
@@ -62,6 +63,18 @@ export default function EntryMetrics() {
     [entries],
   )
 
+  const exportDocumentTitle = 'Bailo entry metrics'
+
+  const handleCsvExport = useCallback(() => {
+    const headers = ['Model ID', 'Model Name', 'Owner']
```
GB907762: These headers already appear in the table. I feel like this is an effort to maintain if you added some more fields in the `MetricsBreakdownTable`. Could you make these consistent so there is one source of truth.

frontend/src/metrics/EntryMetrics.tsx
```diff
@@ -62,6 +63,18 @@ export default function EntryMetrics() {
...
+    const rows = tableData.map((row) => [
+      row.entryId,
+      row.entryName,
+      row.modelOwners.length > 0 ? row.modelOwners.join('; ') : '',
```
GB907762:
```suggestion
      row.modelOwners.join('; ')
```
Also what happens if a `modelOwner` has a semi-colon in the string.

frontend/src/metrics/EntryMetrics.tsx
```diff
@@ -62,6 +63,18 @@ export default function EntryMetrics() {
...
+    ])
+    downloadCsv(exportDocumentTitle, headers, rows)
```
GB907762: A timestamp on the filename would be useful here. And perhaps `exportDocumentTitle` should be kebab case.

frontend/src/metrics/EntryMetrics.tsx
GB907762:
```suggestion
            exportDocumentTitle={exportDocumentTitle }
```

frontend/utils/csvUtils.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (/[",\n]/.test(value)) {
+    return `"${value.replace(/"/g, '""')}"`
+  }
+  return value
+}
```
GB907762, line 18:
```suggestion
 * Escapes a single CSV field. Wraps the value in quotes (and doubles any
 * embedded quotes) if it contains a comma, quote, newline or carriage return.
 */
function escapeCsvField(value: string): string {
  if (/[",\r\n]/.test(value)) {
    return `"${value.replace(/"/g, '""')}"`
  }
  return value
}
```
It's unlikely with the current data but if programmatic access is used a carriage return may occur in the data.

https://www.ietf.org/rfc/rfc4180.txt

Bellow in `toCsvString` You've already used a `CRLF` to delimit new lines.

GB907762 (changes requested):

PE39806 (dismissed):

GB907762 (dismissed):

GB907762 (approved):

PE39806 (approved):

PR #3880, opened 2026-07-22
Bugfix/bai fix number of entries mismatch

ON93754:
- Fixed mismatching counts for schema's
- Improved pie chart UX
- Added array of kind as an additional filter parameter to the endpoint


frontend/src/metrics/components/MetricsBreakdownTable.tsx
```diff
@@ -5,14 +5,20 @@ import { EmptyRow } from 'src/common/table/EmptyRow'
...
   title?: string
   data: ModelBreakdown[]
   isLoading?: boolean
 }
 
+function getLinkRoute(entryKind?: string | null): string {
+  // Return the model route for all entries other than data cards
+  return entryKind ? (entryKind === EntryKind.DATA_CARD ? EntryKind.DATA_CARD : EntryKind.MODEL) : ''
+}
+
```
GB907762: Use `entryKindForRedirect`

GB907762 (changes requested):
Could you also look through the code and identify places `kind` is used. I think this should use `kinds`?

GB907762 (dismissed):

backend/src/connectors/metrics/base.ts
```diff
@@ -32,6 +32,7 @@ import {
 } from './metricUtils.js'
 
 const METRICS_CACHE_TTL = 5 * 60 // 5 minutes
+const MODELS_KIND_FILTER = { kind: { $in: [EntryKind.Model, EntryKind.MirroredModel, EntryKind.UntrustedModel] } }
```
PE39806: I think this `const` is only used once - possibly left over from testing?

frontend/actions/metrics.ts
```diff
@@ -64,6 +64,7 @@ export function useGetNoReleasesPolicyMetrics() {
 export interface ModelBreakdownResponse {
   entryId: string
   entryName: string
+  entryKind: string
```
PE39806:
```suggestion
  entryKind: EntryKindKeys
```
and update `import { ... } from 'types/types'`

frontend/src/metrics/components/MetricsBreakdownTable.tsx
```diff
@@ -31,12 +33,17 @@ export function MetricsBreakdownTable({ title, data, isLoading = false }: Metric
       >
         <TableCell component='th' scope='row'>
           <Typography sx={{ maxWidth: '500px' }}>
-            <Link href={`/model/${row.entryId}`} target='_blank' rel='noopener noreferrer'>
+            <Link
+              href={`/${entryKindForRedirect(row.entryKind as EntryKindKeys)}/${row.entryId}`}
```
PE39806: No need to typecast with updated `ModelBreakdownResponse` above
```suggestion
              href={`/${entryKindForRedirect(row.entryKind)}/${row.entryId}`}
```

frontend/types/types.ts
```diff
@@ -1044,6 +1044,7 @@ export interface PolicyRoleMetrics {
 export interface ModelBreakdown {
   entryId: string
   entryName: string
+  entryKind: string
```
PE39806: Possibly
```suggestion
  entryKind: EntryKindKeys
```

PE39806 (changes requested):

PE39806 (approved):

PR #3882, opened 2026-07-22
Add searchable documentation

GB907762:


frontend/package.json
```diff
@@ -53,10 +56,12 @@
     "react-dom": "19.2.7",
     "react-to-print": "^3.3.0",
     "rehype-highlight": "^7.0.1",
+    "rehype-slug": "^6.0.0",
     "remark-frontmatter": "^5.0.0",
     "remark-gfm": "4.0.1",
     "semver": "^7.8.1",
     "swr": "^2.4.1",
+    "tsx": "^4.23.1",
```
PE39806, line 71: Should this be in `devDependencies`?

frontend/src/wrapper/EntrySearch.tsx
PE39806, line 1: I really like the new docs search, but I wish it didn't replace this search but, rather, added to it. There are a few different ways this could be handled - e.g. splitting the search results into two distinct EntrySearch and DocsSearch sections, or using toggleable search filters to enable/disable searching various parts of the application - and I don't mind which is chosen. The reason for this is that I have seen users make heavy use of the top search bar so I wouldn't be happy about removing this as a feature for them (even with the marketplace search being untouched).

PE39806 (changes requested):

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,427 @@
...
+ */
+function resultHref(result: DocsSearchResult): string {
+  const base = `/docs/${result.slug}`
+  return result.hash ? `${base}#${result.hash}` : base
+}
+
+/**
+ * Build the URL for a page slug (no hash).
+ */
+function pageHref(slug: string): string {
```
ON93754: Could we improve the names of these functions `resultHref` and `pageHref` to something more descriptive?
Perhaps something more like `toResultHref` and `toPageHref` or `generateResultHref` and `generatePageHref` etc?

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,427 @@
...
+      result: DocsSearchResult
+    }
+
+/**
+ * Group flat search results by their parent page slug. Each group starts with a
+ * `page-header` row (using the page-level result if present, otherwise synthesised
+ * from a heading's breadcrumb) followed by any heading children in the order they
+ * were returned by the search index.
+ */
+function groupResults(results: DocsSearchResult[]): DisplayRow[] {
```
ON93754: This function is quite long, could we break some of the logic out into functions to reduce the cognitive load?

Something like this perhaps?

```typescript
function getHeaderMetadata(result: DocsSearchResult) {
  if (result.kind === 'page') {
    return {
      title: result.title,
      breadcrumb: result.breadcrumb,
    }
  }

  const [pageTitle, ...rest] = result.breadcrumb.split(' — ')

  return {
    title: pageTitle || result.slug,
    breadcrumb: rest.join(' — '),
  }
}
```

```typescript
function createPageHeader(
  rows: DisplayRow[],
  result: DocsSearchResult,
): number {
  const { title, breadcrumb } = getHeaderMetadata(result)

  const headerIndex = rows.length

  rows.push({
    kind: 'page-header',
    key: `page:${result.slug || '__root__'}`,
    slug: result.slug,
    title,
    breadcrumb,
    result: result.kind === 'page' ? result : undefined,
  })

  return headerIndex
}
```

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,427 @@
...
+
+  name: 'ShortcutButton',
+  slot: 'Root',
+
+  overridesResolver: (props, styles) => [
+    styles.root,
+    props.color === 'white' && styles.white,
+    props.color === 'black' && styles.black,
+  ],
+})(({ theme, color = 'white' }) => {
```
ON93754:
```suggestion
})(({ theme, color = theme.palette.common.white }) => {
```

frontend/src/docs/useDocsSearchIndex.ts
```diff
@@ -0,0 +1,467 @@
...
+const MAX_HEADING_RESULTS_PER_PAGE = 3
+const SNIPPET_RADIUS = 60
+const MIN_TOKEN_LENGTH = 2
+
+/**
+ * Common English stop words that we don't want to score or highlight on their
+ * own (they inflate scores and produce noisy `<mark>` spans on words like "a",
+ * "is", "the", etc.). They *are* still allowed to participate in phrase matches.
+ */
+const STOP_WORDS = new Set<string>([
```
ON93754: Consider moving these to a constants file

frontend/src/docs/useDocsSearchIndex.ts
```diff
@@ -0,0 +1,467 @@
...
+    .replace(/'/g, '&#39;')
+}
+
+/**
+ * Wrap matches of the query phrase and/or individual tokens in `<mark>` tags.
+ * The phrase is listed first in the alternation so wherever it occurs it gets
+ * highlighted as a single contiguous span; adjacent `<mark>`s separated only by
+ * whitespace are also merged so the visual highlight bar spans that whitespace.
+ */
+function highlight(source: string, terms: QueryTerms): string {
```
ON93754, line 203: Worth some tests for this function?

- Returns escaped source unchanged when there are no terms
- Highlights a single token
- Matching is case-insensitive

frontend/src/docs/useDocsSearchIndex.ts
```diff
@@ -0,0 +1,467 @@
...
+  const tokens = raw.filter((t) => t.length >= MIN_TOKEN_LENGTH && !STOP_WORDS.has(t))
+  return {
+    tokens,
+    phrase,
+    isPhrase: raw.length > 1,
+  }
+}
+
+/** Number of times `needle` occurs in `haystack`, capped at 20 to keep scoring bounded. */
+function countOccurrences(haystack: string, needle: string): number {
```
ON93754, line 270: Worth a few tests

- Returns 0 for an empty needle e.g `countOccurences('hello', '')`
- Returns 0 when needle is not present
- Returns 1 for a single occurrence
- Counts multiple occurrences

frontend/src/docs/useDocsSearchIndex.ts
```diff
@@ -0,0 +1,467 @@
...
+    }
+    if (tokenScore === 0) {
+      return 0
+    }
+    score += tokenScore
+  }
+  return score
+}
+
+function scoreHeading(heading: IndexedHeading, terms: QueryTerms): number {
```
ON93754, line 336: Worth a test to ensure scoring is working as expected?

frontend/src/docs/useDocsSearchIndex.ts
```diff
@@ -0,0 +1,467 @@
...
+  let count = 0
+  let i = 0
+  while ((i = haystack.indexOf(needle, i)) !== -1 && count < 20) {
+    count++
+    i += needle.length
+  }
+  return count
+}
+
+function scorePage(entry: IndexedEntry, terms: QueryTerms): number {
```
ON93754, line 290: Worth a test to ensure scoring is working as expected?

frontend/src/wrapper/EntrySearch.tsx
GB907762: I'll take a look. I think having dialog that's capable of searching the market place and the user docs would be great. Just need to be a bit thoughtful about the UX.

frontend/package.json
GB907762: I don't think the index will compile correctly with a prebuild.

frontend/src/docs/DocsSearch.tsx
GB907762: I think `getRowHref` is clear

frontend/src/docs/useDocsSearchIndex.ts
GB907762: I've removed these. I don't see clear value in having them.

frontend/src/docs/DocsSearch.tsx
GB907762: While I agree the SRP is good practice in many cases to reduce cognitive load, I think in this case over fragmentation would make it hard to track the behaviour and purpose. Particularly if you tried reusing this code. I think the comments and expressive code should help maintain this area of the code base.

I also feel this is an atomic algorithm that shouldn't be divided encase of misuse or unintended regressions.

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+    whiteSpace: 'nowrap',
+
+    '&:hover': {
+      color: foreground,
+      borderColor: foreground,
+      backgroundColor:
+        color === 'black' ? alpha(theme.palette.common.black, 0.08) : alpha(theme.palette.common.white, 0.15),
+    },
+
+    '&:focus-visible': {
```
IR96334: I've not been able to invoke this through **normal** use of the application. Pressing tab only highlights the search bar, then moves onto a different component. Think it can be removed

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+} from '@mui/material'
+import { alpha, styled, useTheme } from '@mui/material/styles'
+import { useRouter } from 'next/router'
+import { KeyboardEvent, ReactElement, useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { Transition } from 'src/common/Transition'
+import useIsMac from 'src/hooks/useIsMac'
+
+import { DocsSearchResult, SearchFilter, useDocsSearchIndex } from './useDocsSearchIndex'
+
+type SearchCategory = SearchFilter
```
IR96334: Same type exists inside ./useDocsSearchIndex although this type includes 'all' while external doesn't.

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+type ResultCategory = Exclude<SearchCategory, 'all'>
```
IR96334: This is basically the SearchCategory in ./useDocsSearchIndex, but it has a different name.

frontend/src/docs/useDocsSearchIndex.ts
```diff
@@ -0,0 +1,626 @@
...
+
+export interface RawIndexEntry {
+  slug: string
+  title: string
+  text: string
+  headings: RawHeading[]
+}
+
+export type DocsSearchResultKind = 'page' | 'heading'
+export type SearchCategory = 'docs' | 'datacards' | 'models'
```
IR96334: As mentioned in DocsSearch comments, the naming is confusing and some types are redundant. Centralise types and make naming more obvious to function.

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+        </Tooltip>
+      )}
+      <Dialog
+        disableRestoreFocus
+        open={open}
+        onClose={closeDialog}
+        slots={{ transition: Transition }}
+        fullWidth
+        maxWidth='lg'
+        scroll='paper'
```
IR96334: don't think this does anything

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+    setQuery('')
+    setCategory('all')
+    setHighlightedIndex(0)
+  }, [])
+
+  const selectCategory = useCallback((nextCategory: SearchCategory) => {
+    setCategory(nextCategory)
+    setHighlightedIndex(0)
+
+    requestAnimationFrame(() => {
```
IR96334: Don't think this does anything.

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+              },
+              textTransform: 'capitalize',
+              height: 'max-content',
+            }}
+          >
+            <SearchIcon />
+          </IconButton>
+        </Tooltip>
+      )}
+      <Dialog
```
IR96334: This file is large, would prefer it to be separated. Move dialog to different component.

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+              pb: 1.5,
+              overflowX: 'auto',
+            }}
+            aria-label='Filter search results'
+          >
+            {CATEGORY_OPTIONS.map((option) => {
+              const selected = category === option.value
+
+              return (
+                <Chip
```
IR96334: We have ChipSelector for this use case. When using ChipSelector I've noticed the prop types are incorrect. Change to: 
```
type PartialChipSelectorProps =
  | {
      multiple: true
      selectedChips: string[]
      onChange: (value: string[]) => void
    }
  | {
      multiple?: false
      selectedChips: string
      onChange: (value: string) => void
    }

type ChipSelectorProps = {
  label?: string
  subheading?: string
  size?: 'small' | 'medium'
  expandThreshold?: number
  chipTooltipTitle?: string
  ariaLabel?: string
  accordion?: boolean
  variant?: 'filled' | 'outlined'
  icon?: ReactElement
  style?: CSSProperties
  options: string[]
  unreachableOptions?: string[]
} & PartialChipSelectorProps
```

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+              })}
+            >
+              <Typography variant='caption'>
+                Models and datacards could not be loaded. Documentation search remains available.
+              </Typography>
+            </Box>
+          )}
+          {isLoading && displayRows.length === 0 ? (
+            <Box sx={{ p: 3 }}>
+              <Typography variant='body2' color='text.secondary'>
```
IR96334: Don't think the color is doing much

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+                {loadingMessage}
+              </Typography>
+            </Box>
+          ) : displayRows.length === 0 ? (
+            <Box sx={{ p: 3 }}>
+              <Typography variant='body2' color='text.secondary'>
```
IR96334: same here

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+                <Box
+                  component='li'
+                  key={section.category}
+                  sx={{
+                    display: 'block',
+                    listStyle: 'none',
+                  }}
+                >
+                  <Typography
+                    component='div'
```
IR96334: Does this do anything?

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+                    variant='overline'
+                    color='text.secondary'
```
IR96334: textSecondary instead?

frontend/src/docs/DocsSearch.tsx
```diff
@@ -0,0 +1,648 @@
...
+
+    requestAnimationFrame(() => {
+      inputRef.current?.focus()
+    })
+  }, [])
+
+  const navigateToRow = useCallback(
+    (row: DisplayRow) => {
+      closeDialog()
+      void router.push(getRowHref(row))
```
IR96334: whats the void for?

IR96334:
I would recommend going through the styling, especially of the styled components, and making sure there aren't redundancies or moot code.

frontend/src/docs/DocsSearch.tsx
GB907762: image

It looks like this. It's useful for accessiblity

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:focus

frontend/src/docs/DocsSearch.tsx
GB907762: [https://mui.com/material-ui/react-dialog/#scrolling-long-content](https://mui.com/material-ui/react-dialog/#scrolling-long-content)

Good spot 'paper' is the default property

frontend/src/docs/DocsSearch.tsx
GB907762: This ensures that you don't have to refocus on the text input when selecting chips.

frontend/src/docs/DocsSearch.tsx
GB907762: Yep. In fact this looks a lot better with a loading component.

frontend/src/docs/DocsSearch.tsx
GB907762: It doesn't do anything visually but it signifies a block layout. Which that title box is. https://developer.mozilla.org/en-US/docs/Glossary/Block-level_content

frontend/src/docs/DocsSearch.tsx
GB907762: Yep good spot!

frontend/src/docs/DocsSearch.tsx
GB907762: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void

Redundant returns a fail or a pass whether the push fulfilled.

frontend/src/docs/DocsSearch.tsx
GB907762: Using the empty blob

frontend/src/docs/DocsSearch.tsx
GB907762: Yes this is intentional so that the filter types can be derived from the categories.

ARADDCC002 (dismissed):

frontend/src/docs/useDocsSearchIndex.ts
```diff
@@ -0,0 +1,578 @@
...
+  DocsSearchQueryTerms as QueryTerms,
+  DocsSearchResult,
+  EntrySearchDocument,
+  IndexedDocsSearchEntry as IndexedEntry,
+  IndexedDocsSearchHeading as IndexedHeading,
+} from 'types/docs'
+import { EntryKindKeys } from 'types/types'
+
+import { flatDirectory } from './directory'
+import rawIndex from './searchIndex.generated.json'
```
PE39806: Add a fallback to this as a fresh bailo clone will not have this file. If an IDE (e.g. VSCode) opens this file then it will try and fail to find it (as the file is gitignored) which might throw a confusing error.

frontend/src/docs/DocsSearchDialog.tsx
```diff
@@ -0,0 +1,510 @@
...
+  DocsSearchPageRow as PageRow,
+  DocsSearchResult,
+  DocsSearchResultSection as ResultSection,
+} from 'types/docs'
+
+import { useDocsSearchIndex } from './useDocsSearchIndex'
+
+type SearchCategory = SearchFilter
+
+const CATEGORY_OPTIONS: DocsSearchCategoryOption[] = [
```
PE39806, line 42: I think I would prefer Docs search results to be after Models & Data cards by default as this makes the new addition slightly closer to the original UX, so eases migration for users. Happy for push back on this though.

frontend/src/docs/useDocsSearchIndex.ts
```diff
@@ -0,0 +1,578 @@
...
+export function searchEntries(
+  entries: EntrySearchDocument[],
+  query: string,
+  category: SearchFilter,
+): DocsSearchResult[] {
+  const matchingCategory = entries.filter((entry) => category === 'all' || entry.category === category)
+
+  if (!query) {
+    return matchingCategory
+      .map((entry) => toEntryResult(entry))
```
PE39806: Since `toEntryResult` may return a score of `0`, is it worth an early `.filter((result) => result.score > 0)` here?

PE39806 (dismissed):

frontend/src/docs/DocsSearchDialog.tsx
GB907762: Yeah this sounds good. I think I'll also make it so the `Help Docs` are the only results initially so it doesn't call to the backend requesting every model and datacard which it does unnecessarily on page rerenders.

frontend/src/docs/useDocsSearchIndex.ts
GB907762: So entries initially I was trying to make a client side based search to score descriptions. Although I forgot to remove the scoring element, when I decided this was additional overhead. 

If we think it's important in future. I'd like to revisit how the server side search works for entries in terms of both pagination and heuristics.

frontend/src/docs/useDocsSearchIndex.ts
GB907762: I've used the public assets folder to avoid this issue. If SWR can't fetch a compiled index it should fall back to no results. Although the initial help pages should cover a variety of use cases.

JRB66955 (approved):

ARADDCC002 (approved):

PR #3886, opened 2026-07-23
Add missing frontend entry descriptions

PE39806:
Breaking changes:
- Renamed `config.uiConfig.untrustedModel.untrustedModelDescription` -> `config.uiConfig.untrustedModel.untrustedModelLongDescription`

Changes:
- Add new `config.uiConfig.untrustedModel.untrustedModelShortDescription`
- Clarify `frontend/pages/entry/new.tsx` wording
- Add missing `frontend/src/entry/CreateEntry.tsx` descriptions
- Add missing `frontend/src/entry/overview/TemplatePage.tsx` `EntryKind.UNTRUSTED_MODEL` case


JRB66955 (dismissed):

backend/config/default.cjs
```diff
@@ -273,7 +273,10 @@ module.exports = {
 
     untrustedModel: {
       enabled: false,
-      untrustedModelDescription: 'These are private only models.',
+      untrustedModelLongDescription:
+        'Untrusted models are are private only models with the registry tab disabled. They have ot yet been verified or validated to the same standard as a regular model.',
```
GB907762:
```suggestion
        'Untrusted models are are private only models with the registry tab disabled. They have not yet been verified or validated to the same standard as a regular model.',
```

infrastructure/helm/bailo/values.yaml
```diff
@@ -276,7 +276,8 @@ config:
 
     untrustedModel:
       enabled: false
-      untrustedModelDescription: 'These are private only models.'
+      untrustedModelLongDescription: 'Untrusted models are are private only models with the registry tab disabled. They have ot yet been verified or validated to the same standard as a regular model.'
```
GB907762:
```suggestion
      untrustedModelLongDescription: 'Untrusted models are are private only models with the registry tab disabled. They have not yet been verified or validated to the same standard as a regular model.'
```

GB907762 (changes requested):
Likely out of scope, so dismiss if needed, but I see we mix second and third person narratives for entries, which don't read very well. Perhaps we could make these all third person?

GB907762 (approved):

ARADDCC002 (approved):

PR #3887, opened 2026-07-24
Rework backend scripts approach

PE39806:
- Add new `backend/src/scripts/scriptHelper.ts` with associated `backend/src/scripts/README.md` and `backend/src/scripts/exampleScript.ts` for easier to manage scripts
  - Safely handle named args
  - Standardise `run` function
  - Add help text
  - Add optional MongoDB connection helper
- Update `backend/src/scripts/generateOpenApiSpecs.ts` to new approach and associated callers
- Make `backend/src/services/schedule/scheduler.ts` lazily load agenda to prevent MongoDB errors from scripts that don't connect to MongoDB
- All "old" scripts can still be run with `npm run script -- <scriptName>`

### DIFF
--- a/backend/src/scripts/README.md
+++ b/backend/src/scripts/README.md
@@ -1,0 +1,95 @@
+# Backend Scripts
+
+Administrative helper scripts for Bailo.
+
+## Running Scripts
+
+Scripts using `defineScript()` should be run directly with `npx tsx`, not via `npm run script` (the legacy runner strips
+named flags).
+
+When running from the host, `defineScript()` automatically detects it is outside Docker and reads the MongoDB connection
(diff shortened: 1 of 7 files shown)


backend/src/scripts/README.md
```diff
@@ -0,0 +1,96 @@
+# Backend Scripts
+
+Administrative helper scripts for Bailo.
+
+## Running Scripts
```
GB907762, line 5: `yargs` has support for the help flag. Would it be more helpful to put some of this content into there?

GB907762 (changes requested):

GB907762 (approved):

JRB66955 (approved):

PR #3892, opened 2026-07-27
docs(root): add further installation instructions

### LINKED ISSUE #3891: Docs: Better onboarding
### Description

I think a more complete walkthrough on how to set this up using docker etc is needed. Using the current installation documentation does not work and it does not include how to set up using docker from scratch.

evenstensberg:
https://github.com/gchq/Bailo/issues/3891

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Follow the instructions below to get a local development environment running. Yo
 To run in development mode (modified files on your host machine will be reloaded into the running application):
 
 > [!NOTE]
-> You will need to install and start docker desktop before running these commands.
+> Ensure Docker and Docker Compose are installed before running these commands.
 
 ```bash
 git clone https://github.com/gchq/Bailo.git && cd Bailo
--- a/README.md
+++ b/README.md
@@ -43,12 +43,18 @@ sharing.
 
 #### Requirements
 
-- Node v26
-- Docker / Docker Compose
+Follow the instructions below to get a local development environment running. You will need:
+
+- Node v26 - [https://nodejs.org/en/download](https://nodejs.org/en/download)
+- Docker / Docker Compose - [https://www.docker.com/get-started/](https://www.docker.com/get-started/)
+
 
 #### Development Build
 
 To run in development mode (modified files on your host machine will be reloaded into the running application):
+
+> [!NOTE]
+> You will need to install and start docker desktop before running these commands.
 
 ```bash
 git clone https://github.com/gchq/Bailo.git && cd Bailo
--- a/README.md
+++ b/README.md
@@ -48,13 +48,11 @@ Follow the instructions below to get a local development environment running. Yo
 - [Node v26](https://nodejs.org/en/download/current)
 - [Docker / Docker Compose](https://www.docker.com/get-started/)
 
-
 #### Development Build
 
 To run in development mode (modified files on your host machine will be reloaded into the running application):
 
-> [!NOTE]
-> Ensure Docker and Docker Compose are installed before running these commands.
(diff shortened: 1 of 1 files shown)


evenstensberg:
Note: would be great with a guide on how to run this without docker too.

README.md
```diff
@@ -43,13 +43,19 @@ sharing.
 
 #### Requirements
 
-- Node v26
-- Docker / Docker Compose
+Follow the instructions below to get a local development environment running. You will need:
+
+- Node v26 - [https://nodejs.org/en/download](https://nodejs.org/en/download)
+- Docker / Docker Compose - [https://www.docker.com/get-started/](https://www.docker.com/get-started/)
```
PE39806: Prefer anchor text link to bare URL
```suggestion
- [Node v26](https://nodejs.org/en/download/current)
- [Docker / Docker Compose](https://www.docker.com/get-started/)
```

README.md
```diff
@@ -43,13 +43,19 @@ sharing.
...
+
 
 #### Development Build
 
 To run in development mode (modified files on your host machine will be reloaded into the running application):
 
+> [!NOTE]
+> You will need to install and start docker desktop before running these commands.
```
JR40159: As this is specific to installing Docker on Windows and Mac I would suggest this should be an update to the [existing requirements section for local Docker development](https://github.com/gchq/Bailo/pull/3892/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L44), making clear the difference in installing Docker for Lunix vs Mac and Windows.

JR40159 (dismissed):
Thank you very much for the PR! One minor suggestion :)

README.md
PE39806: This assumes Docker Desktop is the only supported Docker runtime, but Linux users may be running Docker Engine directly without Docker Desktop.
```suggestion
> Ensure Docker and Docker Compose are installed before running these commands.
```

PE39806 (changes requested):
Thank you for your PR. Your suggestions make the prerequisites section clearer which improves the new developer experience by making setup requirements explicit before the build instructions.

PE39806:
> Note: would be great with a guide on how to run this without docker too.

Thank you for the suggestion. At present, we do not plan to provide a non-Docker development guide.

Bailo is designed around a containerised architecture, and Docker is a core part of the supported development and deployment workflow. In particular:
- Bailo requires multiple services to be run and configured together, including the Frontend, Backend, Nginx, MongoDB, and S3-compatible storage microservices.
- Docker Compose provides a consistent way to orchestrate these services and ensures developers are working in an environment that closely matches supported deployments.
- Certain functionality, such as vulnerability scanning of uploaded artefacts in the `lib/artefactscan` package, relies on container isolation to safely process potentially untrusted code and content.

Maintaining and supporting a non-Docker setup would introduce significant additional complexity and could lead to environments that differ from the configurations we test and support. For these reasons, Docker remains the recommended and supported approach for local development of Bailo.

evenstensberg:
PTAL!

GB907762:
Thanks for taking the time to read and fill gaps in our docs. Hopefully this should aid future contributors. I've ran our formatter, which is a prerequisite to getting a PR accepted.

If you find anything of interest in our current issue tracker feal free to open a future PR. We really appreciate them!

PE39806 (approved):

ARADDCC002 (approved):

evenstensberg:
@GB907762 I think the formatting fix made the note not showing as a note preview should. You'll need the following format:

```text
> NOTE
> actual note
```


https://github.com/gchq/Bailo/commit/eb745d053d576861e4f800f8ed7e7651e0d29b9f

PR #3904, opened 2026-07-28
Feature/bai 2563 new endpoint entries unapproved releases

ON93754:


backend/src/connectors/metrics/base.ts
```diff
@@ -454,6 +458,218 @@ async function calculateModelsMissingReleases(org?: string): Promise<NoReleasesC
...
+  let totalUnapprovedReleases = 0
+
+  for (const model of models) {
+    const unapprovedReleases = unapprovedByModel.get(model.id)
+
+    if (unapprovedReleases === undefined || unapprovedReleases.size === 0) {
+      continue
+    }
+
+    const sortedReleases = [...unapprovedReleases].sort()
```
PE39806: This will not sort semantically (e.g. `['1.0.0', '2.0.0', '10.0.0'].sort()` would give `['1.0.0', '10.0.0', '2.0.0']`)
```suggestion
    const sortedReleases = [...unapprovedReleases].sort((a, b) =>
      semver.compare(a, b),
    )
```
plus `import semver from 'semver'`

backend/src/connectors/metrics/base.ts
```diff
@@ -454,6 +458,218 @@ async function calculateModelsMissingReleases(org?: string): Promise<NoReleasesC
...
+      })
+    }
+
+    const entry = releaseRoleMap.get(key)!
+
+    entry.requiredRoles.add(review.role)
+
+    if (approvedReviewIds.has(review._id.toString())) {
+      entry.approvedRoles.add(review.role)
+    }
```
PE39806, line 593: I think this might get funky if different individuals with the same role give different reviews e.g. for some Release `1.0.0` it could have `Review User1 (MSRO) approved` but also `Review UserB (MSRO) pending`. I'm not quite sure how we resolve this normally but something to check and maybe add a comment here if it's expected behaviour.

backend/src/routes/v3/metrics/getUnapprovedComplianceMetrics.ts
PE39806, line 1: Add new `backend/test/routes/v3/metrics/getUnapprovedComplianceMetrics.spec.ts` file.

PE39806 (changes requested):

backend/test/utils/version.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { describe, expect, test } from 'vitest'
+
+import { sortSemvers } from '../../src/utils/version.js'
+
+describe('utils > version > sortSemvers', () => {
```
PE39806, line 5: I'd like another test to verify that this safely handles duplicate values e.g. `sortSemvers(['1.0.0', '1.0.0'])`.

PE39806 (changes requested):

PE39806 (approved):

ARADDCC002 (approved):

PR #3906, opened 2026-07-28
Validate mirrored model state from combined cards

PE39806:
Update mirrored model state validation to evaluate the merged model card data (`card` + `mirroredCard`) rather than validating only the mirrored model's local card content.

Prior to this PR, model state verification for mirrored models only considered fields present in the mirrored model card. This can incorrectly block state transitions when required fields are satisfied by the source model card.

### DIFF
--- a/backend/src/services/model.ts
+++ b/backend/src/services/model.ts
@@ -626,13 +626,12 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
   }
 
   if (modelDiff.state && model.card) {
-    let valid: boolean
-    if (model.kind === EntryKind.MirroredModel) {
-      const mergedCard = deepMergePreferFirst(model.card.metadata, model.mirroredCard?.metadata || {})
-      valid = (await validateContentAgainstSchema(model.card.schemaId, mergedCard, modelDiff.state)).valid
-    } else {
-      valid = (await validateContentAgainstSchema(model.card.schemaId, model.card.metadata, modelDiff.state)).valid
-    }
+    const card =
+      model.kind === EntryKind.MirroredModel && model.mirroredCard?.metadata
+        ? deepMergePreferFirst(model.card.metadata, model.mirroredCard?.metadata)
+        : model.card.metadata
+    const { valid } = await validateContentAgainstSchema(model.card.schemaId, card, modelDiff.state)
+
     if (!valid) {
       throw BadReq(`Model metadata could not be validated against the schema, for ${modelDiff.state} state.`)
     }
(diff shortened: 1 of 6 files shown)


backend/src/services/model.ts
```diff
@@ -625,7 +626,13 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
   }
 
   if (modelDiff.state && model.card) {
-    const { valid } = await validateContentAgainstSchema(model.card.schemaId, model.card.metadata, modelDiff.state)
+    let valid: boolean
+    if (model.kind === EntryKind.MirroredModel) {
+      const mergedCard = deepMergePreferFirst(model.card.metadata, model.mirroredCard?.metadata || {})
```
JR40159:
```suggestion
    if (model.kind === EntryKind.MirroredModel &&  model.mirroredCard?.metadata) {
      const mergedCard = deepMergePreferFirst(model.card.metadata, model.mirroredCard.metadata)
```

backend/src/services/model.ts
```diff
@@ -625,7 +626,13 @@ export async function updateModel(user: UserInterface, modelId: string, modelDif
   }
 
   if (modelDiff.state && model.card) {
-    const { valid } = await validateContentAgainstSchema(model.card.schemaId, model.card.metadata, modelDiff.state)
+    let valid: boolean
```
JR40159: Remove duplicated calls to `validateContentAgainstSchema` by having a `card` variable here instead of `valid` then add ` const { valid } = await validateContentAgainstSchema(model.card.schemaId, card, modelDiff.state)` above line 636

JR40159: You could also remove the need for the `else` condition by assigning `model.card.metadata` to  `card` as part of the declaration.

backend/src/utils/object.ts
```diff
@@ -23,3 +25,20 @@ export function getPropValue<T = unknown>(source: unknown, path: string): T | un
     return acc != null ? acc[key] : undefined
   }, source)
 }
+
+export function deepMergePreferFirst<T extends object, U extends object>(first: T, second: U): T & U {
```
JR40159: Better names for types than `T` and `U`

JR40159 (changes requested):

backend/src/utils/object.ts
PE39806: Won't do - this follows [TS Generic Parameters](https://www.typescriptlang.org/docs/handbook/2/generics.html#generic-parameter-defaults) which is a pattern used elsewhere in the codebase e.g.
https://github.com/gchq/Bailo/blob/5dc23530a2ddbeb394c57d74f9efa086b8083779/backend/src/utils/array.ts#L1

JR40159: Yeah I know it's a established pattern, I just don't think it's a good one. `T` and `U` aren't the most descriptive and helpful when someone comes to reading it!

GB907762 (changes requested):
Could you check enumerations? Data cards multiselects ect..

It looks like these fail validation

GB907762 (approved):

JR40159 (approved):

PR #3914, opened 2026-07-30
Review status is now a calculated property when getting reviews

ARADDCC002:


backend/src/services/specification.ts
```diff
@@ -316,6 +316,8 @@ export const reviewInterfaceSchema = z.object({
 
   responses: z.array(z.string().optional()),
 
+  status: z.string(),
```
PE39806:
```suggestion
  status: z.string().nullable(),
```

backend/src/services/review.ts
```diff
@@ -54,6 +55,55 @@ export async function findReviews(
     { $lookup: { from: 'v2_models', localField: 'modelId', foreignField: 'id', as: 'model' } },
     { $unwind: { path: '$model' } },
     { $match: { ...(mine && (await findUserInCollaborators(user))) } },
+    // Determine the overall latest decision across all reviewers. Each reviewer's most recent
+    // review response is considered; if any reviewer has requested changes the overall decision
+    // is `request_changes`, otherwise if any reviewer has approved it is `approve`. If neither
+    // applies the field is left unset.
+    {
+      $lookup: {
+        from: 'v2_responses',
```
PE39806, line 64: Probably need to update `ReponseModel` to have indexes `{parentId: 1, kind: 1, entity: 1, createdAt: -1}`.

frontend/types/types.ts
```diff
@@ -599,6 +599,7 @@ type PartialReviewInterface = {
   _id: string
   modelId: string
   role: string
+  status?: string
```
PE39806: Should this be `DecisionKeys`?
```suggestion
  status?: DecisionKeys
```

PE39806 (changes requested):

PE39806 (dismissed):

PE39806 (dismissed):

backend/src/services/specification.ts
```diff
@@ -316,6 +316,8 @@ export const reviewInterfaceSchema = z.object({
 
   responses: z.array(z.string().optional()),
 
+  status: z.string().nullable(),
```
JRB66955: Could this be more specific i.e. only allow specific strings associated with DecisionKeys?

JRB66955 (dismissed):

PE39806 (dismissed):

JR40159 (changes requested):
Frontend review stuff needs updating following https://github.com/gchq/Bailo/pull/3921

PE39806 (approved):

frontend/src/entry/model/reviews/ReleaseAccessRequestReviewSummary.tsx
JR40159, line 1: Is `reviewResponses` now redundant and we can instead use the new `status` property on the review?

JR40159 (changes requested):

JR40159 (approved):

PR #3915, opened 2026-07-30
Feature/bai 2440 draft release implementation

IR96334:


frontend/pages/model/[modelId]/release/new.tsx
```diff
@@ -254,16 +256,28 @@ export default function NewRelease() {
...
+                    </Button>
+                    <Button
+                      variant='contained'
+                      loading={loading}
+                      type='submit'
+                      disabled={!(semver && releaseNotes && isValidSemver(semver) && !isRegistryError)}
+                      sx={{ width: 'fit-content' }}
+                      data-test='createReleaseButton'
+                    >
+                      Create Release
```
Comment (GB907762) [inline frontend/pages/model/[modelId]/release/new.tsx:278]: When you draft a release. Delete it. Create a new one with the semver you get an error like:
` A release using this semver has been deleted. Please use a different semver or contact an admin to restore the deleted release.`

I'd make sure that if something is in draft this doesn't appear.

frontend/src/entry/model/releases/DraftBanner.tsx
```diff
@@ -0,0 +1,85 @@
...
+              <Button
+                variant='outlined'
+                sx={{ borderColor: 'white', color: theme.palette.common.white }}
+                onClick={() => {
+                  setOpen(true)
+                }}
+                disabled={disableButton}
+                loading={isLoading}
+              >
+                <Typography>Publish</Typography>
```
GB907762, line 68: I think this would be better at the end. It forces the owner to review what artefacts and content they've put into it.

frontend/src/entry/model/releases/DraftBanner.tsx
```diff
@@ -0,0 +1,85 @@
...
+              </Button>
+              <ConfirmationDialogue
+                open={open}
+                title='Delete Release'
+                onConfirm={handlePublish}
+                onCancel={() => setOpen(false)}
+                errorMessage={errorMessage}
+                dialogMessage={'Are you sure you want to publish this release? This is irreversable.'}
```
GB907762, line 76: I don't think this is true. You can still delete them.

frontend/src/entry/model/releases/ReleaseDisplay.tsx
```diff
@@ -54,7 +55,11 @@ export default function ReleaseDisplay({
         }}
       >
         <Box sx={{ width: '100%' }}>
-          {reviews.length > 0 && !hideReviewBanner && <ReviewBanner release={release} />}
+          {release.draft ? (
+            <DraftBanner text='This is a draft release' showButton={false} />
```
GB907762, line 59: Can we push drafts to the top so they don't get lost in the list.

frontend/src/entry/model/releases/ReleaseForm.tsx
```diff
@@ -329,27 +318,6 @@ export default function ReleaseForm({
...
-          </>
-        ) : (
-          <FormControl>
-            <FormControlLabel
-              control={<Checkbox size='small' checked={formData.isMinorRelease} onChange={handleMinorReleaseChange} />}
-              label='Minor release - No significant changes, does not require release re-approval'
-            />
-          </FormControl>
-        )}
-      </Stack>
```
GB907762, line 352: I think this is being kept the behaviour of a draft release is slightly different to a minor release.

GB907762 (changes requested):

frontend/src/entry/model/releases/DraftBanner.tsx
IR96334: When I say irreversible, I mean in the context of drafting/releasing. As in you can't re-draft a published release.

backend/src/connectors/authorisation/base.ts
```diff
@@ -249,13 +249,31 @@ export class BasicAuthorisationConnector {
       [ReleaseAction.Export]: ModelAction.Export,
     }
 
+    //If release is draft, map view action to update instead
```
JR40159, line 252: Rather than duplicate the whole action map could you instead have a conditional statement that updates the release view action element of the map to have the model update permission if the release passed to `releases()` is a draft release?

JR40159 (changes requested):

frontend/pages/model/[modelId]/release/new.tsx
Comment (GB907762) [reply frontend/pages/model/[modelId]/release/new.tsx]: Descoped for now

backend/src/connectors/authorisation/base.ts
GB907762: This is linear time where as the current implementation is constant. I'll remove the duplication.

GB907762 (dismissed):

backend/src/services/release.ts
```diff
@@ -213,6 +213,13 @@ export async function updateRelease(user: UserInterface, modelId: string, semver
   }
   const release = await getReleaseBySemver(user, model, semver)
 
+  const isPublish = release.draft && !delta.draft
+  const isDraft = !release.draft && delta.draft
+
+  if (isDraft) {
+    throw BadReq('Can not draft published release.')
```
JR40159:
```suggestion
    throw BadReq('Once a release has been published, it cannot be returned to draft status.')
```

backend/src/services/release.ts
```diff
@@ -213,6 +213,13 @@ export async function updateRelease(user: UserInterface, modelId: string, semver
   }
   const release = await getReleaseBySemver(user, model, semver)
 
+  const isPublish = release.draft && !delta.draft
+  const isDraft = !release.draft && delta.draft
```
JR40159: I wouldn't both with a variable here as it's only used once and the name of the variable is a bit confusing in this context

frontend/pages/model/[modelId]/release/new.tsx
```diff
@@ -254,16 +260,28 @@ export default function NewRelease() {
...
```
Comment (JR40159) [inline frontend/pages/model/[modelId]/release/new.tsx]:
```suggestion
                      Publish release
```

frontend/pages/model/[modelId]/release/new.tsx
```diff
@@ -254,16 +260,28 @@ export default function NewRelease() {
...
+                  <Stack spacing={1} direction='row'>
+                    <Button
+                      variant='outlined'
+                      loading={loading}
+                      onClick={handleDraftRelease}
+                      type='submit'
+                      disabled={!(semver && releaseNotes && isValidSemver(semver) && !isRegistryError)}
+                      sx={{ width: 'fit-content' }}
+                    >
+                      Draft Release
```
Comment (JR40159) [inline frontend/pages/model/[modelId]/release/new.tsx]:
```suggestion
                      Draft release
```

frontend/pages/model/[modelId]/release/new.tsx
Comment (JR40159) [reply frontend/pages/model/[modelId]/release/new.tsx]: Also is there guidance on the difference between drafting and publishing a release?

frontend/src/entry/model/Releases.tsx
JR40159:
```suggestion
                  Create new release
```

JR40159 (changes requested):
If I create a draft release, edit it and then save it, it gets published!

frontend/src/entry/model/releases/ReleaseDisplay.tsx
```diff
@@ -37,7 +38,11 @@ export default function ReleaseDisplay({ model, release, latestRelease, hideFile
         }}
       >
         <Box sx={{ width: '100%' }}>
-          {<ReviewBanner release={release} />}
+          {release.draft ? (
```
JR40159: As with the review banner you could put this conditional in the banner itself to avoid having to duplicate the condition everywhere the banner is used

JR40159 (changes requested):

frontend/src/entry/model/releases/DraftBanner.tsx
JRB66955:
```suggestion
                dialogMessage={'Are you sure you want to publish this release? This is irreversible.'}
```

frontend/pages/model/[modelId]/release/[semver].tsx
```diff
@@ -37,13 +44,50 @@ export default function Release() {
...
+      return
+    }
+    const updatedRelease: UpdateReleaseParams = {
+      modelId: model.id,
+      semver,
+      modelCardVersion: release.modelCardVersion,
+      notes: release.notes,
+      fileIds: [],
+      images: [],
+    }
```
Comment (JRB66955) [inline frontend/pages/model/[modelId]/release/[semver].tsx:59]: If a draft release has a file/image attached to it, then publishing wipes any existing files/images that already existed.

JRB66955 (dismissed):

PE39806 (changes requested):
Update the docs in `frontend/pages/docs/users/models/creating-a-release.mdx`, including a section stating that draft releases are only visible to certain users and that notifications to review are only sent when the draft release is published.

Edit: won't do as this is to be done in a separate ticket.

JR40159:
> Update the docs in `frontend/pages/docs/users/models/creating-a-release.mdx`, including a section stating that draft releases are only visible to certain users and that notifications to review are only sent when the draft release is published.

https://github.com/gchq/Bailo/pull/3915#discussion_r3794701609

frontend/pages/model/[modelId]/release/new.tsx
Comment (JR40159) [reply frontend/pages/model/[modelId]/release/new.tsx]: ?

frontend/src/entry/model/Releases.tsx
JR40159: ?

JR40159 (dismissed):

PE39806 (approved):

JR40159 (approved):

PR #3916, opened 2026-07-30
Fix filter collapse

GB907762:
- Fix the color of the filter
- Collapse only viewable on minimise
- Filter collapsed when minimised on default

### DIFF
--- a/frontend/pages/index.tsx
+++ b/frontend/pages/index.tsx
@@ -17,7 +17,9 @@ import {
   Tabs,
   Tooltip,
   Typography,
+  useMediaQuery,
 } from '@mui/material'
+import { grey } from '@mui/material/colors'
 import { useTheme } from '@mui/material/styles'
 import { useGetPopularEntryTags, useListEntries } from 'actions/entry'
 import { useGetReviewRoles } from 'actions/reviewRoles'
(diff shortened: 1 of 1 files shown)


evenstensberg (approved):
lgtm!

frontend/pages/index.tsx
```diff
@@ -397,10 +403,10 @@ export default function Marketplace() {
               </Button>
               <Box
                 sx={{
-                  backgroundColor: theme.palette.mode === 'light' ? theme.palette.background.paper : 'unset',
+                  backgroundColor: theme.palette.mode === 'light' ? grey[200] : grey[800],
```
ARADDCC002, line 406: This was changed for dark mode. The value `grey[200]` is too bright of a contrast, having no background colour looks better in dark mode IMO.

ARADDCC002 (changes requested):

frontend/pages/index.tsx
evenstensberg: Ideally this project should have a color palette with a pre-defined color scheme.

JR40159: For me, the fact that having no background looks better in dark mode isn't a good enough reason as it's made light mode (the default mode) look worse.

JR40159 (approved):

ARADDCC002 (approved):

PR #3919, opened 2026-07-31
Feature/bai 2607 restrict creation of um

ON93754:


backend/src/services/model.ts
```diff
@@ -52,6 +52,18 @@ function checkMirroredModelRestriction(model: ModelInterface) {
   }
 }
 
+/**
+ * Returns true if the user is authorised to manage untrusted models.
+ */
```
PE39806: This comment is self explanatory from the function name so I'd rather it weren't here to add clutter.

backend/src/utils/__mocks__/config.ts
```diff
@@ -106,6 +106,7 @@ const config: PartialDeep<Config> = {
       },
       userPoolId: '',
       userIdAttribute: '',
+      untrustedModelGroupName: '',
```
PE39806: Also add this to `backend/config/default.cjs` otherwise the value might be `undefined`.

PE39806 (changes requested):

backend/test/connectors/authentication/oauth.spec.ts
```diff
@@ -127,4 +134,82 @@ describe('connectors > authentication > oauth', () => {
...
+
+    expect(result).toBe(false)
+  })
+
+  test('hasRole > admin access grants untrusted model role', async () => {
+    const connector = new OauthAuthenticationConnector()
+    const getEntityMembersSpy = vi
+      .spyOn(connector, 'getEntityMembers')
+      .mockResolvedValueOnce([]) // untrusted model group check fails
+      .mockResolvedValueOnce([user.dn]) // admin group check succeeds
```
PE39806: I don't think this is correct as the `isAdmin` check happens _before_ the UM group check
```ts
const isAdmin = await this.hasGroupMembership(...)
if (isAdmin) {
  return true
}
```
I think you need to remove `.mockResolvedValueOnce([])` and then have
```ts
expect(getEntityMembersSpy).toHaveBeenCalledTimes(1)
expect(getEntityMembersSpy).toHaveBeenCalledWith(
  toEntity('group', config.oauth.cognito.adminGroupName),
)
```

backend/src/connectors/authentication/silly.ts
```diff
@@ -35,6 +35,9 @@ export class SillyAuthenticationConnector extends BaseAuthenticationConnector {
     if (role === Roles.Compliance) {
       return true
     }
+    if (role === Roles.UntrustedModel) {
+      return true
+    }
     return false
```
PE39806: Replace whole check with
```suggestion
    return [Roles.Admin, Roles.Compliance, Roles.UntrustedModel].includes(role)
```

backend/src/services/model.ts
```diff
@@ -73,6 +82,10 @@ export type CreateModelParams = Pick<
 > &
   OptionalCreateModelParams
 export async function createModel(user: UserInterface, modelParams: CreateModelParams) {
+  if (modelParams.kind === EntryKind.UntrustedModel) {
+    await checkUserIsAuthorisedForUntrustedModel(user)
+  }
```
PE39806: Add a test to provide coverage of this.

PE39806 (changes requested):

backend/src/services/model.ts
```diff
@@ -52,6 +52,15 @@ function checkMirroredModelRestriction(model: ModelInterface) {
   }
 }
 
+async function checkUserIsAuthorisedForUntrustedModel(user: UserInterface) {
```
JR40159: As this is an authorisation check, I'd prefer this to be done as part of the `models` function in the base authorisation class. For maintainability it's good to try and keep all the authorisation logic in the same place.

JR40159 (changes requested):

PE39806 (approved):

JR40159 (approved):

PR #3920, opened 2026-07-31
ci: pin all workflows to harden security

### LINKED ISSUE #3913: [Enhancement]: Use sha values in github actions
### To Reproduce

N/A see below:

Many of the workflows in this project uses versions rather than specific sha-sums to harden security.

I.e:

https://github.com/gchq/Bailo/blob/main/.github/workflows/python-publish-test.yml#L20

should use a pinned version.

### Current vs. Expected behavior

See above

### Provide environment information

```bash
N/A
```

### Additional context

_No response_

Medium

evenstensberg:
Closes #3913

### DIFF
--- a/.github/workflows/artefactscan.yml
+++ b/.github/workflows/artefactscan.yml
@@ -15,9 +15,9 @@ jobs:
     outputs:
       relevant: ${{ steps.filter.outputs.relevant }}
     steps:
-      - uses: actions/checkout@v7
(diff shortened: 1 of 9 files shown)


PE39806:
Thank you for putting this together. Unfortunately, dependabot ran automatically so these SHAs are no longer for the correct releases. I have reused some of the SHAs as a useful starter point in #3975 which supersedes this PR with the updated versions.

PR #3921, opened 2026-07-31
Refactor frontend release and access request review content

JR40159:
- [x] Remove reliance on user's role and instead check if any reviews are available for that user.
- [x] Move repeated conditional rendering logic of review banner inside component.
- [x] Rename `ReviewDisplay` to `ReviewStatus` as the component displays the status of a review, not the review itself.
- [x] Use a single component for the identical review information at the bottom of each access request and release in their respective lists.


frontend/actions/review.ts
```diff
@@ -77,6 +77,37 @@ type GetReviewRequestsForModelQuery = {
   open?: boolean
 } & additionalParameters
 
+export function useHeadReviewRequestsForModel({
```
JR40159: Either reduce minimise parameters to what's needed/used for this method or update existing get method to be able to do a head request

frontend/actions/review.ts
```diff
@@ -77,6 +77,27 @@ type GetReviewRequestsForModelQuery = {
...
+  const { data, isLoading, error, mutate } = useSWR<
+    {
+      headers: any
+    },
+    ErrorInfo
+  >(['head', `/api/v2/reviews?${qs.stringify(queryParams)}`], ([, url]: string) => fetcher(url, true))
+
+  return {
+    mutateReviews: mutate,
+    reviewCountHeader: data?.headers['x-count'] ? parseInt(data.headers['x-count']) : 0,
```
ARADDCC002, line 95: Is there a chance that this header might not be a number?

ARADDCC002 (changes requested):

frontend/actions/review.ts
JR40159: Nope, it's something we set as a  number in the backend

frontend/actions/review.ts
```diff
@@ -77,6 +77,27 @@ type GetReviewRequestsForModelQuery = {
...
 
+export function useHeadReviewRequestsForModel({ modelId, semver, accessRequestId }: GetReviewRequestsForModelQuery) {
+  const queryParams = {
+    modelId,
+    ...(semver && { semver }),
+    ...(accessRequestId && { accessRequestId }),
+  }
+  const { data, isLoading, error, mutate } = useSWR<
+    {
+      headers: any
```
PE39806:
```suggestion
      headers: Record<string, string>
```

frontend/src/entry/model/reviews/ReleaseAccessRequestReviewSummary.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { IconButton, Stack, Tooltip, Typography } from '@mui/material'
+import { useGetResponses } from 'actions/response'
+import { useGetReviewRequestsForModel } from 'actions/review'
+import Loading from 'src/common/Loading'
+import ReviewStatus from 'src/entry/model/reviews/ReviewStatus'
+import MultipleErrorWrapper from 'src/errors/MultipleErrorWrapper'
+import { AccessRequestInterface, ReleaseInterface } from 'types/types'
+import { latestReviewsForEachUser } from 'utils/reviewUtils'
+
+export type ReleaseAccessRequestReviewSummaryPops =
```
PE39806: Typo
```suggestion
export type ReleaseAccessRequestReviewSummaryProps =
```

frontend/src/entry/model/reviews/ReleaseAccessRequestReviewSummary.tsx
```diff
@@ -0,0 +1,108 @@
...
+      <>
+        <Stack
+          direction='row'
+          spacing={2}
+          sx={{
+            alignItems: 'center',
+            justifyContent: 'space-between',
+          }}
+        >
+          {reviews.length > 0 && (
```
PE39806: This check is already satisfied just above so should be removed

PE39806 (changes requested):

frontend/pages/model/[modelId]/access-request/[accessRequestId].tsx
```diff
@@ -95,16 +37,10 @@ export default function AccessRequest() {
...
-            isUserReviewsLoading ||
-            isModelLoading ||
-            isCurrentUserLoading ||
-            isReviewResponsesLoading ||
-            isReviewRolesLoading) && <Loading />}
+          {isAccessRequestLoading && <Loading />}
           {accessRequest && (
             <>
-              {userCanReview && <ReviewBanner accessRequest={accessRequest} />}
+              {<ReviewBanner accessRequest={accessRequest} />}
```
Comment (GB907762) [inline frontend/pages/model/[modelId]/access-request/[accessRequestId].tsx]:
```suggestion
              <ReviewBanner accessRequest={accessRequest} />
```

frontend/src/entry/model/accessRequests/AccessRequestDisplay.tsx
```diff
@@ -66,7 +23,7 @@ export default function AccessRequestDisplay({ accessRequest, hideReviewBanner =
         }}
       >
         <Box sx={{ width: '100%' }}>
-          {reviews.length > 0 && !hideReviewBanner && <ReviewBanner accessRequest={accessRequest} />}
+          {<ReviewBanner accessRequest={accessRequest} />}
```
GB907762:
```suggestion
          <ReviewBanner accessRequest={accessRequest} />
```

frontend/src/entry/model/reviews/ReleaseAccessRequestReviewSummary.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const {
+    responses: commentResponses,
+    isResponsesLoading: isCommentResponsesLoading,
+    isResponsesError: isCommentResponsesError,
+  } = useGetResponses([_id])
+  const {
+    responses: reviewResponses,
+    isResponsesLoading: isReviewResponsesLoading,
+    isResponsesError: isReviewResponsesError,
+  } = useGetResponses([...reviews.map((review) => review._id)])
```
GB907762, line 45: I think the top one should fire regardless of comments. Which is an unnecessary request. 
Could you unify these and sort via response.kind?

frontend/src/entry/model/reviews/ReleaseAccessRequestReviewSummary.tsx
```diff
@@ -0,0 +1,108 @@
...
+              <Stack direction='row' spacing={2}>
+                {reviewResponses.length > 0 && (
+                  <Tooltip title='Reviews'>
+                    <Stack direction='row' spacing={1}>
+                      <ListAltIcon color='primary' />
+                      <Typography variant='caption'>{reviewResponses.length}</Typography>
+                    </Stack>
+                  </Tooltip>
+                )}
+                {commentResponses.length > 0 && (
```
GB907762, line 89: This is embedded in the review condition, meaning comment icons aren't rendered when there are no reviews, but there are comments.

frontend/src/entry/model/Releases.tsx
```diff
@@ -31,39 +28,23 @@ export default function Releases({ model, currentUserRoles, readOnly = false }:
...
-      model={model}
-      release={data}
-      latestRelease={getLatestRelease()}
-      hideReviewBanner={
-        !hasRole(
-          currentUserRoles,
-          reviewRoles.map((role) => role.shortName),
-        ) || readOnly
-      }
-    />
```
GB907762, line 48: Be aware this is a breaking change among mirrored models.

Previously you couldn't see the banner in any case but now you can.

This looks expected though?

GB907762 (changes requested):

frontend/src/entry/model/Releases.tsx
JR40159: The reason for this change is because having the check for mirrored models is redundant if you base the presence of the banner on a review existing because reviews aren't mirrored and there's no way to create a review on a mirrored model as you can't create a release for access request manually.

GB907762: Yep good spot. There's also a review in the side bar for mirrored models but it didn't appear on the release page

PE39806 (dismissed):

frontend/src/entry/model/reviews/ReleaseAccessRequestReviewSummary.tsx
JR40159: I can go one better, https://github.com/gchq/Bailo/pull/3914 will move all this logic to the backend! Consequently, I suggest this isn't needed for this PR

frontend/src/entry/model/reviews/ReleaseAccessRequestReviewSummary.tsx
JR40159: Just tested manually and the comment icon shows when there are comments but no reviews

PE39806 (approved):

GB907762 (approved):

ARADDCC002 (approved):

PR #3974, opened 2026-08-03
Add model mirroring integration tests

PE39806:
Previously, model mirroring was not tested at an integration level. This meant that regressions were more likely as we did not check if the functionality works in an automated fashion. To remedy this, I have created a new model mirroring integration test suite using multiple docker compose projects with cypress.

- Runs two Bailo instances (A & B) via `docker compose` with shared network between A's S3 and B's backend services;
- New Cypress suite which:
  - Creates a Model on A with an updated model card description and a release containing 1 file and 1 image;
  - Creates a Mirrored Model on B which points toward the Model on A;
  - Updates the Model on A to point towards Mirrored Model on B and exports the Model on A without any releases, then checks the import succeeded for the Mirrored Model on B;
  - Exports the Model on A with the release, then checks the import succeeded for the Mirrored Model on B (including artefacts).

For manual testing:
```bash
# setup `local.cjs` (preserving backup)
cp backend/config/local.cjs backend/config/local.cjs.bak
cat > backend/config/local.cjs <<'EOT'
module.exports = {
  mongo: {
    uri: 'mongodb://bailoadmin:bailoadmin@mongo:27017/bailo?replicaSet=rs0&authSource=admin',
  },

  s3: {
    credentials: {
      accessKeyId: 'bailoadmin',
      secretAccessKey: 'bailoadmin',
    },
    endpoint: 'http://seaweedfs:8333',
    region: 'ignored',
    forcePathStyle: true,
    rejectUnauthorized: true,
  },

  smtp: {
    enabled: false,
    auth: {
      user: 'mailuser',
      pass: 'mailpass',
    },
    tls: {
      rejectUnauthorized: false,
    },
  },

  logging: {
    file: {
      path: '/home/runner/work/Bailo/Bailo/logs/out.log',
    },
  },

  ui: {
    modelMirror: {
      import: {
        enabled: true,
      },
      export: {
        enabled: true,
        disclaimer: '## Example Agreement \n I agree that this model is suitable for exporting',
      },
    },
  },
  connectors: { artefactScanners: { kinds: [] } },
}
EOT

# docker compose build
INSTANCE_NAME=bailo_a docker compose -f compose.yaml -f compose.instance-a.yaml -p bailo-a build --parallel
INSTANCE_NAME=bailo_b docker compose -f compose.yaml -f compose.instance-b.yaml -p bailo-b build --parallel

# docker compose start
INSTANCE_NAME=bailo_a docker compose -f compose.yaml -f compose.instance-a.yaml -p bailo-a up -d
INSTANCE_NAME=bailo_b docker compose -f compose.yaml -f compose.instance-b.yaml -p bailo-b up -d

# cypress setup
cd frontend
npm i

# run tests
CYPRESS_BASE_URL=http://localhost:8081 npm run cy:run-mirror

# docker compose cleanup
cd ..
INSTANCE_NAME=bailo_b docker compose -f compose.yaml -f compose.instance-b.yaml -p bailo-b down -v --remove-orphans
INSTANCE_NAME=bailo_a docker compose -f compose.yaml -f compose.instance-a.yaml -p bailo-a down -v --remove-orphans

# restore `local.cjs` from backup
rm -f backend/config/local.cjs
mv backend/config/local.cjs.bak backend/config/local.cjs
```


compose.instance-a.yaml
```diff
@@ -0,0 +1,33 @@
+services:
+  nginx:
+    ports:
+      - '8081:8080'
+
+  backend:
+    environment:
+      - NODE_CONFIG_ENV=prod_docker_compose
+
+  minio:
```
GB907762: Should we hold off on using Minio until #3976 is merged?

compose.instance-b.yaml
```diff
@@ -0,0 +1,41 @@
...
+  minio:
+    ports:
+      - '9100:9000'
+      - '9101:9001'
+
+  registry:
+    ports:
+      - '5001:5000'
+
+  mailcrab:
```
GB907762: There doesn't seem to be any use of email in your tests. Could you perhaps check they exist on the peer or remove this container entirely?

compose.instance-b.yaml
```diff
@@ -0,0 +1,41 @@
...
+    networks:
+      default:
+      internal:
+    ports:
+      - 1080:1080
+      - 1025:1025
+    image: marlonb/mailcrab:v1.8.0
```
GB907762:
```suggestion
    image: marlonb/mailcrab:v1.9.0@sha256:fbe35c063cb4c8c169e32f4d91e09ee44459843a005a567cfd3dc9876779e194
```
Inline with #3975

compose.yaml
GB907762, line 1: Do you have any thoughts on using the Standalone container?

Many compose files might be an overhead to maintain?

GB907762 (changes requested):
These tests look quite thorough. Hopefully it should help in tracking regressions.

compose.yaml
PE39806: I would argue that the standalone container is already overhead and I'd love to see it removed - it's already different to the compose `Dockerfile`s which don't use `ubuntu` so is not really representative, whereas this compose stack uses the properly built containers which match prod. Additionally, the way that A's S3 store has to interact with B's backend requires an extra network (which cannot just be shared between all services due to the way the `registry` services resolve) which the standalone container would be insufficient for.

compose.instance-b.yaml
PE39806: Good point. I was having an error at one point with the email trying but failing to send but I think that removing this service and disabling SMTP in config is better.

JRB66955 (approved):

GB907762 (approved):

PR #3976, opened 2026-08-03
Migrate minio to seaweedfs

GB907762:
Update Standalone image and local development to use [SeaweedFS](https://github.com/seaweedfs/seaweedfs)


JRB66955 (dismissed):

compose.yaml
```diff
@@ -36,15 +36,24 @@ services:
     ports:
       - 27017:27017
 
-  minio:
-    image: bitnamilegacy/minio:2025.7.23@sha256:8935e75fa5d11295c17171e4aa49efe390a1193cd7f12e4d21b92af9ffef09d7
+  seaweedfs:
```
PE39806, line 39: Might be worth adding a `healthcheck` e.g. `test: ["CMD", "wget", "-qO-", "http://localhost:8333"]` here. This will mean that dependent services will wait until the S3 API is listening, rather than just the service having started. This will require subsequent `depends_on` to use `condition: service_healthy`

Dockerfile.standalone
```diff
@@ -124,6 +124,6 @@ COPY infrastructure/standalone/supervisord.conf /etc/supervisor/conf.d/superviso
 
 WORKDIR /
 
-EXPOSE 8080 9000 27017
+EXPOSE 8080 8333 23646 27017
```
PE39806: Where does `23646` come from? Documentation includes `8888` and `9333` but not this.

PE39806 (changes requested):

Dockerfile.standalone
GB907762: This is a mistake. Thanks for spotting it. I was trying to get the Admin UI working which runs on `23646` and didn't remove this.

compose.prod.yaml
```diff
@@ -51,8 +51,6 @@ services:
         condition: service_healthy
       registry:
         condition: service_started
-      minio:
-        condition: service_started
```
PE39806, line 55: Does removing this still maintain
```yaml
      seaweedfs:
        condition: service_healthy
```
from `compose.yaml` or does it need to be added here too?

GB907762: I believe this is redundant because of the following. But good spot

[https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/#example](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/#example)

JRB66955 (approved):

PE39806 (approved):

PR #4005, opened 2026-08-04
Added the api for model lifecycle review metrics

ARADDCC002:


backend/src/connectors/metrics/base.ts
```diff
@@ -1207,4 +1292,35 @@ export class BaseMetricsConnector {
...
+    )
+
+    const result: Omit<GetLifecycleComplianceMetricsResponse, 'lastUpdated'> = { global, byOrganisation }
+
+    const lastUpdated = new Date().toISOString()
+
+    setCached(MetricsCacheKeys.NO_RELEASES_COMPLIANCE, {
+      data: result,
+      lastUpdated,
+    })
```
JRB66955: Should we be caching the `weeksUntilDue` as well used to obtain the data/result?

backend/src/routes/v3/metrics/getLifecycleComplianceMetrics.ts
```diff
@@ -0,0 +1,82 @@
...
+import audit from '../../../connectors/audit/index.js'
+import metrics from '../../../connectors/metrics/index.js'
+import { z } from '../../../lib/zod.js'
+import { registerPath } from '../../../services/specification.js'
+import { parse } from '../../../utils/validate.js'
+
+export const getLifecycleComplianceMetricsSchema = z.object({
+  query: z
+    .object({
+      weeksUntilDue: z.string().refine((val) => !Number.isNaN(parseInt(val, 10)), {
```
JRB66955: Currently this isn't as strict as it should be (e.g. `-5` and `8abc` are valid inputs), using `z.coerce.number()` would be better.

backend/src/connectors/metrics/base.ts
```diff
@@ -680,6 +681,90 @@ function buildEntryVolumeCacheKey(interval: EntryVolumeInterval, start: Date, en
...
+  }[]
+}
+
+async function calculateLifecycleComplianceMetrics(
+  weeksUntilDue: number,
+  organisation?: string,
+): Promise<CalculatedLifecycleComplianceMetrics> {
+  const dueDateCutOff = new Date()
+  dueDateCutOff.setDate(dueDateCutOff.getDate() + weeksUntilDue * 7)
+  const openLifecycleReviewsPipeline: PipelineStage[] = [
```
JRB66955, line 703: Minor thing, may be more useful if the pipeline has a `$sort` at the end (e.g. sort by most overdue/nearest due)

backend/src/routes/v3/metrics/getLifecycleComplianceMetrics.ts
```diff
@@ -0,0 +1,82 @@
...
+    .strict(),
+})
+
+export const LifecycleSummarySchema = z.object({
+  count: z.number(),
+})
+
+export const EntryLifecycleMetricsSchema = z.object({
+  entryId: z.string(),
+  dueDate: z.string(),
```
JRB66955:
```suggestion
  dueDate: z.string().datetime(),
```

JRB66955 (changes requested):
All looks to be working fine, just a few tweaks/suggestions.

backend/src/connectors/metrics/base.ts
ARADDCC002: I don't think so - in theory this could be changed frequently and we still want to cache the response.

JRB66955 (dismissed):

backend/src/connectors/metrics/base.ts
```diff
@@ -1207,4 +1297,46 @@ export class BaseMetricsConnector {
...
   }
+
+  async getLifecycleComplianceMetrics(
+    user: UserInterface,
+    weeksUntilDue: number,
+  ): Promise<GetLifecycleComplianceMetricsResponse> {
+    await checkUserIsAuthorised(user)
+
+    const cached = getCached<CachedMetrics<GetLifecycleComplianceMetricsResponse>>(
+      `${MetricsCacheKeys.LIFECYCLE}-${weeksUntilDue}`,
```
PE39806, line 1309: This cache key does not include organisations, but the response includes per-organisation breakdowns
```suggestion
      `${MetricsCacheKeys.LIFECYCLE}-${weeksUntilDue}-${organisationIds.sort().join(',')}`,
```

backend/src/connectors/metrics/base.ts
```diff
@@ -680,6 +681,95 @@ function buildEntryVolumeCacheKey(interval: EntryVolumeInterval, start: Date, en
...
+      $sort: {
+        dueDate: 1,
+      },
+    },
+  ]
+
+  // Filter by the referenced model's organisation when one is provided.
+  if (organisation !== undefined) {
+    openLifecycleReviewsPipeline.push({ $match: { modelOrganisation: organisation } })
+  }
```
PE39806: I think it's better to have this in the pipeline _before_ the last `$project` as this will have fewer results in memory at any time - maybe add
```
{
  $match: {
    ...(organisation && {
      'model.organisation': organisation,
    }),
  },
},
```
after `{ $unwind: '$model' },`

PE39806 (changes requested):

backend/src/connectors/metrics/base.ts
ARADDCC002: Not needed as organisation is filtered in the UI.

PE39806 (dismissed):

backend/src/connectors/metrics/base.ts
```diff
@@ -680,6 +681,97 @@ function buildEntryVolumeCacheKey(interval: EntryVolumeInterval, start: Date, en
...
+    modelOwners: string[]
+  }[]
+}
+
+async function calculateLifecycleComplianceMetrics(
+  weeksUntilDue: number,
+  organisation?: string,
+): Promise<CalculatedLifecycleComplianceMetrics> {
+  const dueDateCutOff = new Date()
+  dueDateCutOff.setDate(dueDateCutOff.getDate() + weeksUntilDue * 7)
```
GB907762: Doesn't this method have a rounding error to midnight?

Consider `humanInterval`

backend/src/connectors/metrics/base.ts
```diff
@@ -1207,4 +1299,46 @@ export class BaseMetricsConnector {
         })) ?? [],
     }))
   }
+
+  async getLifecycleComplianceMetrics(
+    user: UserInterface,
+    weeksUntilDue: number,
```
GB907762, line 1304: Consider a default of `Infinity` to get all models that have a review attached.

backend/src/routes/v3/metrics/getLifecycleComplianceMetrics.ts
```diff
@@ -0,0 +1,80 @@
...
+import audit from '../../../connectors/audit/index.js'
+import metrics from '../../../connectors/metrics/index.js'
+import { z } from '../../../lib/zod.js'
+import { registerPath } from '../../../services/specification.js'
+import { parse } from '../../../utils/validate.js'
+
+export const getLifecycleComplianceMetricsSchema = z.object({
+  query: z
+    .object({
+      weeksUntilDue: z.coerce.number(),
```
GB907762: This is weakly constrained. Negative/Large and floating point numbers are valid. Use natural numbers with a cutoff

backend/test/routes/v3/metrics/getLifecycleComplianceMetrics.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { describe, expect, test, vi } from 'vitest'
+
+import audit from '../../../../src/connectors/audit/__mocks__/index.js'
+import { testGet } from '../../../testUtils/routes.js'
+
+vi.mock('../../../../src/connectors/audit/index.js')
+
+const mockMetricsConnector = vi.hoisted(() => ({
+  getLifecycleComplianceMetrics: vi.fn(() => {}),
```
GB907762: I'd put some real data in here so you can verify it's doing what it says.

backend/src/connectors/metrics/base.ts
```diff
@@ -1207,4 +1299,46 @@ export class BaseMetricsConnector {
...
+    }
+
+    const global = await calculateLifecycleComplianceMetrics(weeksUntilDue)
+
+    const organisationIds = await this.getOrganisationIds()
+
+    const byOrganisation = await Promise.all(
+      organisationIds.map(async (org) => ({
+        organisation: org || 'unset',
+        ...(await calculateLifecycleComplianceMetrics(weeksUntilDue, org)),
```
GB907762, line 1325: When org is falsey it gathers all models regardless of organisation. 

I think we need to look at this, before we release v3. `null` is a safer option and the default on a MUI dropdown

GB907762 (changes requested):

backend/src/connectors/metrics/base.ts
```diff
@@ -697,7 +698,11 @@ async function calculateLifecycleComplianceMetrics(
   organisation?: string,
 ): Promise<CalculatedLifecycleComplianceMetrics> {
   const dueDateCutOff = new Date()
-  dueDateCutOff.setDate(dueDateCutOff.getDate() + weeksUntilDue * 7)
+  const interval = humanInterval(`${weeksUntilDue} weeks`)
+  if (!interval) {
+    throw InternalError('The time interval provided could not be converted to a numerical value.', { weeksUntilDue })
+  }
```
GB907762:
```suggestion
```
Could you just remove this. It will never be hit because of zod.

backend/src/connectors/metrics/base.ts
```diff
@@ -1207,4 +1301,46 @@ export class BaseMetricsConnector {
         })) ?? [],
     }))
   }
+
+  async getLifecycleComplianceMetrics(
+    user: UserInterface,
+    weeksUntilDue: number = Infinity,
```
GB907762: Sorry I don't think I was particularly clear on this. 
Validation on the route fails for every non numeric parameter, so this can't be reach. 

If you think it's useful to get all reviews keep it and modify the endpoint. Otherwise removing this entirely should be fine. 

```suggestion
    weeksUntilDue: number,
```

GB907762 (approved):

PE39806 (approved):

PR #4007, opened 2026-08-05
Model lifecycle metrics UI

ARADDCC002:


frontend/src/metrics/PolicyMetrics.tsx
```diff
@@ -43,14 +51,16 @@ export default function PolicyMetrics() {
...
   const [selectedMetric, setSelectedMetric] = useState<SelectedMetricKindKeys>(SelectedMetricKind.MISSING_ROLES)
-  const selectedData: undefined | PolicyRoleMetrics | BaseNoReleaseMetrics = (() => {
+  const selectedData: undefined | PolicyRoleMetrics | BaseNoReleaseMetrics | BaseLifecycleMetrics = (() => {
     switch (selectedMetric) {
       case SelectedMetricKind.MISSING_ROLES:
         return rolePolicyMetrics
       case SelectedMetricKind.NO_RELEASES:
         return noReleasesPolicyMetrics
       case SelectedMetricKind.UNAPPROVED_RELEASES:
         return noReleasesPolicyMetrics
```
PE39806: Should this be
```suggestion
        return unapprovedReleasesPolicyMetrics
```

frontend/src/metrics/components/OwnerRoleDisplay.tsx
```diff
@@ -0,0 +1,25 @@
+import { useGetEntryRoles } from 'actions/entry'
+import { useMemo } from 'react'
+import Loading from 'src/common/Loading'
+import MessageAlert from 'src/MessageAlert'
+
+export default function OwnerRoleDisplay() {
+  const { entryRoles, isEntryRolesLoading, isEntryRolesError } = useGetEntryRoles()
```
PE39806: I'd prefer that this were a prop passed in rather than a request per use as this is likely to be used in multiple places on a single page, so currently there are a lot of extra calls for the same bit of informaiton.

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
```diff
@@ -0,0 +1,193 @@
...
+          anchorEl={anchorEl}
+          open={open}
+          onClose={handleClose}
+          slotProps={{
+            list: {
+              'aria-labelledby': 'menu-button',
+            },
+          }}
+        >
+          <MenuItem onClick={() => weekFilterOnChange(2)}>
```
PE39806: In the Entries tab, when selecting a filter it automatically closes the popup whereas here it does not. I don't have a preference between which is better, but I do feel we need this to be consistent.

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null)
+  const open = Boolean(anchorEl)
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorEl(event.currentTarget)
+  }
+  const handleClose = () => {
+    setAnchorEl(null)
+  }
+
+  const tableTitle = () => {
```
PE39806: Possible `useMemo` here?

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
```diff
@@ -0,0 +1,193 @@
...
+      case 0:
+        return 'Entries past their due date'
+      default:
+        return 'Entries near or past lifecycle review date'
+    }
+  }
+
+  const overDueWarning = (dueDate: string) => {
+    const today = new Date().setHours(0, 0, 0, 0)
+    const dateToCheck = new Date(dueDate).setHours(0, 0, 0, 0)
```
PE39806: Possibly a candidate for `frontend/utils/dateUtils.ts`?

frontend/src/metrics/PolicyMetrics.tsx
```diff
@@ -3,25 +3,33 @@ import {
...
 
 export const SelectedMetricKind = {
   MISSING_ROLES: 'role',
   NO_RELEASES: 'noReleases',
   UNAPPROVED_RELEASES: 'unapprovedReleases',
+  LIFECYCLE: 'lifecycle',
 } as const
 export type SelectedMetricKindKeys = (typeof SelectedMetricKind)[keyof typeof SelectedMetricKind]
 
+export type WeekFilterOptions = 0 | 2 | 10
```
PE39806: Maybe replace this and the hardcoded references (`0`, `2`, `10`) with some const e.g.
```ts
export const LIFECYCLE_FILTERS = {
  OVERDUE: 0,
  TWO_WEEKS: 2,
  TEN_WEEKS: 10,
} as const
```

PE39806 (changes requested):

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
ON93754: I'd be tempted to have a new function `isOverdue(dueDate: string): boolean` added to `dateUtils.ts` and keep the rendering logic in here?

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
```diff
@@ -0,0 +1,193 @@
...
+    if (dateToCheck <= today) {
```
JRB66955: Should this be `(dateToCheck < today)` instead? As this will mark a review due "today" as overdue.

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
```diff
@@ -0,0 +1,193 @@
...
+            },
+          }}
+        >
+          <MenuItem onClick={() => weekFilterOnChange(2)}>
+            {weekFilter === 2 && (
+              <ListItemIcon>
+                <Check />
+              </ListItemIcon>
+            )}
+            <ListItemText inset={weekFilter !== 2}>Two weeks</ListItemText>
```
GB907762:
```suggestion
            <ListItemText inset={weekFilter !== 2}>2 weeks</ListItemText>
```
I understand that often single digits are expressed as words rather than numeric in formal writing, but this is a statical situation.

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
```diff
@@ -0,0 +1,193 @@
...
+    setAnchorEl(event.currentTarget)
+  }
+  const handleClose = () => {
+    setAnchorEl(null)
+  }
+
+  const tableTitle = () => {
+    switch (weekFilter) {
+      case 2:
+        return 'Entries two weeks until their due date'
```
GB907762:
```suggestion
        return 'Entries 2 weeks until their due date'
```

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
```diff
@@ -0,0 +1,193 @@
...
+            )}
+            <ListItemText inset={weekFilter !== 10}>10 weeks</ListItemText>
+          </MenuItem>
+          <MenuItem onClick={() => weekFilterOnChange(0)}>
+            {weekFilter === 0 && (
+              <ListItemIcon>
+                <Check />
+              </ListItemIcon>
+            )}
+            <ListItemText inset={weekFilter !== 0}>Past due date</ListItemText>
```
GB907762:
```suggestion
            <ListItemText inset={weekFilter !== 0}>Past review date</ListItemText>
```

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
```diff
@@ -0,0 +1,193 @@
...
+          </MenuItem>
```
GB907762: Could you use a map and use a type for better maintenance and extension.

frontend/src/metrics/PolicyMetrics.tsx
GB907762: I've noticed this removes the filter if there are no entries on that filter. This is problematic.

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
ARADDCC002: It's not massively needed as the change is based on a UI filter rather than data fetched via the UI, but certainly doesn't hurt to add it.

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
ARADDCC002: > I'd be tempted to have a new function `isOverdue(dueDate: string): boolean` added to `dateUtils.ts` and keep the rendering logic in here?

Yes I think this approach is better. The utils file isn't for a functions that render anything, but re-usable data functions.

PE39806 (dismissed):

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const tableTitle = useMemo(() => {
+    switch (weekFilter) {
+      case WeekFilterOptions.TWO_WEEKS:
+        return 'Entries 2 weeks until their due date'
+      case WeekFilterOptions.TEN_WEEKS:
+        return 'Entries 10 weeks until their due date'
+      case WeekFilterOptions.OVERDUE:
+        return 'Entries past their due date'
+      default:
+        return 'Entries near or past lifecycle review date'
```
GB907762: Language and clarity improvements:
```suggestion
      case WeekFilterOptions.TWO_WEEKS:
        return 'Models to review in 2 weeks or less'
      case WeekFilterOptions.TEN_WEEKS:
        return 'Models to review in 10 weeks or less'
      case WeekFilterOptions.OVERDUE:
        return 'Models past their review date'
      default:
        return 'Models near or past lifecycle review date'
```
I also don't know whether it's misleading to say something is `X days before a due date` But also show results that are overdue?

frontend/src/metrics/PolicyLifecycleMetricsCharts.tsx
```diff
@@ -0,0 +1,194 @@
...
+        <Button
+          id='menu-button'
+          aria-label='filter-menu-button'
+          onClick={handleClick}
+          sx={{ width: 'fit-content' }}
+          variant='outlined'
+          size='small'
+          endIcon={<ArrowDropDown />}
+        >
+          Filter by missing role weeks until review is due
```
GB907762:
```suggestion
          Filter weeks until review is due
```

GB907762 (changes requested):

GB907762 (approved):

PE39806 (approved):

PR #4008, opened 2026-08-06
Improvements to releases with long notes

ARADDCC002:


frontend/src/common/ExpandableTypography.tsx
```diff
@@ -1,18 +1,31 @@
...
   children: string
   maxLength?: number
+  showMarkdown?: boolean
+  showMoreDirection?: 'column' | 'row' | 'row-reverse' | 'column-reverse'
 }
 
-export default function ExpandableTypography({ children: text, maxLength = 100, ...props }: ExpandableTypographyProps) {
+export default function ExpandableTypography({
+  children: text,
+  maxLength = 100,
```
GB907762, line 14: Consider a `useRef` along with a `useLayoutEffect` to evaluate and change the width and height of the element.

frontend/src/common/ExpandableTypography.tsx
```diff
@@ -1,18 +1,31 @@
...
+  showMarkdown = false,
```
GB907762, line 15: Markdown renders images tabels ect..

This causes problematic functionality when truncating these elements.

Look into the AST parser or evaluating the scroll height.

GB907762 (changes requested):

PE39806 (changes requested):
I think this current approach is flawed because markdown syntax length does not directly equate to the rendered content length. Something like `[Documentation](https://gchq.github.io/Bailo/docs/users/models/model-card/importing-model-card-text)` is a lot of raw characters but is very short when rendered, whereas `<svg width="800" height="600" viewBox="0 0 126.829 101.739"><use xlink:href="/emptyBlobLight.svg" fill="red"/></svg>` renders something huge.

Instead, I think it would be better to generate a plain-text preview from parsed markdown AST and measure the rendered text - not the markdown source.

This could be achieved with something like:
```ts
import React from 'react'
import ReactMarkdown from 'markdown-to-jsx'

function extractText(node: React.ReactNode): string {
  if (node == null || typeof node === 'boolean') {
    return ''
  }

  if (typeof node === 'string' || typeof node === 'number') {
    return String(node)
  }

  if (Array.isArray(node)) {
    return node.map(extractText).join(' ')
  }

  if (React.isValidElement(node)) {
    const type = node.type

    // Preserve image presence in preview.
    if (type === 'img') {
      return node.props.alt || '[image]'
    }

    // Ignore non-visible content.
    if (type === 'script' || type === 'style') {
      return ''
    }

    return extractText(node.props.children)
  }

  return ''
}

export function getMarkdownPreview(markdown: string, maxLength: number): string {
  const rendered = <ReactMarkdown>{markdown}</ReactMarkdown>

  const text = extractText(rendered)
    .replace(/\s+/g, ' ')
    .trim()

  if (text.length <= maxLength) {
    return text
  }

  return `${text.slice(0, maxLength).trimEnd()}...`
}
```
with
```ts
const previewText = useMemo(
  () => getMarkdownPreview(text, maxLength),
  [text, maxLength]
)

{expanded ? (
  <MarkdownDisplay>{text}</MarkdownDisplay>
) : (
  <Typography {...props}>{previewText}</Typography>
)}
```
This approach likely builds the full React element tree for the markdown so could be made more efficient by using a package like `markdown-to-jsx` or `remark` for AST without the extra React stuff, but the above should be good enough.

frontend/src/common/ExpandableTypography.tsx
GB907762, line 1: Consider making another component for `ExapndableMarkdown`. You've already largely decoupled these components. They both have specific properties and behaviour.

frontend/src/common/ExpandableTypography.tsx
```diff
@@ -1,24 +1,44 @@
...
+  showMoreDirection = 'row',
+  ...props
```
GB907762, line 14: Could you be a bit more well defined over what is parsed in here. I know this is not a change specific in your PR but could cause confusion/unnecessary handling.

frontend/src/common/ExpandableTypography.tsx
```diff
@@ -1,24 +1,44 @@
...
+    () => (showMarkdown ? getMarkdownPreview(text, maxLength) : null),
+    [maxLength, showMarkdown, text],
+  )
+  const isTruncated = showMarkdown ? markdownPreview?.truncated : text.length > maxLength
+
+  if (isTruncated) {
     return (
-      <Stack sx={{ mb: 1, alignItems: 'center' }} direction={expanded ? 'column' : 'row'} spacing={1}>
-        <Typography {...props}>{expanded ? text : `${text.slice(0, maxLength).trimEnd()}...`}</Typography>
+      <Stack sx={{ mb: 1, alignItems: 'center' }} direction={expanded ? 'column' : showMoreDirection} spacing={1}>
```
GB907762: `alignItems: 'center'` centres some markdown elements . This misaligns styles from this expandable design to the readonly version seen in the single view version of releases

frontend/src/common/ExpandableTypography.tsx
```diff
@@ -1,24 +1,44 @@
...
+          <Typography {...props}>{expanded ? text : `${text.slice(0, maxLength).trimEnd()}...`}</Typography>
+        )}
         <Button size='small' onClick={() => setExpanded(!expanded)}>
           {expanded ? 'Show less' : 'Show more'}
         </Button>
       </Stack>
     )
   } else {
-    return <Typography {...props}>{text}</Typography>
+    return showMarkdown ? <MarkdownDisplay>{text}</MarkdownDisplay> : <Typography {...props}>{text}</Typography>
```
GB907762: Is it worth making this a link to the release page instead? 

I can see if you have many expanded releases this can fill up the page quickly.

frontend/utils/markdownUtils.ts
```diff
@@ -0,0 +1,159 @@
...
+    return truncatedItems
+  }
+
+  const truncateNode = (node: ASTNode): ASTNode | null => {
+    switch (node.type) {
+      case RuleType.text:
+      case RuleType.codeInline:
+      case RuleType.codeBlock:
+      case RuleType.footnoteReference: {
+        const truncated = truncateText(node.text, remaining)
```
GB907762, line 94: Be aware `truncateText` has a different behaviour to `truncateNode`.  So this allows stuff like images to render in a footer.

frontend/utils/markdownUtils.ts
```diff
@@ -0,0 +1,159 @@
...
+      case RuleType.table: {
+        const header = node.header.map(truncateChildren).filter((cell) => cell.length > 0)
+        const cells = node.cells
+          .map((row) => row.map(truncateChildren).filter((cell) => cell.length > 0))
+          .filter((row) => row.length > 0)
+
+        return header.length > 0 || cells.length > 0 ? { ...node, header, cells } : null
+      }
+      case RuleType.image:
+        return { ...node, type: RuleType.htmlBlock, text: '_[image]_', tag: 'div' }
```
GB907762: I think we should remove this and display the image. Could provide important context like diagrams ect.

GB907762 (changes requested):

frontend/src/common/ExpandableTypography.tsx
ARADDCC002: I'd rather it showed the full release notes when you select "view more", but happy to change it in the future if others disagree with me!

frontend/utils/markdownUtils.ts
ARADDCC002: I think it's fine to ignore for now and resolve if it becomes a notable problem in the future.

frontend/utils/markdownUtils.ts
ARADDCC002: I've now displayed the image, but it's set to a height of 400.

frontend/utils/markdownUtils.ts
GB907762: Yep this was more about displaying an image as `[image]` in the main body but as a full image in the footer

GB907762 (dismissed):

JRB66955 (dismissed):

PE39806 (changes requested):
I'm not sure what but if making the release notes be our `README.md` files then the badges become incredibly large, causing significant horizontal overflow and adding horizontal scrollbars.

Edit: pretty sure it must be `frontend/utils/markdownUtils.ts` > `case RuleType.image:` > `height=400`. Maybe this should be `style="max-height:400px;max-width:100%"`?

I also think adding something like
```tsx
img: {
  component: (props: any) => <img style={{ maxHeight: '100%', maxWidth: '100%' }} {...props} />,
},
```
to `frontend/src/common/MarkdownDisplay.tsx` is a good idea to prevent images overflowing when viewing a release in full.

PE39806 (dismissed):

GB907762 (approved):

PE39806 (approved):

PR #4035, opened 2026-08-10
Fix initial redirect

GB907762:
Ensure path safety with tests. Route users to api/login/callback

Callback redirect is redundant in our config.

### DIFF
--- a/backend/test/connectors/authentication/oauth.spec.ts
+++ b/backend/test/connectors/authentication/oauth.spec.ts
@@ -1,4 +1,5 @@
-import { Request, Response } from 'express'
+import express, { Request, Response } from 'express'
+import supertest from 'supertest'
 import { describe, expect, test, vi } from 'vitest'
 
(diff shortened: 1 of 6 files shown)


JR40159 (dismissed):

backend/test/connectors/authentication/oauth.spec.ts
```diff
@@ -74,6 +75,50 @@ describe('connectors > authentication > oauth', () => {
     expect(router.stack).toMatchSnapshot()
   })
 
+  test.each([
```
PE39806, line 78: Can you add cases to check encoded path traversal e.g.
```ts
/%2F%2Fmalicious.example
/\evil
/%5C%5Cevil
/%0d%0aLocation:https://evil.example
```

PE39806 (changes requested):

JR40159 (approved):

PE39806 (approved):

PR #4036, opened 2026-08-10
Decrease Docker image size

PE39806:
- Switch `backend/docs` build to use `alpine` image variant
- Update various `.dockerignore` files to reduce files copied in
- Simplify a few install steps to be quicker & smaller
- Explicit `prod` target where previously implicit
- Reorder some install steps to reduce docker layer cache invalidation

Contributes towards #3284


JRB66955 (dismissed):

lib/python/.dockerignore
GB907762, line 1: We have much of the stuff removed in our `gitignore` that I don't think should be there. 

Do you have any thoughts on this?

lib/python/.dockerignore
```diff
@@ -1,146 +1,39 @@
...
+.coverage
+htmlcov/
+*.lcov
 
-# VS Code
+# IDE
 .vscode/
 
-#resnet
-resnet50_weights.pth
```
GB907762, line 143: This is an artefact produced by one of the notebooks.

Is it necessary to mount this to the container?

GB907762 (changes requested):

lib/python/.dockerignore
PE39806: Good point - I had thought it was a legacy item but it should still be included as it is still producible.

lib/python/.dockerignore
PE39806: I'll update the various `.gitignore`s to align with these changes.

JRB66955 (approved):

GB907762 (approved):

PR #4037, opened 2026-08-11
Fix/model card diff

GB907762:
- Update model selector with a debounce of a quarter of a second to prevent unnecessary requests. 
- Fix `onDelete` so that when the model autocompletes are removed the query is changed. 
- When you select a new model it defaults to the latest version. 
- Use unified view of card snapshots to remove the ambiguity around mirrored models states and changes
- When comparing mirrored and non-mirrored models Bailo compares the read-only information in a mirrored model card with the editable information on a non-mirrored model. For single source of truth.


JRB66955 (approved):

frontend/src/entry/overview/EntryCardSnapshotSelector.tsx
```diff
@@ -0,0 +1,215 @@
...
+          component='button'
+          type='button'
+          variant='outlined'
+          disabled={disabled}
+          onClick={() => setOpen(true)}
+          aria-haspopup='dialog'
+          sx={{
+            p: 2,
+            height: '100%',
+            textAlign: 'left',
```
ARADDCC002, line 139: This should default to `left` anyway so can be removed.

frontend/src/entry/overview/EntryCardSnapshotSelector.tsx
```diff
@@ -0,0 +1,215 @@
...
+            <Table stickyHeader aria-label='Card revision' sx={{ minWidth: 550 }}>
+              <TableHead>
+                <TableRow>
+                  <TableCell>Version</TableCell>
+                  <TableCell>Created By</TableCell>
+                  <TableCell>Created At</TableCell>
+                </TableRow>
+              </TableHead>
+              <TableBody>
+                {[...snapshots].reverse().map((snapshot) => {
```
ARADDCC002, line 175: This map should be moved outside of the main render

ARADDCC002 (changes requested):

frontend/src/entry/overview/EntryCardSnapshotSelector.tsx
GB907762: The default for a button is center.

frontend/src/entry/overview/EntryCardSnapshotSelector.tsx
GB907762: The snapshot is memorised anyway. I can't see any meaningful impact in moving this outside the main render as this a relatively inexpensive mapping.

ARADDCC002 (approved):

PR #4047, opened 2026-08-12
Added deployment assessment schema management

ARADDCC002:
* Minimal and core deployment assessment schemas added
* Ability for admins to view and add deployment assessment schemas via admin panel
* `modelSelector` widget added


backend/config/default.cjs
```diff
@@ -150,6 +150,14 @@ module.exports = {
         reviewRoles: ['msro'],
       },
     ],
+    deploymentAssessments: [
+      {
+        name: 'Minimal Deployment Assessment Schema v1',
+        id: 'minimal-deployment-assessment-schema-v1',
+        description: 'TBC',
```
GB907762:
```suggestion
        description: 'A minimal deployment assessment capturing the deployment name and details.',
```

backend/src/services/schema.ts
```diff
@@ -292,6 +292,18 @@ export async function addDefaultSchemas() {
...
+    log.info({ name: schema.name, reference: schema.id }, `Ensuring schema ${schema.id} exists`)
+    const modelSchema = new SchemaModel({
+      ...schema,
+      kind: SchemaKind.DeploymentAssessment,
+      active: true,
+      hidden: false,
+    })
+    await SchemaModel.deleteOne({ id: schema.id })
+    await modelSchema.save()
+  }
```
GB907762: We have a bunch of these, for access requests, models, ect.. Could we flat map them. So that we don't have to copy and paste this code block three times.

backend/src/scripts/example_schemas/core_deployment_assessment_schema.json
```diff
@@ -0,0 +1,39 @@
...
+          "title": "Who is the risk owner attached to this deployment assessment?",
+          "type": "string",
+          "widget": "entitySelector"
+        },
+        "riskOwnerJustification": {
+          "title": "Justify why the risk owner has been assigned",
+          "type": "string"
+        },
+        "entryList": {
+          "title": "List all models assigned to this deployment assessment",
```
GB907762: I think this should be outside the schema for better searching and atomicity in the database. And shouldn't be a widget.

backend/src/scripts/example_schemas/core_deployment_assessment_schema.json
```diff
@@ -0,0 +1,39 @@
...
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "minLength": 1,
```
GB907762:
```suggestion
          "minItems": 1,
```

backend/src/scripts/example_schemas/minimal_deployment_assessment_schema.json
GB907762, line 1: `backend/src/scripts/example_schemas/minimal_deployment_assessment.json` -> `backend/src/scripts/example_schemas/minimal_deployment_assessment_schema.json`

backend/src/scripts/example_schemas/minimal_deployment_assessment.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
GB907762:
```suggestion
  "$schema": "https://json-schema.org/draft/2020-12/schema",
```

backend/src/scripts/example_schemas/minimal_deployment_assessment_schema.json
```diff
@@ -0,0 +1,26 @@
...
+          "description": "This will be used to distinguish your deployment assessment.",
+          "type": "string"
+        },
+        "deploymentDate": {
+          "title": "When is this deployment due to start?",
+          "type": "string",
+          "format": "date"
+        }
+      },
+      "required": ["deploymentSummary", "deploymentDate"],
```
GB907762, line 20:
```suggestion
      "required": ["name", "deploymentDate"],
```

backend/src/scripts/example_schemas/minimal_deployment_assessment.json
```diff
@@ -0,0 +1,26 @@
...
+          "title": "When is this deployment due to start?",
+          "type": "string",
+          "format": "date"
+        }
+      },
+      "required": ["deploymentSummary", "deploymentDate"],
+      "additionalProperties": false
+    }
+  },
+  "required": ["overview"],
```
GB907762:
```suggestion
  "required": ["about"],
```

backend/src/scripts/example_schemas/core_deployment_assessment_schema.json
GB907762, line 1: What's the justification for having two schemas?

You aren't using this one in the code.

backend/src/scripts/example_schemas/core_deployment_assessment_schema.json
```diff
@@ -0,0 +1,39 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
GB907762:
```suggestion
  "$schema": "http://json-schema.org/draft/2020-12/schema",
```

GB907762 (changes requested):

backend/src/scripts/example_schemas/core_deployment_assessment_schema.json
ARADDCC002: Sorry I appreciate this isn't clear from looking at just this code, these "core" properties will be stored on the base model itself, which means they can be properly indexed and used throughout the application. This core schema is so that we can generate a page on the RJSF form. 

The reason for using a widget is so that we can have unique behaviour in the future for how we display the models in the list, as there's thoughts about having hover displays for each chip.

backend/src/scripts/example_schemas/core_deployment_assessment_schema.json
ARADDCC002: The deployment assessment will have some base properties which can be seen on the mongoose model, such as name, riskOwner and entryList. These are important properties that the application _needs_ in order to use deployment assessments. 

The minimal schema, like we do for other things, is for _all_ other properties that can be entirely custom. 

The idea behind this approach is that we can generate one form using both schemas - the core one will be the same for every single instance of Bailo, the minimal one can be changed and swapped out to include whatever information that instance/organisation needs.

GB907762: This approach will create extra overhead. To maintain consistency and durability of the database, I would suggest create a mongo document with the core information and have a relationship on a card in the way we have on models.

backend/src/scripts/example_schemas/core_deployment_assessment_schema.json
GB907762: See https://github.com/gchq/Bailo/pull/4047#discussion_r3773470124

backend/src/scripts/example_schemas/minimal_deployment_assessment_schema.json
ARADDCC002: "deploymentSummary" is correct I've just put it in the wrong place in the actual question

JR40159 (approved):

JR40159 (approved):

backend/src/services/schema.ts
```diff
@@ -38,9 +38,56 @@ export async function searchSchemas(
...
+          title: 'List all models assigned to this deployment assessment',
+          type: 'array',
+          items: {
+            type: 'string',
+          },
+          minItems: 1,
+          widget: 'modelSelector',
+        },
+      },
+      required: ['name', 'riskOwner', 'entryList'],
```
PE39806: Possibly
```suggestion
      required: ['name', 'riskOwner', 'riskOwnerJustification', 'entryList'],
```

frontend/src/MuiForms/EntitySelector.tsx
```diff
@@ -180,12 +200,17 @@ export default function EntitySelector({
       ) : compare.inMirroredCompare && currentValue.length ? (
         <InlineDiff from={formatEntityValue(compare.compareFromState)} to={currentValueString} />
       ) : (
+        currentValue &&
         currentValue.length > 0 && (
```
PE39806, line 204: Are both `currentValue.length` still safe or does there need to first be some `Array.isArray(currentValue)`?

backend/src/services/schema.ts
```diff
@@ -256,42 +307,25 @@ export async function updateSchema(user: UserInterface, schemaId: string, diff:
...
+    const schemaModel = new SchemaModel({
       ...schema,
-      kind: SchemaKind.DataCard,
+      kind,
       active: true,
       hidden: false,
     })
     await SchemaModel.deleteOne({ id: schema.id })
-    await dataCardSchema.save()
+    await schemaModel.save()
```
PE39806: Maybe we should use `await SchemaModel.replaceOne(..., { upsert: true })` here as it's atomic?

PE39806 (changes requested):

frontend/src/MuiForms/EntitySelector.tsx
ARADDCC002: It does still work for strings

JR40159 (approved):

PE39806 (approved):

GB907762 (approved):

PR #4048, opened 2026-08-13
Restrict all access UM to role

GB907762:


ARADDCC002 (dismissed):

backend/test/connectors/authorisation/base.spec.ts
```diff
@@ -181,7 +181,7 @@ describe('connectors > authorisation > base', () => {
     })
   })
 
-  test('model > create untrusted model without the untrusted model role', async () => {
+  test('model > manage untrusted model without the untrusted model role', async () => {
```
PE39806: Please add a test case to also cover `mockAuthentication.hasRole.mockResolvedValueOnce(true)` for the management of an UM when the user does have permission.

PE39806 (changes requested):

JR40159 (dismissed):

PE39806 (approved):

JR40159 (approved):

PR #4050, opened 2026-08-17
Move & extract CI/CD files

PE39806:
Refactor GitHub workflows to extract inline files, and move CI/CD only compose files to declutter root dir.


.github/workflows/build.yml
```diff
@@ -456,54 +404,14 @@ jobs:
...
-          }
-          EOT
+          cp .github/resources/config/local.cjs.template backend/config/local.cjs
+          if [ "${{ matrix.job_name }}" = "cross_instance" ]; then
+            sed -i 's|{{SMTP_ENABLED}}|enabled: false,|' backend/config/local.cjs
+            sed -i "s|{{EXTRA_CONNECTORS}}|connectors: { artefactScanners: { kinds: [] } },|" backend/config/local.cjs
+          else
+            sed -i 's|{{SMTP_ENABLED}}||' backend/config/local.cjs
+            sed -i 's|{{EXTRA_CONNECTORS}}||' backend/config/local.cjs
+          fi
```
GB907762, line 411: Would it be better to use environment variables to configure these, instead of templates? 
Such that you have two files one for cross instance and otherwise?

[https://github.com/node-config/node-config/wiki/Configuration-Files#multi-instance-deployments](https://github.com/node-config/node-config/wiki/Configuration-Files#multi-instance-deployments)

PE39806: I'll split it into two files but won't use an env var as we can just set the name in the CI/CD `cp` step.

GB907762 (approved):

JR40159 (approved):

PR #4053, opened 2026-08-17
Allow users to set lifecycle review dates that are in the past

PE39806:
Allows users to set a lifecycle review in the past so that they can migrate existing model cards to Bailo.


JR40159 (dismissed):

frontend/src/entry/EntryOverviewDetails.tsx
```diff
@@ -110,135 +121,149 @@ export default function EntryOverviewDetails({ entry, mutateEntry }: Organisatio
...
+            )}
+          </Stack>
+          {entry.kind !== EntryKind.DATA_CARD && entry.card && (
+            <Stack spacing={1}>
+              <Typography color='primary' sx={{ fontWeight: 'bold' }}>
+                Model card review
+              </Typography>
+              {updateEntryPermission.hasPermission && reviews.length === 0 && (
+                <>
+                  <DatePicker
```
GB907762, line 166: Consider having a minimum review date in the frontend. At the moment 1900 is accessible.

GB907762 (changes requested):

JR40159 (approved):

GB907762 (approved):

PR #4055, opened 2026-08-18
Fix show more on expandable typography to use webkit clamp instead of string truncation

GB907762:


JR40159 (changes requested):
Add some tests for this component

ARADDCC002 (approved):

JR40159 (approved):

PR #4056, opened 2026-08-18
Create a deployment assessment using the frontend

JR40159:


frontend/pages/deployment-assessment/new.tsx
```diff
@@ -0,0 +1,148 @@
...
+    const data = getStepsData(splitSchema, true)
+    const res = await postDeploymentAssessment(schemaId, data, false)
+
+    if (!res.ok) {
+      setErrorText(await getErrorMessage(res))
+      setSubmitButtonLoading(false)
+      return
+    }
+
+    router.push('/deployments')
```
GB907762: Invalid URL.

frontend/pages/deployment-assessments/new.tsx
```diff
@@ -0,0 +1,148 @@
...
+  })
+  if (error) {
+    return error
+  }
+
+  if (!schemaId) {
+    return (
+      <>
+        <Title text='Select a schema' />
+        <SchemaSelect
```
GB907762: I know it's not a problem with this PR specifically but the `SchemaButton`s share a loading state which causes all schema buttons to spin at the same time.

frontend/pages/deployment-assessment/new.tsx
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  if (!schemaId) {
+    return (
+      <>
+        <Title text='Select a schema' />
+        <SchemaSelect
+          schemaKind={SchemaKind.ACCESS_REQUEST}
+          onSchemaSelect={handleSchemaSelect}
+          backHref='/deployments'
```
GB907762:
```suggestion
          backHref='/deployment-assessment'
```

frontend/src/wrapper/SideNavigation.tsx
```diff
@@ -160,7 +160,7 @@ export default function SideNavigation({
                   badgeCount={reviewCount}
                 />
                 <NavMenuItem
-                  href='/deployments'
+                  href='/deployment-assessment'
```
GB907762: I understand why this was done. Although might it be better to clarify the language around deployment assessments?

Usually you should avoid noun phrases in REST and instead split the hierarchy, this ensures they are atomic and we can build on top of them in the future. For example:

- `/deployments` A collection of deployments (plural to denote a collection)
- `/deployments/assessment/{id}` When you're viewing a deployment assessment
- `/deployments/inference/{id}` - Easy to build in the future with this approach

frontend/pages/deployment-assessments/new.tsx
```diff
@@ -0,0 +1,148 @@
...
+    router.push(`/deployment-assessments/new?schemaId=${selected.id}`)
+  }
+
+  async function onSaveDraft() {
+    setErrorText('')
+    setDraftSuccessText('')
+    setDraftButtonLoading(true)
+
+    if (!schemaId) {
+      setErrorText('Please wait until the page has finished loading before attempting to save.')
```
IR96334, line 49: As isLoading (in SchemaFormPage) is checked before rendering button to save draft, this _should_ never be possible.

frontend/pages/deployment-assessments/new.tsx
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  async function onSubmit() {
+    setErrorText('')
+    setDraftSuccessText('')
+    setSubmitButtonLoading(true)
+    setFormValidationErrorState(false)
+
+    if (!schemaId) {
+      setErrorText('Please wait until the page has finished loading before attempting to submit.')
```
IR96334, line 75: Same with above

frontend/src/schemas/SchemaSelect.tsx
```diff
@@ -57,24 +60,24 @@ export default function SchemaSelect({ schemaKind, entry }: SchemaSelectProps) {
 
   const accessRequestCallback = useCallback(
     async (newSchema: SchemaInterface) => {
-      setLoading(true)
-      router.push(`/model/${entry.id}/access-request/new?schemaId=${newSchema.id}`)
+      setLoadingSchemaId(newSchema.id)
+      router.push(`/model/${entry!.id}/access-request/new?schemaId=${newSchema.id}`)
```
IR96334: Is `!` supposed to be here?

JR40159: Now that this is also being re-used for DAs, `entry` is now optional

frontend/pages/deployment-assessments/new.tsx
JR40159: Following this pattern: https://github.com/gchq/Bailo/blob/ad707da6a1060ee232d16ba8b1ac22513e9f9bab/frontend/pages/model/%5BmodelId%5D/access-request/new.tsx#L62-L66

It's more of a type safety thing as `schemaId` could be undefined.

frontend/src/schemas/SchemaSelect.tsx
IR96334: This relationship between DA and no entry should be programatic, something like:
`type SchemaSelectProps = {
  schemaKind: Exclude<SchemaKindKeys, typeof SchemaKind.DEPLOYMENT_ASSESSMENT>
  entry?: EntryInterface
} & { schemaKind: typeof SchemaKind.DEPLOYMENT_ASSESSMENT; entry?: never }`

IR96334: After further investigation, this would take a small re-factor. It would end up looking similar to EditableRelease on main, where you separate the entry-specific functionality into its own component.

frontend/src/entry/overview/FormEditPage.tsx
```diff
@@ -59,6 +59,31 @@ export type RouterQueryParams = {
   isEdit?: boolean
 }
 
+export function removeEmptyValues(value) {
```
IR96334: Don't forget to move this to formUtils

frontend/src/schemas/SchemaSelect.tsx
JR40159: Should now be updated

backend/src/routes/v3/deploymentAssessment/postDeploymentAssessment.ts
```diff
@@ -6,23 +6,38 @@ import { z } from '../../../lib/zod.js'
...
+    })
+    .strict(),
+  z
     .object({
+      name,
       schemaId: deploymentAssessmentSchemaIdSchema,
-      metadata: deploymentAssessmentMetadataSchema,
-      draft: deploymentAssessmentDraftSchema.optional().default(true),
+      metadata: deploymentAssessmentMetadataSchema.optional(),
+      draft: z.literal(true).optional().default(true),
```
PE39806: Prefer `z.coerce.boolean()` as this is how we use this elsewhere

backend/src/services/deploymentAssessment.ts
```diff
@@ -154,32 +167,28 @@ export async function createDeploymentAssessment(user: UserInterface, params: Cr
...
-    throw BadReq('Deployment assessment metadata could not be validated against the schema.', { errors })
-  }
-
-  const metadata = params.metadata as DeploymentAssessmentMetadata
-  const { name, riskOwner, modelIds } = metadata.overview
-
-  if (!params.draft && !riskOwner) {
-    throw BadReq('Deployment risk owner is required')
-  }
+  if (params.metadata) {
```
PE39806: Add another check to throw an error when `params.metadata` is unset and this is not a `draft`.

PE39806 (changes requested):

backend/src/routes/v3/deploymentAssessment/postDeploymentAssessment.ts
JR40159: I used `z.literal` here so the schema can differentiate between `true` and `false`

backend/src/services/deploymentAssessment.ts
JR40159: Would this not be duplicating the validation implemented in the zod schema?

PE39806: My concern is that this function could be called by another internal function in the service layer rather than coming through the `routes` layer, in which case the zod schema wouldn't be validated. Same applies to other fields that are only optional while in `draft` state.

backend/src/services/deploymentAssessment.ts
```diff
@@ -31,10 +30,23 @@ export interface SearchDeploymentAssessmentsParams {
...
-  metadata: unknown
+export interface SearchDeploymentAssessmentsParams {
+  schemaId?: string
+  modelIds?: string[]
+  riskOwner?: string
+  createdBy?: string
+  createdAfter?: string
+  createdBefore?: string
+  draft?: boolean
+  search?: string
```
PE39806: Duplicate definition

backend/src/services/deploymentAssessment.ts
```diff
@@ -4,12 +4,11 @@ import { QueryFilter } from 'mongoose'
...
 import authorisation from '../connectors/authorisation/index.js'
-import DeploymentAssessmentModel, {
-  DeploymentAssessmentInterface,
-  DeploymentAssessmentMetadata,
-} from '../models/DeploymentAssessment.js'
+import { z } from '../lib/zod.js'
+import DeploymentAssessmentModel, { DeploymentAssessmentInterface } from '../models/DeploymentAssessment.js'
 import ModelModel, { EntryKind, EntryVisibility, SystemRoles } from '../models/Model.js'
 import { UserInterface } from '../models/User.js'
+import { deploymentAssessmentSchema } from '../routes/v3/deploymentAssessment/postDeploymentAssessment.js'
```
PE39806: Avoid importing from `routes` if possible as this creates a circular import, which often causes hard-to-debug issues with our vitest suite.

frontend/src/MuiForms/ModelSelector.tsx
```diff
@@ -94,6 +110,10 @@ export default function ModelSelector({
       .join('\n')
   }
 
+  if (currentValue.length === 1 && currentValue[0] === null) {
+    onChange([])
+  }
```
PE39806, line 115: Should probably be inside a `useEffect` rather than directly in the render body.

frontend/src/MuiForms/ModelSelector.tsx
```diff
@@ -38,12 +39,27 @@ export default function ModelSelector({
...
   const [modelListQuery, setModelListQuery] = useState('')
   const theme = useTheme()
+  const uiConfig = useContext(UiConfigContext)
 
   const {
     entries: models,
     isEntriesLoading: isModelsLoading,
     isEntriesError: isModelsError,
-  } = useListEntries(EntryKind.MODEL)
+  } = useListEntries(
```
PE39806, line 48: Potentially out of scope but I think we should swap to take an options object, rather than having so many params where many are `undefined`.

infrastructure/helm/bailo/templates/bailo/bailo.configmap.yaml
```diff
@@ -244,10 +244,10 @@ data:
           untrustedModelShortDescription: '{{ .Values.config.ui.untrustedModel.untrustedModelShortDescription }}',
           fileUploadGuidance: '{{ .Values.config.ui.untrustedModel.fileUploadGuidance }}',
         },
-      },
 
-      deploymentAssessments: {
-        deployableModelState: '{{ .Values.config.deploymentAssessments.deployableModelState }}',
+        deploymentAssessments: {
+          deployableModelState: '{{ .Values.config.deploymentAssessments.deployableModelState }}',
```
PE39806: This should be moved to be under `ui`, which also requires updating `values.yaml`.

PE39806 (changes requested):

backend/src/services/deploymentAssessment.ts
JR40159: The type I have implemented for `createDeploymentAssessment` that is created from the zod schema conditionally type metadata based on the draft property which should provide protection _if_ other internal functions call this method. In general, I'd always prefer to keep things simple and implement for the requirements we have rather than requirements that could happen in the future

backend/src/routes/v3/deploymentAssessment/postDeploymentAssessment.ts
```diff
@@ -6,23 +6,38 @@ import { z } from '../../../lib/zod.js'
...
   registerPath,
 } from '../../../services/specification.js'
 import { parse } from '../../../utils/validate.js'
 
-export const postDeploymentAssessmentSchema = z.object({
-  body: z
+const name = z
+  .string()
+  .min(1, 'You must provide a deployment assessment name')
+  .openapi({ example: 'Just A Rather Very Intelligent System' })
```
JRB66955:
```suggestion
const name = z
  .string()
  .trim()
  .min(1, 'You must provide a deployment assessment name')
  .openapi({ example: 'Just A Rather Very Intelligent System' })
```

I think we should trim whitespace at the start/end otherwise " " would be a valid deployment assessment name (see https://zod.dev/api?id=strings#strings)

frontend/src/MuiForms/ModelSelector.tsx
JR40159: Happy to however, I took this pattern from elsewhere in the codebase:
https://github.com/gchq/Bailo/blob/184ac9d8fc7744ac09f93703b06ae99eb640eb19/frontend/src/MuiForms/MultipleDropdown.tsx#L58-L64

infrastructure/helm/bailo/templates/bailo/bailo.configmap.yaml
JR40159: It is under `ui`...?

PE39806: `.Values.config.deploymentAssessments.deployableModelState` -> `.Values.config.ui.deploymentAssessments.deployableModelState`

JR40159: Yep, updated that. Just confirming that it's already in the `ui` section of the configmap

backend/src/services/specification.ts
```diff
@@ -115,6 +115,31 @@ export const deploymentAssessmentSummarySchema = z.object({
...
+      name: z
+        .string()
+        .min(1, 'You must provide a deployment assessment name')
+        .openapi({ example: 'Just A Rather Very Intelligent System' }),
+      schemaId: deploymentAssessmentSchemaIdSchema,
+      metadata: deploymentAssessmentMetadataSchema.optional(),
+      draft: z.literal(true).optional().default(true),
+    })
+    .strict(),
+])
```
PE39806, line 137: Could slightly reduce duplication with
```suggestion
const baseDeploymentAssessmentSchema = z.object({
  name: z
    .string()
    .min(1, 'You must provide a deployment assessment name')
    .openapi({ example: 'Just A Rather Very Intelligent System' }),
  schemaId: deploymentAssessmentSchemaIdSchema,
})

export const deploymentAssessmentSchema = z.discriminatedUnion('draft', [
  baseDeploymentAssessmentSchema
    .extend({
      metadata: deploymentAssessmentMetadataRequiredSchema,
      draft: z.literal(false),
    })
    .strict(),
  baseDeploymentAssessmentSchema
    .extend({
      metadata: deploymentAssessmentMetadataSchema.optional(),
      draft: z.literal(true).optional().default(true),
    })
    .strict(),
])
```

JR40159: I knew this was coming :D

frontend/src/MuiForms/ModelSelector.tsx
JR40159: I'd say out of scope personally. I tried to keep the change to the entry search functionality to a minimum. I also think in the near future there will be a v3 version of this endpoint and corresponding actions function which could address this.

backend/src/models/DeploymentAssessment.ts
```diff
@@ -15,6 +14,7 @@ export interface DeploymentAssessmentMetadata {
 
 export interface DeploymentAssessmentInterface {
   id: string
+  name: string
```
PE39806, line 17: Do we want `name` to be editable? If so, update `backend/src/routes/v3/deploymentAssessment/patchDeploymentAssessment.ts` & `updateDeploymentAssessment`

frontend/pages/deployment-assessments/new.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const data = getStepsData(splitSchema, true)
+    const res = await postDeploymentAssessment(name, schemaId, removeEmptyValues(data), true)
+
+    if (!res.ok) {
+      setErrorText(await getErrorMessage(res))
+      setDraftButtonLoading(false)
+      return
+    }
+
+    setDraftSuccessText('Draft saved successfully.')
```
PE39806: I like this way of creating a draft and staying on it, but I think it would be good to also display a link to the view DA page (`/deployments/<da-id>`) when a draft is saved. Otherwise, there's no indicator what this DA's unique ID is.

PE39806: On second thoughts, repeatedly clicking "Save draft" sends a new POST request every time so multiple DAs can be accidentally created. Either this should POST once and then redirect, or the first should be a POST with subsequent saves (and the submit) being PATCHes.

PE39806 (changes requested):

frontend/pages/deployment-assessments/new.tsx
```diff
@@ -0,0 +1,153 @@
...
+      onSubmit={onSubmit}
+      submitButtonLoading={submitButtonLoading}
+      formValidationErrorState={formValidationErrorState}
+      errorText={errorText}
+      onSaveDraft={onSaveDraft}
+      draftButtonLoading={draftButtonLoading}
+      draftSuccessText={draftSuccessText}
+      disableActions={!name ? 'Please enter a deployment assessment name' : undefined}
+    >
+      <LabelledInput fullWidth label='Deployment Assessment Name' htmlFor='deployment-assessment-name'>
```
ARADDCC002, line 141: I think that this input should be below the "Select a different schema" button.

ARADDCC002 (changes requested):

backend/src/services/schema.ts
PE39806:
```suggestion
        required: ['riskOwner', 'justification', 'modelIds'],
```

PE39806 (changes requested):

frontend/pages/deployment-assessments/new.tsx
JR40159: I'm not sure how easy this would be to do with the shared `SchemaFormPage`. Any ideas? My thinking was trying to keep the "Select a different schema" button and the form together on the page

PE39806: As the back button is not a form control in `frontend/src/schemas/SchemaFormPage.tsx`, I think it would be safe to hoist it out of `!hideForm` and it put first inside `Paper`, followed by `children`. Something like
```tsx
<Paper sx={{ mx: 'auto', my: 4, p: 4 }}>
  <Stack spacing={4}>
    <Link href={backHref}>
      <Button sx={{ width: 'fit-content' }} startIcon={<ArrowBack />}>
        {backLabel}
      </Button>
    </Link>
    {children}
    {!hideForm && (
      <>
        <JsonSchemaForm ... />
        <Stack spacing={1} sx={{ alignItems: 'flex-end' }}>...</Stack>
      </>
    )}
  </Stack>
</Paper>
```

frontend/pages/deployment-assessments/new.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const data = getStepsData(splitSchema, true)
+    const res = await postDeploymentAssessment(name, schemaId, data, false)
+
+    if (!res.ok) {
+      setErrorText(await getErrorMessage(res))
+      setSubmitButtonLoading(false)
+      return
+    }
+
+    router.push('/deployment-assessments')
```
PE39806: Prefer this to redirect to the DA page using the reponse's ID

frontend/pages/deployment-assessments/new.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    if (!res.ok) {
+      setErrorText(await getErrorMessage(res))
+      setDraftButtonLoading(false)
+      return
+    }
+
+    setDraftButtonLoading(false)
+
+    router.push('/deployment-assessments')
```
PE39806: Prefer this to redirect to the DA page using the reponse's ID

PE39806 (changes requested):

frontend/pages/deployment-assessments/new.tsx
JR40159: It wasn't there 5 mins ago....

PE39806 (dismissed):

ARADDCC002 (dismissed):

ARADDCC002 (approved):

PR #4070, opened 2026-08-18
Prevent docker and user ownership errors on build

PE39806:
Prevents various build permission issues from `npm run build` in `frontend` for build artefacts that docker may generate (and are therefore often owned by root user).

### DIFF
--- a/compose.override.yaml
+++ b/compose.override.yaml
@@ -13,6 +13,7 @@ services:
     volumes:
       - ./frontend:/app
       - /app/node_modules/@next # use container-built @next SWC binaries (musl), not the host's (e.g. glibc)
+      - /app/.next # prevent `uncaughtException [Error: EACCES: permission denied, open '.../Bailo/frontend/.next/...']` on `npm run build`
     build:
       target: dev
 
--- a/frontend/tsconfig.json
+++ b/frontend/tsconfig.json
@@ -22,5 +22,5 @@
     "noImplicitAny": false
   },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "vitest.config.mts", "**/*.mdx"],
-  "exclude": ["node_modules"]
+  "exclude": ["node_modules", "public/docs-search-index.generated.json"]
 }
--- a/frontend/.dockerignore
+++ b/frontend/.dockerignore
@@ -25,6 +25,7 @@ node_modules/
 # Next.js build output
 .next
 out
+next-env.d.ts
 
 # cache
 cache


frontend/tsconfig.json
```diff
@@ -22,5 +22,5 @@
     "noImplicitAny": false
   },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "vitest.config.mts", "**/*.mdx"],
-  "exclude": ["node_modules"]
+  "exclude": ["node_modules", "public/docs-search-index.generated.json"]
```
GB907762, line 25: Could you explain why this is necessary. On the surface the documentation index is not captured in the include array?

compose.override.yaml
```diff
@@ -13,6 +13,7 @@ services:
     volumes:
       - ./frontend:/app
       - /app/node_modules/@next # use container-built @next SWC binaries (musl), not the host's (e.g. glibc)
+      - /app/.next # prevent `uncaughtException [Error: EACCES: permission denied, open '.../Bailo/frontend/.next/...']` on `npm run build`
```
GB907762, line 16: Not sure but it looks like this anonymous volume prevents access to scripts like `npm run build` from the non-docker user. Is this something we want?

GB907762 (changes requested):

frontend/tsconfig.json
PE39806: It fixes the following permission error for a fresh repo:

```console
$ git clone git@github.com:gchq/Bailo.git bailo-test
$ cd bailo-test
$ npm i
$ npm run certs
$ docker compose build --parallel --no-cache
$ docker compose up -d
$ docker compose down -v --remove-orphans
$ cd frontend
$ npm run build
npm notice run prebuild
npm notice run tsx scripts/build-docs-search-index.mjs
node:fs:2914
    return binding.writeFileUtf8(
                   ^

Error: EACCES: permission denied, open '.../bailo-test/frontend/public/docs-search-index.generated.json'
    at writeFileSync (node:fs:2914:20)
    at writeSearchIndex (file://.../bailo-test/frontend/scripts/docs-search-index.mjs:230:3)
    at main (file://.../bailo-test/frontend/scripts/build-docs-search-index.mjs:17:3)
    at file://.../bailo-test/frontend/scripts/build-docs-search-index.mjs:23:3
    at ModuleJob.run (node:internal/modules/esm/module_job:569:25)
    at async node:internal/modules/esm/loader:650:26
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
  errno: -13,
  code: 'EACCES',
  syscall: 'open',
  path: '.../bailo-test/frontend/public/docs-search-index.generated.json'
}

Node.js v26.7.0
```

I'm not sure why this file is caught by `npm run build` but this is the fix.

PE39806:
This doesn't seem to actually fix the issue I had hoped it would. Closing.

PR #4071, opened 2026-08-18
Add warning if model state is unset

JR40159:


PE39806 (dismissed):

frontend/src/common/EntrySelect.tsx
```diff
@@ -75,7 +80,28 @@ export default function EntrySelect({
...
+                  : {}
+              }
+              renderValue={(value: string) =>
+                value ? (
+                  value
+                ) : (
+                  <Stack direction='row' spacing={0.5} sx={{ alignItems: 'center' }}>
+                    <em>Unset</em>
+                    {showWarningWhenUnset && (
+                      <Tooltip title={`No ${labelLowerCase} has been set`}>
```
IR96334, line 90: As whole stack has "warning"-ness maybe wrap tooltip around whole stack instead of just symbol

PE39806 (dismissed):

frontend/src/common/EntrySelect.tsx
```diff
@@ -75,7 +80,28 @@ export default function EntrySelect({
               value={value ?? ''}
               onChange={handleSelectOption}
               displayEmpty
-              renderValue={(value: string) => (value ? value : <em>Unset</em>)}
+              sx={
+                showWarningWhenUnset && !value
+                  ? {
+                      '& .MuiOutlinedInput-notchedOutline': { borderColor: theme.palette.warning.main },
+                      '&:hover .MuiOutlinedInput-notchedOutline': { borderColor: theme.palette.warning.dark },
```
GB907762: `theme.palette.warning.dark` is not a theme that exists on Bailo. 

This component already has styling that covers hovering (grey and dark grey) and errors (red). I don't think conveys much additional information users.

frontend/src/common/EntrySelect.tsx
```diff
@@ -75,7 +80,28 @@ export default function EntrySelect({
...
+              }
+              renderValue={(value: string) =>
+                value ? (
+                  value
+                ) : (
+                  <Stack direction='row' spacing={0.5} sx={{ alignItems: 'center' }}>
+                    <em>Unset</em>
+                    {showWarningWhenUnset && (
+                      <Tooltip title={`No ${labelLowerCase} has been set`}>
+                        <WarningIcon color='warning' fontSize='small' />
```
GB907762: The native MUI alerts use an aria alert role to signal to screen readers to read this first thing on the page. Please add:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role

frontend/src/common/EntrySelect.tsx
```diff
@@ -89,7 +115,29 @@ export default function EntrySelect({
...
+            {value ? (
+              <Typography>{value}</Typography>
+            ) : showWarningWhenUnset ? (
+              <Tooltip title={`No ${labelLowerCase} has been set`}>
+                <Stack
+                  direction='row'
+                  spacing={0.5}
+                  sx={{
+                    alignItems: 'center',
+                    backgroundColor: alpha(theme.palette.warning.main, 0.1),
```
GB907762, line 119: The contrast is very low and I don't feel highlighting especially helps when users scan the page. I think the icon and the tooltip you've provided works well.

GB907762 (commented):
Do you think this type of highlighting needs to be available to non model owners?
And does this need to be available on data cards?

JR40159:
> Do you think this type of highlighting needs to be available to non model owners? And does this need to be available on data cards?

Fair point, initially at least I would like this warning to appear on the model for everyone as it adds an extra incentive to the model owners to update the state. I also think this is valid for data cards as they too should have a state set.

GB907762 (approved):

PE39806 (approved):

PR #4072, opened 2026-08-18
Restrict lifecyle future date

PE39806:
Add a (configurable) maximum date that a new lifecycle review can be set to. Also adds a migration script to enforce max date which reports any models which were changed by the migration script.


backend/src/migrations/033_limit_lifecycle_future_date.ts
```diff
@@ -0,0 +1,30 @@
...
+  const reviews = await ReviewModel.find({ kind: ReviewKind.Lifecycle, dueDate: { $gt: maxDate } })
+
+  const modifiedReviews: { modelId: string; previousDueDate: Date }[] = []
+  for (const review of reviews) {
+    // safe cast as we know the date exists from the earlier query
+    const previousDueDate = review.dueDate as Date
+    review.dueDate = maxDate
+    await review.save()
+    modifiedReviews.push({ modelId: review.modelId, previousDueDate })
+  }
```
GB907762, line 23: You may want to use a `bulkWrite`? Although performance might be minor, so feel free to disregard.

frontend/src/entry/EntryOverviewDetails.tsx
```diff
@@ -170,6 +174,7 @@ export default function EntryOverviewDetails({ entry, mutateEntry }: Organisatio
                       setReviewDate(newValue)
                     }}
                     minDate={dayjs('2000/01/01')}
+                    maxDate={increaseCurrentDateByHumanInterval(uiConfig.lifecycle.maxReviewInterval)}
```
GB907762: Is it worth setting the default date to something for example the max date?

backend/test/services/v3/review.spec.ts
GB907762, line 1: Is it worth mocking `config.ui.lifecycle.maxReviewInterval`? and using [vi.setSystemTime](https://vitest.dev/api/vi.html#vi-setsystemtime) along with human interval to be explicit about what the maximum time is?

backend/src/services/v3/review.ts
```diff
@@ -115,6 +122,12 @@ export async function createLifecycleReview(
   modelId: string,
   dueDate: Date,
 ): Promise<ReviewInterface> {
+  if (!isLifecycleReviewDateValid(dueDate)) {
```
GB907762: There's already this validation at the route layer. Do we need this on the service layer?

GB907762 (changes requested):
https://github.com/gchq/Bailo/blob/main/frontend/src/common/ReviewWithComment.tsx#L143-L149
Update these inline

https://github.com/gchq/Bailo/blob/main/backend/src/routes/v3/review/postReviewResponse.ts
I've noticed that we don't have the same validation with this endpoint. Is this something we should think about?

backend/config/default.cjs
```diff
@@ -284,6 +284,10 @@ module.exports = {
     llmImport: {
       enabled: false,
     },
+
+    lifecycle: {
+      maxReviewInterval: '1 year',
```
JR40159: It might make more sense for the default value not change current functionality, ie allow any date in the future. This is mainly due to the migration script.

JR40159 (changes requested):

frontend/src/entry/EntryOverviewDetails.tsx
PE39806: Won't do - defaulting to the current date encourages users to set a date but not for them to be lazy and set it to the furthest date possible.

backend/src/migrations/033_limit_lifecycle_future_date.ts
PE39806: Won't do to maintain readability and because performance is likely not a significant issue.

backend/config/default.cjs
```diff
@@ -284,6 +284,10 @@ module.exports = {
     llmImport: {
       enabled: false,
     },
+
+    lifecycle: {
+      maxReviewInterval: '',
```
GB907762, line 289: Because the default is now removed dates can no longer be set in the future.

GB907762 (changes requested):

backend/config/default.cjs
PE39806: So busy focussing on the backend I forgot to check the frontend...

GB907762 (approved):

JR40159 (changes requested):
Need to record which model IDs have been updated

PE39806:
> Need to record which model IDs have been updated

It does in MongoDB `bailo.migrations` e.g.
```js
{
  _id: ObjectId('6a8710017c11f1eb4c277240'),
  name: '033_limit_lifecycle_future_date.ts',
  metadata: {
    maxDate: ISODate('2027-08-20T14:32:33.069Z'),
    modifiedReviews: [
      {
        modelId: 'q-j61vfs',
        previousDueDate: ISODate('2030-01-01T00:00:00.000Z')
      }
    ]
  },
  createdAt: ISODate('2026-08-20T14:32:33.133Z'),
  updatedAt: ISODate('2026-08-20T14:32:33.133Z'),
  __v: NumberInt('0')
}
```

JR40159 (approved):

PR #4076, opened 2026-08-19
Fix model schema selected SWR cache mismatch

PE39806:
SWR cache key used when invalidating after schema selection didn't match the one used by the model page, so the page kept serving stale data showing the "Create Model Card" dialogue. Switched to `useGetModel` so both pages share the same cache key and the model page updates immediately after a schema is selected.

### DIFF
--- a/frontend/src/schemas/SchemaSelect.tsx
+++ b/frontend/src/schemas/SchemaSelect.tsx
@@ -13,7 +13,7 @@ import {
   Stack,
   Typography,
 } from '@mui/material'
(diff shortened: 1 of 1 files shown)


ARADDCC002 (dismissed):

GB907762 (changes requested):
Data cards seem to contain the same bug.

We should also probably work to resolve these pieces of tech debt towards Version 3?

GB907762 (approved):

GB907762 (approved):

ARADDCC002 (approved):

PR #4082, opened 2026-08-20
Stroom Config updates

JRB66955:


infrastructure/helm/README.md
```diff
@@ -440,7 +440,7 @@ The following tables describe selected configuration options available in `value
 | `connectors.artefactScanners.scanTimeoutMs` | Number of milliseconds waited until the microservice is considered unresponsive | `60000` |
 | `instrumentation.enabled` | Enable OpenTelemetry instrumentation | `false` |
 | `instrumentation.debug` | Enable instrumentation debugging | `false` |
-| `stroom.logOnlyMode` | Enable Stroom audit logging | `true` |
+| `stroom.sendEvents` | Enable Stroom audit logging | `true` |
 | `stroom.feed` | Name of the Stroom feed | `feed` |
```
PE39806: This has been removed

infrastructure/helm/bailo/values.yaml
```diff
@@ -326,7 +326,7 @@ instrumentation:
   debug: false
 
 stroom:
-  logOnlyMode: true
+  sendEvents: true
   feed: 'feed'
```
PE39806: Removed

PE39806 (changes requested):

PE39806 (approved):

backend/src/utils/config.ts
```diff
@@ -194,6 +193,7 @@ export interface Config {
     xmlns: string
     schemaLocation: string
     version: string
+    headers: Record<string, string>
```
GB907762, line 196: Could you use this instead?
[https://nodejs.org/api/globals.html#class-headers](https://nodejs.org/api/globals.html#class-headers)

GB907762 (changes requested):

backend/src/utils/config.ts
JRB66955: Absolutely could, but I'm not sure this would be a worthwhile change given we are just passing these defined headers straight into the fetch request in `stroom.ts`, and it would mean extra work defining it within the yaml/extracting it from the class (rather than just the plain object defined at the moment) - I may be missing something though?

GB907762 (approved):

PR #4084, opened 2026-08-21
Improve unsaved changes popup and coverage

PE39806:
- Replace `window.confirm` browser alerts with MUI `ConfirmationDialogue` for unsaved changes warnings, providing a consistent in-app dialogue experience
- Extend unsaved changes protection to previously unguarded pages: new entry, new release, new access request and model settings/details
- Extend `useUnsavedChange.spec.ts` unit tests and add `PageWithTabs` component test
- Replace deprecated `FormEvent` with `SyntheticEvent`

TODO: check more tabs & pages which may benefit from this feature


frontend/pages/model/[modelId]/access-request/new.tsx
```diff
@@ -128,7 +146,7 @@ export default function NewAccessRequest() {
                 </Link>
                 <JsonSchemaForm
                   splitSchema={splitSchema}
-                  setSplitSchema={setSplitSchema}
+                  setSplitSchema={handleFormChange}
```
Comment (GB907762) [inline frontend/pages/model/[modelId]/access-request/new.tsx]: Not sure if it's because there's a debounce in here. But whether the popup appears seems to oscillate. 

- Create an AR. Type something in a field
- Try to go back, click confirm
- Select a schema and repeat the same

frontend/src/hooks/useUnsavedChanges.ts
GB907762, line 1: There's some bad behaviour if the native window navigation is used. The hook seems to ignore it until you've navigated off this page.

frontend/src/hooks/useUnsavedChanges.ts
```diff
@@ -1,53 +1,70 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-deprecated -- required for cross-browser beforeunload support
       return (e.returnValue = warningText)
     }
 
     const getComparableUrl = (url: string) => {
       const [pathname, queryString = ''] = url.split('?')
       const params = new URLSearchParams(queryString)
-      // Ignore various query params when determining if the URL has meaningfully changed
       params.delete('page')
       params.delete('requiredByModelState')
```
GB907762: I think there's a slight bug in this where if you change the state while in edit mode on the model card, it will fallback to the saved data.

Disregard, if you feel this is out of scope. It's not something introduced on this change.

frontend/src/common/PageWithTabs.tsx
GB907762, line 1: This doesn't get triggered on the model card review. There's also a bug that the `Model card review` depopulates but the `Organisation` and the `State` don't.

GB907762 (changes requested):
Looks like a good start. Could you also consider uploading files?

If you click another page while uploading the connection is severed and the file is not uploaded.

PE39806:
This is becoming too complex with the browser navigation. Closing, but I think there are useful parts that are worth pulling out into another PR for some of the newly checked edit pages.

PR #4086, opened 2026-08-21
Create DA GET API, update audit and auth

PE39806:
- Add new `GET /api/v3/deployment-assessments/{deploymentAssessmentId}` endpoint
- ~~Change `POST /api/v3/deployment-assessments` (plural) to `POST /api/v3/deployment-assessment` (singular)~~ reverted
- Move some specification `const`s to relevant file
- Add new audit event
- Add new auth for DAs
- Update tests

### DIFF
--- a/backend/src/connectors/authorisation/base.ts
+++ b/backend/src/connectors/authorisation/base.ts
@@ -1,4 +1,5 @@
 import { AccessRequestDoc } from '../../models/AccessRequest.js'
+import { DeploymentAssessmentDoc } from '../../models/DeploymentAssessment.js'
 import { FileInterface } from '../../models/File.js'
 import { EntryKind, EntryVisibility, ModelDoc } from '../../models/Model.js'
 import { ReleaseDoc, ReleaseInterface } from '../../models/Release.js'
@@ -21,6 +22,8 @@ import {
   AccessRequestAction,
   AccessRequestActionKeys,
   ActionLookup,
+  DeploymentAssessmentAction,
+  DeploymentAssessmentActionKeys,
   FileAction,
   FileActionKeys,
   ImageAction,
@@ -94,6 +97,14 @@ export class BasicAuthorisationConnector {
     action: AccessRequestActionKeys,
   ) {
     return (await this.accessRequests(user, model, [accessRequest], action))[0]
+  }
+
+  async deploymentAssessment(
+    user: UserInterface,
+    deploymentAssessment: DeploymentAssessmentDoc,
+    action: DeploymentAssessmentActionKeys,
+  ) {
+    return (await this.deploymentAssessments(user, [deploymentAssessment], action))[0]
   }
 
   async response(user: UserInterface, response: ResponseDoc, action: ResponseActionKeys) {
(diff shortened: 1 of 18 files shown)


backend/src/routes/v3/deploymentAssessment/getDeploymentAssessment.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const getDeploymentAssessmentSchema = z.object({
+  params: z.object({
+    deploymentAssessmentId: z.string(),
+  }),
+})
+
+registerPath({
+  method: 'get',
+  path: '/api/v3/deployment-assessment/{deploymentAssessmentId}',
```
GB907762: Should we work towards Rest convention with v3 (plurals on a collection)? Or should we just deem this out of scope for now?

backend/src/routes/v3/deploymentAssessment/getDeploymentAssessment.ts
```diff
@@ -0,0 +1,54 @@
...
+      content: {
+        'application/json': {
+          schema: z.object({
+            deploymentAssessment: deploymentAssessmentInterfaceSchema,
+          }),
+        },
+      },
+    },
+  },
+})
```
GB907762:
```suggestion
  },
  'v3',
)
```

backend/src/routes/v3/deploymentAssessment/postDeploymentAssessment.ts
```diff
@@ -5,60 +5,23 @@ import audit from '../../../connectors/audit/index.js'
...
     })
     .strict(),
 })
 
 registerPath(
   {
     method: 'post',
-    path: '/api/v3/deployment-assessments',
+    path: '/api/v3/deployment-assessment',
     tags: ['deployment assessments'],
```
GB907762, line 25:
```suggestion
    tags: ['deployment assessment'],
```

backend/src/services/deploymentAssessment.ts
```diff
@@ -61,6 +61,15 @@ async function validateModels(modelIds: string[]) {
   }
 }
 
+export async function getDeploymentAssessmentById(deploymentAssessmentId: string) {
+  const deploymentAssessment = await DeploymentAssessmentModel.findOne({ id: deploymentAssessmentId })
```
GB907762, line 67: Have a look at #4090. I've made it so you can't view DAs in draft unless you're the author. 

Thought being, this should encourage authors to submit DAs with the full information since they can only be viewed by themselves and there's no external pressure to submit a draft. 

https://github.com/gchq/Bailo/pull/4090/changes#diff-52b3855698cbda1bd470d70ccc5cb879dbeef643f2016a2d3552722bdf334e54R127

GB907762 (changes requested):

backend/src/routes/v3/deploymentAssessment/getDeploymentAssessment.ts
PE39806: Let's follow REST convention for `v3` - I had made this change to be in line with existing approach but we should really use best practices and `v3` is a good chance to do that.

backend/src/routes/v3/deploymentAssessment/postDeploymentAssessment.ts
PE39806: Switched both files to have this as plural - I'm not sure if this is correct best practice but it matches the plural URL (REST convention).

GB907762: These are just tags to group endpoints on swagger. As long as it's consistent.

backend/src/routes/v3/deploymentAssessment/getDeploymentAssessment.ts
ARADDCC002: I think if that's the direction we want to take (which seems sensible) then now is a good time to start.

backend/src/connectors/authorisation/base.ts
GB907762, line 1: Very very minor. So feel to disregard. 

I think `You do not have permission to <action> this Deployment Assessment` Doesn't give enough context it may prompt users to seek permission to a DA?

Something better might be along the lines:
`You must be the risk owner or the author of this Deployment Assessment to <action> it in draft`

Alternatively throw a 404 without the context?

GB907762 (changes requested):

ARADDCC002 (dismissed):

GB907762 (approved):

ARADDCC002 (approved):

PR #4087, opened 2026-08-21
Enable GitHub workflows on `v3` branch

PE39806:


ARADDCC002 (dismissed):

JRB66955 (dismissed):

GB907762 (dismissed):
Happy with this. But as a consideration to future proof, could use `release/**/*`

Since this is the sort of automation conventional branches lend themselves to https://conventionalbranch.org/#specification

ARADDCC002 (approved):

PR #4088, opened 2026-08-21
Deployment assessment creation notifications

ARADDCC002:


backend/src/services/smtp/smtp.ts
```diff
@@ -92,6 +93,71 @@ export async function getApprovedAccessRequests(modelId: string) {
...
+      { title: 'Deployment assessment name', data: deployment.metadata.overview.name },
+      { title: 'Deployment assessment ID', data: deployment.id },
+      {
+        title: 'Created By',
+        data: creatorName,
+      },
+    ],
+    [
+      { name: 'Open Deployment assessment', url: `${appBaseUrl}/deployments/${encodeURIComponent(deployment.id)}` },
+      { name: 'See Deployment assessments', url: `${appBaseUrl}/deployments` },
```
GB907762: Just be aware this will take you to the `Needs Action` tab this will show all the reviews you all the DAs you need to review (for DROs) and to make changes to (for Authors).

Which looks correct in this case.

backend/src/services/smtp/smtp.ts
```diff
@@ -92,6 +93,71 @@ export async function getApprovedAccessRequests(modelId: string) {
...
+  deployment: DeploymentAssessmentInterface,
+  model: ModelInterface,
+  creatorName: string,
+) {
+  if (!config.smtp.enabled) {
+    log.info('Not sending email due to SMTP disabled')
+    return
+  }
+
+  const modelUrl = `${appBaseUrl}/${encodeURIComponent(model.kind)}` + `/${encodeURIComponent(model.id)}`
```
GB907762: Mirrored models can be attached to DAs as well

backend/src/services/smtp/smtp.ts
```diff
@@ -92,6 +93,71 @@ export async function getApprovedAccessRequests(modelId: string) {
...
+
+  const emailContent = buildEmail(
+    `A model you manage has been included in the deployment assessment ${deployment.metadata.overview.name}`,
```
GB907762:
```suggestion
    `A deployment assessment has been created for a model you own`,
```

backend/src/services/smtp/smtp.ts
```diff
@@ -92,6 +93,71 @@ export async function getApprovedAccessRequests(modelId: string) {
...
+      { title: 'Deployment Assessment ID', data: deployment.id },
+      { title: 'Model Name', data: model.name },
+      { title: 'Model ID', data: model.id },
+      {
+        title: 'Created By',
+        data: creatorName,
+      },
+    ],
+    [
+      { name: 'Open Deployment Assessment', url: `${appBaseUrl}/deployments/${encodeURIComponent(deployment.id)}` },
```
GB907762: I think the language here suggests that you want an owner to create a DA I think `View` is clearer. Although I understand that `Open` is used in the rest of the codebase.

```suggestion
      { name: 'View Deployment Assessment', url: `${appBaseUrl}/deployments/${encodeURIComponent(deployment.id)}` },
```

backend/src/services/smtp/smtp.ts
```diff
@@ -92,6 +93,71 @@ export async function getApprovedAccessRequests(modelId: string) {
...
+      { name: 'See Deployment Assessments', url: `${appBaseUrl}/deployments` },
```
GB907762:
```suggestion
      { name: 'View Deployment Assessments', url: `${appBaseUrl}/deployments?tab=all-assessments?models=${model.id}` },
```
This query will eventually display all DAs on a model, which is important for that model owner.

Also `See` is a passive verb. `View` implies that it's intentional

backend/src/services/deploymentAssessment.ts
```diff
@@ -80,6 +134,10 @@ export async function createDeploymentAssessment(user: UserInterface, params: Cr
   const metadata = params.metadata as DeploymentAssessmentMetadata
   const { name, riskOwner, modelIds } = metadata.overview
 
+  if (!params.draft && !riskOwner) {
+    throw BadReq('Deployment risk owner is required')
+  }
```
GB907762, line 155: This is probably a bad type, that you've tried to validate?
If you look in schema itself it should be clear that risk owners and all core properties are required. This is just repetition on the service layer.

backend/src/services/deploymentAssessment.ts
```diff
@@ -106,5 +164,9 @@ export async function createDeploymentAssessment(user: UserInterface, params: Cr
     throw error
   }
 
+  if (!params.draft && riskOwner) {
+    await notifyDeploymentStakeholders(riskOwner, modelIds ?? [], deploymentAssessment)
```
GB907762, line 190: Again I think this a bad type `modelIds` shouldn't need a null coalescence operator. Since these are defined in the schema as having at least one model and being a required property.

GB907762 (changes requested):

backend/src/services/smtp/smtp.ts
ARADDCC002: I've added `entryKindForRedirect` to the backend in a new router utils file. It's possible this is a duplicate and we already have it in the backend, but I can't seem to find it

backend/src/services/deploymentAssessment.ts
ARADDCC002:
```
export interface DeploymentAssessmentMetadata {
  overview: {
    name: string
    riskOwner?: string
    justification?: string
    modelIds?: string[]
    [key: string]: unknown
  }
  [key: string]: unknown
}
```
They're optional, because aside from `name`, every other field is supported as optional because it can be in draft state.

We could in theory have conditional typing to see if `draft` is false, and if so, these fields are mandatory.

backend/src/services/deploymentAssessment.ts
ARADDCC002: As explained in the other comment, they're optional because of draft state. We'd to have some form of conditional typing, or we could enforce these four properties whilst in draft, but I don't particularly like that.

backend/src/services/deploymentAssessment.ts
GB907762: I think this is probably a better solution as it should apply to all services like `POST`, `PATCH` and `GET`?

GB907762 (dismissed):

backend/src/utils/routerUtils.ts
```diff
@@ -0,0 +1,18 @@
+import { EntryKind, EntryKindKeys } from '../models/Model.js'
+
+export const entryKindForRedirect = (entryKind: EntryKindKeys) => {
+  switch (entryKind) {
+    case EntryKind.Model:
+      return 'model'
+    case EntryKind.MirroredModel:
+      return 'model'
+    case EntryKind.UntrustedModel:
```
PE39806, line 7:
```suggestion
    case EntryKind.Model:
    case EntryKind.MirroredModel:
    case EntryKind.UntrustedModel:
```

backend/src/utils/routerUtils.ts
```diff
@@ -0,0 +1,18 @@
...
+      return 'model'
+    case EntryKind.DataCard:
+      return 'data-card'
+    case EntryKind.MirroredDataCard:
```
PE39806, line 10:
```suggestion
    case EntryKind.DataCard:
    case EntryKind.MirroredDataCard:
```

backend/src/services/smtp/smtp.ts
```diff
@@ -92,6 +94,71 @@ export async function getApprovedAccessRequests(modelId: string) {
...
+      { title: 'Model Name', data: model.name },
+      { title: 'Model ID', data: model.id },
+      {
+        title: 'Created By',
+        data: creatorName,
+      },
+    ],
+    [
+      { name: 'View Deployment Assessment', url: `${appBaseUrl}/deployments/${encodeURIComponent(deployment.id)}` },
+      { name: 'View Deployment Assessments', url: `${appBaseUrl}/deployments?tab=all-assessments?models=${model.id}` },
```
PE39806:
```suggestion
      { name: 'View Deployment Assessments', url: `${appBaseUrl}/deployments?tab=all-assessments&models=${model.id}` },
```

backend/src/services/deploymentAssessment.ts
```diff
@@ -61,6 +63,58 @@ async function validateModels(modelIds: string[]) {
...
 
+async function notifyDeploymentStakeholders(
+  riskOwner: string,
+  modelIds: string[],
+  deploymentAssessment: DeploymentAssessmentInterface,
+): Promise<void> {
+  try {
+    const models = await ModelModel.find({
+      id: { $in: modelIds },
+    })
```
PE39806: I think we only need the POJO here
```suggestion
    const models = await ModelModel.find({
      id: { $in: modelIds },
    }).lean()
```

PE39806 (changes requested):

PE39806 (dismissed):

GB907762 (dismissed):

PE39806 (approved):

GB907762 (approved):

PR #4090, opened 2026-08-21
Add search filter for deployment assessments

GB907762:
Add a search filter for deployment assessments. 

This includes filters on:
- Schemas
- Models
- DROs
- Authors
- When the DA was created
- Drafts
- Partial case insensitive matches on the name

*Drafts are not searchable to non-authors, to encourage accurate documentation of deployments*

### DIFF
--- a/backend/src/services/deploymentAssessment.ts
+++ b/backend/src/services/deploymentAssessment.ts
@@ -1,3 +1,5 @@
+import { escapeRegExp } from 'lodash-es'
+
 import authentication from '../connectors/authentication/index.js'
 import DeploymentAssessmentModel, {
   DeploymentAssessmentInterface,
@@ -12,6 +14,17 @@ import { BadReq, Conflict } from '../utils/error.js'
 import { convertStringToId } from '../utils/id.js'
 import { isMongoServerError } from '../utils/mongo.js'
 import { getSchemaById, validateContentAgainstSchema } from './schema.js'
+
+export interface SearchDeploymentAssessmentsParams {
+  schemaId?: string
+  modelIds?: string[]
+  riskOwner?: string
+  createdBy?: string
+  createdAfter?: string
+  createdBefore?: string
+  draft?: boolean
+  search?: string
+}
 
 export type CreateDeploymentAssessmentParams = Pick<DeploymentAssessmentInterface, 'schemaId' | 'draft'> & {
   metadata: unknown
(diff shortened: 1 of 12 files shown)


backend/src/services/deploymentAssessment.ts
```diff
@@ -108,3 +121,39 @@ export async function createDeploymentAssessment(user: UserInterface, params: Cr
 
   return deploymentAssessment
 }
+
+export async function searchDeploymentAssessments(user: UserInterface, params: SearchDeploymentAssessmentsParams) {
+  const query: Record<string, unknown> = {
+    $and: [{ $or: [{ draft: false }, { draft: true, createdBy: user.dn }] }],
```
PE39806: Should this also include if the current user is the risk owner for drafts?
```suggestion
    $and: [{ $or: [{ draft: false }, { draft: true, createdBy: user.dn }, { draft: true, 'metadata.overview.riskOwner': user.dn }] }],
```

backend/src/services/deploymentAssessment.ts
```diff
@@ -108,3 +121,39 @@ export async function createDeploymentAssessment(user: UserInterface, params: Cr
...
+    query.createdAt = {
+      ...(params.createdAfter && { $gte: new Date(params.createdAfter) }),
+      ...(beforeDate && { $lt: beforeDate }),
+    }
+  }
+  if (params.draft !== undefined) {
+    query.draft = params.draft
+  }
+  if (params.search) {
+    query['metadata.overview.name'] = { $regex: escapeRegExp(params.search), $options: 'i' }
```
PE39806: Maybe also search over the justification? Or split the current `search` param into multiple params if we want to keep this more fine grained.
```suggestion
    query.$and.push({
      $or: [
        { 'metadata.overview.name': regex },
        { 'metadata.overview.justification': regex },
      ],
    })
```

backend/src/services/deploymentAssessment.ts
```diff
@@ -108,3 +121,39 @@ export async function createDeploymentAssessment(user: UserInterface, params: Cr
...
+  if (params.modelIds?.length) {
+    query['metadata.overview.modelIds'] = { $all: params.modelIds }
+  }
+  if (params.riskOwner) {
+    query['metadata.overview.riskOwner'] = params.riskOwner
+  }
+  if (params.createdBy) {
+    query.createdBy = params.createdBy
+  }
+  if (params.createdAfter || params.createdBefore) {
```
PE39806, line 225: Probably should include some `new Date(params.createdAfter) > new Date(params.createdBefore)` check if both params are present.

backend/src/services/deploymentAssessment.ts
```diff
@@ -108,3 +121,39 @@ export async function createDeploymentAssessment(user: UserInterface, params: Cr
 
   return deploymentAssessment
 }
+
+export async function searchDeploymentAssessments(user: UserInterface, params: SearchDeploymentAssessmentsParams) {
+  const query: Record<string, unknown> = {
```
PE39806: Stricter type (`import { ..., QueryFilter } from 'mongoose'`)
```suggestion
  const query: QueryFilter<DeploymentAssessmentInterface> = {
```

PE39806 (changes requested):

backend/src/services/deploymentAssessment.ts
GB907762: This is already covered in the route. I'd like to not repeat this validation if I can help it.

backend/src/services/deploymentAssessment.ts
GB907762: I'll just go through the auth connector for this?

It may be worth having a look at this because it would offload the authorisation form the backend to the database layer which is quicker but I'll use the connector for now.

ARADDCC002 (approved):

IR96334 (approved):

backend/src/services/deploymentAssessment.ts
```diff
@@ -192,3 +206,41 @@ export async function createDeploymentAssessment(user: UserInterface, params: Cr
...
 }
+
+export async function searchDeploymentAssessments(user: UserInterface, params: SearchDeploymentAssessmentsParams) {
+  const query: QueryFilter<DeploymentAssessmentInterface> = {}
+
+  if (params.schemaId) {
+    query.schemaId = params.schemaId
+  }
+  if (params.modelIds?.length) {
+    query['metadata.overview.modelIds'] = { $all: params.modelIds }
```
IR96334, line 217: is the purpose of this filter to find DAs that have at least one of the models or all? If the the prior, maybe use $in?

PE39806 (approved):

PE39806 (approved):

backend/src/services/deploymentAssessment.ts
GB907762: It's to match all modelId's on a DA. This is standard for query parameters on arrays where they use an intersection rather than a union.

PR #4092, opened 2026-08-24
Fix metrics count drift and schema soft-delete miscounting

PE39806:
Usage metrics global counts (`withAccessRequest`, `withReleases`, `entries`, etc.) were computed by summing per-org results over a fixed config list - models with unlisted org values were silently excluded, causing drift against the breakdown endpoint which queries globally.
`SchemaModel` was the only model without the soft-delete plugin, so `calculateSchemaBreakdown`'s `$lookup` to `v2_models` included soft-deleted models in schema counts, also deflating the "None" bucket. The soft-delete plugin's `pre('aggregate')` hook did not filter deleted documents from `$lookup` target collections.

- Compute global metrics independently via instead of summing per-org results
- Add `softDeletionPlugin` to `SchemaModel` with necessary service changes (`replaceOne` + upsert instead of delete-then-recreate)
 - Inject `deleted: { $ne: true }` into `$lookup` soft-delete plugin sub-pipelines (both simple-form and pipeline-form)
- Add `deleteOne` instance method to soft-delete plugin
- Parallelise global, per-org, and user count computation in `getUsageMetrics`
- Add soft-delete plugin test coverage
- Replace some test `.toMatchSnapshot` with `.toEqual` to avoid stale snapshots from any changes to `backend/test/testUtils/setupMongooseModelMocks.ts`


JRB66955 (approved):

Labeeb2339 (commented):
I reviewed the metrics query changes, SchemaModel plugin adoption, and the aggregate hook. Computing the global row independently avoids losing organisations that are absent from the configured list, while `countDocuments({ id: { $in: ids } })` keeps relation counts tied to existing models. The lookup rewrite preserves scalar `localField`/`foreignField` joins used by the repository and injects the soft-delete predicate into both simple and pipeline forms. The schema overwrite/default paths now use upserted replacements rather than a delete gap, and the added plugin tests cover the new behavior. All reported CI checks are green; I found no additional issue in this diff.

ARADDCC002 (approved):

PR #4094, opened 2026-08-25
Add Deployment Assessment update endpoint

PE39806:
As a Deployment Assessment's `riskOwner` or `createdBy` user, I want to update an existing Deployment Assessment so that Deployment Assessment editing can be supported.

- Add a new `PATCH /api/v3/deployment-assessment/{deploymentAssessmentId}` endpoint
- If the DA is a draft then allow saving with missing fields but do not allow saving invalid/extra fields
- If the DA is not a draft (including updating the current state to no longer be a draft) then all required/mandatory fields must be populated
- Only authorised users can update the DA

### DIFF
--- a/backend/src/routes/v3/deploymentAssessment/patchDeploymentAssessment.ts
+++ b/backend/src/routes/v3/deploymentAssessment/patchDeploymentAssessment.ts
@@ -1,0 +1,80 @@
+import { Request, Response } from 'express'
+
+import { AuditInfo } from '../../../connectors/audit/Base.js'
+import audit from '../../../connectors/audit/index.js'
+import { z } from '../../../lib/zod.js'
+import { DeploymentAssessmentInterface } from '../../../models/DeploymentAssessment.js'
+import { updateDeploymentAssessment } from '../../../services/deploymentAssessment.js'
+import {
+  deploymentAssessmentDraftSchema,
+  deploymentAssessmentInterfaceSchema,
+  deploymentAssessmentMetadataSchema,
+  registerPath,
+} from '../../../services/specification.js'
+import { parse } from '../../../utils/validate.js'
+
+export const patchDeploymentAssessmentSchema = z.object({
+  body: z
+    .object({
+      metadata: deploymentAssessmentMetadataSchema.optional(),
+      draft: deploymentAssessmentDraftSchema.optional(),
+    })
+    .refine(
+      (obj) => {
+        for (const val of Object.values(obj)) {
+          if (val !== undefined) {
+            return true
+          }
+        }
+        return false
+      },
+      {
+        message: 'Body must have at least one property defined.',
+      },
+    ),
+  params: z.object({
+    deploymentAssessmentId: z.string(),
+  }),
+})
+
+registerPath(
+  {
+    method: 'patch',
+    path: '/api/v3/deployment-assessments/{deploymentAssessmentId}',
+    tags: ['deployment assessments'],
+    description: 'Get a deployment assessment.',
+    schema: patchDeploymentAssessmentSchema,
+    responses: {
+      200: {
+        description: 'A deployment assessment.',
+        content: {
+          'application/json': {
+            schema: z.object({
+              deploymentAssessment: deploymentAssessmentInterfaceSchema,
+            }),
+          },
+        },
+      },
+    },
+  },
+  'v3',
+)
+
+interface GetDeploymentAssessmentResponse {
+  deploymentAssessment: DeploymentAssessmentInterface
+}
+
+export const patchDeploymentAssessment = [
+  async (req: Request, res: Response<GetDeploymentAssessmentResponse>): Promise<void> => {
+    req.audit = AuditInfo.UpdateDeploymentAssessment
+    const { body, params } = parse(req, patchDeploymentAssessmentSchema)
+
+    const deploymentAssessment = await updateDeploymentAssessment(req.user, params.deploymentAssessmentId, body)
+
+    await audit.onUpdateDeploymentAssessment(req, deploymentAssessment)
+
+    res.json({
+      deploymentAssessment,
+    })
+  },
+]
--- a/backend/src/routes/v3/deploymentAssessment/patchDeploymentAssessment.ts
+++ b/backend/src/routes/v3/deploymentAssessment/patchDeploymentAssessment.ts
@@ -42,7 +42,7 @@ registerPath(
     method: 'patch',
     path: '/api/v3/deployment-assessments/{deploymentAssessmentId}',
     tags: ['deployment assessments'],
-    description: 'Get a deployment assessment.',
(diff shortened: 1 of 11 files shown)


backend/src/routes/v3/deploymentAssessment/patchDeploymentAssessment.ts
```diff
@@ -0,0 +1,80 @@
...
+          if (val !== undefined) {
+            return true
+          }
+        }
+        return false
+      },
+      {
+        message: 'Body must have at least one property defined.',
+      },
+    ),
```
GB907762, line 36: Surely this is covered by the service layer? 

I know in existing endpoints we don't have very good route validation for this, but is it worth having no additional properties. In the schema validation this is already covered. But you can freely add properties that have no effect.

backend/src/services/deploymentAssessment.ts
```diff
@@ -79,6 +80,83 @@ async function validateModels(modelIds: string[]) {
...
+  }
+
+  // The schema check above cannot narrow the type, so guard the properties this service relies on.
+  if (!isDeploymentAssessmentMetadata(metadata)) {
+    throw BadReq('Deployment assessment metadata has an invalid overview.', {
+      schemaId,
+      property: findInvalidMetadataProperty(metadata),
+    })
+  }
+
```
GB907762: I don't think this is the correct way to type guard. I'd rather use proper `DeploymentAssessmentInterface['metadata']` instead of an unknown type. 

And then create a conditional type that switches on the draft property? Otherwise this might be used on any writable action and cause unnecessary validation.

backend/src/services/deploymentAssessment.ts
```diff
@@ -200,10 +251,49 @@ export async function createDeploymentAssessment(user: UserInterface, params: Cr
...
+  const auth = await authorisation.deploymentAssessment(user, deploymentAssessment, DeploymentAssessmentAction.Update)
+  if (!auth.success) {
+    throw Forbidden(auth.info, { userDn: user.dn, deploymentAssessmentId })
+  }
+
+  const metadata = await validateDeploymentAssessment(
+    deploymentAssessment.schemaId,
+    diff.metadata ?? deploymentAssessment.metadata,
+    diff.draft ?? deploymentAssessment.draft,
```
GB907762, line 236: For publishing `draft->non-draft`. If you an existing schema saved, but you only patch part of it, this evaluates the request body and not the schema that will be evaluated in the database.

It means say I want to only patch a model because I have none in draft. I have to patch the whole schema.

backend/src/services/deploymentAssessment.ts
```diff
@@ -200,10 +251,49 @@ export async function createDeploymentAssessment(user: UserInterface, params: Cr
...
+  )
+
+  if (diff.metadata !== undefined) {
+    deploymentAssessment.metadata = metadata
+    deploymentAssessment.markModified('metadata')
+  }
+  if (diff.draft !== undefined) {
+    if (!deploymentAssessment.draft && diff.draft) {
+      throw BadReq('Cannot convert a released deployment assessment back to a draft.')
```
GB907762:
```suggestion
      throw BadReq('Cannot convert a submitted deployment assessment back to a draft.')
```
Might be a better characterisation.

backend/src/services/deploymentAssessment.ts
```diff
@@ -200,10 +251,49 @@ export async function createDeploymentAssessment(user: UserInterface, params: Cr
...
+    }
+    deploymentAssessment.draft = diff.draft
```
GB907762, line 247: Not something for this PR, but is it worth thinking about when you patch the metadata (make a change to the form) this has an effect on the review status on #4102

GB907762 (changes requested):

backend/src/routes/v3/deploymentAssessment/patchDeploymentAssessment.ts
PE39806: The service layer will happily accept an empty `diff` `{}` which will then essentially just perform the same as the GET endpoint but with extra auth checks. This `refine` prevents an empty `body` being passed in which there is no sense in processing.

backend/src/services/deploymentAssessment.ts
PE39806: `metadata` is all-or-nothing - this is the same as the existing `backend/src/routes/v2/model/accessRequest/patchAccessRequest.ts` approach. If the query only sets part of the `metadata` then that is treated as the entire new metadata and changes are not merged in any way. If the query does not specify `metadata` in the `body` then it _will_ pull from the DB.

Other DA properties (`schemaId`, `id`, `createdBy` etc.) are not in the `metadata` and are only initialised when creating the DA.

backend/src/services/deploymentAssessment.ts
```diff
@@ -79,6 +78,40 @@ async function validateModels(modelIds: string[]) {
...
 }
 
+async function validateDeploymentAssessment(
+  schemaId: DeploymentAssessmentInterface['schemaId'],
+  metadata: DeploymentAssessmentInterface['metadata'],
+  draft: DeploymentAssessmentInterface['draft'],
+): Promise<DeploymentAssessmentMetadata> {
+  const schema = await getSchemaById(schemaId)
+  if (schema.hidden) {
+    throw BadReq('Cannot create a deployment assessment using a hidden schema.', { schemaId })
```
JRB66955: Should this error message be more generic as this function is used by both `updateDeploymentAssessment` and `createDeploymentAssessment`? Or can we assume that this can branch can never be hit when `updateDeploymentAssessment` is run?

backend/src/services/deploymentAssessment.ts
GB907762: I think the standard is to support nested updates in a patch. But I'm not fussed either way.

backend/src/services/deploymentAssessment.ts
PE39806: Good point. I think there's a wider consideration around how to handle schemas changing have `hidden: true` - should existing DAs still be valid? My gut tells me the easiest approach is to only do this `.hidden` check on DA creation and allow existing DAs to continue to be edited even if the schema changes to have `hidden: true`.
Does that sound sensible?

backend/src/services/deploymentAssessment.ts
PE39806: Let's save proper https://www.rfc-editor.org/info/rfc7396/ compliance for a future piece of work.

backend/src/services/deploymentAssessment.ts
JRB66955: Sounds sensible to me :)

JRB66955 (approved):

GB907762 (approved):

ARADDCC002 (approved):

PR #4102, opened 2026-08-25
Add review and comment endpoints to deployment assessment review

GB907762:
### Changes
- Add `POST /api/v3/deployment-assessments/{deploymentAssessmentId}/review` for DRO decision. 
- Add `Add POST /api/v3/deployment-assessments/{deploymentAssessmentId}/comments` to post general comments about a DA.
- Update `GET /api/v3/deployment-assessments` with a state query param, to filter by review stage. 
- Add reviews and review comments to `GET /api/v3/deployment-assessments/{deploymentAssessmentId}` to the response schema.
- Add auditing
- Add test coverage
- Use role display name for DRO to configure what this should show as on the UI.

### Things to test
- Drafts are not reviewable but can have comments.
- Check that everyone can make comments unless the DA is in draft. In which case only DROs and Authors.
- Only risk owners can submit a review.
- Check states in the  `GET /api/v3/deployment-assessments`  endpoint. It should match the most recent (rejection, approval, needs changes).


backend/src/services/deploymentAssessment.ts
```diff
@@ -145,6 +164,115 @@ export async function getDeploymentAssessmentById(user: UserInterface, deploymen
...
+
+  const responses = [...comments, ...reviews.flatMap(({ responses: reviewResponses }) => reviewResponses)].sort(
+    (first, second) => new Date(first.createdAt).getTime() - new Date(second.createdAt).getTime(),
+  )
+  const latestDecision = responses.findLast(({ kind }) => kind === ResponseKind.Review)?.decision
+
+  return {
+    deploymentAssessment,
+    responses,
+    state: deriveDeploymentAssessmentState(deploymentAssessment, latestDecision),
```
PE39806, line 228: `state` can be set as `undefined` through this. Possibly prefer conditional spread e.g. `...(state && { state })`.

backend/src/services/deploymentAssessment.ts
```diff
@@ -242,5 +381,50 @@ export async function searchDeploymentAssessments(user: UserInterface, params: S
...
+      ])
+    : []
+  const decisionsByAssessmentId = new Map(latestDecisions.map(({ _id, decision }) => [_id, decision]))
+
+  const searchResults: DeploymentAssessmentSearchResult[] = authorisedAssessments.map((assessment) => {
+    if (assessment.draft !== false) {
+      return assessment
+    }
+
+    const state = deriveDeploymentAssessmentState(assessment, decisionsByAssessmentId.get(assessment.id))
```
PE39806, line 489: As above with `undefined`.

backend/src/services/smtp/smtp.ts
```diff
@@ -345,7 +345,7 @@ async function notifyRole(review: ReviewInterface, title: string, fields: Info[]
     { name: 'See Reviews', url: `${appBaseUrl}/review?category=access` },
   ])
 
-  await dispatchEmailToModelRole(review.modelId, review.role, emailContent)
+  await dispatchEmailToModelRole(review.modelId!, review.role, emailContent)
```
PE39806: I'm not keen on this and would, instead, prefer some type narrowing like
```
if (review.kind === ReviewKind.DeploymentAssessment) {
  throw BadReq('Deployment assessment reviews are handled separately.', { reviewId })
}
```
maybe in a reusable helper function given this is present in several places.

backend/src/services/deploymentAssessment.ts
```diff
@@ -242,5 +381,50 @@ export async function searchDeploymentAssessments(user: UserInterface, params: S
...
+        {
+          $lookup: {
+            from: 'v2_responses',
+            let: { reviewId: '$_id' },
+            pipeline: [
+              {
+                $match: {
+                  $expr: { $eq: ['$parentId', '$$reviewId'] },
+                  kind: ResponseKind.Review,
+                  deleted: { $ne: true },
```
PE39806: I think this isn't needed due to `schema.pre('aggregate', ...` in `backend/src/models/plugins/softDeletePlugin.ts`

backend/src/services/deploymentAssessment.ts
```diff
@@ -145,6 +164,115 @@ export async function getDeploymentAssessmentById(user: UserInterface, deploymen
...
+  const deploymentAssessment = await getDeploymentAssessmentById(user, deploymentAssessmentId)
+  const [comments, reviews] = await Promise.all([
+    ResponseModel.find({ parentId: deploymentAssessment._id, kind: ResponseKind.Comment }),
+    ReviewModel.aggregate<{ responses: ResponseInterface[] }>([
+      { $match: { deploymentAssessmentId, kind: ReviewKind.DeploymentAssessment, deleted: { $ne: true } } },
+      {
+        $lookup: {
+          from: 'v2_responses',
+          let: { reviewId: '$_id' },
+          pipeline: [{ $match: { $expr: { $eq: ['$parentId', '$$reviewId'] }, deleted: { $ne: true } } }],
```
PE39806: Same again for `deleted`.

backend/src/services/deploymentAssessment.ts
```diff
@@ -145,6 +164,115 @@ export async function getDeploymentAssessmentById(user: UserInterface, deploymen
...
+  user: UserInterface,
+  deploymentAssessmentId: string,
+): Promise<DeploymentAssessmentDetails> {
+  const deploymentAssessment = await getDeploymentAssessmentById(user, deploymentAssessmentId)
+  const [comments, reviews] = await Promise.all([
+    ResponseModel.find({ parentId: deploymentAssessment._id, kind: ResponseKind.Comment }),
+    ReviewModel.aggregate<{ responses: ResponseInterface[] }>([
+      { $match: { deploymentAssessmentId, kind: ReviewKind.DeploymentAssessment, deleted: { $ne: true } } },
+      {
+        $lookup: {
```
PE39806: Is this `$lookup` similar enough to `searchDeploymentAssessments` to make a reusable helper?

backend/src/services/deploymentAssessment.ts
```diff
@@ -29,12 +35,25 @@ export interface SearchDeploymentAssessmentsParams {
...
   metadata: unknown
 }
 
+export interface DeploymentAssessmentDetails {
+  deploymentAssessment: DeploymentAssessmentDoc
+  responses: ResponseInterface[]
+  state?: DeploymentAssessmentStateKeys
+}
+
+const deploymentRiskOwnerRole = 'dro'
```
PE39806: We define `SystemRoles` in `backend/src/models/Model.ts` rather than the service layer. Maybe this should do the same?

PE39806 (changes requested):

backend/src/services/deploymentAssessment.ts
GB907762: Good spot!

backend/src/services/deploymentAssessment.ts
GB907762: Is this a system role or a review role?

Also this is not going to be tied to a model unlike ARs and Releases. I'll make it a review role for now.

backend/src/services/deploymentAssessment.ts
GB907762: This is the top level route service which uses `res.json` from express https://expressjs.com/en/5x/api/response/#resjson

This method uses `JSON.stringify` which removes any undefined property from the object.

PE39806 (dismissed):

backend/src/routes/v3/deploymentAssessment/getDeploymentAssessment.ts
```diff
@@ -37,21 +43,15 @@ registerPath(
   'v3',
 )
 
-interface GetDeploymentAssessmentResponse {
-  deploymentAssessment: DeploymentAssessmentInterface
-}
-
 export const getDeploymentAssessment = [
-  async (req: Request, res: Response<GetDeploymentAssessmentResponse>): Promise<void> => {
+  async (req: Request, res: Response): Promise<void> => {
```
JRB66955, line 47: Should a new type to replace `GetDeploymentAssessmentResponse` here be created instead of just simply deleting it?

GB907762: I've removed this to keep the typing consistent (This is what it is for every other endpoint). In the future we may choose to move to this approach?

JRB66955: Didn't realise that this is consistent - makes sense, we can decide if we want to start typing this for other endpoints in the future,

PE39806 (dismissed):

backend/config/default.cjs
```diff
@@ -118,6 +118,13 @@ module.exports = {
       description: 'Reviewer',
       systemRole: 'owner',
     },
+    {
```
JR40159: If this has been done to enable customising the display name then I'd suggest adding DRO to the existing display name config is far more light weight than adding a dynamic/review role. Also creating a reviewing role would need to make the DRO core DA schema question dynamic and also ensure that a DRO cannot be added to a model

JR40159 (changes requested):

PE39806 (dismissed):

backend/src/services/deploymentAssessment.ts
```diff
@@ -172,6 +191,128 @@ export async function getDeploymentAssessmentById(user: UserInterface, deploymen
...
+          },
+        },
+        { $sort: { createdAt: -1 } },
+      ],
+      as: 'responses',
+    },
+  }
+}
+
+function deriveDeploymentAssessmentState(
```
JR40159, line 197: I think this makes sense, returning the review state with the item that's being reviewed. However, I think this should more more generic so it could be reused in future. As such, I think this function should be called `deriveReviewState` and reside in the review service. The database call this functionality uses also supports this as it doesn't look like it belongs in this service.

backend/src/services/review.ts
```diff
@@ -368,15 +368,15 @@ export async function findReviewRole(user: UserInterface, shortName: string) {
 }
 
 export async function findReviewRoles(schemaIds?: string[]): Promise<ReviewRoleDoc[]> {
-  const mongoQuery: QueryFilter<ReviewRoleInterface> = {}
+  const mongoQuery: QueryFilter<ReviewRoleInterface> = { shortName: { $ne: 'dro' } }
```
JR40159, line 404: This isn't using review roles any more?

backend/src/services/deploymentAssessment.ts
```diff
@@ -322,5 +474,34 @@ export async function searchDeploymentAssessments(user: UserInterface, params: S
   const deploymentAssessments = await DeploymentAssessmentModel.find(query).sort({ draft: -1, updatedAt: -1 })
   const auths = await authorisation.deploymentAssessments(user, deploymentAssessments, DeploymentAssessmentAction.View)
 
-  return deploymentAssessments.filter((_, i) => auths[i].success)
+  const authorisedAssessments = deploymentAssessments.filter((_, i) => auths[i].success)
+  const assessmentIds = authorisedAssessments.filter(({ draft }) => draft === false).map(({ id }) => id)
+  const latestDecisions = assessmentIds.length
+    ? await ReviewModel.aggregate<{ _id: string; decision?: DecisionKeys }>([
```
JR40159, line 453: Does executing this mongo query belong in the deploymentAssessment service?

JR40159 (changes requested):

backend/src/services/deploymentAssessment.ts
GB907762: For now I think it is better to leave it like this. Since this could have undefined behaviour with the other review types. If you see in the grouping this groups via the deployment assessment id which is not a property of other review types.

JRB66955 (dismissed):

backend/src/services/deploymentAssessment.ts
JR40159: The function in the review service doesn't need to be generic, eg `getLatestDecisionsForDeploymentAssessments`. As it's purpose is getting review data I think it would make more sense in the review service and simplify this search DA function. In the review service we already have function specific to things like releases and access requests.

JR40159 (changes requested):

backend/src/services/deploymentAssessment.ts
GB907762: I think we should separate it out as it's not useful to the semantics of the business logic of DAs.

backend/src/services/v3/response.ts
```diff
@@ -20,8 +20,10 @@ export async function respondToReview(
   dueDate?: Date,
 ) {
   const review = getModelReview(await findReviewById(user, reviewId))
-  if (response.decision === Decision.Approve) {
-    validateLifecycleReview(review, dueDate)
+  if (review.kind === ReviewKind.Lifecycle && response.decision === Decision.Approve && dueDate) {
+    if (dueDate.getTime() <= Date.now() || !isLifecycleReviewDateValid(dueDate)) {
+      throw BadReq('Due date of next review is invalid.')
```
ARADDCC002: Minor, but might want to add `dueDate` to the error.

ARADDCC002 (dismissed):

JR40159 (approved):

ON93754 (approved):

PR #4103, opened 2026-08-25
added da page added permissions for das

ARADDCC002:

PR #4106, opened 2026-08-26
Delete Deployment Assessment

PE39806:
As a Deployment Assessment `riskOwner` or `createdBy` user, I want to delete a DA so that I can remove DAs that are no longer required.

-  New `DELETE /api/v2/deployment-assessment/{deploymentAssessmentId}` backend endpoint
    - Only `riskOwner` or `createdBy` user can trigger, others get permission denied
    - Delete the DA plus any associated reviews & responses
    - Update audit connectors
- Implement frontend deletion
    - New `DeletionConfirmationDialogue` component (matches existing entry deletion functionality)
    - Replace DA deletion modal with `DeletionConfirmationDialogue`
    - Call new backend endpoint

Depends on #4103 and #4102

Also fixes:
- `/deployments` -> `/deployment-assessments` for some paths
- `metadata.overview.name` -> `name` for DAs


frontend/src/common/DeletionConfirmationDialogue.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { ReactNode, useEffect, useState } from 'react'
+import { Transition } from 'src/common/Transition'
+import useNotification from 'src/hooks/useNotification'
+import MessageAlert from 'src/MessageAlert'
+import { getErrorMessage } from 'utils/fetcher'
+
+type DeletionConfirmationDialogueProps = {
+  open: boolean
+  title: string
+  /** Closes the dialogue. Called when the user cancels and after a successful deletion. */
```
ARADDCC002: This isn't needed

frontend/src/common/DeletionConfirmationDialogue.tsx
```diff
@@ -0,0 +1,146 @@
...
+  onClose: () => void
+  /** Performs the delete request. The dialogue owns loading, error, notification and redirect. */
```
ARADDCC002: This isn't needed

frontend/src/common/DeletionConfirmationDialogue.tsx
```diff
@@ -0,0 +1,146 @@
...
+  onDelete: () => Promise<Response>
+  /** When provided, the user must type this exact string before the delete button is enabled. */
+  confirmationText?: string
+  dialogMessage?: string
+  /** Success snackbar text. Omit to send no notification. */
```
ARADDCC002: This isn't needed

ARADDCC002 (dismissed):

sylvesterkaczmarek (commented):
There is a transaction/authorisation race worth tightening here. `removeDeploymentAssessment(..., session)` is called inside `useTransaction`, but `getDeploymentAssessmentById()` performs its `findOne` without that session. Both the View check and the subsequent Delete check therefore run against a document read outside the transaction snapshot.

Because `metadata.overview.riskOwner` is mutable and is part of `isNamedUser` for delete permission, a concurrent update can revoke the caller between that read and the transactional delete, while this transaction still commits using the stale authorised document. Could the lookup be made session-aware and the authorisation decision be performed on the document read inside the same transaction? A session/race regression test would be useful as well.

backend/src/services/deploymentAssessment.ts
```diff
@@ -409,6 +436,14 @@ export async function updateDeploymentAssessment(
 
   await deploymentAssessment.save()
 
+  if (isBeingSubmitted && metadata.overview && metadata.overview.riskOwner) {
+    await notifyDeploymentStakeholders(
+      metadata.overview.riskOwner,
```
GB907762: We're using a list of owners now. I don't think email notifications were updated as part of this change. #4116. Might be something to descope from this PR and tackle in a new one.

backend/src/routes/v2/model/deleteModel.ts
GB907762, line 1: What happens to models when you delete them? I'd assume they have to be removed from the DA?

GB907762 (changes requested):

backend/src/routes/v2/model/deleteModel.ts
PE39806: Out of scope for this PR, but I'll add a TODO for a future PR to pick up.

ARADDCC002 (approved):

GB907762 (approved):

PR #4108, opened 2026-08-27
Display schema id in schemas page

PE39806:
As a Bailo admin, I want to be able to differentiate schemas by their ID as it is possible to give different schemas identical names & descriptions, so that I can tell otherwise identical schemas apart.

Also invalidates the stale schema cache in the backend to ensure updated values are displayed correctly.


frontend/src/common/ValueDisplay.tsx
```diff
@@ -0,0 +1,17 @@
...
+import MarkdownDisplay from 'src/common/MarkdownDisplay'
+
+interface ValueDisplayProps {
+  value?: string
+  richText?: boolean
+  emptyText?: string
+}
+
+/** Renders a value as plain text or markdown, falling back to an italic placeholder when unset. */
+export default function ValueDisplay({ value, richText = false, emptyText = 'Empty' }: ValueDisplayProps) {
```
ARADDCC002: What's the rationale behind having this as a separate component rather than rendering this inside `<LabelledValue />`? Is it to make it more generic so that the label can accept any value display?

I'm not sure I like that if that _is_ the case as it could makes things look inconsistent across the app, whereas the point of this change is to improve consistency.

ARADDCC002 (approved):

JRB66955 (approved):

PR #4116, opened 2026-08-28
Change riskOwner from single string to array of strings

JR40159:


backend/src/services/deploymentAssessment.ts
```diff
@@ -38,16 +38,18 @@ export interface SearchDeploymentAssessmentsParams {
...
-async function validateRiskOwner(riskOwner: string) {
-  const { kind, value } = fromEntity(riskOwner)
-  if (kind !== 'user' || !value) {
-    throw BadReq('The risk owner must be a valid user entity.', { riskOwner })
-  }
+async function validateRiskOwner(riskOwners: string[]) {
+  for (const riskOwner of riskOwners) {
+    const { kind, value } = fromEntity(riskOwner)
+    if (kind !== 'user' || !value) {
+      throw BadReq('The risk owner must be a valid user entity.', { riskOwner })
```
GB907762, line 45: Maybe not something for now. But you might want to reflect this in the entity selector on the frontend so that groups cannot be added?

GB907762 (approved):

ARADDCC002 (approved):

backend/src/services/deploymentAssessment.ts
ARADDCC002: Because this was existing behaviour I think it's fine to do in another PR

JRB66955 (approved):

PR #4117, opened 2026-08-28
Dev/remove schema and test duplication

JR40159:


GB907762 (approved):

ARADDCC002 (approved):

backend/src/services/specification.ts
```diff
@@ -121,6 +121,7 @@ const baseDeploymentAssessmentSchema = z.object({
   schemaId: deploymentAssessmentSchemaIdSchema,
 })
 
+//Here
```
JRB66955, line 124: This comment needed?

sylvesterkaczmarek (commented):
This refactor weakens the `creates an assessment with a generated ID and authenticated creator` regression. `params` now retains `id` and `createdBy` from `testDeploymentAssessment`, and those values exactly match the mocked generated ID and authenticated user, so an implementation that accidentally passed caller-supplied values through would still satisfy `toHaveBeenCalledWith(params)`. Please omit the server-owned fields from `params` and keep the explicit expectation for the generated `id` and authenticated `createdBy`.

PR #4118, opened 2026-08-28
Add model auth and reuse function for create DA

JR40159:


GB907762 (changes requested):
This is great! Could we have a test?

JRB66955 (approved):

JRB66955 (changes requested):
I love when GitHub doesn't show me merge conflicts and doesn't allow me to dismiss my review so this is a workaround

sylvesterkaczmarek (approved):
The model View authorisation is now in the shared deployment-assessment validation path, so both create and update reject models the caller cannot view, and the added tests cover both paths. The refactor also removes the duplicated validation logic cleanly. Approval is for the current patch; the branch still needs its merge conflict resolved.

ARADDCC002 (approved):

GB907762 (approved):

JRB66955 (approved):

PR #4176, opened 2026-09-02
Feature/bai 2692 edit da

IR96334:
Edit DA and Publish Draft DA


sylvesterkaczmarek (commented):
`originalSplitSchema` is not actually a snapshot: `handleEdit()` stores the same step objects, while `setStepState()` shallow-copies only the array and mutates `steps[index].state`. Answer-only edits can therefore mutate both old/new values and make `answersChanged` false, silently skipping the PATCH. Deep-clone the edit snapshot and add an answer-only save test.

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
```diff
@@ -30,6 +34,19 @@ export default function DeploymentAssessment() {
     return error
   }
 
+  async function handlePublish() {
+    if (deploymentAssessment) {
+      setIsLoading(true)
+      const response = await patchDeploymentAssessment(deploymentAssessment.id, undefined, false)
+      if (!response.ok) {
+        setPatchErrorMessage(await getErrorMessage(response))
```
Comment (PE39806) [inline frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:50]: If this is `Bad Request: Deployment assessment metadata could not be validated against the schema.` then the errors should ideally be highlighted directly in the RJSF as otherwise this isn't clear.

frontend/src/deployments/EditableDeploymentAssessmentForm.tsx
```diff
@@ -103,6 +136,77 @@ export default function EditableDeploymentAssessmentForm({
...
+                  value={newName}
+                  onChange={(event) => setNewName(event.target.value)}
+                  size='small'
+                />
+              ) : (
+                <>
+                  <Typography>{deploymentAssessment ? deploymentAssessment.name : 'Loading...'}</Typography>
+                  <CopyToClipboardButton
+                    textToCopy={deploymentAssessment.name}
+                    notificationText='Copied deployment assessment ID to clipboard'
```
PE39806: Copies the name but says it copies the ID.

frontend/src/deployment-assessments/EditableDeploymentAssessmentForm.tsx
```diff
@@ -62,16 +78,32 @@ export default function EditableDeploymentAssessmentForm({
...
 
-        if (!isValid) {
-          setIsLoading(false)
-          return
-        }
+      const answersChanged = getChangedFields(oldData, data).length > 0
+      const nameChanged = deploymentAssessment.name !== newName
+      const nothingChanged = !answersChanged && !nameChanged
+
+      if (nothingChanged) {
```
PE39806, line 89: This isn't being updated correctly so no changes are saved.

frontend/src/deployments/EditableDeploymentAssessmentForm.tsx
```diff
@@ -103,6 +136,77 @@ export default function EditableDeploymentAssessmentForm({
     setUnsavedChanges(isEdit)
   }, [isEdit, setUnsavedChanges])
 
+  const formHeading = useMemo(
+    () => (
+      <>
+        {schema && (
+          <Stack>
+            <Typography
```
PE39806: Could these use `LabelledValue` and/or `LabelledInput`?

PE39806 (changes requested):

frontend/src/deployment-assessments/EditableDeploymentAssessmentForm.tsx
```diff
@@ -103,6 +147,65 @@ export default function EditableDeploymentAssessmentForm({
...
+                  open={schemaInformationOpen}
+                  schema={schema}
+                  onClose={() => setSchemaInformationOpen(false)}
+                />
+              </Stack>
+            </LabelledValue>
+            <LabelledValue label='Created by'>
+              <UserDisplay dn={deploymentAssessment.createdBy} />
+            </LabelledValue>
+            <LabelledInput label='Name' fullWidth required={isEdit}>
```
PE39806, line 160: Super long names e.g. `SupercalifragilisticexpialidociousLlanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogochSupercalifragilisticexpialidociousLlanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch` cause horizontal overflow issues.

PE39806, line 160: I think that the name should be above the Schema and Created by fields as it is more important. I also think it should be in larger text as it is a title, unlike the other fields.

frontend/src/deployment-assessments/EditableDeploymentAssessmentForm.tsx
```diff
@@ -103,6 +147,65 @@ export default function EditableDeploymentAssessmentForm({
...
+              <Stack
+                direction='row'
+                sx={{
+                  alignItems: 'left',
+                }}
+              >
+                {isEdit ? (
+                  <TextField
```
PE39806, line 168: It would be nice if this TextField were full width.

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
Comment (PE39806) [inline frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:50]: This error message should be cleared if changes are saved to the DA in the interim (maybe in `onIsEditChange` call)

PE39806 (dismissed):

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
```diff
@@ -40,43 +57,15 @@ export default function DeploymentAssessment() {
...
-                      size='small'
-                      onClick={() => {}}
-                      data-test='publishButton'
-                    >
-                      Publish
-                    </Button>
-                  </Stack>
-                </Paper>
-              )}
+              <DraftBanner
```
Comment (ARADDCC002) [inline frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:69]: We need to add/change a proper to change the wording of the dialog, it currently says `Are you sure you want to publish this release` which is incorrect.

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
```diff
@@ -30,6 +34,19 @@ export default function DeploymentAssessment() {
     return error
   }
 
+  async function handlePublish() {
```
Comment (ARADDCC002) [inline frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:45]: When it fails the validation it displays a standard generic schema validation error message. 

It might be helpful if we do UI validation on the form before it actually publishes, so it shows which fields need to be filled in.

ARADDCC002 (dismissed):

ARADDCC002:
Are we sure we want to enforce DROs being added even for draft assessments? I was under the impression that we only needed the assessment name.

sylvesterkaczmarek (commented):
Another edit-state edge case: cancelling calls `resetForm()` for the schema answers but never restores `newName`. If I rename an assessment, cancel, then enter edit mode again, the field still contains the cancelled name and a later save can persist it unintentionally. Could `handleCancel()` reset `newName` to `deploymentAssessment.name` too?

sylvesterkaczmarek (commented):
One build issue I noticed: the new form imports `getChangedFields` and `KeyedMutator` via literal `node_modules/...` specifiers. Normal Node/Next resolution treats that as a package/path named `node_modules`, not the installed package entry point, so these imports can fail outside a very specific filesystem layout. Could these come from `@rjsf/utils` and `swr` directly?

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
Comment (IR96334) [reply frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:45]: Descoped this to new piece of work.

IR96334:
> Are we sure we want to enforce DROs being added even for draft assessments? I was under the impression that we only needed the assessment name.

This was bug with RJSF where if array field has minItems of 1 it saturates the array with null.

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
Comment (JR40159) [inline frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:69]: The title of the dialog box when you press publish is "Deletion Confirmation"

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
```diff
@@ -40,43 +57,15 @@ export default function DeploymentAssessment() {
...
-                      onClick={() => {}}
-                      data-test='publishButton'
-                    >
-                      Publish
-                    </Button>
-                  </Stack>
-                </Paper>
-              )}
+              <DraftBanner
+                errorMessage={patchErrorMessage}
```
Comment (JR40159) [inline frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:70]: This error message seems to persist after the dialog is closed

JR40159 (changes requested):

backend/src/services/deploymentAssessment.ts
```diff
@@ -316,10 +316,10 @@ export async function createDeploymentAssessment(
       throw BadReq('Deployment assessment metadata could not be validated against the schema.', { errors })
     }
 
-    if (metadata.overview.riskOwner && metadata.overview.riskOwner.length > 0) {
+    if (metadata.overview && metadata.overview.riskOwner && metadata.overview.riskOwner.length > 0) {
```
JR40159:
```suggestion
    if (metadata.overview?.riskOwner && metadata.overview.riskOwner.length > 0) {
```

backend/src/services/deploymentAssessment.ts
```diff
@@ -316,10 +316,10 @@ export async function createDeploymentAssessment(
...
       await validateRiskOwner(metadata.overview.riskOwner)
     }
-    if (metadata.overview.modelIds && metadata.overview.modelIds.length > 0) {
+    if (metadata.overview && metadata.overview.modelIds && metadata.overview.modelIds.length > 0) {
```
JR40159:
```suggestion
    if (metadata?.overview.modelIds && metadata.overview.modelIds.length > 0) {
```

backend/src/services/schema.ts
```diff
@@ -396,6 +396,7 @@ export async function validateContentAgainstSchema(
 ) {
   const schema = await getSchemaById(schemaId, options.modelState)
   const jsonSchema = options.draft ? makeSchemaOptional(schema.jsonSchema) : schema.jsonSchema
+
```
JR40159: ?

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
Comment (JR40159) [reply frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:45]: Or the frontend could parse the error response it gets from the backend to find which fields have an issue. But agree this should be out of scope

JR40159 (changes requested):

JR40159 (commented):
Nice is a nice to have but it could be worth using https://github.com/gchq/Bailo/blob/main/frontend/src/hooks/useNotification.tsx to send a display a success notification when a draft has been published successfully

sylvesterkaczmarek (approved):
Rechecked the current diff against my two earlier comments. `handleCancel()` now restores `newName` from `deploymentAssessment.name`, and the form imports `getChangedFields` and `KeyedMutator` from `@rjsf/utils` and `swr` rather than literal `node_modules/...` paths. Those concerns are resolved; no blocker from me.

JR40159 (approved):

GB907762 (approved):

PR #4178, opened 2026-09-02
Add all deployment assessments page

GB907762:


sylvesterkaczmarek (commented):
`EntityAutocomplete` only surfaces the special 413 error case. This refactor removes `EntitySelector`'s previous handling for every other directory error, so a 401/500/network failure now looks like an empty autocomplete and can block selection with no explanation. Surface non-413 `isUsersError` values and add a regression for a failed entity lookup.

frontend/src/deployment-assessments/DeploymentAssessmentsList.tsx
PE39806, line 1: I think we should rename `‎frontend/src/deployments/` -> `‎frontend/src/deployment-assessments/` for consistency with `frontend/pages`.

frontend/src/deployment-assessments/DeploymentAssessmentSummaryCard.tsx
PE39806, line 1: As above

frontend/src/deployment-assessments/DeploymentAssessmentSummaryCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+            >
+              <Typography component='h2' variant='h6' color='primary' noWrap>
+                {assessment.name}
+              </Typography>
+            </Link>
+            <CopyToClipboardButton
+              textToCopy={assessment.id}
+              notificationText='Copied deployment assessment ID to clipboard'
+              ariaLabel='copy deployment assessment ID to clipboard'
+            />
```
PE39806, line 130: It doesn't make sense to me that this copies the ID when it is beside the name. I know this is the case in the current DA page but #4176 changes this to copy the name.

frontend/src/deployment-assessments/DeploymentAssessmentsList.tsx
```diff
@@ -0,0 +1,211 @@
...
+              selectedModelIds={filters.modelIds}
+              onChange={(modelIds) => setFilters({ modelIds })}
+            />
+          </Stack>
+          <Box sx={{ width: { xs: '100%', sm: 280 } }}>
+            <EntityAutocompleteFilter
+              label='Risk owner'
+              value={filters.riskOwner}
+              onChange={(riskOwner) => setFilters({ riskOwner })}
+            />
```
PE39806, line 123: This works, but it would be nice if it could be restricted to exclude groups as groups cannot be risk owners.

frontend/src/deployments/DeploymentAssessmentSummaryCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+}
+
+const stateLabels: Record<DeploymentAssessmentStateKeys, string> = {
+  [DeploymentAssessmentState.NEEDS_REVIEW]: 'Needs review',
+  [DeploymentAssessmentState.REJECTED]: 'Rejected',
+  [DeploymentAssessmentState.CHANGES_REQUESTED]: 'Changes requested',
+  [DeploymentAssessmentState.APPROVED]: 'Approved',
+}
+
+function AssessmentStateChip({ assessment }: Pick<DeploymentAssessmentSummaryCardProps, 'assessment'>) {
```
PE39806: It would be cool if these were clickable and updated the filter, like how clicking on the model ID updates the filter.

PE39806 (changes requested):

frontend/src/deployment-assessments/DeploymentAssessmentSummaryCard.tsx
GB907762: Yep agreed. I'll use the same pattern we use for models.

frontend/src/deployment-assessments/DeploymentAssessmentsList.tsx
GB907762: https://github.com/gchq/Bailo/pull/4178/changes#diff-383710ec5173de8780cd75f56c7499ed2809dabb1d81a6a7a29ca3a46476238fR158

Yeah agreed. I've put a TODO here. 

The backend validation will fail when you create one. But I'm tempted to tackle it as a different PR since this should effect other things?

frontend/src/hooks/useDeploymentAssessmentFilters.ts
```diff
@@ -0,0 +1,116 @@
...
+  }, [router])
+
+  return { filters, setFilters, resetFilters }
+}
+
+export function statusToApiFilters(status?: DeploymentAssessmentStatusKeys): {
+  draft?: boolean
+  state?: DeploymentAssessmentStateKeys
+} {
+  if (status === DeploymentAssessmentStatus.DRAFT) {
```
PE39806, line 109: Is it guaranteed that a draft cannot also have a review? If not, we need to change this behaviour.

frontend/src/deployments/DeploymentAssessmentSummaryCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+  )
+}
+
+export default function DeploymentAssessmentSummaryCard({
+  assessment,
+  returnTo,
+  selectedModelIds,
+  onSelectedModelIdsChange,
+}: DeploymentAssessmentSummaryCardProps) {
+  const { schemas } = useGetSchemas(SchemaKind.DEPLOYMENT_ASSESSMENT)
```
PE39806: This lookup should probably be done on the parent as this approach will invoke a call for every DA summary card, even when they have the same schema.

frontend/src/common/CreatedDateRangeFilter.tsx
```diff
@@ -0,0 +1,49 @@
...
+  onCreatedBeforeChange: (value?: string) => void
+}
+
+export default function CreatedDateRangeFilter({
+  createdAfter,
+  createdBefore,
+  onCreatedAfterChange,
+  onCreatedBeforeChange,
+}: CreatedDateRangeFilterProps) {
+  const hasError = Boolean(createdAfter && createdBefore && createdAfter > createdBefore)
```
PE39806: I know that string comparison on `YYYY-MM-DD` formatted dates works but I would prefer a more robust approach e.g.
```suggestion
  const hasError = Boolean(createdAfter && createdBefore && dayjs(createdAfter).isAfter(dayjs(createdBefore)))
```
We may already have a helper/util for this?

PE39806 (changes requested):

frontend/src/deployment-assessments/DeploymentAssessmentsList.tsx
PE39806: That sounds sensible. Just add another TODO here please!

frontend/src/hooks/useDeploymentAssessmentFilters.ts
GB907762: Yes it's guaranteed. A DA cannot be reviewed from draft.

frontend/src/deployments/DeploymentAssessmentSummaryCard.tsx
GB907762: This is a fair point. Although SWR will handle same requests without the need to revalidate, so from a request perspective this is performant. 

I do think this is inefficient though since the name is the only important thing to display. The browser doesn't need to compute all schemas to get this. Instead I'll use the `useGetSchema` hook.

ARADDCC002 (dismissed):

PE39806 (approved):

ARADDCC002 (approved):

PR #4181, opened 2026-09-03
DA Needs Action tab

PE39806:
As a risk owner or creator I would like to view DAs that need my attention. This will allow me to focus on relevant DAs so I can action them in good time.

- Add new query param to backend GET DAs endpoint and include review status in response
- Populate placeholder "Needs action" tab
- Update `UserAvater` and `UserDisplay` to optionally highlight current user


sylvesterkaczmarek (commented):
The aggregation slices `responses` to one element before sorting by `responses.createdAt`, so any later response has already been discarded. That can report an older decision/review date and put an assessment in the wrong Needs Action state. `checkReleaseApproved` already sorts the response array before selecting index 0; could we do the same here and add a multiple-response regression?

JRB66955 (dismissed):

PE39806:
> The aggregation slices `responses` to one element before sorting by `responses.createdAt`, so any later response has already been discarded. That can report an older decision/review date and put an assessment in the wrong Needs Action state. `checkReleaseApproved` already sorts the response array before selecting index 0; could we do the same here and add a multiple-response regression?

Thanks for looking closely at this - but I don't think the ordering is lost here. The sort isn't after the slice; it's inside the `$lookup` sub-pipeline, so it runs first:

```
pipeline: [
  { $match: { $expr: { $eq: ['$parentId', '$$reviewId'] }, kind: ResponseKind.Review } },
  { $sort: { createdAt: -1 } },      // <- ordered here
],
as: 'responses',
...
{ $set: { responses: { $slice: ['$responses', 1] } } },   // <- so this keeps the newest
```

The second sort does something different: an assessment can have several review rounds, so after `$unwind` there's one row per round. `$sort: { 'responses.createdAt': -1 }` + `$group`/`$first` then picks the newest across rounds. Latest-within-round, then latest-across-rounds.

`checkReleaseApproved` needs `$sortArray` only because its `$lookup` uses the `localField`/`foreignField` form, which has no sub-pipeline to sort in. The closer precedent is `getResponses` in the same file, which uses this same as here.

JRB66955 (dismissed):

frontend/utils/deploymentAssessmentUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  switch (state) {
+    case DeploymentAssessmentState.NeedsReview:
+      return { label: 'Awaiting review', colour: 'default' }
+    case DeploymentAssessmentState.ChangesRequested:
+      return { label: 'Changes requested', colour: 'warning' }
+    case DeploymentAssessmentState.Rejected:
+      return { label: 'Rejected', colour: 'error' }
+    default:
```
GB907762: From my testing I can't see approved DAs on the tab. Might be redundant?

frontend/src/deployment-assessments/NeedsAction.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <div>
+          <ChipSelector
+            label='Status'
+            multiple
+            options={statusOptions}
+            selectedChips={selectedStatuses}
+            onChange={setSelectedStatuses}
+            size='small'
+            chipTooltipTitle={(status) => `Filter by status: ${status}`}
+          />
```
GB907762: I like that this component is being reused.
I think it's worth unifying these maybe not in this PR but #4178 uses individual selection.

frontend/src/deployment-assessments/DeploymentAssessmentItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+    <ListItem
+      disablePadding
+      sx={(theme) => ({
+        borderLeft: '6px solid',
+        borderLeftColor:
+          displayState.colour === 'default' ? theme.palette.divider : theme.palette[displayState.colour].main,
+      })}
+    >
+      <ListItemButton
+        component={Link}
```
GB907762: I'm not a big fan of using a link as the top level component, we don't have this anywhere else. But again we can work to unify these.

frontend/src/common/UserAvatar.tsx
GB907762, line 1: So risk owners use the `UserDisplay` but the created by uses this component. Is it worth making these consistent?

backend/src/routes/v3/deploymentAssessment/getDeploymentAssessments.ts
GB907762, line 1: It looks like deleted DAs are still accessible.

GB907762 (dismissed):

sylvesterkaczmarek (commented):
One API semantics issue: `needsAction=false` is accepted by the query schema, but both checks treat it the same as the parameter being absent, so the endpoint returns all assessments rather than assessments that do not need action. Could this distinguish explicit `false` from `undefined`, as the existing `draft` filter does?

ON93754 (dismissed):

sylvesterkaczmarek (commented):
Rechecked the current `needsAction=false` path. It now distinguishes false from undefined, but the query-level `$nor` is broader than the actual action predicate: it excludes every assessment the user created or is a risk owner for, including approved assessments that do not need action and should therefore be returned for `needsAction=false`. The final `needsUserAction()` filter only runs for true, so it cannot correct that over-filtering. Could the false case be derived from the same full action-state predicate as the true case, with a regression for an approved assessment owned/created by the user?

backend/src/services/deploymentAssessment.ts
```diff
@@ -226,6 +228,32 @@ function deriveDeploymentAssessmentState(
   }
 }
 
+/** Risk owners are persisted in entity form, but bare DNs are matched too for legacy and seeded data. */
+function isRiskOwner(deploymentAssessment: Pick<DeploymentAssessmentInterface, 'metadata'>, user: UserInterface) {
```
JR40159: I don't see the need for this function- there is no legacy data and the seeded data should be in the expected format

backend/src/services/deploymentAssessment.ts
```diff
@@ -489,9 +517,36 @@ export async function searchDeploymentAssessments(user: UserInterface, params: S
...
+    andClauses.push({
+      $or: [{ name: search }, { 'metadata.overview.justification': search }],
+    })
+  }
+
+  if (params.needsAction !== undefined) {
+    const actionConditions = [
+      {
+        'metadata.overview.riskOwner': {
+          $in: [toEntity('user', user.dn), user.dn],
```
JR40159: Again, not sure why we're support multiple formats of entity's here

backend/src/services/deploymentAssessment.ts
```diff
@@ -526,19 +581,30 @@ export async function searchDeploymentAssessments(user: UserInterface, params: S
         { $set: { responses: { $slice: ['$responses', 1] } } },
         { $unwind: { path: '$responses', preserveNullAndEmptyArrays: false } },
         { $sort: { 'responses.createdAt': -1 } },
-        { $group: { _id: '$deploymentAssessmentId', decision: { $first: '$responses.decision' } } },
+        {
```
JR40159, line 595: How does this relate to the needs action stuff?

backend/src/services/deploymentAssessment.ts
```diff
@@ -526,19 +581,30 @@ export async function searchDeploymentAssessments(user: UserInterface, params: S
...
-    return Object.assign(assessment, { state })
+    const { decision, reviewedAt } = latestDecisionsByAssessmentId.get(assessment.id) ?? {}
+    const state = deriveDeploymentAssessmentState(assessment, decision)
+    return Object.assign(assessment, { state, ...(reviewedAt && { reviewedAt }) })
   })
 
-  return searchResults.filter(({ state }) => !params.state || state === params.state)
+  return searchResults
+    .filter(({ state }) => !params.state || state === params.state)
+    .filter((assessment) => !params.needsAction || needsUserAction(assessment, user))
```
JR40159: Rather than do some filtering as part of the Mongo query and some using js here, why not update `userNeedsActoin` to return a query filter that then gets added to the Mongo query?

JR40159 (changes requested):

backend/src/services/deploymentAssessment.ts
GB907762: This is a call to action. So it might be useful to infer when a DA was last reviewed. Although this has been subsequently removed form the UI.

Would it be best to have a last review date for DROs and authors?

backend/src/services/deploymentAssessment.ts
GB907762: I suppose. Although there's pros and cons to separating this out. 

Using this predicate in this way ensures that a DA needing action and not needing action are mutually exclusive. It also makes this reusable. Say you want to determine who to copy into an email for reviewing a DA.

backend/src/services/deploymentAssessment.ts
GB907762: Yep good spot

backend/src/services/deploymentAssessment.ts
JR40159: Probably best to leave the last review date out for now as in the future we will likely be added a re-review date to DA approvals (similar to access requests) so it probably best to leave last review stuff until then to make sure it's consistent and doesn't cause confusion

JRB66955 (approved):

JR40159 (approved):

PR #4182, opened 2026-09-03
Migrate Python to uv with lock

### LINKED ISSUE #4080: Migrate & consolidate Python packages structure
### Problem

The 3 Python directories (`lib/python/`, `lib/artefactscan_api/`, `backend/docs/`) each manage dependencies independently with inconsistent tooling. Dev dependencies (black, pre-commit, pytest) are duplicated across directories with no single source of truth. No lockfiles exist, so transitive dependencies are invisible for SBOM generation. Linting coverage is inconsistent (pylint in `lib/python/` only, none for `artefactscan` or `docs`). The `lib/python/pyproject.toml` is largely pylint config.

### Proposed Solution

Adopt **uv** (Astral) as the Python package manager with workspace support, and **ruff** (Astral) as the unified linter/formatter replacing black, isort, pylint, pycln, and pyupgrade. Standardise all 3 directories on `pyproject.toml` with a root workspace config, shared dev dependency groups, and a single `uv.lock` capturing all transitive dependencies for SBOM.

### Alternatives

- pip-tools: mature but no workspace/monorepo support - does not solve shared dev dep duplication
- Poetry: non-PEP 621 metadata conflicts with existing Flit build system in `lib/python/`
- PDM: no Dependabot support
- Keep current linting stack: would require duplicating >100 lines of pylint config across directories

Low

---

### Requirements

- [x] Migrate `lib/artefactscan_api/` and `backend/docs/` from `requirements.txt` to `pyproject.toml`; update Dockerfiles and CI workflows
- [x] Replace black + isort + pylint + pycln + pyupgrade with ruff; add root `ruff.toml`; update `.pre-commit-config.yaml`; update `lib/python/pyproject.toml`
- [ ] Introduce uv workspaces with root `pyproject.toml`; generate `uv.lock` for SBOM; consolidate shared dev deps via `[dependency-groups]`; migrate all CI workflows from pip to uv; replace 3 Dependabot pip entries with 1 uv entry
- [ ] Add CI guardrail to verify `uv export` matches committed `requirements.txt`

PE39806:
Closes #4080

## Motivation

The three Python projects (`lib/python`, `lib/artefactscan_api`, `backend/docs`) each managed dependencies differently and none had a lockfile so transitive dependencies were invisible to SBOM tooling, and Docker builds installed wheels with no integrity verification.

## Changes

- **uv** - each project is now an independent uv project with a committed `uv.lock`. `requirements.txt` deleted; Dockerfiles and CI install from the lock.
- **Reproducible builds** - locks pin all transitive deps *with hashes*; the old `requirements.txt` files had none. CI uses `uv sync --locked`; publishing uses `uv build`
- **Version ranges** - `==` pins replaced with lower bounds no upper caps (the lock provides reproducibility; caps would only constrain downstream consumers of `bailo`). Dev deps moved to PEP 735 `[dependency-groups]` so they no longer ship in published metadata
- **Dependabot** - 3 pip entries -> 1 uv entry with cross-directory grouping, so a shared dep opens one PR not three. `increase-if-necessary` keeps the tested floors from being ratcheted upward
- **`uv-lock.yml`** - checks all three lockfiles and auto-fixes stale ones on PRs, covering the dependabot bug where `pyproject.toml` is bumped without relocking

uv-lock-flow

## GitHub App setup completed

A [GitHub App](https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/registering-a-github-app) has been registered for `uv-lock.yml`, needed because `GITHUB_TOKEN` commits [don't retrigger workflows](https://docs.github.com/en/actions/concepts/security/github_token) - without it the other checks would never re-run against an auto-fixed lockfile.

- User-owned App (Org-owned would be preferable but not viable just now), **Contents: Read and write** only, webhook disabled, installed on this repo
- `UV_LOCK_APP_CLIENT_ID` added to Actions variables
- `UV_LOCK_APP_PRIVATE_KEY` added to **both** Actions and Dependabot secrets - dependabot-triggered runs can't read Actions secrets, which is the main case this workflow covers
- Token minted at runtime via [`actions/create-github-app-token`](https://github.com/actions/create-github-app-token)

## Verification

- `ruff check` and `ruff format --check` clean across all three projects
- `uv sync --locked` resolves on every CI Python version (bailo 3.10-3.14, artefactscan 3.10-3.12, docs 3.12-3.14)
- artefactscan `prod` image serves `/info` returning the `pyproject.toml` version; `dev` target serves with source bind-mounted
- `backend` sphinx stage builds and the rendered HTML carries the real `bailo` version
- `uv build` produces a wheel with the correct metadata and, notably, no `test` extra


sylvesterkaczmarek (commented):
Security concern: the workflow mints a `contents:write` App token and persists it through checkout before running `uv lock` on PR-controlled `pyproject.toml` files. Lock resolution can execute dependency/build metadata code while that write-capable credential is already available. Could the lock/check phase use read-only credentials and mint the App token only for the final same-repo push step?

PR #4183, opened 2026-09-03
Added My Assessments tab

ON93754:
- Added new "My Assessments" tab showing swim lane columns for each status of deployment assessment
- Assessments on this page are read only, and will sit in their respective column
- Added the ability to 'hide' columns, which will persist in browser local storage

### DIFF
--- a/frontend/src/deployment-assessments/MyAssessments.tsx
+++ b/frontend/src/deployment-assessments/MyAssessments.tsx
@@ -2,12 +2,12 @@ import { Box, Chip, Container, Typography } from '@mui/material'
 import { useGetDeploymentAssessments } from 'actions/deploymentAssessments'
 import { useEffect, useMemo, useState } from 'react'
 import renderQueryState from 'src/common/renderQueryState'
-import { SwimLaneColumn } from 'src/deployments/components/SwimLaneColumn'
+import { SwimLaneColumn } from 'src/deployment-assessments/components/SwimLaneColumn'
 import {
   getHiddenDeploymentAssessmentColumns,
   saveHiddenDeploymentAssessmentColumns,
 } from 'src/storage/userPreferences'
-import { DeploymentAssessmentInterface, DeploymentAssessmentState } from 'types/types'
+import { DeploymentAssessmentState, DeploymentAssessmentSummary } from 'types/types'
 
 const COLUMNS = [
   { key: 'in_draft', label: 'In Draft' },
@@ -17,7 +17,7 @@ const COLUMNS = [
   { key: DeploymentAssessmentState.Approved, label: 'Approved' },
 ]
 
-function getColumnKey(assessment: DeploymentAssessmentInterface): string {
+function getColumnKey(assessment: DeploymentAssessmentSummary): string {
   return assessment.draft ? 'in_draft' : assessment.state
 }
 
@@ -32,11 +32,10 @@ export default function MyAssessments() {
   }, [hiddenColumnKeys])
 
   const columns = useMemo(() => {
-    const grouped = new Map<string, DeploymentAssessmentInterface[]>(COLUMNS.map((column) => [column.key, []]))
+    const grouped = new Map<string, DeploymentAssessmentSummary[]>(COLUMNS.map((column) => [column.key, []]))
     for (const assessment of deploymentAssessments ?? []) {
       grouped.get(getColumnKey(assessment))?.push(assessment)
     }
-
     return COLUMNS.map((column) => ({
       ...column,
       items: grouped.get(column.key) ?? [],
--- a/frontend/src/deployment-assessments/MyAssessments.tsx
+++ b/frontend/src/deployment-assessments/MyAssessments.tsx
@@ -1,8 +1,9 @@
 import { Box, Button, Chip, Container, Stack, Typography } from '@mui/material'
-import { alpha, Theme, useTheme } from '@mui/material/styles'
+import { Theme, useTheme } from '@mui/material/styles'
 import { useGetDeploymentAssessments } from 'actions/deploymentAssessments'
-import { useEffect, useMemo, useState } from 'react'
+import { useContext, useEffect, useMemo, useState } from 'react'
 import renderQueryState from 'src/common/renderQueryState'
+import CurrentUserContext from 'src/contexts/currentUserContext'
 import { SwimLaneColumn } from 'src/deployment-assessments/components/SwimLaneColumn'
 import {
   getHiddenDeploymentAssessmentColumns,
@@ -10,23 +11,23 @@ import {
 } from 'src/storage/userPreferences'
 import { DeploymentAssessmentState, DeploymentAssessmentSummary } from 'types/types'
 
-interface SwimLaneColumn {
+interface SwimLaneColumnConfig {
   key: string
   label: string
   color: string
 }
 
-function getColumnConfig(theme: Theme): SwimLaneColumn[] {
+function getColumnConfig(theme: Theme): SwimLaneColumnConfig[] {
   return [
     { key: 'in_draft', label: 'In Draft', color: theme.palette.grey[200] },
-    { key: DeploymentAssessmentState.NeedsReview, label: 'Needs Review', color: alpha(theme.palette.info.main, 0.2) },
+    { key: DeploymentAssessmentState.NeedsReview, label: 'Needs Review', color: theme.palette.grey[200] },
     {
       key: DeploymentAssessmentState.ChangesRequested,
       label: 'Changes Requested',
-      color: alpha(theme.palette.warning.main, 0.2),
+      color: theme.palette.grey[200],
     },
-    { key: DeploymentAssessmentState.Rejected, label: 'Rejected', color: alpha(theme.palette.error.main, 0.2) },
-    { key: DeploymentAssessmentState.Approved, label: 'Approved', color: alpha(theme.palette.success.main, 0.2) },
+    { key: DeploymentAssessmentState.Rejected, label: 'Rejected', color: theme.palette.grey[200] },
+    { key: DeploymentAssessmentState.Approved, label: 'Approved', color: theme.palette.grey[200] },
   ]
 }
 
@@ -36,8 +37,9 @@ function getColumnKey(assessment: DeploymentAssessmentSummary): string {
 
 export default function MyAssessments() {
   const theme = useTheme()
+  const currentUser = useContext(CurrentUserContext)
   const { deploymentAssessments, isDeploymentAssessmentsLoading, isDeploymentAssessmentsError } =
-    useGetDeploymentAssessments()
+    useGetDeploymentAssessments({ createdBy: currentUser.dn })
 
   const [hiddenColumnKeys, setHiddenColumnKeys] = useState(getHiddenDeploymentAssessmentColumns)
 
--- a/frontend/src/deployment-assessments/MyAssessments.tsx
+++ b/frontend/src/deployment-assessments/MyAssessments.tsx
@@ -1,4 +1,5 @@
-import { Box, Chip, Container, Typography } from '@mui/material'
+import { Box, Button, Chip, Container, Stack, Typography } from '@mui/material'
+import { alpha, Theme, useTheme } from '@mui/material/styles'
 import { useGetDeploymentAssessments } from 'actions/deploymentAssessments'
 import { useEffect, useMemo, useState } from 'react'
 import renderQueryState from 'src/common/renderQueryState'
@@ -9,21 +10,34 @@ import {
 } from 'src/storage/userPreferences'
 import { DeploymentAssessmentState, DeploymentAssessmentSummary } from 'types/types'
 
-const COLUMNS = [
-  { key: 'in_draft', label: 'In Draft' },
-  { key: DeploymentAssessmentState.NeedsReview, label: 'Needs Review' },
-  { key: DeploymentAssessmentState.ChangesRequested, label: 'Changes Requested' },
-  { key: DeploymentAssessmentState.Rejected, label: 'Rejected' },
-  { key: DeploymentAssessmentState.Approved, label: 'Approved' },
-]
+interface SwimLaneColumn {
+  key: string
+  label: string
+  color: string
+}
+
+function getColumnConfig(theme: Theme): SwimLaneColumn[] {
+  return [
+    { key: 'in_draft', label: 'In Draft', color: theme.palette.grey[200] },
+    { key: DeploymentAssessmentState.NeedsReview, label: 'Needs Review', color: alpha(theme.palette.info.main, 0.2) },
+    {
+      key: DeploymentAssessmentState.ChangesRequested,
+      label: 'Changes Requested',
+      color: alpha(theme.palette.warning.main, 0.2),
+    },
+    { key: DeploymentAssessmentState.Rejected, label: 'Rejected', color: alpha(theme.palette.error.main, 0.2) },
+    { key: DeploymentAssessmentState.Approved, label: 'Approved', color: alpha(theme.palette.success.main, 0.2) },
+  ]
+}
 
 function getColumnKey(assessment: DeploymentAssessmentSummary): string {
   return assessment.draft ? 'in_draft' : assessment.state
 }
 
 export default function MyAssessments() {
+  const theme = useTheme()
   const { deploymentAssessments, isDeploymentAssessmentsLoading, isDeploymentAssessmentsError } =
-    useGetDeploymentAssessments({})
+    useGetDeploymentAssessments()
 
   const [hiddenColumnKeys, setHiddenColumnKeys] = useState(getHiddenDeploymentAssessmentColumns)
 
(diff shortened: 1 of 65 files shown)


frontend/src/storage/userPreferences.ts
```diff
@@ -46,3 +48,20 @@ export const saveDisplayFormStats = (newValue: boolean) => {
   preferences.displayFormStats = newValue
   saveUserPreferences(preferences)
 }
+
+/**
+ * Get the columns to be hidden in the My Assessments tab.
```
PE39806: @ARADDCC002 isn't keen on self explanatory comments and this can be inferred from the function name

frontend/src/storage/userPreferences.ts
```diff
@@ -46,3 +48,20 @@ export const saveDisplayFormStats = (newValue: boolean) => {
...
+ */
+export const getHiddenDeploymentAssessmentColumns = (): string[] => {
+  const preferences: UserPreferences = getUserPreferences()
+  return preferences?.hiddenDeploymentAssessmentColumns ?? []
+}
+
+/**
+ * Save the columns to be hidden in the My Assessments tab.
```
PE39806: As above

PE39806 (dismissed):
I hit an error when trying to view the new content
```
## Error Type
Runtime TypeError

## Error Message
Cannot read properties of undefined (reading 'includes')


    at UserDisplay (src/common/UserDisplay.tsx:42:8)

## Code Frame
  40 |   const ref = useRef<HTMLDivElement>(null)
  41 |   const { userInformation, isUserInformationLoading, isUserInformationError } = useGetUserInformation(
> 42 |     dn.includes(':') ? dn.split(':')[1] : dn,
     |        ^
  43 |   )
  44 |
  45 |   const popoverEnter = () => {

Next.js version: 16.3.0 (Turbopack)
```

You likely need to copy in the same changes from https://github.com/gchq/Bailo/pull/4181/changes#diff-824ee34c5aacf20c233d49398f6d96e8388e13de8b5a6318cd60b066cef6953e

sylvesterkaczmarek (commented):
`MyAssessments` currently calls `useGetDeploymentAssessments({})`, and the backend's empty query returns every assessment the user is authorised to view; it does not implicitly scope to the current user. This makes the new 'My assessments' tab effectively 'All visible assessments'. Please pass the intended current-user filter (`createdBy`, `riskOwner`, or the required union) and cover the case where another user's assessment is also visible.

frontend/src/deployments/MyAssessments.tsx
```diff
@@ -0,0 +1,104 @@
...
+  { key: DeploymentStates.Approved, label: 'Approved' },
+]
+
+function getColumnKey(assessment: DeploymentAssessmentInterface): string {
+  return assessment.draft ? 'in_draft' : assessment.state
+}
+
+export default function MyAssessments() {
+  const { deploymentAssessments, isDeploymentAssessmentsLoading, isDeploymentAssessmentsError } =
+    useGetDeploymentAssessments({})
```
ARADDCC002: `filters` is an optional argument, so you should be able to remove `{}`

frontend/src/deployment-assessments/components/SwimLaneAssessmentCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+  return Array.isArray(owner) ? owner[0] : owner
+}
+
+export function SwimLaneAssessmentCard({ assessment }: SwimLaneAssessmentCardProps) {
+  const owner = getFirstOwner(assessment.owner)
+  const deployer = assessment.createdBy
+
+  return (
+    <Card variant='outlined'>
```
GB907762: Is there any other information that could be useful to see. Potentially models?

frontend/src/deployment-assessments/components/SwimLaneAssessmentCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <CardContent sx={{ p: 1.5, '&:last-child': { pb: 1.5 } }}>
+        <Typography component='h3' variant='body2' sx={{ fontWeight: 600, mb: 0.5 }}>
+          {assessment.name}
```
GB907762: Does this need to be a link to the individual page? Inline with other patterns within the app.

frontend/src/deployment-assessments/components/SwimLaneAssessmentCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+          color='text.secondary'
+          sx={{
+            mb: 1.5,
+            display: '-webkit-box',
+            WebkitLineClamp: 3,
+            WebkitBoxOrient: 'vertical',
+            overflow: 'hidden',
+          }}
+        >
+          {assessment.justification}
```
GB907762: To me it's not clear what this text is. Does it need a label?

frontend/src/deployment-assessments/components/SwimLaneColumn.tsx
```diff
@@ -0,0 +1,58 @@
...
+        display: 'flex',
+        flexDirection: 'column',
+        gap: 1,
+      }}
+    >
+      <Box sx={{ display: 'flex', alignItems: 'center', justifyContent: 'space-between' }}>
+        <Typography
+          component='h2'
+          variant='caption'
+          sx={{ fontWeight: 700, textTransform: 'uppercase', letterSpacing: 0.3 }}
```
GB907762: Could you use title case instead of uppercase to keep consistent with the rest of app, and use this within the following JS expression.

frontend/src/deployment-assessments/components/SwimLaneColumn.tsx
```diff
@@ -0,0 +1,58 @@
...
+        </Box>
+      </Box>
+      <Box sx={{ display: 'flex', flexDirection: 'column', gap: 0.5, pt: 1 }}>
+        {assessments.map((assessment) => (
+          <SwimLaneAssessmentCard key={assessment.id} assessment={assessment} />
+        ))}
+        {assessments.length === 0 && (
+          <Typography component='p' variant='caption' color='text.secondary'>
+            No assessments
+          </Typography>
```
GB907762: Could you use an empty blob.

frontend/src/deployment-assessments/MyAssessments.tsx
```diff
@@ -0,0 +1,103 @@
...
+  { key: DeploymentAssessmentState.Approved, label: 'Approved' },
+]
+
+function getColumnKey(assessment: DeploymentAssessmentSummary): string {
+  return assessment.draft ? 'in_draft' : assessment.state
+}
+
+export default function MyAssessments() {
+  const { deploymentAssessments, isDeploymentAssessmentsLoading, isDeploymentAssessmentsError } =
+    useGetDeploymentAssessments({})
```
GB907762:
```suggestion
    useGetDeploymentAssessments()
```
This is a default property.

frontend/src/deployment-assessments/MyAssessments.tsx
```diff
@@ -0,0 +1,103 @@
...
+            Hidden columns:
+          </Typography>
+          {hiddenColumns.map((column) => (
+            <Chip
+              key={column.key}
+              label={column.label}
+              size='small'
+              variant='outlined'
+              onDelete={() => showColumn(column.key)}
+            />
```
GB907762, line 120: Could you reuse the state chips on the `All assessments` page?

GB907762 (changes requested):

JR40159 (changes requested):
1. Make the number of items in each column more visible but putting it after the title, not on the right had side
2. Change the icon for minimise, it's currently a close symbol
3. I'd prefer minimise resulting in a thinner swim-lane with all the items hidden, the count still visible and title still visible with an expand button
4. Add an expand all button if at least one column is collapsed
5. Add a collapse empty columns button 
6. Make the deployment assessments clickable

ARADDCC002:
* The "hide" icon should probably be either a dash or perhaps an eye icon instead of a cross and it looks more like a "close" or "delete".
* (optional) the items being the same height would be more visually consistent
* I'm not sure what `Deployer` and `Owner` mean, they sound like the same person, by owner do you mean Risk Owner?
* Some colours on the items/columns would be nice.
  * Draft - keep as grey 
  * Review - info
  * Changes - warning
  * Rejected - error
  * Approved - success

frontend/src/deployment-assessments/components/SwimLaneAssessmentCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export function SwimLaneAssessmentCard({ assessment }: SwimLaneAssessmentCardProps) {
+  const owner = getFirstOwner(assessment.owner)
+  const deployer = assessment.createdBy
+
+  return (
+    <Card variant='outlined'>
+      <CardContent sx={{ p: 1.5, '&:last-child': { pb: 1.5 } }}>
+        <Typography component='h3' variant='body2' sx={{ fontWeight: 600, mb: 0.5 }}>
```
IR96334: Overflow issue on large names:
Use something like:
``          sx={{ fontWeight: 600, mb: 0.5, overflow: 'hidden', textOverflow: 'ellipsis' }}``

frontend/src/deployment-assessments/MyAssessments.tsx
ON93754: Having made a lot of changes to the code there are now enough differences in functionality that I think it needs it own component

frontend/src/deployment-assessments/MyAssessments.tsx
```diff
@@ -0,0 +1,103 @@
...
+      )}
+      <Box sx={{ overflowX: 'auto', py: 2 }}>
+        <Box
+          sx={{
+            display: 'flex',
+            gap: 1,
+            width: '100%',
+          }}
+        >
+          {visibleColumns.map((column) => (
```
IR96334, line 132: On a smaller viewport width, the columns disappear entirely and aren't scrollable. This seems partially due to the `display: 'flex'` on line 86.

The swim lane view is very good for larger screens, but demands too much screenwidth to look good on smaller screens.

One possible solution is to do a check, e.g:
` const isLarge = useMediaQuery(theme.breakpoints.up('lg'))`
and then switch a different view. For example, instead of swim lanes it could be one column in which you can change the key of the lane by a menu. 

Or more simply, you could just forgo this trifle: use a stack and switch the direction so the lanes stack on top of each other, though viewing will require scrolling.

frontend/src/deployment-assessments/components/SwimLaneAssessmentCard.tsx
IR96334: Fix overflow with justification

frontend/src/deployment-assessments/components/SwimLaneAssessmentCard.tsx
```diff
@@ -0,0 +1,135 @@
...
+          <Typography
+            variant='body2'
+            sx={{
+              display: '-webkit-box',
+              WebkitLineClamp: 2,
+              WebkitBoxOrient: 'vertical',
+              overflow: 'hidden',
+            }}
+          >
+            {assessment.justification}
```
IR96334: When no Justification, either get rid of justification title or show is empty.

frontend/src/deployment-assessments/components/SwimLaneColumn.tsx
```diff
@@ -0,0 +1,64 @@
...
+          component='h2'
+          variant='caption'
+          sx={{ fontWeight: 700, textTransform: 'capitalize', letterSpacing: 0.3 }}
+        >
+          {`${title} (${assessments.length})`}
+        </Typography>
+        <Box sx={{ display: 'flex', alignItems: 'center', gap: 0.5 }}>
+          <Tooltip title={`Hide ${title} column`}>
+            <IconButton size='small' onClick={onHide} sx={{ p: 0.25 }} aria-label={`Hide ${title} column`}>
+              <MinimizeIcon sx={{ fontSize: 16 }} />
```
IR96334: minus symbol is not centered

frontend/src/deployment-assessments/MyAssessments.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  const queryState = renderQueryState([isDeploymentAssessmentsError], false)
+  if (queryState) {
+    return queryState
+  }
+
+  return (
+    <Container maxWidth={false} disableGutters sx={{ px: 2 }}>
+      <Stack direction='row' spacing={1} sx={{ pt: 2, flexWrap: 'wrap', gap: 1 }}>
+        <Button
```
IR96334, line 87: I'm not actually too fond of the combination of the buttons (hide and show). I find its a bit unnecessarily convoluted for 5 columns. Also you can hide empty columns, and while in this "mode" you can add non-empty columns to the hide list. I preferred the previous incarnation as it was simple and clean.

frontend/src/deployment-assessments/MyAssessments.tsx
IR96334: This is still relevant

frontend/src/deployment-assessments/components/SwimLaneColumn.tsx
IR96334: Also, prefer the X from previous commits. This looks more like a minimise, not remove.

frontend/src/deployment-assessments/MyAssessments.tsx
```diff
@@ -0,0 +1,144 @@
...
+} from 'src/storage/userPreferences'
+import { DeploymentAssessmentState, DeploymentAssessmentSummary } from 'types/types'
+
+interface SwimLaneColumn {
+  key: string
+  label: string
+  color: string
+}
+
+function getColumnConfig(theme: Theme): SwimLaneColumn[] {
```
IR96334: These colours are a little garish and not in keeping with the Bailo theme.

frontend/src/deployment-assessments/MyAssessments.tsx
```diff
@@ -0,0 +1,144 @@
...
+}
+
+function getColumnKey(assessment: DeploymentAssessmentSummary): string {
+  return assessment.draft ? 'in_draft' : assessment.state
+}
+
+export default function MyAssessments() {
+  const theme = useTheme()
+  const { deploymentAssessments, isDeploymentAssessmentsLoading, isDeploymentAssessmentsError } =
+    useGetDeploymentAssessments()
```
JRB66955: Is this doing any filtering? Current read is this isn't - which means that this displays _all_ assessments, not just "mine".

frontend/src/deployment-assessments/MyAssessments.tsx
```diff
@@ -0,0 +1,144 @@
...
+import { useEffect, useMemo, useState } from 'react'
+import renderQueryState from 'src/common/renderQueryState'
+import { SwimLaneColumn } from 'src/deployment-assessments/components/SwimLaneColumn'
+import {
+  getHiddenDeploymentAssessmentColumns,
+  saveHiddenDeploymentAssessmentColumns,
+} from 'src/storage/userPreferences'
+import { DeploymentAssessmentState, DeploymentAssessmentSummary } from 'types/types'
+
+interface SwimLaneColumn {
```
JRB66955: Could be renamed? Just due to also importing a `SwimLaneColumn` component in the same file - may be overly cautious though.

frontend/src/deployment-assessments/components/SwimLaneAssessmentCard.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const models = assessment.models?.join(', ') ?? ''
+  const returnTo = '/deployment-assessments?tab=my-assessments'
+
+  return (
+    <Card variant='outlined' sx={{ minHeight: 140 }}>
+      <CardContent sx={{ p: 1.5, '&:last-child': { pb: 1.5 } }}>
+        <Link
+          href={{
+            pathname: `/deployment-assessments/${assessment.id}`,
+            query: returnTo ? { returnTo } : undefined,
```
JRB66955:
```suggestion
            query: { returnTo }
```

Think we can avoid the null check because its already defined above as a string value.

JRB66955 (changes requested):

frontend/src/deployment-assessments/components/SwimLaneAssessmentCard.tsx
ON93754: Currently justification is a mandatory field so it will always be there, but this will also be getting removed in a future PR so i think its fine to leave it for now

frontend/src/deployment-assessments/components/SwimLaneColumn.tsx
ON93754: Removed this from the scope of the ticket, new ticket created to tackle improved UX

frontend/src/deployment-assessments/MyAssessments.tsx
ON93754: Removed this from the scope of the ticket, new ticket created to tackle improved UX

frontend/src/deployment-assessments/MyAssessments.tsx
ON93754: Removed this from the scope of the ticket, new ticket created to tackle improved UX

JRB66955 (dismissed):

IR96334 (dismissed):

JR40159 (dismissed):

GB907762 (approved):

frontend/src/deployment-assessments/components/SwimLaneAssessmentCard.tsx
```diff
@@ -0,0 +1,135 @@
...
+            sx={{
+              display: '-webkit-box',
+              WebkitLineClamp: 2,
+              WebkitBoxOrient: 'vertical',
+              overflow: 'hidden',
+            }}
+          >
+            {assessment.justification}
+          </Typography>
+        </Box>
```
GB907762: Minor but could you remove this in view of #4204

GB907762 (changes requested):

GB907762 (approved):

PR #4184, opened 2026-09-03
Deployment assessments tab on model page

ARADDCC002:


sylvesterkaczmarek (commented):
Hi, I think the tab disable condition is missing the untrusted-model case. `disabledText` has a message specifically for `UNTRUSTED_MODEL`, but `disabled` is only `!entry.card`, so an untrusted model with a card gets an enabled Deployment assessments tab. Should this mirror Access requests with `!entry.card || entry.kind === EntryKind.UNTRUSTED_MODEL`?

frontend/pages/model/[modelId]/index.tsx
```diff
@@ -56,6 +57,17 @@ export default function Model() {
                   ? 'Access requests are not available for untrusted models.'
                   : 'Select a schema to view this tab.',
             },
+            {
+              title: 'Deployment assessments',
+              path: 'deployments',
+              view: <DeploymentAssessments model={entry} />,
+              datatest: 'deploymentAssessmentsTab',
+              disabled: !entry.card,
```
Comment (GB907762) [inline frontend/pages/model/[modelId]/index.tsx]: Do you think this should be disabled based on if a model card schema has not been selected, or based on the criteria of a model being in a deployable state and public?

frontend/pages/model/[modelId]/index.tsx
```diff
@@ -56,6 +57,17 @@ export default function Model() {
...
+              disabledText:
+                entry.kind === EntryKind.UNTRUSTED_MODEL
+                  ? 'Deployment assessments are not available for untrusted models.'
+                  : 'Select a schema to view this tab.',
```
Comment (GB907762) [inline frontend/pages/model/[modelId]/index.tsx]: Instead of UMs. Use private models and `config.ui.deploymentAssessments.deployableModelState`?

GB907762 (changes requested):

GB907762 (changes requested):
Do we also need to consider hiding drafts, on the model tab?

frontend/pages/model/[modelId]/index.tsx
Comment (ARADDCC002) [reply frontend/pages/model/[modelId]/index.tsx]: I think that having a schema is the bare minimum needed. If a user goes to a model that has  no schema and sees the tab as active and selects "Create deployment assessment" and then doesn't see that model in the list of available models I think it will get confusing. 

We're waiting on confirmation on what state a model can be in before it can be attached to a deployment, so possibly once that's finalised we can then update this.

ARADDCC002:
> Do we also need to consider hiding drafts, on the model tab?

I was under the impression drafts were not visible outside of the author of that assessment, unless I'm mistaken?

frontend/pages/model/[modelId]/index.tsx
Comment (ARADDCC002) [reply frontend/pages/model/[modelId]/index.tsx]: Looking at your comment below we can ignore my last comment, I didn't know we had `uiConfig.deploymentAssessments.deployableModelState` available, that makes it much easier to disable

GB907762 (approved):

ON93754 (approved):

JRB66955 (approved):

PR #4186, opened 2026-09-04
Added review ui for deployment assessments

ARADDCC002:


JR40159 (changes requested):
When I create a new assessment as by saving a draft I get `Not Found: The deployment assessment does not have a review round.`

frontend/actions/review.ts
```diff
@@ -14,6 +15,13 @@ import { ErrorInfo, fetcher } from '../utils/fetcher'
 
 const emptyReviewList = []
 
+function toSafePathId(value: string): string {
```
ON93754: Could we have a couple of unit tests for this function?

frontend/src/deployment-assessments/AssessmentStateChip.tsx
```diff
@@ -0,0 +1,86 @@
...
+        }}
+      />
+    )
+  }
+
+  const state = assessment.state
+  const isSelected = selectable && selectedState === state
+
+  const palette = {
+    [DeploymentAssessmentState.NEEDS_REVIEW]: {
```
ON93754: We'll need to decide collectively which format we want, and do it consistently everywhere, as currently we are doing a mix of `DeploymentAssessmentState.NEEDS_REVIEW` and `DeploymentAssessmentState.NeedsReview`

frontend/types/types.ts
```diff
@@ -583,6 +583,7 @@ export const Decision = {
   RequestChanges: 'request_changes',
   Approve: 'approve',
   Undo: 'undo',
+  Reject: 'reject',
```
ON93754, line 586: Similar, in some of these structures we are using CamelCase e.g `Reject` and in others we are using CAPSLOCK like `DEPLOYMENTS` below

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
```diff
@@ -116,6 +127,14 @@ export default function DeploymentAssessment() {
                     onIsEditChange={setIsEdit}
                   />
                 )}
+                <ReviewComments
+                  identifier={deploymentAssessment.id}
+                  parentId={deploymentAssessment._id}
+                  kind={ReviewKind.DEPLOYMENTS}
+                  isEdit={isEdit}
+                  mutator={mutateDeploymentAssessment}
+                  entryId={''}
```
Comment (ON93754) [inline frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx]: I think you can just do `entryID=''`

backend/src/services/deploymentAssessment.ts
```diff
@@ -231,13 +230,27 @@ export async function getDeploymentAssessmentDetails(
...
+    return {
+      deploymentAssessment,
+      responses: [],
+    }
+  }
+
+  const latestReview = await getLatestDeploymentAssessmentReview(deploymentAssessmentId)
+  const responses = await ResponseModel.find({ parentId: latestReview._id })
+
+  const latestDecision = responses
```
ON93754: Worth making this a function/adding a comment? There is a lot going on here

frontend/src/reviews/DeploymentAssessmentReviewItem.tsx
```diff
@@ -0,0 +1,70 @@
...
+            spacing={1}
+            sx={{
+              alignItems: 'center',
+              justifyContent: 'center',
+            }}
+          >
+            <NotificationsNoneOutlinedIcon sx={{ fontSize: 'medium' }} color='warning' />
+            <Typography variant='subtitle2' sx={{ fontStyle: 'italic' }} component='p'>
+              This deployment assessment needs to be reviewed by the deployment risk owner.
+            </Typography>
```
GB907762: In similar reviews this message is replaced once a decision has been made such as
`Approved by Deployment Risk Owner`

frontend/pages/deployment-assessments/[deploymentAssessmentId]/review.tsx
```diff
@@ -0,0 +1,130 @@
...
+            >
+              <Link href={`/deployment-assessments/${deploymentAssessmentId}`}>
+                <Button sx={{ width: 'fit-content' }} startIcon={<ArrowBack />}>
+                  Back to deployment assessment
+                </Button>
+              </Link>
+              <Typography variant='h6' component='h1' color='primary'>
+                {deploymentAssessment ? `Reviewing deployment assessment ${deploymentAssessment.name}` : 'Loading...'}
+              </Typography>
+              <Button onClick={() => setIsAssessmentDialogOpen(true)}>View full deployment assessment</Button>
```
Comment (GB907762) [inline frontend/pages/deployment-assessments/[deploymentAssessmentId]/review.tsx]: I know at the moment we separate out these into different pages for most reviews. Do you have any opinion on making the comments on the DA page reverse chronological so a DRO can review at in the same view as this instead of having a popup?

frontend/actions/review.ts
```diff
@@ -14,6 +15,13 @@ import { ErrorInfo, fetcher } from '../utils/fetcher'
...
+  if (!/^[A-Za-z0-9-]+$/.test(value)) {
+    throw new Error('Invalid deployment assessment ID')
+  }
+  return encodeURIComponent(value)
+}
```
GB907762: What's the purpose of this clause. We don't have this for access requests models ect..

backend/src/services/deploymentAssessment.ts
```diff
@@ -231,13 +230,27 @@ export async function getDeploymentAssessmentDetails(
...
+    .filter((r) => r.kind === ResponseKind.Review)
+    .sort((a, b) => new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime())
+    .at(0)?.decision as DecisionKeys | undefined
```
GB907762: The queried documents from `getLatestDeploymentAssessmentReview` already return in reverse chronological order. I think this sorting is redundant?

GB907762 (changes requested):

JR40159 (changes requested):
Would be good if the risk owner could complete the review while also being able to view the deployment assessment at the same time

frontend/src/deployment-assessments/AssessmentStateChip.tsx
ARADDCC002: Frontend uses all upper case, while backend tends to use the latter approach. I agree we should standardise it, but out of scope for here I think.

frontend/types/types.ts
ARADDCC002: I agree that the inconsistency isn't good, but I think we should resolve this all at once in a different PR if you're ok with that?

backend/src/services/deploymentAssessment.ts
ARADDCC002: Added a comment

frontend/actions/review.ts
ARADDCC002: We do have it for review IDs as well (we had a similar function which I've since removed and used this one and made it generic). 

Ideally we can re-use it as I've moved it into `utils/stringUtils`.

ARADDCC002:
> Would be good if the risk owner could complete the review while also being able to view the deployment assessment at the same time

This is something that we can do in the future when we change how the review UI works.

GB907762 (dismissed):

backend/src/services/deploymentAssessment.ts
```diff
@@ -260,13 +259,26 @@ export async function getDeploymentAssessmentDetails(
...
   const deploymentAssessment = await getDeploymentAssessmentById(user, deploymentAssessmentId)
-  const responses = await getResponses(deploymentAssessment._id)
-  const latestDecision = responses.findLast(({ kind }) => kind === ResponseKind.Review)?.decision
+
+  if (deploymentAssessment.draft === true) {
+    return {
+      deploymentAssessment,
+      responses: [],
+    }
+  }
```
GB907762, line 268: Could we fix this to properly initialise on creation.

GB907762 (changes requested):
Could you also investigate using a popup for review comments instead of expanding the cards. Since this could be quite jarring?

sylvesterkaczmarek (commented):
`getDeploymentAssessmentDetails()` no longer selects the latest review decision. `ResponseModel.find(...)` is unsorted and the code takes `.filter(...).at(0)`, so after a reviewer changes their decision an earlier response can continue to determine the assessment state. The previous path sorted responses and used the last review response. Could we sort review responses by `createdAt` descending (or otherwise select the latest) and add a regression for approve -> request changes/reject?

sylvesterkaczmarek (commented):
Deployment-assessment reviews bypass the normal visibility check when `mine=false`. For this kind the pipeline skips `matchUserToDro`, then `findReviews()` returns the aggregation directly instead of authorising each joined deployment assessment, so any caller of `/api/v2/reviews?kind=deploymentAssessment&mine=false` can enumerate review records (and the joined assessment data) they may not be allowed to view through the deployment-assessment endpoint. Could we apply deployment-assessment view authorisation before returning these results and add a non-DRO/non-viewer regression?

frontend/src/Form/EditableFormHeading.tsx
```diff
@@ -65,28 +65,19 @@ export default function EditableFormHeading({
...
-                onClick={onEdit}
-                data-test='editFormButton'
-                disabled={isRegistryError}
-                startIcon={<Create />}
-              >
-                {editButtonText}
-              </Button>
+              <Tooltip title={editButtonText}>
+                <IconButton data-test='editFormButton' onClick={onEdit}>
+                  <Create />
```
IR96334, line 70: Make this primary colour (purple) as it looks almost "deactivated" or unclickable.

frontend/src/common/ReviewWithComment.tsx
```diff
@@ -110,36 +118,35 @@ export default function ReviewWithComment({
...
+                  option.role === value.role
+                }
+                onChange={onChange}
+                value={reviewRequest}
+                getOptionLabel={(option) => getRoleDisplayName(option.role, entryRoles)}
+                options={reviews}
+                renderInput={(params) => <TextField {...params} label='Select your role' size='small' />}
+              />
+            )}
+            <RichTextEditor
```
IR96334, line 141: Do we want a max character limit?

frontend/src/entry/model/reviews/ReviewBanner.tsx
```diff
@@ -49,6 +78,34 @@ export default function ReviewBanner({ release, accessRequest }: ReviewBannerPro
...
+      <Stack
+        direction='row'
+        spacing={2}
+        sx={{
+          justifyContent: 'space-between',
+          alignItems: 'center',
+          width: '100%',
+        }}
+      >
+        <Stack direction='row' spacing={1}>
```
IR96334, line 92: This stack is squashed when the DA name is long. Could add a minWidth to stack to stop this, e.g. 250px?

frontend/src/common/ReviewWithComment.tsx
ARADDCC002: I think we'd need to discuss this further as it would require backend changes to make this effective.

frontend/src/entry/model/reviews/ReviewBanner.tsx
IR96334: Still seems to be squashed against the long DA name

sylvesterkaczmarek (commented):
Re-reviewed current `95f63293`. The deployment-assessment authorization issue I raised is still present. In `findReviews()`, `mine=false` skips `matchUserToDro()`, and deployment-assessment results are returned directly at the end without an `authorisation.deploymentAssessments(..., View)` check. The commits since my earlier review are frontend/UI changes and do not change that backend path. A caller can therefore still enumerate deployment-assessment review records they are not authorised to view through the assessment endpoint. This remains a blocker from my review.

ARADDCC002:
> Re-reviewed current `95f63293`. The deployment-assessment authorization issue I raised is still present. In `findReviews()`, `mine=false` skips `matchUserToDro()`, and deployment-assessment results are returned directly at the end without an `authorisation.deploymentAssessments(..., View)` check. The commits since my earlier review are frontend/UI changes and do not change that backend path. A caller can therefore still enumerate deployment-assessment review records they are not authorised to view through the assessment endpoint. This remains a blocker from my review.

Reviews do not need need to be private as we use them to determine the "review status" of items such as releases and deployment assessments. The `matchUSerToDro()` check is specifically for determining which reviews are assigned to the `currentUser`.

sylvesterkaczmarek:
Thanks, that makes sense for the review/status record itself. My concern is slightly narrower: for deployment-assessment reviews the aggregation joins the full `deploymentAssessment` document and, when `mine=false`, returns that joined result directly. A normal deployment-assessment read goes through `DeploymentAssessmentAction.View`. If the joined assessment document is also intentionally visible to any authenticated caller, I am happy to withdraw this blocker. If not, I think this path should either project only the review/status fields it needs or apply the existing deployment-assessment view authorisation before returning. The concern is the joined assessment payload rather than privacy of the review record itself.

PE39806 (changes requested):
A couple of minor issues:
- sometimes the review popup can be "too far" to the right which adds a horizontal scrollbar
- reactions on comments don't trigger a re-render (but refreshing the page shows that the request completed)

frontend/src/deployment-assessments/DeploymentAssessmentSummaryCard.tsx
JR40159, line 1: Is still think on we should use the review banner on the DA summary/preview view. I think consistency is really important for good UX and to use the review banner everywhere apart from here seems a bit odd

JR40159 (changes requested):

PE39806 (dismissed):

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
```diff
@@ -56,51 +98,124 @@ export default function DeploymentAssessment() {
...
+              </>
+            )}
+          </Paper>
+          {isReviewPanelShown && (
+            <Popper
+              open={reviewPopoverOpen}
+              anchorEl={anchorEl}
+              sx={{
+                maxWidth: '450px',
+                width: '100%',
```
Comment (GB907762) [inline frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:185]: I can't see where it is but there's some styling on this component that prevents the toolbar from becoming full screen and only fills the view port inside the popper, could be here.

GB907762 (changes requested):

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
Comment (ARADDCC002) [reply frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:185]: Definitely an issue but I don't see it as being big enough to be a blocker for now.

backend/src/services/deploymentAssessment.ts
```diff
@@ -262,13 +261,26 @@ export async function getDeploymentAssessmentDetails(
...
+
+  // Get the latest review and its corresponding responses to determine the state of this deployment assessment
+  const latestReview = await getLatestDeploymentAssessmentReview(deploymentAssessmentId)
+  const responses = await ResponseModel.find({ parentId: [latestReview._id, deploymentAssessment._id] })
+
+  const latestDecision = responses.filter((r) => r.kind === ResponseKind.Review).at(0)?.decision as
```
JRB66955, line 280: Based on the filtering (and `.at(0)`) should we have some kind of sort from the mongo query? For example, add a `.sort({ createdAt: -1 })`?

frontend/actions/deploymentAssessments.ts
```diff
@@ -91,13 +92,15 @@ export function useGetDeploymentAssessment(deploymentId?: string) {
...
+      state: DeploymentAssessmentStateKeys
+      responses?: ResponseInterface[]
     },
     ErrorInfo
   >(deploymentId ? `/api/v3/deployment-assessments/${deploymentId}` : null, fetcher)
 
   return {
     mutateDeploymentAssessment: mutate,
-    deploymentAssessment: data?.deploymentAssessment,
+    deploymentAssessment: data && { ...data?.deploymentAssessment, state: data?.state, respones: data?.responses },
```
JRB66955:
```suggestion
    deploymentAssessment: data && { ...data?.deploymentAssessment, state: data?.state, responses: data?.responses },
```

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
```diff
@@ -56,51 +98,124 @@ export default function DeploymentAssessment() {
...
+                    )}
+                  </Stack>
+                  <ReviewComments
+                    identifier={deploymentAssessment.id}
+                    parentId={deploymentAssessment._id}
+                    kind={ReviewKind.DEPLOYMENTS}
+                    isEdit={isEdit}
+                    mutator={mutateDeploymentAssessment}
+                    entryId=''
+                    responseList={deploymentAssessment.respones}
```
Comment (JRB66955) [inline frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx]:
```suggestion
                    responseList={deploymentAssessment.responses}
```

JRB66955 (changes requested):

backend/src/services/deploymentAssessment.ts
ARADDCC002: This was discussed previously, and it's already sorted at an earlier point from what I can remember

PE39806 (dismissed):

JR40159 (dismissed):

JR40159 (approved):

PE39806 (approved):

PR #4188, opened 2026-09-07
feat: add scan completion webhook notifications

### LINKED ISSUE #3772: Scan completion webhook notifications
### Problem

It would be good to have a way to enable automatons when an artefact (file or Docker image) has a clean set of scans.
For example, pull a Docker image when all scans are clean.

### Proposed Solution

When all scans have completed on an artefact send a webhook detailing the results of the scans. This will then enable other systems to take appropriate action.
Bailo already has partial support for webhooks and this would require extending the current support by adding a new webhook event kind.

### Alternatives

Currently this can be done manually by refreshing or polling an artefact scan results.

Low

sylvesterkaczmarek:
##### Checklist

- [ ] `npm run test` succeeds both unit and integration testing.

Backend artefact scanning and webhooks; webhook user documentation.

### Description of change

Adds a `scanComplete` webhook event for model artefacts.

- Files emit one event after all configured file scanners reach a final state.
- Docker images emit one event only after every image layer has finished scanning.
- Payloads include scanner state, summary, scanner version and last-run time; image results also include the layer digest.
- Raw scanner `additionalInfo` is intentionally excluded from webhook payloads.
- Existing scanning endpoints remain fire-and-forget.
- Webhook API examples and user documentation now include the new event.

Closes #3772.

### Validation

- Regression tests fail on `main` before the change and pass after it.
- Targeted backend tests: 27/27 passed.
- Backend Prettier, ESLint and TypeScript build passed.
- Frontend Prettier, ESLint and production build passed.
- Frontend unit tests: 159/159 passed with `TZ=UTC`.
- Documentation style checks passed using the same GNU `realpath` behaviour as CI/Linux.
- Full backend suite with `TZ=UTC`: 1,416/1,417 tests passed. The only failure was an unrelated existing v3 review-response route test returning 404 once; that test passed 5/5 when rerun in isolation.
- `git diff --check` passed.

### DIFF
--- a/backend/src/models/Webhook.ts
+++ b/backend/src/models/Webhook.ts
@@ -8,6 +8,7 @@ export const WebhookEvent = {
   CreateReviewResponse: 'createReviewResponse',
   CreateAccessRequest: 'createAccessRequest',
   ImportModel: 'importModel',
+  ScanComplete: 'scanComplete',
 } as const
 export type WebhookEventKeys = (typeof WebhookEvent)[keyof typeof WebhookEvent]
 
--- a/backend/src/services/scan.ts
+++ b/backend/src/services/scan.ts
@@ -8,6 +8,7 @@ import { FileInterface } from '../models/File.js'
 import { ImageRef } from '../models/Release.js'
 import ScanModel, { ArtefactKind, ArtefactKindKeys } from '../models/Scan.js'
 import { UserInterface } from '../models/User.js'
+import { WebhookEvent } from '../models/Webhook.js'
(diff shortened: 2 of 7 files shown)


JR40159:
Thank you for your contribution.

As we cannot verify that this submission meets the requirements set out in our updated contribution guidelines regarding human authorship, oversight, and accountability, we are closing this PR. If you would like the change to be considered, please re-submit it in accordance with the updated guidance and PR template (#4215).

Thank you for your understanding.

Issue #2348, opened 2025-06-27
Allow retrieval of detailed scan output


GCHQDeveloper36:
## Is your feature request related to a problem? Please describe.

I'd like to see more detailed output from the scanners, even if the scan is successful.

## Describe the solution you would like

For each scan invocation, I want to be able to see as much detail as possible that was provided by the tool. This might be the actual log file from the tool, or a more structured response where the tool is more tightly integrated into Bailo.

The system should be able to be configured globally by default to allow any collaborator to view the detailed scan results, but allow a configuration to require a 'minimum' role - either of the entry, or a global system role.

## Describe alternatives you have considered

Any output that wasn't considered an 'issue' is lost after a scan has been executed, so there is no current alternative to getting output from a successful scan.

ARADDCC002:
> This sounds like a good idea. Do you think that adding `details?: any[]` to `ScanInterface` (and its inherited types) in https://github.com/gchq/Bailo/tree/main/backend/src/models/Scan.ts which a scanner then populates would suffice? This would allow putting in `strings` if just going for raw logs and `objects` for anything structured. A simple `switch case` on `toolName` or even just the type of `details` could be used to format the `details` items when presented via the frontend.

I'm not personally a massive fan of having `any[]` there if we can avoid it. I appreciate it might be a bit restrictive, but I'd rather we decided on a good pattern and type it accordingly. Scanning isn't something I'm hugely clued up on so I'm not in a position to definitively decide on what type to use (as I don't know what's the most useful), but I think it's important to type it so that users know exactly what it's going to accept.

Surely string is OK? It would allow for users to enter markdown etc.

GCHQDeveloper36:
Agreed on trying to keep 'some' structure in place. Maybe you could namespace it by the tool name? So something like:

```
details?: ScanDetail[]
problems?: ProblemDetail[]
...
ProblemDetail: {
  name: enum ['clamav', 'modelscan', ...],
  severity: ['Critical', 'High', 'Medium', 'Low', 'Info', 'Unknown'],
  detail: string
}
...
ScanDetail: {
  name: enum ['clamav', 'modelscan', ...],
  output: string[], // raw output e.g. stdout
  [key: string]: any
}

ClamAvDetail extends ScanDetail: {
  name: 'clamav',
  viruses: ...
}
```

Example severity levels can be seen [here](https://docs.gitlab.com/user/application_security/vulnerabilities/severities/) and how they might map between different contexts.

[...]

PR #4189, opened 2026-09-07
feat: expose detailed file scan output

sylvesterkaczmarek:
##### Checklist

Backend file scan API specification; frontend file scanning results and documentation.

### Description of change

Exposes the detailed scanner output Bailo already stores for file scans and makes it available from the file scanning UI.

- Retains `additionalInfo` in the documented file scan response schema instead of stripping it from the public contract.
- Adds an expandable `Detailed scanner output` section to individual scan results when structured scanner output is available.
- Keeps the normal scan summary compact; detailed output is only shown on demand.
- Updates the file scanning documentation with the new detail view.
- Keeps the existing file-view authorisation boundary unchanged.

This addresses the detailed-output retrieval and display gap described in #2348. A separate configurable minimum-role threshold for detailed scan output is not introduced by this change.

### Validation

- Regression tests fail on `main` before the change and pass after it.
- Targeted backend file route tests: 3/3 passed.
- New frontend `ScanResultDetail` regression test passed.
- Full backend suite: 1,415/1,415 passed with `TZ=UTC`.
- Full frontend suite: 160/160 passed with `TZ=UTC`.
- Exact root `npm run lint`, `npm run build`, and `npm run test` passed.
- `npm run style` made no changes.
- Documentation style checks passed.
- `git diff --check` passed.

### DIFF
--- a/backend/src/services/specification.ts
+++ b/backend/src/services/specification.ts
@@ -204,6 +204,7 @@ export const fileWithScanInterfaceSchema = z.object({
         scannerVersion: true,
         state: true,
         summary: true,
+        additionalInfo: true,
         lastRunAt: true,
         _id: true,
         id: true,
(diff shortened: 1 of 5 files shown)


JR40159:

sylvesterkaczmarek:
Hi @JR40159, thanks for clarifying the updated guidance. I may be misunderstanding the reason for the closures.

I personally opened these pull requests, reviewed the changes before submitting them, and take responsibility for the contributions under my account. If the issue is that the PRs pre-date the updated template, I am happy to update or re-submit them as required. I just want to make sure I follow the intended process correctly.

PR #4190, opened 2026-09-07
feat: download release files as archive

sylvesterkaczmarek:
Fixes #3322.

##### Checklist

Release file downloads, file authorisation, release assets UI and user documentation.

### Description of change

Adds a bulk download option for all files attached to a release.

The new API endpoint streams release files as a `tar.gz` archive using the existing file download path rather than buffering whole files in memory. Download permission is checked for every file before streaming starts, archive entry names are flattened and de-duplicated to avoid unsafe or conflicting paths, and individual file download auditing is preserved.

The release files accordion now includes a **Download all files** action, with matching user documentation.

Regression coverage verifies archive contents, permission preflight, safe entry naming, stream failures and the frontend download link. Full validation passed with 1,419 backend tests and 160 frontend tests.


JR40159:

PR #4191, opened 2026-09-07
feat: add access request discovery

### LINKED ISSUE #2346: Discoverability of access requests
## Is your feature request related to a problem? Please describe.

As a user I can't find models I have access to already.

## Describe the solution you would like

Instead of the reviewer-centric stance of the review pages, as a consumer of models I should be able to discover the status of access requests rather than going to each individual model. In particular, I should be able to find models that I have got access to - not just access requests I've created, but access requests that I am named on the access list for (and where I am a member of a group that is named on the access list).

Updating the search filters on the marketplace might be an easy first starting place without introducing a whole new section (similar to 'my roles').

Longer term, a way of seeing the access requests and their overall status (open, in progress, completed) would be better. I can then additionally filter by metadata of the access requests (text search over 'name', date range search for 'end date', entity search for access list, etc).

## Describe alternatives you have considered

Currently users have to go to each individual model's access request tab, and figuring out where 'their' access requests are.

sylvesterkaczmarek:
##### Checklist

Access management, frontend navigation and user documentation.

### Description of change

Adds a consumer-facing **Your access** page so users can find access requests they created or that include them or one of their groups as an access entity.

The page links each request back to its model and reuses the existing access-request review status display. This also fixes two gaps in the existing global access-request search path:

- `mine=true` now includes requests created by the current user as well as requests targeting one of their entities;
- the frontend model filter now sends the API's `modelId` query key instead of `modelIds`.

Regression tests demonstrate both search defects before the fix, and the new consumer list has frontend coverage. Documentation now points users to the new navigation item.

Closes #2346.

Validation:

- backend: 1,415 tests passed
- frontend: 161 tests passed
- root `npm run check-style`, `npm run lint`, `npm run build` and `npm run test` passed
- changed files pass CSpell

### DIFF
--- a/backend/src/services/accessRequest.ts
+++ b/backend/src/services/accessRequest.ts
@@ -171,9 +171,14 @@ export async function findAccessRequests(
   }
 
   if (mine) {
-    query['metadata.overview.entities'] = {
-      $in: await authentication.getEntities(user),
-    }
+    query.$or = [
+      { createdBy: user.dn },
+      {
+        'metadata.overview.entities': {
+          $in: await authentication.getEntities(user),
+        },
+      },
+    ]
   }
 
   const stages: PipelineStage[] = [{ $match: query }]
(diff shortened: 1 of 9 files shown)


JR40159:

PR #4192, opened 2026-09-08
Send email notifications when a DA is reviewed

JRB66955:


backend/src/services/smtp/smtp.ts
```diff
@@ -169,6 +169,91 @@ export async function notifyDeploymentModelOwners(
...
+) {
+  if (!config.smtp.enabled) {
+    log.info('Not sending email due to SMTP disabled')
+    return
+  }
+
+  let emailTitle: string
+
+  if (decision === 'reject') {
+    emailTitle = `A deployment assessment you own has been rejected`
```
JR40159, line 185: It could be worth setting `actionRequired` to `true`

backend/src/services/smtp/smtp.ts
```diff
@@ -169,6 +169,91 @@ export async function notifyDeploymentModelOwners(
...
+  } else if (decision === 'request_changes') {
+    emailTitle = `A deployment assessment you own has had changes requested`
```
JR40159, line 187: It could be worth setting `actionRequired` to `true`

backend/src/services/smtp/smtp.ts
```diff
@@ -169,6 +169,91 @@ export async function notifyDeploymentModelOwners(
...
+  } else {
+    emailTitle = `A deployment assessment has been reviewed with decision: ${decision}`
+  }
+
+  const emailContent = buildEmail(
+    `${emailTitle}`,
```
JR40159:
```suggestion
    emailTitle,
```

backend/src/services/smtp/smtp.ts
```diff
@@ -169,6 +169,91 @@ export async function notifyDeploymentModelOwners(
...
+      {
+        name: 'Open Deployment Assessment',
+        url: `${appBaseUrl}/deployment-assessments/${encodeURIComponent(deployment.id)}`,
+      },
+      { name: 'View Deployment Assessments', url: `${appBaseUrl}/deployment-assessments` },
+    ],
+    false,
+  )
+
+  await dispatchEmail([toEntity('user', deployment.createdBy)], await emailContent)
```
JR40159:
```suggestion
  const emailContent = await buildEmail(
    `${emailTitle}`,
    [
      { title: 'Deployment Assessment Name', data: deployment.name },
      { title: 'Deployment Assessment ID', data: deployment.id },
      {
        title: 'Reviewed By',
        data:
          (await authentication.getUserInformation(toEntity('user', assessmentReviewer))).name || assessmentReviewer,
      },
    ],
    [
      {
        name: 'Open Deployment Assessment',
        url: `${appBaseUrl}/deployment-assessments/${encodeURIComponent(deployment.id)}`,
      },
      { name: 'View Deployment Assessments', url: `${appBaseUrl}/deployment-assessments` },
    ],
    false,
  )

  await dispatchEmail([toEntity('user', deployment.createdBy)], emailContent)
```

JR40159: I have no idea why this is a convention in this file. It's not a big deal but it gets me every time

backend/src/services/smtp/smtp.ts
```diff
@@ -169,6 +169,91 @@ export async function notifyDeploymentModelOwners(
   await dispatchEmail(entities, await emailContent)
 }
 
+export async function notifyDeploymentAssessmentCreator(
```
JR40159:
```suggestion
export async function notifyReviewResponseForDeploymentAssessment(
```

backend/src/services/smtp/smtp.ts
```diff
@@ -169,6 +169,91 @@ export async function notifyDeploymentModelOwners(
...
+      },
+    ],
+    [
+      {
+        name: 'Open Deployment Assessment',
+        url: `${appBaseUrl}/deployment-assessments/${encodeURIComponent(deployment.id)}`,
+      },
+      { name: 'View Deployment Assessments', url: `${appBaseUrl}/deployment-assessments` },
+    ],
+    false,
```
JR40159: In previous comments, I thought this should be conditionally set for this method but maybe this should always be true based on where the function is called?

backend/src/services/smtp/smtp.ts
```diff
@@ -169,6 +169,91 @@ export async function notifyDeploymentModelOwners(
   await dispatchEmail(entities, await emailContent)
 }
 
+export async function notifyDeploymentAssessmentCreator(
+  deployment: DeploymentAssessmentInterface,
+  decision: string,
```
JR40159: Could use the type here to ensure this method is only used for Reject and RequestChanges decisions

JR40159 (changes requested):

backend/src/services/smtp/smtp.ts
JRB66955: I've set the `actionRequired` boolean within the `buildEmail` section instead of within this if/else block.

backend/src/services/smtp/smtp.ts
JRB66955: I've set the `actionRequired` boolean within the `buildEmail` section instead of within this if/else block.

backend/src/services/smtp/smtp.ts
JR40159:
```suggestion
export async function notifyRiskOwnerOfDeploymentAssessment(
```

backend/src/services/smtp/smtp.ts
```diff
@@ -125,7 +125,7 @@ export async function notifyDeploymentRiskOwner(
     true,
   )
 
-  await dispatchEmail([riskOwner], await emailContent)
+  await dispatchEmail([riskOwner], emailContent)
 }
 
 export async function notifyDeploymentModelOwners(
```
JR40159:
```suggestion
export async function notifyModelOwnersOfDeploymentAssessment(
```

backend/src/services/smtp/smtp.ts
```diff
@@ -166,7 +166,92 @@ export async function notifyDeploymentModelOwners(
...
+      },
+      { name: 'View Deployment Assessments', url: `${appBaseUrl}/deployment-assessments` },
+    ],
+    true,
+  )
+
+  await dispatchEmail([toEntity('user', deployment.createdBy)], emailContent)
+}
+
+export async function notifyModelDevelopers(
```
JR40159:
```suggestion
export async function notifyModelOwnersOfDeploymentApproval(
```

backend/src/services/deploymentAssessment.ts
```diff
@@ -324,9 +329,63 @@ export async function reviewDeploymentAssessment(
...
+
   return response
 }
 
+async function notifyDeploymentAssessmentReviewed(
+  deploymentAssessment: DeploymentAssessmentInterface,
+  decision: Exclude<DecisionKeys, 'undo'>,
+): Promise<void> {
+  try {
+    const assessmentReviewer = deploymentAssessment.metadata.overview?.riskOwner?.toString() ?? ''
```
JR40159: `riskOwner` is an array (it should probably be called `riskOwners` now) and as such there can be multiple reviewers of a DA. `notifyDeploymentAssessmentReviewed` should probably take in the user who has reviewed the DA

backend/src/services/deploymentAssessment.ts
```diff
@@ -324,9 +329,63 @@ export async function reviewDeploymentAssessment(
...
+        deploymentAssessmentId: deploymentAssessment.id,
+        decision,
+      },
+      'Failed to send deployment assessment review notification',
+    )
+  }
+}
+
+async function getModelDevelopers(deploymentAssessment: DeploymentAssessmentInterface) {
+  const models = await ModelModel.find({
```
JR40159: It could be worth adding a `getModelsByIds` function to the model service that takes an array of ids.

backend/src/services/deploymentAssessment.ts
```diff
@@ -324,9 +329,63 @@ export async function reviewDeploymentAssessment(
...
+    id: { $in: deploymentAssessment.metadata.overview?.modelIds ?? [] },
+  }).lean()
+
+  const modelDevelopers = models.map((model) => ({
```
JR40159, line 385: This looks very similar to `getRoleEntities`

JR40159 (changes requested):

backend/src/services/deploymentAssessment.ts
JRB66955: I've added a `getModelsByIds` and a `getModelsByIdsNoAuth` to have something similar along with `getModelById/getModelByIdNoAuth` - with some basic tests as well.

sylvesterkaczmarek (commented):
`assessmentReviewer` is already an entity string (`toEntity('user', user.dn)`), but `notifyReviewResponseForDeploymentAssessment()` wraps it with `toEntity('user', ...)` again before calling `getUserInformation`. That produces values like `user:user:alice`, so reviewer lookup fails and the notification path is caught/logged instead of sending. Could we pass `assessmentReviewer` directly here and add a regression that exercises the real entity format?

sylvesterkaczmarek (approved):
Rechecked `76f5902c`. The deployment-assessment reviewer is now passed through as the existing entity string and consumed directly by `getUserInformation`, so the double `user:` prefix failure I raised is removed. The corresponding tests exercise that entity form and CI is green. No remaining blocker from me.

backend/src/services/smtp/smtp.ts
```diff
@@ -166,7 +166,91 @@ export async function notifyDeploymentModelOwners(
...
+      },
+    ],
+    [
+      {
+        name: 'Open Deployment Assessment',
+        url: `${appBaseUrl}/deployment-assessments/${encodeURIComponent(deployment.id)}`,
+      },
+      { name: 'View Deployment Assessments', url: `${appBaseUrl}/deployment-assessments` },
+    ],
+    true,
```
GB907762, line 209: Don't rely on this but possibly could use `needsUserAction` to evaluate this?

backend/src/services/deploymentAssessment.ts
```diff
@@ -314,19 +324,72 @@ export async function reviewDeploymentAssessment(
...
+      case Decision.Approve: {
+        const modelDevelopers = await getModelDevelopers(deploymentAssessment)
+        for (const { model, developers } of modelDevelopers) {
+          await notifyModelOwnersOfDeploymentApproval(
+            developers,
+            deploymentAssessment,
+            model,
+            deploymentAssessment.createdBy,
+          )
+        }
```
GB907762, line 365: Could you wrap this in a `Promise.all` for efficiency and redundancy on uncaught errors.

backend/src/services/model.ts
```diff
@@ -958,3 +958,38 @@ export function getRoleEntities<T extends string>(
...
+  kinds?: EntryKindKeys[] | EntryKindKeys,
+): Promise<ModelDoc[]> {
+  const models = await getModelsByIdsNoAuth(modelIds, kinds)
+
+  for (const model of models) {
+    const auth = await authorisation.model(user, model, ModelAction.View)
+    if (!auth.success) {
+      throw Forbidden(auth.info, { userDn: user.dn, modelId: model.id })
+    }
+  }
```
GB907762, line 974: To avoid a loop, could you use `authorisation.models`?

backend/src/services/model.ts
```diff
@@ -958,3 +958,38 @@ export function getRoleEntities<T extends string>(
...
+  const kindArray = kinds ? (Array.isArray(kinds) ? kinds : [kinds]) : undefined
+
+  const models = await ModelModel.find({
+    id: { $in: modelIds },
+    ...(kindArray && { kind: { $in: kindArray } }),
+  })
+
+  if (models.length === 0) {
+    throw NotFound('The requested entries were not found.', { modelIds })
+  }
```
GB907762, line 992: It looks like this indicates all models that are requested?

Which means this could attempt to owners on models that don't exist. In this specific case one out of 3 could fail and cause bad behaviour. Instead could you find the compliment of the found models and throw them in this error context?

GB907762 (changes requested):

backend/src/services/model.ts
JRB66955: I've taken a stab at replacing this NotFound check with something more fancy, with some filtering to extract out the list of modelIds not found - worth checking if this is fine?

sylvesterkaczmarek (changes requested):
Re-reviewed current `8816c02a`. The missing-ID complement check itself looks sound, but this head is not ready yet. The new model-service tests are still calling `getModelsByIdsNoAuth` / `getModelsByIds` with object placeholders rather than `string[]`, which is why the unit job currently fails with `modelIds.filter is not a function` / `models.map is not a function`. More importantly, the switch to `authorisation.models()` also changes `getModelsByIds()` from all-or-nothing authorization to silently filtering unauthorized entries. The previous implementation and the existing bad-authorisation test expect a `Forbidden` failure. Please keep the batched authorisation call but preserve that behavior when any requested model is unauthorized, and update the tests to pass real ID arrays and mock `authorisation.models`; a partial-missing-ID regression would also pin the new complement logic.

backend/src/services/smtp/smtp.ts
JRB66955: Not sure this is necessary at the moment given I think we want this to be true for both decision states this function covers (rejected/changes requested) - could revisit in future but unsure we need it specifically at this time.

JR40159 (dismissed):

sylvesterkaczmarek (changes requested):
Re-reviewed current `be414b6d`. The ID-array test failures from my previous review are fixed and CI is green, but the authorization behavior change remains. `getModelsByIds()` still returns `models.filter((_, i) => auths[i].success)`, and the new test now explicitly expects an unauthorized requested model to be silently omitted. That changes the previous all-or-nothing contract: callers that requested N concrete model IDs can now receive fewer models without any `Forbidden` error. Please keep the batched `authorisation.models()` call, but fail when any requested model is unauthorized rather than filtering it out, and pin that behavior with the regression.

GB907762 (dismissed):

JR40159 (approved):

sylvesterkaczmarek (changes requested):
Rechecked current `8aaed388`. My authorization blocker is still present: `getModelsByIds()` still returns `models.filter((_, i) => auths[i].success)`, so an explicitly requested unauthorized model is silently omitted rather than preserving the previous all-or-nothing `Forbidden` behavior. The latest change is a merge from `v3` and does not alter this path. Please fail when any batched authorization result is unsuccessful and keep that contract pinned with the regression.

PE39806 (approved):

sylvesterkaczmarek:
This has merged. I opened #4247 to keep the `getModelsByIds()` partial-authorization behavior tracked separately rather than leaving the review concern buried in the merged PR.

PR #4208, opened 2026-09-09
HTML export for deployment assessments

ARADDCC002:


backend/src/services/export.ts
```diff
@@ -0,0 +1,157 @@
...
+      } else {
+        output += outdent`\n\n
+            ${obj}
+        `
+      }
+      break
+    default:
+      throw new Error(
+        `One of the types within this schema has not been implemented in the export method.  Received type ${(schema as any).type}`,
+      )
```
GB907762, line 153: Does this need a Bailo error to return to the user?

backend/src/services/export.ts
```diff
@@ -0,0 +1,157 @@
...
+        output += outdent`\n\n
+          ${'#'.repeat(depth + 1)} Entry #${count + 1}
+        `
+        output = recursiveRender(value, schema.items, output, depth + 1)
+      }
+      break
+    }
+    case 'number':
+      // We can add a description like this, but I felt it overkill:
+      // ${schema.description ? `> ${schema.description}` : ''}
```
GB907762, line 102:
```suggestion
```
Agreed don't use a quote block.

backend/src/routes/v3/deploymentAssessment/getDeploymentAssessmentHtml.ts
```diff
@@ -0,0 +1,50 @@
...
+      content: {
+        'application/html': {
+          schema: {
+            type: 'string',
+          },
+        },
+      },
+    },
+  },
+})
```
GB907762, line 36:
```suggestion
registerPath(
  {
    method: 'get',
    path: '/api/v3/deployment-assessments/{deploymentAssessmentId}/html',
    tags: ['deployment assessments'],
    description: 'Get a specific version of a deployment assessment as HTML.',
    schema: getDeploymentAssessmentHtmlSchema,
    responses: {
      200: {
        description: 'Deployment assessment HTML.',
        content: {
          'application/html': {
            schema: {
              type: 'string',
            },
          },
        },
      },
    },
  },
  'v3',
)
```
- Use v3
- `model card` -> `deployment assessment`
- Use consistent tagging

GB907762 (dismissed):

sylvesterkaczmarek (commented):
The new deployment-assessment export feeds assessment metadata through Showdown and then inserts the result into `{{{body}}}`. Showdown does not sanitize raw HTML or unsafe URL schemes by default, so author-supplied markup such as `` or `[link](javascript:...)` survives into the served HTML document. Could we sanitize the converted body before templating (or otherwise enable the project’s untrusted-input hardening) and add an export regression for raw HTML/unsafe links?

backend/src/services/deploymentAssessmentExport.ts
JR40159, line 1: Is it worth combining this with the `modelCardExport` so that both functions to export the cards to html are in one file?

backend/src/services/export.ts
JR40159, line 1: The name `export` is probably too generic considering that Bailo supports different types of export. Also maybe these functions are more suitable to a utils file rather than a service?

backend/src/services/deploymentAssessmentExport.ts
```diff
@@ -0,0 +1,24 @@
...
+  const schema = await getSchemaById(deploymentAssessment.schemaId)
+
+  let output = outdent`
+      # ${deploymentAssessment.name}\n
+    `
+  output = recursiveRender(deploymentAssessment.metadata, schema.jsonSchema as Fragment, output)
+  const converter = new showdown.Converter()
+  converter.setFlavor('github')
+  const body = converter.makeHtml(output)
+  const sanitizedBody = sanitizeHtml(body)
```
JR40159: Could this also be added to the model card equivalent?

JR40159 (changes requested):

backend/src/services/export.ts
JR40159: Could we call this service `htmlExport` to denote the specific type of export?

JR40159 (changes requested):

JR40159 (dismissed):

sylvesterkaczmarek (approved):
Re-reviewed current `b48a2af5`. The export-sanitization issue I raised is resolved: both model-card and deployment-assessment Markdown are converted with Showdown, then passed through `sanitizeHtml()` before insertion into the Handlebars template. That removes the raw HTML/event-handler and unsafe-link path I was concerned about. CI is green. No remaining blocker from me.

backend/src/routes/v3/deploymentAssessment/getDeploymentAssessmentHtml.ts
PE39806, line 1: Add a new test file for this file.

backend/src/utils/export.ts
PE39806, line 1: This should also have some tests.

PE39806 (changes requested):

PE39806 (approved):

sylvesterkaczmarek (approved):
Re-reviewed current `bfc6e12d`. The sanitization path I previously checked is unchanged: generated Markdown is converted and passed through `sanitizeHtml()` before templating. The new commit adds route and export utility coverage, and the current CI suite is green. No remaining blocker from me.

JR40159 (approved):

PR #4214, opened 2026-09-10
Fix Python downloads with reserved filename characters

### LINKED ISSUE #4213: Python downloads misinterpret reserved characters in filenames
Reproduced on `79cf11cea0` in the Python client.

`Client.get_download_by_filename("test-model", "1.0.0", "model#final.bin")` interpolates the filename directly into the request URL. `#` starts a fragment, `?` starts a query string, and a literal percent escape such as `%23` is interpreted as an escape rather than part of the filename. The resulting request does not address the intended download route/file.

This affects both standard agents and `TokenAgent`, including calls through `Release.download()`.

Expected: encode the filename as one URL path segment before passing it to the agent, without changing download streaming, timeout, or authentication.

I have regression tests using real Requests request preparation and requests-mock. Six reserved-character cases fail on the current code; plain, space-containing, and Unicode filenames provide six passing controls. I am preparing the fix.

sylvesterkaczmarek:
## Summary
Fixes #4213.

Percent-encode download filenames as a single path segment before constructing the URL. This preserves literal `#`, `?`, and percent escapes for both standard and token authentication, including `Release.download()`.

Streaming, timeout, and authentication behaviour are unchanged. No dependencies added.

## Validation
- 12 request-level regression cases, including reserved characters, plain names, spaces, and Unicode. Six cases failed before the fix.
- Full Python unit suite: **139 passed**; integration/MLflow tests remain deselected by the existing configuration.
- Ruff lint and formatting checks passed.

### DIFF
--- a/lib/python/src/bailo/core/client.py
+++ b/lib/python/src/bailo/core/client.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from io import BytesIO
 from json import JSONDecodeError
 from typing import Any
+from urllib.parse import quote
 
 import requests
 
@@ -527,6 +528,7 @@ class Client:
         :param filename: The filename trying to download from
         :return: Response object
         """
+        filename = quote(filename, safe="")
         if isinstance(self.agent, TokenAgent):
             return self.agent.get(
                 f"{self.url}/v2/token/model/{model_id}/release/{semver}/file/{filename}/download",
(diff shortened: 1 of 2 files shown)


PE39806 (approved):

GB907762 (approved):

sylvesterkaczmarek:
Thanks for reviewing and merging this.

PR #4217, opened 2026-09-10
Rollback entity selector

GB907762:
### Description of change
- Rolled back entity selector 
- Update swim lanes to use multiple DROs

### Checklist

### DIFF
--- a/frontend/src/MuiForms/EntitySelector.tsx
+++ b/frontend/src/MuiForms/EntitySelector.tsx
@@ -1,64 +1,37 @@
 import { Box, Chip, Stack, Typography } from '@mui/material'
+import Autocomplete from '@mui/material/Autocomplete'
 import { useTheme } from '@mui/material/styles'
+import TextField from '@mui/material/TextField'
 import { Registry, RJSFSchema } from '@rjsf/utils'
-import { useCallback, useEffect, useMemo, useState } from 'react'
+import { debounce } from 'lodash-es'
+import { KeyboardEvent, SyntheticEvent, useCallback, useMemo, useState } from 'react'
 import CompareField from 'src/common/CompareField'
-import EntityAutocomplete from 'src/common/EntityAutocomplete'
 import InlineDiff from 'src/common/InlineDiff'
 import UserDisplay from 'src/common/UserDisplay'
 import getCompareFieldState from 'src/hooks/useCompareField'
 import { EntityObject } from 'types/types'
 
-import { useGetCurrentUser } from '../../actions/user'
+import { useGetCurrentUser, useListEntities } from '../../actions/user'
 import Loading from '../common/Loading'
 import MessageAlert from '../MessageAlert'
 
 interface EntitySelectorProps {
   label?: string
   required?: boolean
-  value: Array<string | null | undefined> | string | null | undefined
-  onChange: (newValue: string[] | string) => void
+  value: string[]
+  onChange: (newValue: string[]) => void
   registry?: Registry
   rawErrors?: string[]
   id: string
   schema: RJSFSchema
 }
 
-function isEmptyEntry(entry: string | null | undefined): boolean {
-  return entry === null || entry === undefined || entry === ''
-}
-
-export function getEntitySelectorValue(
-  selectedEntities: EntityObject[] | EntityObject | null | undefined,
-  isMultiple: boolean,
-): string[] | string {
-  if (isMultiple) {
-    const entities = (Array.isArray(selectedEntities) ? selectedEntities : [selectedEntities]).filter(
-      (entity): entity is EntityObject => entity !== null && entity !== undefined,
-    )
-    return entities.map((entity) => `${entity.kind}:${entity.id}`)
-  }
-
-  if (!selectedEntities) {
+function formatEntityValue(entities: string[] | undefined): string {
+  if (!entities || entities.length === 0) {
     return ''
   }
-
-  const entity = Array.isArray(selectedEntities) ? selectedEntities[0] : selectedEntities
-  if (!entity) {
-    return ''
-  }
-
-  return `${entity.kind}:${entity.id}`
-}
-
-function formatEntityValue(value: string[] | string | undefined): string {
-  if (!value || value.length === 0) {
-    return ''
-  }
-
-  const entities = Array.isArray(value) ? value : [value]
-
   return entities
+    .filter(Boolean)
     .map((entity) => {
       const [, entityId] = entity.split(':')
       return entityId ?? entity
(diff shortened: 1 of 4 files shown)


sylvesterkaczmarek (commented):
When an assessment has multiple risk owners, `assessment.owner?.join(", ")` produces one combined entity string and passes it to `UserDisplay`. `UserDisplay` resolves a single entity by splitting on `:`, so e.g. `user:alice, user:bob` is looked up as `alice, user` and the owners render incorrectly. Could we render one `UserDisplay` per owner instead of joining the entity IDs first?

JRB66955 (changes requested):
Don't seem to be able to submit a DA with multiple risk owners due to the schema - could we change the schema to accommodate this change? (i.e. remove `maxItems: 1,` and re-word the question from `Who is the risk owner attached....` to `Who are the risk owner(s) attached....`)

GB907762:
> Don't seem to be able to submit a DA with multiple risk owners due to the schema - could we change the schema to accommodate this change? (i.e. remove `maxItems: 1,` and re-word the question from `Who is the risk owner attached....` to `Who are the risk owner(s) attached....`)

I think for now this is purely accommodating the default schema. You can test by changing `maxItems` if you wish. In future we plan to pull out the core fields so they can be configured.

JR40159 (changes requested):
For clarity references to `riskOwner` in the repository should be renamed to `riskOwners`

JR40159 (changes requested):
https://github.com/gchq/Bailo/blob/af2544eb9c3637b6eddb6ce8330c2c22f2a0f05e/backend/src/routes/v3/deploymentAssessment/getDeploymentAssessments.ts#L20 Perhaps should be updated to be an array as with modelIds

GB907762:
So this is for retrieving DAs with a specific risk Owner. I agree. Although I don't think this pertains to any breaking behaviour to using multiple DROs. So I'm going to descope and do as a separate change

JR40159 (dismissed):

ARADDCC002 (dismissed):
Approved, but my only comment would be if we can change the wording of the DRO selection question to be:

`Who are the risk owners attached to this deployment assessment?`

It's pretty minor, though so happy for you to ignore.

GB907762:
> Who are the risk owners attached to this deployment assessment?

This is a very good spot. I was cautious of doing it here. Since the in built schema uses only one risk owner, and it wouldn't make semantic sense. And would be changed with a different schema.

sylvesterkaczmarek (commented):
Re-reviewed current `dd816a0e`. The multi-owner rendering issue I raised is resolved: `UserRow` now renders one `UserDisplay` per owner instead of joining entity IDs into one invalid value. There is still one failing unit test on this head, but the failure is a stale expectation in `test/services/schema.spec.ts` that still expects `riskOwner`; runtime/schema output now correctly uses `riskOwners`. Please update that test expectation so the unit suite is green. My original rendering blocker is otherwise resolved.

backend/src/routes/v3/deploymentAssessment/getDeploymentAssessments.ts
PE39806, line 20: Should this be updated to `riskOwners: coerceArray(z.array(z.string().min(1))).optional()`, and the correcsponding search updated?

backend/config/default.cjs
```diff
@@ -278,7 +278,7 @@ module.exports = {
       owner: 'Owner',
       contributor: 'Contributor',
       consumer: 'Consumer',
-      riskOwner: 'Deployment Risk Owner',
+      riskOwners: 'Deployment Risk Owner',
```
PE39806: Update `backend/test/services/roles.spec.ts:31`, `backend/test/services/schema.spec.ts:30`, `frontend/types/types.ts:79` and all inherited uses (e.g. `frontend/pages/deployment-assessments/index.tsx:39`) in line with this change.

PE39806 (changes requested):

backend/src/routes/v3/deploymentAssessment/getDeploymentAssessments.ts
GB907762: Yeah there's a couple of comments on this question.
https://github.com/gchq/Bailo/pull/4217#issuecomment-5665378921
From my perspective this PR is purely to allow the users to update and add multiple risk owners?

backend/config/default.cjs
GB907762: I think I made a mistake committing this. it pertains to the name you get on the frontend so this is configurable. I'll role this back.

sylvesterkaczmarek (approved):
Re-reviewed current `920f4506`. The multi-owner rendering issue I raised remains resolved, and the stale schema expectation from my last review now uses `riskOwners`; the full current CI suite is green. The remaining discussion about whether the search filter should accept multiple risk owners is separate from my earlier blocker. No remaining blocker from me.

backend/src/services/schema.ts
```diff
@@ -70,7 +70,7 @@ function prefixDeploymentAssessmentWithSummary(jsonSchema: JsonSchema) {
         title: 'Details',
         type: 'object',
         properties: {
-          riskOwner: {
+          riskOwners: {
             title: 'Who is the risk owner attached to this deployment assessment?',
```
PE39806, line 74:
```suggestion
            title: 'Who are the risk owners attached to this deployment assessment?',
```

backend/test/services/schema.spec.ts
```diff
@@ -132,10 +132,11 @@ describe('services > schema', () => {
...
     expect(result[0].jsonSchema.properties).toEqual({
       overview: expect.objectContaining({
+        additionalProperties: false,
         title: 'Details',
-        required: ['riskOwner', 'modelIds'],
+        required: ['riskOwners', 'modelIds'],
         properties: expect.objectContaining({
-          riskOwner: expect.objectContaining({
+          riskOwners: expect.objectContaining({
             title: 'Who is the risk owner attached to this deployment assessment?',
```
PE39806, line 140:
```suggestion
            title: 'Who are the risk owners attached to this deployment assessment?',
```

PE39806 (changes requested):

PE39806 (approved):

backend/src/services/schema.ts
PE39806: Ignore as already discussed above.

backend/test/services/schema.spec.ts
PE39806: Ignore as already discussed above.

frontend/src/MuiForms/EntitySelector.tsx
```diff
@@ -167,39 +122,74 @@ export default function EntitySelector({
...
+        <InlineDiff from={formatEntityValue(compare.compareFromState)} to={currentValueString} />
       ) : (
-        <Typography component='span' sx={{ fontStyle: 'italic', color: theme.palette.customTextInput.main }}>
-          None
-        </Typography>
+        currentValue.length > 0 && (
+          <Box sx={{ overflowX: 'auto', p: 1 }}>
+            <Stack spacing={1} direction='row'>
+              {currentValue.map((entity) => (
+                <Chip label={<UserDisplay dn={entity} />} key={entity} sx={{ width: 'fit-content' }} />
```
IR96334, line 188: Currently getting bug where the currentValue is [undefined], when on DA edit page. This means the entity value supplied to UserDisplay is undefined. There is a fix that deals with empty entityselector content that could fix this if you merge in v3.

sylvesterkaczmarek (approved):
Rechecked current 79f882d7. The later EntitySelector fix addresses the edit/view regression raised after my previous approval: undefined entries are filtered before UserDisplay rendering, the empty state renders as Unanswered, and the regression covers a mixed undefined/valid value. My earlier multi-owner concern remains resolved on the current head.

PE39806 (approved):

ON93754 (approved):

PR #4221, opened 2026-09-10
Create linked access requests for multiple models

### LINKED ISSUE #2347: Create access requests for multiple models
## Is your feature request related to a problem? Please describe.

If I want to get access to many models, I have to create many access requests where the content is likely otherwise the same.

## Describe the solution you would like

A way in the UI to create an access request that covers multiple entries. Behind the scenes, these should still be multiple access requests with the appropriate reviewers, but there should be a way of linking them together for discoverability. Reviewers should be aware that an access request is part of a 'grouped' request. Reviewers & submitters should be able to get a list of all access requests in that 'grouped' request. Long term they should be able to respond to multiple requests in bulk, but for now individual responses is fine.

## Describe alternatives you have considered

API-based approach isn't suitable for many users

sylvesterkaczmarek:
### Description of change

Fixes #2347.

Create linked access requests for up to 20 local models from one form. Each model keeps its own authorisation checks, request, reviewers and decision. Request and review pages link to other group members the viewer can access.

Creation returns explicit per-model results. Successful requests remain saved when another target fails; the UI shows both outcomes and prevents resubmitting the whole group. Existing single-model requests are unchanged.

### Validation

1,443 backend tests and 172 frontend tests pass, including 42 new service, API and UI cases. Both production builds, TypeScript, changed-file ESLint/Prettier, documentation style and OpenAPI generation pass. Docker, Kubernetes, live MongoDB and Cypress integration checks were not run.

### Checklist

### DIFF
--- a/backend/src/models/AccessRequest.ts
+++ b/backend/src/models/AccessRequest.ts
@@ -20,6 +20,7 @@ export interface AccessRequestMetadata {
 export interface AccessRequestInterface {
   id: string
   modelId: string
+  groupId?: string
 
   schemaId: string
   metadata: AccessRequestMetadata
@@ -41,6 +42,7 @@ const AccessRequestSchema = new Schema<AccessRequestDoc>(
   {
     id: { type: String, unique: true, required: true },
     modelId: { type: String, required: true },
+    groupId: { type: String },
 
     schemaId: { type: String, required: true },
     metadata: { type: Schema.Types.Mixed, required: true },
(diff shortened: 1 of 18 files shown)


sylvesterkaczmarek:
Rebased cleanly onto current main and rebuilt as one SSH-verified commit. Focused backend regressions pass (29 tests across 2 files), focused frontend regressions pass (13 tests), and full lint plus production builds pass on the current dependency graph. The PR is now 0 commits behind and mergeable. Fresh workflows are awaiting maintainer approval.

sylvesterkaczmarek:
Refreshed again after main advanced. Head is now SSH-verified 5342f4bb45, 0 behind and mergeable. Fresh validation in the real checkout passes: 29/29 focused backend tests, 13/13 focused frontend tests, full lint, and both production builds. A transient 404 seen once in a parallel worktree run was not reproducible; the exact partial-failure case returns 200 with the expected body, and the full focused route/service set passes cleanly. Ready for review.

PR #4226, opened 2026-09-14
Add nullable deployable state

GB907762:
### Description of change

Allow for deployment state enforcement to be disabled.

### Checklist


ARADDCC002 (approved):

PE39806 (approved):
I am happy with this as-is.
In the future, I think it would be worth us switching to `deployableModelStates: string[]` to allow multiple states to be chosen (logical OR). The empty array `[]` would be equivalent to this newly handled `null` state.

PR #4227, opened 2026-09-15
Standardise test config mock

PE39806:
### Description of change

`backend/test` had three competing ways to mock `utils/config.ts`:

1. The global default mock
2. Per-file `vi.mock` factories that replaced it wholesale
3. Ad-hoc `vi.spyOn`/direct mutation

Each per-file mock restated values the default already had and drifted independently. This masked the following bugs:

- `clients/artefactScan.spec.ts` mocked `artefactscan.enabled`/`connection.host` (keys not on `Config`) so its snapshot had `"undefined://undefined:undefined/info"` baked in. Corrected to `"http://127.0.0.1:8081/info"` (only snapshot changed)
- `routes/registry/registryAuth.spec.ts` - 11 tests passed only because `req.query.service` and `config.registry.service` were both `undefined`. Fixed, plus a new explicit missing-`service` test
- `clients/stroom.spec.ts` had a `vi.mock` with a wrong relative path that did nothing

These mixed approached have been replaced with a unified approach:

- `setTestConfig(overrides)` in `test/testUtils/setupTestConfig.ts` - deep-merges a `PartialDeep<Config>` into the globally mocked config, reverted automatically after every test. Registered via a `vi.mock` factory so the reference survives `vi.resetModules()`
- `src/utils/__mocks__/config.ts` extended to cover every top-level `Config` key, tracking `config/default.cjs` except where a test fixture needs otherwise (marked inline)
- ESLint guard (`no-restricted-syntax`, `test/**`) bans per-file mocking of `utils/config.js` and spying on config properties, pointing at `setTestConfig()`
- Migrated all affected test files
- Updated `AGENTS.md`

### Checklist


backend/src/utils/__mocks__/config.ts
```diff
@@ -1,15 +1,34 @@
 import { PartialDeep } from '../../types/types.js'
 import { Config } from '../config.js'
 
+/**
+ * Default config for every test, applied by `test/testUtils/setupTestConfig.ts`.
+ * Values track `config/default.cjs` except where marked `Fixture`. Keep every top-level key of
+ * `Config` present, as the per-test revert only restores keys listed here.
+ */
```
GB907762: Would it be more appropiate to do a merge with `default.cjs` here?

I can see this may cause regressions?

backend/test/testUtils/setupTestConfig.ts
```diff
@@ -1,3 +1,23 @@
...
+import type { Config } from '../../src/utils/config.js'
+
+// A factory keeps this reference stable across `vi.resetModules()`
+const testConfig = cloneDeep(defaultTestConfig)
+
+vi.mock('../../src/utils/config.js', () => ({ __esModule: true, default: testConfig }))
+
+/** Merges overrides into the mocked config, reverted after each test. Assign directly to unset a value. */
+export function setTestConfig(overrides: PartialDeep<Config>) {
+  mergeWith(testConfig, overrides, (_target, source) => (Array.isArray(source) ? source : undefined))
```
GB907762: Might it be safer to use node configs merge utils rather than lodash. I think the customiser might be flakey in some sanarios?

GB907762 (changes requested):

backend/src/utils/__mocks__/config.ts
PE39806: Good point. Our current mock config does diverge from `default.cjs` in a few ways which are generally just to minimise longer strings etc. so this change will introduce several changes in tests & snapshots, but I agree is worthwhile.

JRB66955 (approved):

GB907762 (approved):

PR #4239, opened 2026-09-15
Highlight empty required fields for draft DA

PE39806:
### Description of change

As a draft DA owner, I want to clearly see which DA fields are missing so that I can understand what needs to be added so that I can publish the DA.

When clicking "Publish" for a draft DA, a frontend check is performed on the entered data. Missing required fields are then highlighted. The validation warning is subsequently hidden when the required values are populated.

Note that, as a happy accident, these changes also affect the display of missing fields for Access Request creation, so ensure there are no regressions there. Previously the missing fields were not highlighted in an AR when trying to create a new AR. This PR does not introduce any changes when editing an AR.

### Checklist


frontend/utils/formUtils.ts
```diff
@@ -205,29 +205,65 @@ export function setStepsData(
...
+export function validateForm(step: StepNoRender) {
+  return validateStep(step).errors.length === 0
+}
+
+/** Maps the dotted path of each field failing validation, e.g. `overview.name`, to its message. */
+export function getInvalidFields(step: StepNoRender): Map<string, string> {
+  return new Map(
+    validateStep(step).errors.map(({ path, name, argument }) => [
+      (name === 'required' ? [...path, argument] : path).join('.'),
+      name === 'required' ? 'This field is required' : 'This field is incomplete',
```
GB907762: Because this fails validation it doesn't automatically indicates that it's incomplete. For example you could be entering a string when a number is required ect.

frontend/pages/deployment-assessments/new.tsx
```diff
@@ -90,7 +90,7 @@ export default function NewDeploymentAssessment() {
       const isValid = validateForm(step)
 
       if (!isValid) {
-        setErrorText('Please make sure that all sections have been completed.')
+        setErrorText('Please make sure that all required sections have been completed.')
```
GB907762: This denotes a validation error. Could be things like types being wrong. I'm not sure this fits?

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
```diff
@@ -35,13 +38,41 @@ export default function DeploymentAssessment() {
...
+    }
+
+    return getStepsFromSchema(schema, {}, [], deploymentAssessment?.metadata).every(validateForm)
+  }, [schema, deploymentAssessment?.metadata])
+
+  useEffect(() => {
+    if (isPublishable) {
+      setShowValidationErrors(false)
+    }
+  }, [isPublishable])
```
Comment (GB907762) [inline frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:55]: This seems to look to prevent the user reconfirming to publish a draft?

Is this expected

GB907762 (changes requested):

frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx
Comment (PE39806) [reply frontend/pages/deployment-assessments/[deploymentAssessmentId]/index.tsx:55]: I don't quite follow? If a draft DA initially fails validation, then this code will block the popup which does prevent publishing. If the DA is updated then `isPublishable` is recomputed (`isPublishable`'s `useMemo` is triggered by dep `deploymentAssessment?.metadata`) and, if `isPublishable`, will "unlock" the popup which allows publishing.

sylvesterkaczmarek (commented):
`withoutClearedAnswers()` treats two equivalent empty-array states differently. `[]` returns `undefined`, so a required array is absent and fails validation, but `[""]` is mapped and filtered to `[]` and is not collapsed again; a required array with no `minItems` then passes. That means adding an item and clearing it can make an otherwise empty required field publishable. Please collapse the filtered array to `undefined` when it ends up empty, and add a regression for a cleared array item.

sylvesterkaczmarek (approved):
Rechecked current `b8034be8`. The cleared-array case is fixed: arrays are collapsed to `undefined` after cleared items are removed, so `[]`, `[""]`, and `["", ""]` all validate as unanswered when the field is required. The regression covers those cases directly, and CI is green. My earlier blocker is resolved.

JRB66955 (approved):

sylvesterkaczmarek (approved):
Rechecked current 5b699332 after the RJSF/AJV validation refactor. The required-array edge case from my earlier review remains covered: the custom validator treats cleared arrays as unanswered, and the regressions explicitly reject [], [""], and ["", ""] while accepting a real answered item. CI is green. No remaining blocker from my review.

JRB66955 (approved):

PR #4246, opened 2026-09-17
Updated deployment assessment schema prefix

ARADDCC002:
### Description of change

Updated the prefix of questions sent when using a deployment assessment schema

### Checklist


sylvesterkaczmarek (commented):
The required-field merge reads from the wrong schema node in both branches. The guard checks `properties.modelOverview.required` / `properties.signOff.required`, but the true branch then pulls `properties.details.required`, so any schema-specific required fields in those sections are discarded and the generated schema keeps only `modelIds` / `riskOwner`. Please read the corresponding section’s `required` array and add a regression where each section already has another required field.

ARADDCC002:
> The required-field merge reads from the wrong schema node in both branches. The guard checks `properties.modelOverview.required` / `properties.signOff.required`, but the true branch then pulls `properties.details.required`, so any schema-specific required fields in those sections are discarded and the generated schema keeps only `modelIds` / `riskOwner`. Please read the corresponding section’s `required` array and add a regression where each section already has another required field.

This seems to be working as expected. If I upload a schema with a question inside `modelOverview` and set it as required, I can see both the new field and the prefixed field inside the required array for that section.

backend/src/connectors/authorisation/base.ts
```diff
@@ -351,7 +351,7 @@ export class BasicAuthorisationConnector {
       deploymentAssessments.map(async (deploymentAssessment) => {
         const isNamedUser =
```
ON93754, line 352: Optional - Might be overkill, but worth considering if this will be re-used elsewhere in the deployment assessments code? Possibly a util?

ON93754 (approved):

backend/src/connectors/authorisation/base.ts
ARADDCC002: I think this is something we can do in the future if we reuse it, but good suggestion!

sylvesterkaczmarek (approved):
Rechecked current 45a066af. The required-field merge now reads modelOverview.required and signOff.required independently and prepends modelIds/riskOwner without discarding existing schema-specific requirements. The current test/CI matrix is green. This resolves the issue I raised.

PE39806 (approved):

PR #4248, opened 2026-09-17
Remove added review creation on the review role additions

GB907762:
### Description of change

The condition `(!Array.isArray(validReviews) || validReviews.length === 0)` is always true because find produces either an instance of the array if a predicate is true for an element or `undefined` otherwise.

This means if you create a new role in Bailo (such as DRO) additional reviews are created for all of the original review documents. 

### Checklist


JRB66955 (approved):

ARADDCC002 (approved):

ON93754 (approved):

sylvesterkaczmarek (commented):
The production caller in schema.ts invokes addReviewsForNewRole(...) without awaiting it. Since this helper performs awaited queries and saves, the schema update can finish before the reviews are created, and a rejection becomes an unhandled promise. The new unit test calls the helper directly, so it does not cover that path. Could the caller await this helper and add a schema-update regression so review creation is complete before the update returns?

PR #4250, opened 2026-09-18
Feature/bai 2735 create da

IR96334:
### Description of change

Added e2e tests for deployment assessment feature.

### Checklist
Tests:
- [x] Creation of DA
- [x] Editing of DA
- [x] Deletion of DA
- [x] Drafting of DA
- [x] Editing of DA
- [x] Publishing of DA


sylvesterkaczmarek (commented):
Both create flows use cy.get('[data-test=createButton') with the closing ] missing. That is an invalid CSS attribute selector, so these tests fail before they can click the create button. It should be [data-test=createButton].

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
GB907762, line 1: The `DOWN` invocations look flaky to me. Could you replace and use the menu item key.

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
```diff
@@ -0,0 +1,209 @@
+import 'cypress-axe'
+
```
GB907762: Not needed. This is an a11y dependency.

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
```diff
@@ -0,0 +1,209 @@
...
+const schemaId = 'minimal-deployment-assessment-schema-v1'
+
+const modelName = 'DATestModel'
+
+let DAid = ''
+let draftDAid = ''
+
+const newDAValuesObject = {
+  name: 'newTestDAName',
```
GB907762, line 9: Could you use full camel case in naming. For example `DA->deploymentAssessment`

GB907762 (changes requested):

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
```diff
@@ -0,0 +1,209 @@
...
+    },
+    about: {
+      deploymentSummary: 'This is still a deployment',
+      deploymentDate: '2098-05-01',
+    },
+  },
+}
+
+describe('Deployment Assessment Suite', () => {
+  //Creates model to be attached to DA, sets state to Production
```
PE39806: Prefer a space between comment marker `//` and beginning of text
```suggestion
  // Creates model to be attached to DA, sets state to Production
```

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
```diff
@@ -0,0 +1,209 @@
...
+    cy.get('[data-test=draftDeploymentAssessmentButton]').click()
+    cy.wait('@postDA').then((intercept) => {
+      expect(intercept.response?.statusCode).to.eq(201)
+      draftDAid = intercept.response?.body.deploymentAssessment.id
+      expect(intercept.response?.body.deploymentAssessment.draft).to.eq(true)
+    })
+
+    cy.url().should('contain', `deployment-assessments/${draftDAid}`)
+
+    //Erroneously attempts to publish with unfinished
```
PE39806:
```suggestion
    // Erroneously attempts to publish with unfinished
```

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
```diff
@@ -0,0 +1,209 @@
...
+      expect(intercept.response?.body.message).to.equal('Successfully removed deployment assessment.')
+    })
+    cy.location('pathname').should('eq', '/deployment-assessments')
+  })
+
+  it('Create a draft deployment assessment', () => {
+    cy.intercept('POST', '/api/v3/deployment-assessments').as('postDA')
+    cy.intercept('PATCH', `/api/v3/deployment-assessments/*`).as('patchDraftDA')
+
+    //Creates draft DA
```
PE39806:
```suggestion
    // Creates draft DA
```

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
```diff
@@ -0,0 +1,209 @@
...
+    cy.url().should('contain', `deployment-assessments/${draftDAid}`)
+
+    //Erroneously attempts to publish with unfinished
+    cy.get('[data-test=draftBanner]').within(() => cy.get('[data-test=publishDraftButton]').click())
+    cy.get('[data-test=confirmButton]').click()
+    cy.wait('@patchDraftDA').then((intercept) => {
+      expect(intercept.response?.statusCode).to.eq(400)
+    })
+
+    //Edit form and publish
```
PE39806:
```suggestion
    // Edit form and publish
```

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
```diff
@@ -0,0 +1,209 @@
...
+    })
+
+    cy.get('[data-test=draftBanner]').within(() => cy.get('[data-test=publishDraftButton]').click())
+    cy.get('[data-test=confirmButton]').click()
+    cy.wait('@patchDraftDA').then((intercept) => {
+      expect(intercept.response?.statusCode).to.eq(200)
+      expect(intercept.response?.body.deploymentAssessment.draft).to.eq(false)
+    })
+
+    //Draft banner should no longer exist
```
PE39806:
```suggestion
    // Draft banner should no longer exist
```

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
```diff
@@ -0,0 +1,209 @@
...
+    })
+
+    cy.url().should('contain', `deployment-assessments/${draftDAid}`)
+
+    //Erroneously attempts to publish with unfinished
+    cy.get('[data-test=draftBanner]').within(() => cy.get('[data-test=publishDraftButton]').click())
+    cy.get('[data-test=confirmButton]').click()
+    cy.wait('@patchDraftDA').then((intercept) => {
+      expect(intercept.response?.statusCode).to.eq(400)
+    })
```
PE39806, line 166: Add a check that the draft/publish banner is still present:
```ts
cy.get('[data-test=draftBanner]').should('exist')
cy.contains('required').should('be.visible')
```

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
```diff
@@ -0,0 +1,209 @@
...
+    cy.url().should('contain', `deployment-assessments/${DAid}`)
+  })
+
+  it('edit new deployment assessment', () => {
+    cy.intercept('GET', `/api/v3/deployment-assessments/${DAid}`).as('getDA')
+    cy.intercept('PATCH', `/api/v3/deployment-assessments/${DAid}`).as('patchDA')
+    cy.visit(`/deployment-assessments/${DAid}`)
+    cy.wait('@getDA').wait(500)
+    cy.get('[data-test=editFormButton]').click()
+    cy.get('[value=TestDAName]').clear().type(newDAValuesObject.name)
```
PE39806: Prefer
```suggestion
    cy.get('[data-test=editNameInput]').clear().type(newDAValuesObject.name)
```

PE39806 (changes requested):

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
IR96334: I understand your point, but I think this is fine for a few reasons.
Using internal component attributes is less readable and less obvious as to what its actually selecting, whereas the keyboard presses denote purpose and are a valid, even nominal use of component, and will/should remain functional throughout MUI versions.
It is an internal component, so its internal attributes can change between versions and mean little to the reader, thus I would like to avoid it where we can.

frontend/src/common/PageWithTabs.tsx
```diff
@@ -193,6 +193,7 @@ export default function PageWithTabs({
               variant='contained'
               onClick={actionButtonOnClick}
               startIcon={actionButtonIcon ? actionButtonIcon : <></>}
+              data-test='createButton'
```
ARADDCC002: The `PageWithTabs` component is designed to be generic, as such the button can technically have any kind of action attached to it. Whilst `createButton` might make sense for your e2e test suite, it might not serve the same function for another, so I think we should have a more generic name for it.

frontend/src/schemas/SchemaFormPage.tsx
```diff
@@ -80,6 +80,7 @@ export default function SchemaFormPage({
                             onClick={onSaveDraft}
                             loading={draftButtonLoading}
                             disabled={submitButtonLoading || actionsDisabled}
+                            data-test='draftDeploymentAssessmentButton'
```
ARADDCC002, line 83: I think this component is generic so a more generic data-test name would be better. Same for line 98.

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
```diff
@@ -0,0 +1,209 @@
+import 'cypress-axe'
+
+const schemaId = 'minimal-deployment-assessment-schema-v1'
+
+const modelName = 'DATestModel'
+
+let DAid = ''
+let draftDAid = ''
```
ARADDCC002: VERY minor, but `draftDAId`

frontend/cypress/e2e/bailo/deployment-assessment.cy.ts
```diff
@@ -0,0 +1,209 @@
...
+    cy.log('Upload new model and set schema via API')
+    cy.request('POST', 'http://localhost:8080/api/v2/models', {
+      name: modelName,
+      kind: 'model',
+      description: 'This is a test',
+      visibility: 'public',
+    }).then((response) => {
+      expect(response.status).to.eq(200)
+      expect(response.body.model).to.have.property('name', modelName)
+      newDAValuesObject.metadata.overview.modelIds = [response.body.model.id]
```
ARADDCC002: This path will need changing since I changed the structure of the core DA properties returned by the `getSchema` API.

You'll need to do this in various places.

ARADDCC002 (changes requested):

JRB66955 (dismissed):

sylvesterkaczmarek (changes requested):
Rechecked current 264c23cd. The malformed create selector I raised is fixed, but the latest patch introduces two new blockers.

1. frontend/cypress/e2e/bailo/deployment-assessment.cy.ts no longer imports anything, so TypeScript treats it as a global script. Its top-level schemaId/modelName now collide with make-and-review-access-request.cy.ts, and current CI fails with TS2451 redeclaration errors. Please keep the spec a module (for example retain an import or add export {}) so its top-level names are scoped to the file.

2. EntityAutocomplete now calls setTimeout(() => {}, 5000) inside onChange. That schedules an empty callback and returns immediately, so it cannot stabilize or delay the selection flow; it only adds an unrelated timer to production code. Please remove it and make any necessary test synchronization explicit in Cypress instead.

The selector issue itself is resolved, but the current head does not build.

sylvesterkaczmarek (changes requested):
Rechecked current 9530ee00. The two issues from my previous review are resolved: the Cypress spec is a module again via export {}, and the no-op production setTimeout has been removed.

One blocker remains in the new test suite itself. The current end_to_end job runs deployment-assessment.cy.ts with 0 passing / 4 failing. The first create flow reaches the Sign-Off section but never finds [data-test=entitySelector] input at line 96; the edit/delete failures then cascade because deploymentAssessmentId was never created, and the draft flow hits the same missing risk-owner control at line 195.

Could the DA test be aligned with the risk-owner control that actually renders for this schema/current user state, and then rerun independently enough that the later cases do not only fail as fallout from a failed create? Once that spec itself is green, I do not have a remaining issue from the earlier review.

frontend/src/MuiForms/DateSelector.tsx
```diff
@@ -63,6 +63,7 @@ export default function DateSelector({ onChange, value, label, registry, require
           onChange={handleChange}
           format='DD-MM-YYYY'
           sx={{ '.MuiInputBase-input': { p: '10px' } }}
+          data-test='dateSelectorInput'
```
PE39806: This works, but a more Stable" approach that I have found is to do something like
```ts
slotProps={{
  textField: {
    slotProps: {
      htmlInput: { 'data-test': 'dateSelectorInput'} as React.InputHTMLAttributes<HTMLInputElement>,
    },
  },
}}
```
This can then be safely interacted with by some
```ts
cy.get('[data-test=dateSelectorInput]').click()
cy.focused().type('01012099, { delay: 80, force: true })
```
and checked with
```ts
cy.get('[data-test=dateSelectorInput]').should('have.value', '01/01/2099')
```
Your method works so don't feel the need to swap over, but this is a helpful way to check that the typed value is actually present.

sylvesterkaczmarek (changes requested):
Rechecked current `9530ee00`. The end-to-end job is still failing specifically in the new deployment-assessment spec: both create paths time out at `[data-test=entitySelector] input`. The edit and delete failures are downstream because the first create never succeeds, so `deploymentAssessmentId` is never populated. This is consistent with the earlier review note that the core DA property structure changed: the test still assumes the old sign-off/risk-owner rendering. Please update the DA test to the current core schema/widget path and rerun the end-to-end job before merge.

sylvesterkaczmarek (changes requested):
Rechecked current b4477602 and the latest end_to_end run. The DA spec is still 0/4 passing. The first create now force-clicks Sign-Off and waits 20 seconds, but [data-test=entitySelector] still never appears; the draft flow still times out on [data-test=entitySelector] input. The edit/delete failures remain downstream because the create never produces deploymentAssessmentId.

So the remaining issue is not timing. The test still appears to be addressing a risk-owner control that this schema/current user state does not render. Aligning the test with the actual current sign-off schema/widget path, rather than extending waits, is still needed before this spec is green.

